PR #45, opened 2024-06-03
Test on Python 3.10 for Intel Gaudi

tiran:
SynapseAI / HabanaLabs 1.15 for Intel Gaudi is only available for Python 3.10.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -13,6 +13,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
+          - "3.10"
           - "3.11"
           - "3.12"
 
@@ -39,6 +40,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
+          - "3.10"
           - "3.11"
           - "3.12"
         test-script:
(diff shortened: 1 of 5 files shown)


tox.ini
```diff
@@ -1,6 +1,6 @@
 [tox]
 minversion = 3.2.0
-envlist=py,linter
+envlist=py{310,311,312},linter
```
dhellmann, line 3: I've been sticking to just 1 version of python locally to speed things up. Can we leave this just `py` and rely on CI to run the matrix?

dhellmann (commented):
Good catch, thanks!

.github/workflows/test.yaml
```diff
@@ -39,6 +40,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version:
+          - "3.10"
```
dhellmann, line 43: It would be good to update the `.mergify.yaml` file to require that these new jobs pass, too.

tox.ini
tiran: makes sense!

dhellmann (approved):

PR #49, opened 2024-06-04
Vendor Rust crates into source distributions

tiran:
`vendor_rust()` takes an existing sdist tar ball and creates a new sdist with all Rust crates vendored. This allows offline compilation without downloading any Rust crates at build time.

Switch from `tomli` to `toml`. The `tomli` package does not have write support.

I have successfully wheel builds with vendored crates for

- `cryptography-42.0.7.tar.gz`
- `maturin-1.5.1.tar.gz`
- `pydantic_core-2.18.4.tar.gz`
- `setuptools-rust-1.9.0.tar.gz`

with `CARGO_NET_OFFLINE=true python3 -m build -w`.


src/fromager/vendor_rust.py
```diff
@@ -0,0 +1,192 @@
...
+        cfg = {}
+
+    cfg.update(CARGO_CONFIG)
+
+    with open(config_toml, "w", encoding="utf-8") as f:
+        toml.dump(cfg, f)
+
+
+def vendor_rust(
+    sdist: pathlib.Path,
```
dhellmann, line 134: We should work out how to integrate this into the flow.

Today we have

- download-source
- prepare-source
  - unpack
  - patch
- prepare-build
- build-wheel

That leaves the patched source tree with build artifacts in it on the disk, and leaves the original sdist unmodified.

I think we want instead something like

- download-source
- prepare-source
  - unpack
  - patch
  - vendor rust
- re-roll-sdist (to preserve a copy of our patched sources with the vendored rust code, before any build artifacts are written, so we can distribute the sources we use for the build)
- prepare-build
- build-wheel

If that's the flow, then I think for input we'll want this function to take the path to the source root instead of the sdist file.

src/fromager/vendor_rust.py
```diff
@@ -0,0 +1,192 @@
...
+    args = ["cargo", "vendor", f"--manifest-path={manifests[0]}"]
+    for manifest in manifests[1:]:
+        args.append(f"--sync={manifest}")
+    args.append(project_dir / VENDOR_DIR)
+    subprocess.check_call(args, stdout=subprocess.DEVNULL, env={})
+
+    # Remove large, unused archive and lib files. They are only
+    # used on Windows, macOS, and iOS. This makes the vendor bundle up to
+    # 60% smaller.
+    for pattern in VENDOR_FILTER_PATTERNS:
```
dhellmann, line 85: Should we make the pattern configurable, eventually? Are we likely to want to have it be part of the per-package customization interface?

src/fromager/vendor_rust.py
```diff
@@ -0,0 +1,192 @@
...
+    msrv: typing.Optional[Version] = None,
+):
+    """Modify Cargo.toml
+
+    pydantic-core has an unnecessary high MSRV.
+    """
+    with open(fileame, "r", encoding="utf-8") as f:
+        cfg = toml.load(f)
+
+    modified = False
```
dhellmann, line 103: I'm trying to keep project-specific knowledge out of the core bits, and provide an override hook or other mechanism for a build system to inject that sort of knowledge based on the packages it wants to build. In this case, for setting the rust level, I could see us setting a global config in the settings file, or I could see a patch being applied to the package.

src/fromager/vendor_rust.py
```diff
@@ -0,0 +1,192 @@
...
+
+    with tempfile.TemporaryDirectory() as tmp:
+        tmpdir = pathlib.Path(tmp)
+        with tarfile.open(sdist) as tar:
+            # scan for Cargo.toml before extracting files
+            for filename in tar.getnames():
+                if os.path.basename(filename) == "Cargo.toml":
+                    break
+            else:
+                raise NoCargoToml(sdist)
```
dhellmann, line 155: Maybe instead of an exception, just return a flag indicating that no work was done?

src/fromager/vendor_rust.py
```diff
@@ -0,0 +1,192 @@
...
+            # filter argument was added in 3.10.12, 3.11.4, 3.12.0
+            tar.extractall(path=tmpdir, filter="data")
+
+        project_name = sdist.name[: -len(".tar.gz")]
+        project_dir = tmpdir / project_name
+        if not project_dir.is_dir():
+            raise FileNotFoundError(project_dir)
+
+        manifests = list(project_dir.glob("**/Cargo.toml"))
+        for manifest in manifests:
```
dhellmann, line 165: An info log message here would provide good progress output.

dhellmann (commented):
This is cool, thank you!

I left some notes about how it might be integrated in an inline comment. Let me know what you think, maybe you had a different idea in mind.

One global comment is I'm trying to log fairly verbosely. Debug messages are usually written to a file. Info messages go to the console as progress messages, indicating that some action (especially anything that might take time) is starting.

tiran (commented):
I have addressed your suggestions and updated the code.

I also added a minimal Rust project with pyo3 and setuptools-rust to `e2e` directory for future testing.

src/fromager/vendor_rust.py
tiran: The function now returns true or false instead of raising an exception.

src/fromager/vendor_rust.py
dhellmann: OK, if it's not going to be variable keeping it in the source makes sense. We can always make it configurable later, if we need to.

src/fromager/wheels.py
```diff
@@ -15,6 +15,9 @@ def build_wheel(ctx, req, sdist_root_dir, build_env):
     if not builder:
         builder = default_build_wheel
     extra_environ = overrides.extra_environ_for_pkg(ctx.envs_dir, req.name, ctx.variant)
+    # TODO: refactor?
+    # Build Rust without network access
+    extra_environ["CARGO_NET_OFFLINE"] = "true"
```
dhellmann, line 20: I wonder if this should go in `default_build_wheel`? Maybe it's better to do it here so any overrides also use the same setting without having to remember to add it themselves.

dhellmann (approved):
Thank you!

PR #55, opened 2024-06-06
convert cli processing to use click

dhellmann:
Restructure the step-by-step commands in anticipation of adding some
new all-in-one commands that provide a higher-level interface.


dhellmann:
@mergifyio refresh

PR #76, opened 2024-06-14
allowing specified options for variants

Gregory-Pereira:
While learning the codebase and playing around I thought it would be a good idea to restrict `VARIANT` options and keep them inline with the backend.

### DIFF
--- a/src/fromager/commands/bootstrap.py
+++ b/src/fromager/commands/bootstrap.py
@@ -24,8 +24,8 @@ def _get_requirements_from_args(toplevel, requirements_file):
 
 
 @click.command()
-@click.option('--variant', default='cpu',
-              help='the build variant name')
+@click.option('--variant', default='cpu', help='the build variant name',
+              type=click.Choice(['cpu', 'cuda-ubi9', 'gaudi-ubi9', 'rocm-ubi9']))
 @click.option('-r', '--requirements-file', multiple=True,
               help='pip requirements file')
 @click.argument('toplevel', nargs=-1)


src/fromager/commands/bootstrap.py
```diff
@@ -24,8 +24,8 @@ def _get_requirements_from_args(toplevel, requirements_file):
 
 
 @click.command()
-@click.option('--variant', default='cpu',
-              help='the build variant name')
+@click.option('--variant', default='cpu', help='the build variant name',
+              type=click.Choice(['cpu', 'cuda-ubi9', 'gaudi-ubi9', 'rocm-ubi9']))
```
dhellmann, line 28: These are the variants for what we're building for RHEL AI, but other people may want to build other variants.

I had considered looking for variants defined in the settings file, since that would be under the control of the user running fromager. What do you think about adding a section to the settings file or trying to derive it from that file in some other way?

For example, right now the file has something like 

```
pre_built:
  variant:
    - package1
    - package2
```

It might be better to flip the nesting levels of `pre_built` and `variant`, I don't know?

Gregory-Pereira: Fantastic idea! I didn't consider open source concerns so the code should derive from the settings file and this will also allow us to still set this. Ill proceed with this but raise this again if I come across a better implementation. In the mean time I will convert this to a draft until I can re-implement this, starting with sequence download.

dhellmann:
We can't use this approach, so I'm going to close this PR. We talked about validating using the config file at one point, but let's talk again before spending time on implementation.

PR #85, opened 2024-06-16
adding download sequence command

Gregory-Pereira:
Implements `download-sequence` command and corresponding tests. Uses a `build-order.json` compiled for CPU for `simplejson==3.19.2`. Later on enablement / testing should be added for multiple sdist servers.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -62,6 +62,10 @@ pull_request_rules:
           - check-success=e2e (3.11, 1.75, rust_vendor)
           - check-success=e2e (3.12, 1.75, rust_vendor)
 
+          - check-success=e2e (3.10, 1.75, download_sequence)
+          - check-success=e2e (3.11, 1.75, download_sequence)
+          - check-success=e2e (3.12, 1.75, download_sequence)
+
           - '-draft'
 
           - or:
--- a/src/fromager/commands/download_sequence.py
+++ b/src/fromager/commands/download_sequence.py
@@ -1,0 +1,30 @@
+import logging
+import json
+
+import click
+from packaging.requirements import Requirement
+
+from .. import sources
+
+logger = logging.getLogger(__name__)
+
+
+@click.command()
+@click.argument('build_order_file')
+@click.argument('sdist_server_url')
+@click.pass_obj
+def download_sequence(wkctx, build_order_file, sdist_server_url):
+    """Download a sequence of source distributions in order.
+
+    BUILD_ORDER_FILE is the build-order.json files to build
+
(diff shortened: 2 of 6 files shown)


.mergify.yml
```diff
@@ -62,6 +62,10 @@ pull_request_rules:
           - check-success=e2e (3.11, 1.75, rust_vendor)
           - check-success=e2e (3.12, 1.75, rust_vendor)
 
+          - check-success=e2e (3.10, 1.75, download_sequence)
```
dhellmann, line 65: It's not a big deal, but I've been trying to keep these in alphabetical order based on the test name.

src/fromager/commands/download_sequence.py
```diff
@@ -0,0 +1,30 @@
...
+    SDIST_SERVER_URL is the URL for a PyPI-compatible package index hosting sdists
+
+    Performs the equivalent of the 'step download-source-archive' command for each item in
+    the build order file.
+
+    """
+    with open(build_order_file, 'r') as f:
+        for entry in json.load(f):
+            req = Requirement(f"{entry['dist']}=={entry['version']}")
+            sources.download_source(wkctx, req, [sdist_server_url])
```
dhellmann, line 30: If this doesn't already print any progress information, we may want to change https://github.com/python-wheel-build/fromager/blob/main/src/fromager/sources.py#L35 to use info level instead of debug. That will ensure someone running this sees some info about what's going on.

src/fromager/commands/download_sequence.py
```diff
@@ -0,0 +1,30 @@
...
+    BUILD_ORDER_FILE is the build-order.json files to build
+
+    SDIST_SERVER_URL is the URL for a PyPI-compatible package index hosting sdists
+
+    Performs the equivalent of the 'step download-source-archive' command for each item in
+    the build order file.
+
+    """
+    with open(build_order_file, 'r') as f:
+        for entry in json.load(f):
```
dhellmann, line 28: The loop is nice and straightforward. It may be fairly slow, though. Did you give any thought to parallelizing it? `futures.ThreadPoolExecutor()` is an easy way to do that. There's an example in https://gitlab.com/redhat/rhel-ai/wheels/pipeline-tool/-/blob/main/src/pipeline_tool/jobs.py?ref_type=heads#L58

Let's take this as-is, and consider whether we ought to make that enhancement in another PR.

dhellmann (approved):
I had a couple of nits inline, but this looks good so let's merge it and iterate.

.mergify.yml
Gregory-Pereira: Understood Ill update this if I make another file touching this file.

PR #86, opened 2024-06-16
WIP: Create build order step

Gregory-Pereira:
Implementing ability to create build order files without downloading source or building wheels.

### DIFF
--- a/src/fromager/commands/bootstrap.py
+++ b/src/fromager/commands/bootstrap.py
@@ -8,7 +8,7 @@ from .. import sdist, server
 logger = logging.getLogger(__name__)
 
 
-def _get_requirements_from_args(toplevel, requirements_file):
+def get_requirements_from_args(toplevel, requirements_file):
     to_build = []
     to_build.extend(toplevel)
     for filename in requirements_file:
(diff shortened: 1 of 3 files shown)


dhellmann:
Is the idea that the user would feed it a requirements file with everything already in the right order?

Gregory-Pereira:
The concept behind this PR was to enable just the creation of the `build-order.json` as a step, that didn't actually do the downloading and the building. In retrospect I am not actually sure this is even possible, as `build-order.json`s are variant specific. I also didn't know if dependency resolution is possible without downloading the sdists and building them, which the more I think about it probably is not possible.

dhellmann:
> The concept behind this PR was to enable just the creation of the `build-order.json` as a step, that didn't actually do the downloading and the building. In retrospect I am not actually sure this is even possible, as `build-order.json`s are variant specific. I also didn't know if dependency resolution is possible without downloading the sdists and building them, which the more I think about it probably is not possible.

Ah, right, the process for determining the dependencies involves 90% of the work of building the wheels themselves, so the bootstrap process does it all. And you're also correct that the results are variant-specific. Some of the build scripts dynamically determine dependencies based on settings fed into the build, like which accelerator to optimize for.

Gregory-Pereira:
Understood, in that case I will close this PR.

PR #88, opened 2024-06-17
let the resolver filter the type of distribution

### LINKED ISSUE #87: pre-built wheel download incorrectly downloads sdists
When bootstrapping `hydra-core` the resolver finds 2 links for the latest version:

```
    <a href="https://files.pythonhosted.org/packages/6d/8e/07e42bc434a847154083b315779b0a81d567154504624e181caf2c71cd98/hydra-core-1.3.2.tar.gz#sha256=8a878ed67216997c3e9d88a8e72e7b4767e81af37afb4ea3334b269a4390a824">hydra-core-1.3.2.tar.gz</a>
    <br/>
    <a href="https://files.pythonhosted.org/packages/c6/50/e0edd38dcd63fb26a8547f13d28f7a008bc4a3fd4eb4ff030673f22ad41a/hydra_core-1.3.2-py3-none-any.whl#sha256=fa0238a9e31df3373b35b0bfb672c34cc92718d21f81311d8996a16de1141d8b" data-dist-info-metadata="sha256=399046cbf9ae7ebab8dfd009e2b4f748212c710a0e75ca501a72bbb2d456e2e7" data-core-metadata="sha256=399046cbf9ae7ebab8dfd009e2b4f748212c710a0e75ca501a72bbb2d456e2e7">hydra_core-1.3.2-py3-none-any.whl</a>
```

Note that the sdist uses `-` and the wheel uses `_` in the dist name portion of the filename.

When the resolver sorts those the sdist comes out first and ends up being reported as the wheel file.

The flag to the resolver is `sdist_only` and that is set to False, but what we really want is a flag that says whether to return a wheel or an sdist so that we only get one or the other type of output.

dhellmann:
Change the resolver code to have explicit options for whether to include
sdist and wheel distributions as candidates.

Rename sources.resolve_sdist() to sources.resolve_dist() to reflect the
scope change.

Add requests_mock testing dependency.

Fixes #87

### DIFF
--- a/src/fromager/resolver.py
+++ b/src/fromager/resolver.py
@@ -159,9 +159,11 @@ def get_metadata_for_wheel(url):
 
 
 class PyPIProvider(ExtrasProvider):
-    def __init__(self, only_sdists=False, sdist_server_url='https://pypi.org/simple/'):
+    def __init__(self, include_sdists=True, include_wheels=True,
+                 sdist_server_url='https://pypi.org/simple/'):
         super().__init__()
-        self.only_sdists = only_sdists
+        self.include_sdists = include_sdists
+        self.include_wheels = include_wheels
         self.sdist_server_url = sdist_server_url
 
     def identify(self, requirement_or_candidate):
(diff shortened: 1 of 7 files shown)


src/fromager/resolver.py
```diff
@@ -184,13 +186,21 @@ def find_matches(self, identifier, requirements, incompatibilities):
...
+            if candidate.version in bad_versions:
+                continue
+            # Skip versions that do not match the requirement
+            if not all(candidate.version in r.specifier for r in requirements):
+                continue
+            # Only include sdists if we're asked to
+            if self.include_sdists and not candidate.is_sdist:
+                continue
+            # Only include wheels if we're asked to
+            if self.include_wheels and candidate.is_sdist:
```
Gregory-Pereira, line 201: Are there other things on PYPI that might be neither wheel nor sdist?

src/fromager/sources.py
```diff
@@ -73,7 +78,7 @@ def resolve_sdist(req, sdist_server_url, only_sdists=True):
 
 def default_download_source(ctx, req, sdist_server_url):
     "Download the requirement and return the name of the output path."
-    url, version = resolve_sdist(req, sdist_server_url)
+    url, version = resolve_dist(req, sdist_server_url, include_wheels=False)
```
Gregory-Pereira, line 81: since we are refactoring concept of sdist --> dist we should probably also explicitly say `include_sdists=true`, no?

Gregory-Pereira (commented):
/lgtm just 2 questions

src/fromager/resolver.py
dhellmann: There's an old format called "egg" but it's obsolete so if we did find any we wouldn't want to use them.

src/fromager/sources.py
dhellmann: sure, that's reasonable, I'll post another PR with that

PR #93, opened 2024-06-17
when bootstrapping, report origin of each requirement

dhellmann:
The CLI takes as input requirements or files with lists of requirements. This adds the filename as part of the build-order output and log messages so it is easier to understand where a requirement was fed into the system.

### DIFF
--- a/src/fromager/commands/bootstrap.py
+++ b/src/fromager/commands/bootstrap.py
@@ -10,7 +10,7 @@ logger = logging.getLogger(__name__)
 
 def _get_requirements_from_args(toplevel, requirements_file):
     to_build = []
-    to_build.extend(toplevel)
+    to_build.extend(('toplevel', t) for t in toplevel)
     for filename in requirements_file:
         with open(filename, 'r') as f:
             for line in f:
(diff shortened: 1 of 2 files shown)


Gregory-Pereira (approved):
feature looks good to me. Probably a stupid question but I was wondering about is usage flow. From my understanding wouldnt the most likely use of this be passing multiple requirements or `build-order.json` files? I would expect we might see:
```tree
├── requirement-one
│   └── requirements.txt
└── requirement-two
    └── requirements.txt
```
In which case the filenames would be the same right? Or am I misunderstanding something? Does this mean by default we should start naming build-orders by top-level package name?

dhellmann:
> feature looks good to me. Probably a stupid question but I was wondering about is usage flow. From my understanding wouldnt the most likely use of this be passing multiple requirements or `build-order.json` files? I would expect we might see:
> 
> ```
> ├── requirement-one
> │   └── requirements.txt
> └── requirement-two
>     └── requirements.txt
> ```
> 
> In which case the filenames would be the same right? Or am I misunderstanding something? Does this mean by default we should start naming build-orders by top-level package name?

I should have included sample output in the PR. The output includes the full filename, so it looks like

```
packaging: * handling container-dependencies/deepspeed/requirements.txt requirement packaging []
```

Gregory-Pereira:
Sounds great, thanks for the clarification. Merge whenever you are comfortable.

PR #95, opened 2024-06-19
filter out pre-builts

Gregory-Pereira:
Outcome of discussion yesterday. Must check pre-builts and filter them out of source upload.


src/fromager/commands/download_sequence.py
```diff
@@ -26,5 +26,7 @@ def download_sequence(wkctx, build_order_file, sdist_server_url):
     """
     with open(build_order_file, 'r') as f:
         for entry in json.load(f):
+            if entry["prebuilt"] == True:
```
dhellmann, line 29: I think the linter is going to want

```suggestion
            if entry["prebuilt"]:
```

Gregory-Pereira: Sounds good I will make that change. Any clue why the other tests are breaking? I didn't think this change would have any impact on `build_order`, `build_steps` or `boostrap`? Is CI just broken in general right now?

dhellmann: https://github.com/python-wheel-build/fromager/pull/97 should fix those other jobs.

dhellmann (approved):

PR #97, opened 2024-06-20
make e2e jobs more robust

dhellmann:
Some of the e2e jobs had hard-coded expectations for version numbers
of dependencies of packages involved in the tests. Those versions will
change over time as new releases are created, so this commit reworks
the tests to be more flexible about the version numbers and removes
the dependencies that have been dropped from the test packages.


Gregory-Pereira (approved):
/lgtm, thanks!

PR #98, opened 2024-06-21
feature: make a new sdist after running prepare_source

### LINKED ISSUE #67: Add step to create updated source archive
After prepare source step, make a new sdist

dhellmann:
Add a step command to build an sdist for a set of sources.

Update the build and build-sequence commands to produce an sdist as
well as a wheel when building something.

Update the bootstrap command to create sdists for everything that is
bootstrapped.

Fixes #67

### DIFF
--- a/src/fromager/commands/step.py
+++ b/src/fromager/commands/step.py
@@ -48,8 +48,6 @@ def prepare_source(wkctx, dist_name, dist_version):
 
     DIST_VERSION is the version to process
 
-    SDIST_SERVER_URL is the URL for a PyPI-compatible package index hosting sdists
-
     """
     req = Requirement(f"{dist_name}=={dist_version}")
     sdists_downloads = pathlib.Path(wkctx.sdists_repo) / "downloads"
@@ -64,6 +62,27 @@ def prepare_source(wkctx, dist_name, dist_version):
     # FIXME: Does the version need to be a Version instead of str?
     source_root_dir = sources.prepare_source(wkctx, req, source_filename, dist_version)
     print(source_root_dir)
+
+
+@step.command()
+@click.argument("dist_name")
+@click.argument("dist_version")
+@click.pass_obj
+def build_sdist(wkctx, dist_name, dist_version):
+    """build a new source distribution for the package
+
+    DIST_NAME is the name of a distribution
+
+    DIST_VERSION is the version to process
+
+    The source distribution is placed in the `sdists-repo/builds`
+    directory.
+
+    """
+    req = Requirement(f"{dist_name}=={dist_version}")
+    source_root_dir = _find_source_root_dir(wkctx.work_dir, req, dist_version)
+    sdist_filename = sources.build_sdist(wkctx, req, source_root_dir)
+    print(sdist_filename)
 
 
 def _find_source_root_dir(work_dir, req, dist_version):
--- a/src/fromager/sources.py
+++ b/src/fromager/sources.py
@@ -12,7 +12,7 @@ from urllib.parse import urlparse
 import requests
 import resolvelib
 
-from . import overrides, resolver, vendor_rust
+from . import dependencies, overrides, resolver, tarballs, vendor_rust
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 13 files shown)


dhellmann:
This is triggering more errors in our bootstrapping process because some packages expect to have their sdist built from a git checkout not a directory that is itself an unpacked sdist.

dhellmann:
@tiran am I right in believing that just taring the directory up after we patch it is OK for creating a new source distribution? There's nothing in the source distribution that lists the files that are part of the source, right? The archive contents themselves are used to determine that?

tiran:
You have to re-generate the `PKG-INFO` file if you modify any project metadata like project version, dependencies, or versions of dependencies.

tiran:
Are you having trouble with packages that use setuptools-scm to detect version from VCS? You can force setuptools-scm to use a version with `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_${NORMALIZED_DIST_NAME}` env var, https://setuptools-scm.readthedocs.io/en/stable/config/#environment-variables .

dhellmann:
The package that caused trouble used meson-python for builds. The error was

```
meson-python: building source distribution in /work/bootstrap-output/work-dir/meson_python-0.16.0/meson_python-0.16.0
['/usr/bin/python3.11', '/usr/local/lib/python3.11/site-packages/pyproject_hooks/_in_process/_in_process.py', 'build_sdist', '/tmp/tmp48aif8xy'] failed with + meson setup /work/bootstrap-output/work-dir/meson_python-0.16.0/meson_python-0.16.0 /work/bootstrap-output/work-dir/meson_python-0.16.0/meson_python-0.16.0/.mesonpy-az7vui48 -Dbuildtype=release -Db_ndebug=if-release -Db_vscrt=md --native-file=/work/bootstrap-output/work-dir/meson_python-0.16.0/meson_python-0.16.0/.mesonpy-az7vui48/meson-python-native-file.ini
The Meson build system
Version: 1.4.1
Source dir: /work/bootstrap-output/work-dir/meson_python-0.16.0/meson_python-0.16.0
Build dir: /work/bootstrap-output/work-dir/meson_python-0.16.0/meson_python-0.16.0/.mesonpy-az7vui48
Build type: native build
Project name: meson-python
Project version: 0.16.0
Host machine cpu family: x86_64
Host machine cpu: x86_64
Program python found: YES (/usr/bin/python3.11)
Build targets in project: 0

meson-python 0.16.0  

  User defined options
    Native files: /work/bootstrap-output/work-dir/meson_python-0.16.0/meson_python-0.16.0/.mesonpy-az7vui48/meson-python-native-file.ini
    buildtype   : release
    b_ndebug    : if-release
    b_vscrt     : md

Found ninja-1.10.2 at /usr/local/bin/ninja
+ meson dist --allow-dirty --no-tests --formats gztar
Dist currently only works with Git or Mercurial repos
```

dhellmann:
> You have to re-generate the `PKG-INFO` file if you modify any project metadata like project version, dependencies, or versions of dependencies.

There doesn't appear to be an interface in `pyproject_hooks` just to generate the metadata for an sdist (there is one for generating wheel metadata). 

We're not doing build versioning, yet, so we shouldn't be modifying any version numbers. When we do, fromager could just edit `PKG-INFO` directly.

It's icky, but if we change dependencies for something using a patch, we could use the same patch to modify `PKG-INFO`.

So, I think we're probably OK for now, as long as we acknowledge those limitations.

dhellmann:
@tiran , unless you think this approach is fundamentally wrong I'd like to merge it and do a new release this week.

src/fromager/commands/step.py
```diff
@@ -66,6 +64,27 @@ def prepare_source(wkctx, dist_name, dist_version):
...
+    DIST_VERSION is the version to process
+
+    The source distribution is placed in the `sdists-repo/builds`
+    directory.
+
+    """
+    req = Requirement(f"{dist_name}=={dist_version}")
+    source_root_dir = _find_source_root_dir(wkctx.work_dir, req, dist_version)
+    sdist_filename = sources.build_sdist(wkctx, req, source_root_dir)
+    print(sdist_filename)
```
tiran, line 85: debug print?

```suggestion
    return sdist_filename
```

src/fromager/sources.py
```diff
@@ -210,3 +210,45 @@ def _default_prepare_source(ctx, req, source_filename, version):
...
+    # The format argument is specified based on
+    # https://peps.python.org/pep-0517/#build-sdist.
+    with tarfile.open(sdist_filename, "x:gz", format=tarfile.PAX_FORMAT) as sdist:
+        sdist.add(sdist_root_dir, arcname=sdist_root_dir.name)
+    return sdist_filename
+
+
+def pep517_build_sdist(ctx, req, sdist_root_dir):
+    """Use the PEP 517 API to build a source distribution from a modified source tree."""
+    extra_environ = overrides.extra_environ_for_pkg(ctx.envs_dir, req.name, ctx.variant)
```
tiran, line 248: Can we include setuptools-scm override here?
```
normalized = re.sub(r"[\._-]+", "_", name).upper()
extra_environ[f"SETUPTOOLS_SCM_PRETEND_VERSION_FOR_{normalized}"] = version
```

src/fromager/sources.py
```diff
@@ -210,3 +210,45 @@ def _default_prepare_source(ctx, req, source_filename, version):
...
+    #
+    # For cases where the PEP 517 approach works, use
+    # pep517_build_sdist().
+    sdist_filename = ctx.sdists_builds / (sdist_root_dir.name + ".tar.gz")
+    if sdist_filename.exists():
+        sdist_filename.unlink()
+    # The format argument is specified based on
+    # https://peps.python.org/pep-0517/#build-sdist.
+    with tarfile.open(sdist_filename, "x:gz", format=tarfile.PAX_FORMAT) as sdist:
+        sdist.add(sdist_root_dir, arcname=sdist_root_dir.name)
```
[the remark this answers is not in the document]
dhellmann: Done, thanks for the pointer to that code!

src/fromager/commands/step.py
dhellmann: The commands print the thing they create as output so that scripts can save the values to do something with them.

src/fromager/sources.py
dhellmann: We could. I'm not sure whether we should put that in the core or not, though. Not everything uses setuptools-scm, or needs the variable set, right?

tiran (approved):

Issue #68, opened 2024-06-09
Rebuilds


dhellmann:
We need to be able to track rebuilds of packages. One way to do that is to set the build number part of the version.

tiran:
Small correction: The build number is not part of the version. The build number is only part of the wheel file name and not part of the package metadata.

[File name convention](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#file-name-convention)
> **build tag**
> Optional build number. Must start with a digit. Acts as a tie-breaker if two wheel file names are the same in all other respects (i.e. name, version, and other tags). Sort as an empty tuple if unspecified, else sort as a two-item tuple with the first item being the initial digits as an int, and the second item being the remainder of the tag as a str.
>
> A common use-case for build numbers is rebuilding a binary distribution due to a change in the build environment, like when using the manylinux image to build distributions using pre-release CPython versions.
> 
> Build numbers are not a part of the distribution version and thus are difficult to reference externally, especially so outside the Python ecosystem of tools and standards. A common case where a distribution would need to referenced externally is when resolving a security vulnerability.
>
> Due to this limitation, new distributions which need to be referenced externally should not use build numbers when building the new distribution. Instead a new distribution version should be created for such cases.

The version number can include a [post release number](https://packaging.python.org/en/latest/specifications/version-specifiers/#post-releases). The post release number is part of the version of a package and therefore in the package metadata. The post number is visible in "pip list".

We can use either post release number or a combination of post release number + local version track CVE fixes. We could also include a custom file in the package metadata with release notes.

shubhbapna:
related to #343 #316

PR #104, opened 2024-06-26
Change post and local version of a wheel file

tiran:
A new function `change_wheel_version()` changes the post and local elements of the version of a wheel file. The function creates a new file with new file name, new `dist-info` directory, and updated `METADATA` file.

See: #68

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,6 +31,7 @@ classifiers = [
 requires-python = ">=3.10"
 
 dependencies = [
+    "auditwheel",
     "click>=8.1.7",
     "html5lib",
     "packaging",
--- a/src/fromager/wheelversion.py
+++ b/src/fromager/wheelversion.py
@@ -1,0 +1,136 @@
+# SPDX-License-Identifier: Apache-2.0
+"""Change wheel version"""
+
+import pathlib
+from email.generator import BytesGenerator
+from email.message import Message
+from email.parser import BytesParser
+from email.policy import EmailPolicy, compat32
+
+from auditwheel.wheeltools import InWheelCtx
+from packaging.tags import Tag
+from packaging.utils import BuildTag, NormalizedName, parse_wheel_filename
+from packaging.version import LocalType, Version
+
+__all__ = ("change_wheel_version",)
+
+
+def join_wheel_filename(
+    name: NormalizedName,
+    version: Version,
+    buildtag: BuildTag,
+    tags: frozenset[Tag],
+) -> str:
+    """Make a wheel file name from its components"""
+    parts = [str(name), str(version)]
+    if buildtag:
+        parts.append(f"{buildtag[0]}{buildtag[1]}")
+    parts.extend(str(tag) for tag in tags)
+    return "-".join(parts) + ".whl"
+
+
+def read_pkg_info(pkg_info_path: pathlib.Path) -> Message:
+    """Read a package info file"""
+    with pkg_info_path.open("rb") as f:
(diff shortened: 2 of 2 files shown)


dhellmann:
This changes the wheel after it's built. As part of producing our product builds, we need to also deliver the source that went into the build. The toolchain for doing that looks at the versions and grabs both the source and the wheel. So, I think we have to be able to update the source distribution version info, too. To do that, do we have to do more than rename the file to use the same version?

tiran:
A [modern source distribution](https://packaging.python.org/en/latest/specifications/source-distribution-format/) has the version in:

- the file name (`name-version.tar.gz`)
- the first level directory (`name-version`)
- the `PKG-INFO` file

At a bare minimum we have to update these three places to create a new sdist post version. There is also a legacy sdist format (generated by distutil's sdist). I don't know if any of our packages use the legacy format.

However there is more work to do if we want our `shrubbery-42.0.post1.tar.gz` to generate a `shrubbery-42.0.post1-py3-none-any.whl` without an additional step. The output version is defined by the build system and build backend. AFAIK the build backend does not necessarily use the version information from the sdist metadata. Instead the version number can be all over the place:

- hard coded in pyproject.toml
- hard coded in setup.cfg
- hard coded in setup.py
- dynamic provider in pyproject.toml (e.g. setuptools-scm, read from a Python attribute)
- dynamic attribute in setup.cfg, e.g. `version = attr: package.__version__`
- custom code in setup.py that gets the version from another place like importing or exec-ing code.
- other shenanigans

tiran:
I suggest that we do not attempt to fix the sdist -> whl version problem for now. That's going to be too much hand holding for packages.

- we create `shrubbery-42.0.post1.tar.gz` from original `shrubbery-42.0.tar.gz` + patches. We only change name, directory name, and `PKG-INFO` to contain `42.0.post1` version.
- upload  `shrubbery-42.0.post1.tar.gz`
- build system fetches the new  `shrubbery-42.0.post1.tar.gz` sdist and builds it. This will result in  `shrubbery-42.0-py3-none-any.whl` without `post1` in the version, because we didn't fix the version in the build system.
- build system uses the new code from this PR to correct the version of the wheel to  `shrubbery-42.0.post1-py3-none-any.whl`

For GPU and vendor-specific packages like PyTorch, we also include the GPU arch and version in the file name, e.g. `torch-2.3.1+cuda125-cp311-cp311-linux_x86_64.whl`. We could either teach our system to map wheels with local versions (`+cuda125`) to sdist without local version. Or we could upload the same sources under two different names, e.g. `torch-2.3.1+cuda125.tar.gz` and `torch-2.3.1+rocm60.tar.gz` with same content, except for root directory name and `PKG-INFO`.

dhellmann:
I like the idea of doing the minimal work to update the sdist and wheel separately. We can improve on that later, if needed.

I'm not sure we need the variant tags in the versions. Within the build system everything about the variants is segregated into separate indexes and images already, so we may not need to add the info to the versions, too. When we figure out how to track the post version number, maybe it will become clear how to include that info and if it's straightforward and doesn't interfered with package selection then we can do it.

Right now, I'm assuming we need to potentially support a different post version for each package we build. Potentially for each variant, too. We could track that in the overrides/settings.yaml or some other file, but it feels like it could become unwieldy so I'm not sure that's what we want.

dhellmann:
I think #325 makes this obsolete?

tiran:
I agree.

And if we ever need the feature, then it should be implemented as part of wheel unpack and pack process.

PR #105, opened 2024-06-26
feature: let overrides do version resolution

### LINKED ISSUE #103: override function to control version resolution
The default behavior is to use PyPI.org to resolve a requirement specification to pick a package version. In cases where someone is building from a fork, that may not produce a valid version. Users should be able to override the version resolution algorithm.

dhellmann:
Some overrides are for doing things like building from a fork of a
project with fixes that aren't upstream. In those cases, the builder
might have to determine the version in some way other than looking at
pypi.org. Provide a hook for that, and include a sample implementation
that uses tags on github.com.

Resolves #103

### DIFF
--- a/src/fromager/resolver.py
+++ b/src/fromager/resolver.py
@@ -249,22 +249,24 @@ class PyPIProvider(ExtrasProvider):
 class GitHubTagProvider(ExtrasProvider):
     def __init__(
         self,
-        organization,
-        repo,
+        organization: str,
+        repo: str,
     ):
         super().__init__()
         self.organization = organization
         self.repo = repo
         self.client = github.Github()
 
-    def identify(self, requirement_or_candidate):
+    def identify(self, requirement_or_candidate: Requirement | Candidate) -> str:
         return canonicalize_name(requirement_or_candidate.name)
 
-    def get_extras_for(self, requirement_or_candidate):
+    def get_extras_for(
+        self, requirement_or_candidate: Requirement | Candidate
+    ) -> tuple[str]:
         # Extras is a set, which is not hashable
         return tuple(sorted(requirement_or_candidate.extras))
 
-    def get_base_requirement(self, candidate):
+    def get_base_requirement(self, candidate: Candidate) -> Requirement:
         return Requirement(f"{candidate.name}=={candidate.version}")
 
     def get_preference(self, identifier, resolutions, candidates, information, **kwds):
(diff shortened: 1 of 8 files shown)


src/fromager/resolver.py
```diff
@@ -243,3 +244,64 @@ def is_satisfied_by(self, requirement: Requirement, candidate: Candidate) -> boo
...
+        repo = self.client.get_repo(f"{self.organization}/{self.repo}")
+
+        requirements = list(requirements[identifier])
+        bad_versions = {c.version for c in incompatibilities[identifier]}
+
+        # Need to pass the extras to the search, so they
+        # are added to the candidate at creation - we
+        # treat candidates as immutable once created.
+        candidates = []
+        for tag in repo.get_tags():
```
tiran, line 283: Not every project uses tags that conform to PEP 440. Also I was surprised that packaging's `Version` class can handle `Version("v0.1.0")`. I suggest to make tag name to version more flexible, e.g. with a user-supplied function `fn(tag: str) -> Version`?

src/fromager/resolver.py
```diff
@@ -243,3 +244,64 @@ def is_satisfied_by(self, requirement: Requirement, candidate: Candidate) -> boo
...
+class GitHubTagProvider(ExtrasProvider):
+    def __init__(
+        self,
+        organization,
+        repo,
+    ):
+        super().__init__()
+        self.organization = organization
+        self.repo = repo
+        self.client = github.Github()
```
tiran, line 258: GitHub has rate limiting for anon access based on IP addresses. Is there a way to pass credentials in case our infra runs into rate limiting?

tiran (commented):
I have added two comments. The rest looks good. Also I'm missing type annotation. We can add them later, though.

src/fromager/resolver.py
dhellmann: I've found tokens easy to use, so we could add that with a small environment variable check.

src/fromager/resolver.py
dhellmann: That would be a good enhancement. I'll file a ticket to follow up with that.

dhellmann: #117

src/fromager/resolver.py
dhellmann: #118

tests/test_resolver.py
```diff
@@ -65,3 +65,282 @@ def test_provider_choose_sdist():
...
+  "stargazers_count": 1,
+  "watchers_count": 1,
+  "language": "Python",
+  "has_issues": true,
+  "has_projects": true,
+  "has_downloads": true,
+  "has_wiki": false,
+  "has_pages": false,
+  "has_discussions": false,
+  "forks_count": 4,
```
Gregory-Pereira, line 155: This will fail with more forks yes?

dhellmann: This is a mock response. It's used to provide a fake API endpoint so that we always get consistent data, and so the tests don't have to talk to GitHub.com. Most of the response isn't actually used, but it's needed to make the GitHub client work properly so I included it all.

tiran (approved):

PR #106, opened 2024-06-27
sort dependencies before recursing for consistent output

dhellmann:
The dependencies are a set, and we might get them in different orders from
different invocations. Sorting the values makes the processing order
consistent and the build order output from run to run is therefore easier
to compare.


Gregory-Pereira (approved):
/lgtm

We discussed this [here](https://redhat-internal.slack.com/archives/C079FE5H94J/p1719524121874859)

PR #108, opened 2024-06-29
document netrc configuration for private repository access

### LINKED ISSUE #107: Allow for private index auth through Fromager
This behaviour around auth-ing to private indexes seems to fit better in the upstream code. I want the ability to authenticate against a variety of own private indexes and store credentials for them. In terms of implementations Im currently looking at a few options:

## 1. Keyring
[project link](https://pypi.org/project/keyring/).
The idea of the keyring implementation is that there would be an auth module with 3 commands,`add`, `list` and `remove` (with maybe `test` as the 4th, which would just see if you can list the packages at the root of the simple index with the provided creds). 

### Implementation

Starting with the adding functionality, there are 4 values that would be stored in the keystore. Firstly there would be an `fromager_indexes` indexes array which would be implemented as a string of comma separated values for the index names (ex: `fromager_indexes="pypi,gitlab-private-sdist"`). This would track all the indexes, and this variable would be updated as indexes enter or leave. This would be configured the first time running fromager, and if there is a collision (which there should not based on naming convention) it would ask you to overwrite that value in the keystore.

The next 4 values are index specific, and they are derived from `index_name` as input from user input: `f{index_name}_index_url`, `f{index_name}_index_user`, and `f{index_name}_index_pass`. The index_name would be tracked in the `fromager_indexes` discussed above.  In the same fashion when someone wants to add an index, it would check those 3 variables for collisions, and if none of them had one it would add all 3.

In the same way remove would first look through `fromager_indexes` to see if the index exists, and if so it would use that value to unset the other 3 values. List would just dump the output of the `fromager_indexes` formatted.

### Tradeoffs

The pros of this implementation style is that it would be very secure, leveraging the underlying OS built in secret storage functionality, but the downside is its implemented differently on each platform. Since were not doing anything particularly complicated I think this won't present issues but its possible ¯\_(ツ)_/¯.

## 2. ENV variables 

It is possible to save these values as env values in bash. This implementation would look very similar to the keyring implementation, where we save and update the 4 values (3 per index, 1 for all). I am not the biggest fan of this idea because lets say someone overwrites one of these variables or they manage their ENV poorly and all of a sudden it would not work. To me the keyring implementation to this is more robust, if a little more work.

We could also update `fromager` to call [env variables to pip itself](https://pip.pypa.io/en/stable/topics/configuration/#environment-variables). However I would assume that these private index auth params which get built into the index url itself in the conf file would be exposed in logs which is not ideal. 

## 3. conf files

Pip supports 3 layers of configuraiton files (`global`, `user`, and`site`) as well as specifying a configuration file that’s loaded last, and whose values override the values set in the aforementioned files via the env variable: `PIP_CONFIG_FILE` ([see docs on this](https://pip.pypa.io/en/stable/topics/configuration/#pip-config-file).

According to the docs "Each file read overrides any values read from previous files, so if the global timeout is specified in both the global file and the per-user file then the latter value will be used.", so I am not sure if we might run into collisions if something is set in the global like to use pypi and then we specify other private indexes in the conf file passed as `PIP_CONFIG_FILE` - would it allow or not allow using the pypi index?

The general concept of this implementation is that `fromager` manages its own configuration file in `work-dir` that then gets used in its calls by passing the `PIP_CONFIG_FILE` env.

## closing notes

Let me know if I am missing anything, and I would love to get your thoughts @tiran and @dhellmann

dhellmann:
Resolves #107

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,26 @@ environment and prepared source, compiling any extensions using the
 appropriate override environment settings (refer to
 [customization](docs/customization.md) for details about overrides).
 
+## Using private registries
+
+Fromager uses the [requests](https://requests.readthedocs.io) library and `pip`
+at different points for talking to package registries. Both support
+authenticating to remote servers in various ways. The simplest way to integrate
+the authentication with fromager is to have a
+[netrc](https://docs.python.org/3/library/netrc.html) file with a valid entry
+for the host. The file will be read from `~/.netrc` by default. Another location
+can be specified by setting the `NETRC` environment variable.
+
+For example, to use a gitlab package registry, use a [personal
+access
+token](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html#create-a-personal-access-token)
+as documented in [this
+issue](https://gitlab.com/gitlab-org/gitlab/-/issues/350582):
+
+```plaintext
+machine gitlab.com login oauth2 password $token
+```
+
 ## Additional docs
 
 * [Package build customization instructions](docs/customization.md)


README.md
```diff
@@ -113,6 +113,22 @@ environment and prepared source, compiling any extensions using the
 appropriate override environment settings (refer to
 [customization](docs/customization.md) for details about overrides).
 
+## Using private registries
+
+Fromager uses the [requests](https://requests.readthedocs.io) library
+and `pip` at different points for talking to package registries. Both
+support authenticating to remote servers in various ways. The simplest
+way to integrate the authentication with fromager is to have a
+`~/.netrc` file with a valid entry for the host. For example, to use a
```
tiran, line 122: pip and requests also support `NETRC` env var. Should we document this, too?

dhellmann: Makes sense, I'll look at that.

@Gregory-Pereira if we can set a variable, maybe we don't need to move files around in the build job?

dhellmann: Updated

tiran (approved):

dhellmann:
@mergifyio refresh

PR #111, opened 2024-07-01
Configure mypy checker in tox and GHA

tiran:
- add `mypy` checker to tox
- configure mypy to ignore packages without type annotations and type stubs
- add mypy checker to GHA (ignoring errors)


dhellmann (approved):
I notice the new job isn't required in the mergify settings. Let's do that separately, to give the rest of us a chance to learn how to make mypy happy.

PR #112, opened 2024-07-01
Define custom click params for path and versions

tiran:
The new parameter type `ClickPath` is a subclass of `click.Path` that returns a pathlib.Path instead of a path-like object. All path paramemter now use `ClickPath` for handling input.

`PackageVersion` is a string input field that converts a version string to a packaging `Version` class with error checking.


dhellmann (approved):
Nice improvements, thanks!

PR #113, opened 2024-07-01
Convert Path to str for external_commands.run()

tiran:
`external_commands.run()` accepts a `list[str]`. Convert all occurances of `pathlib.Path` to `str`.


dhellmann (approved):
I'm surprised that's not a method of Path.

PR #116, opened 2024-07-01
look for patches in subdirectories

### LINKED ISSUE #114: improve patch management
We have encountered a case where the same wheel is being built from different source archives for different variants. Those source archives use a different path name as the source root when unpacked, but they have a common prefix (one is `package-version-variant` the other is `package-version`). That means that any patches for `package-version` will be applied to `package-version-variant`, even if they should not be.

We can at the very least move patches into subdirectories based on the source directory name. That will not be a backwards-compatible change.

We could also introduce an optional variant part of the path (`patches/variant/source-root`). It's not clear that's absolutely needed (it doesn't do anything for the case at hand, and we haven't encountered other cases where it's needed).

dhellmann:
Add support for organizing patches in subdirectories
instead of just prefixed with the source directory
name to avoid collisions when we have different source
directory names for different variants of the same
package.

Fixes #114

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -36,7 +36,7 @@ organized with a subdirectory per variant.
 Environment files are named using the [canonical distribution
 name](#canonical-distribution-names) and the suffix `.env`.
 
-```
+```console
 $ tree overrides/envs/
 overrides/envs/
 ├── cpu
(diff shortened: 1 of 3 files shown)


docs/customization.md
```diff
@@ -36,7 +36,7 @@ organized with a subdirectory per variant.
 Environment files are named using the [canonical distribution
 name](#canonical-distribution-names) and the suffix `.env`.
 
-```
+```console
```
Gregory-Pereira, line 39: are these meant to be bash? they include both bash commands and log.

Gregory-Pereira (approved):
/lgtm just one question

docs/customization.md
dhellmann: I thought console was for that case? I honestly don't know, but this stopped the linter from complaining.

tiran: +1 `console` is an alias for `ShellSession`, which is designed to high-light interactive session in a Unix shell.

tiran (approved):

docs/customization.md
Gregory-Pereira: Wonderful, thank you both!

PR #119, opened 2024-07-01
expand test coverage for fromager.dependencies

dhellmann:
Add some unit tests that pull the dependencies of fromager itself.

### DIFF
--- a/tests/test_dependencies.py
+++ b/tests/test_dependencies.py
@@ -1,6 +1,15 @@
-import pytest
+import functools
+import itertools
+import pathlib
+import shutil
+import typing
 
-from fromager import dependencies
+import pytest
+from packaging.requirements import Requirement
+
+from fromager import context, dependencies
+
+_fromager_root = pathlib.Path(__file__).parent.parent
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


tests/test_dependencies.py
```diff
@@ -32,7 +37,57 @@
...
+    build_system: dict[str, list[str]] | dict[str, str | list[str] | None],
+    expected_results: dict[str, typing.Any] | dict[str, str | list[str] | None],
+):
     pyproject_toml = {"build-system": build_system}
     actual = dependencies.get_build_backend(pyproject_toml)
     assert expected_results == actual
+
+
+def test_get_build_system_dependencies(tmp_context: context.WorkContext):
+    fromager_root = pathlib.Path(os.getcwd())
```
tiran, line 50: This assumes that the tests are always executed from the base directory. I prefer to use `__file__` instead of CWD as reference point. First parent is the `tests` directory, grandparent is the root directory. You could make this a module global, too.
```suggestion
    fromager_root = pathlib.Path(__file__).parent.parent
```

tiran (changes requested):

tests/test_dependencies.py
dhellmann: Done

dhellmann:
This version is leaving debris in the root dir now when tests run, let me resolve that before we merge it.

tiran (approved):

PR #120, opened 2024-07-02
Fix canonical distribution names

marco-pag:
Fix the canonical distribution names section in the docs and update the canonicalize command.

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -296,9 +296,8 @@ distribution, wheel, and python package names are represented in
 filenames and requirements specifications. To standardize and ensure
 that build customizations are recognized correctly, we always use a
 canonical version of the name, computed using
-`packaging.utils.canonicalize_name()` and then replacing hyphens (`-`)
-with underscores (`_`). For convenience, the `canonicalize`
-command will print the correct form of a name.
+`packaging.utils.canonicalize_name()`. For convenience, the
+`canonicalize` command will print the correct form of a name.
 
 ```
 $ tox -e cli -- canonicalize flit-core
--- a/src/fromager/overrides.py
+++ b/src/fromager/overrides.py
@@ -77,10 +77,7 @@ def extra_environ_for_pkg(
 
 
 def pkgname_to_override_module(pkgname: str) -> str:
-    canonical_name = canonicalize_name(pkgname)
-    module_name = canonical_name.replace("-", "_")
-    return module_name
-
+    return canonicalize_name(pkgname)
 
 def find_override_method(distname: str, method: str) -> typing.Callable:
     """Given a distname and method name, look for an override implementation of the method.


dhellmann:
I forget why it was important to replace `-` with `_`, so let me dig a bit to find it before we merge this change.

dhellmann:
I think it had to do with the fact that the sdist filename took the value from the packaging settings without modifying it at all, so it could come out different at different times, and be different from the wheel itself which _does_ force canonicalization. That means it's impossible to ensure we get the same name all the time unless we make the extra transformation. Since fromager relies so much on looking up settings and plugins by using the package name, we need some way to force it into a  consistent format.

tiran:
https://packaging.python.org/en/latest/specifications/name-normalization/ defines the normalization rules. 

> The name should be lowercased with all runs of the characters ., -, or _ replaced with a single - character. This can be implemented in Python with the re module:

dhellmann:
> https://packaging.python.org/en/latest/specifications/name-normalization/ defines the normalization rules.
> 
> > The name should be lowercased with all runs of the characters ., -, or _ replaced with a single - character.

OK. That's not what I was seeing, I think. I'll revisit this and see if we can use `-` instead of `_`.

tiran:
PyPI isn't consistent. There is standard on one hand. And there is legacy support on the other hand.

PyPI are a ton of legacy packages that have a `.` in the package name and sdist name. For example most of the Zope 3 packages like https://pypi.org/project/zope.interface use legacy names. https://pypi.org/project/zope-interface (dash) redirects to  https://pypi.org/project/zope.interface (dot).

dhellmann:
> PyPI isn't consistent. There is standard on one hand. And there is legacy support on the other hand.
> 
> For example PyPI are a ton of legacy packages that have a `.` in the package name and sdist name. For example most of the Zope 3 packages like https://pypi.org/project/zope.interface use legacy names. https://pypi.org/project/zope-interface (dash) redirects to https://pypi.org/project/zope.interface (dot).

Right. I had to pick something. It looks like I picked the wrong thing. I probably picked that based on a bad example, but I don't remember any more.

In any case, any change would be backwards-incompatible, so I want to take good care to understand how much will break in the downstream builds before we make it.

dhellmann:
> > PyPI isn't consistent. There is standard on one hand. And there is legacy support on the other hand.
> > For example PyPI are a ton of legacy packages that have a `.` in the package name and sdist name. For example most of the Zope 3 packages like https://pypi.org/project/zope.interface use legacy names. https://pypi.org/project/zope-interface (dash) redirects to https://pypi.org/project/zope.interface (dot).
> 
> Right. I had to pick something. It looks like I picked the wrong thing. I probably picked that based on a bad example, but I don't remember any more.
> 
> In any case, any change would be backwards-incompatible, so I want to take good care to understand how much will break in the downstream builds before we make it.

A conversation earlier today made me realize that the reason for replacing `-` with `_` has to do with needing to be able to use those names for python package names for the override plugin entry point registration, and wanting to use the same naming everywhere so users don't have to know when to use different types of canonicalization.

dhellmann:
I don't think we will want this, unfortunately. It's going to be hard to make it backwards compatible.

PR #124, opened 2024-07-02
fix: pass environment variables from caller to build steps

### LINKED ISSUE #115: pass environment variables from caller to build steps
Is it possible to have fromager pass the value of MAX_JOBS into build processes?  This would be great for systems with a lot of cpus and memory.

shubhbapna:
fixes #115 

Instead of initializing `extra_environ` as an empty dictionary we can initialize it with whatever environment variables the current process has access to. This keeps the flexibility of overriding these variables later on using the .env per package files or plugins.

### DIFF
--- a/src/fromager/overrides.py
+++ b/src/fromager/overrides.py
@@ -1,6 +1,5 @@
 import itertools
 import logging
-import os
 import pathlib
 import typing
 
@@ -72,7 +71,7 @@ def extra_environ_for_pkg(
 
     By default all packages have access to the environment variables in the current process
     """
-    extra_environ = dict(os.environ)
+    extra_environ = {}
 
     pkgname = pkgname_to_override_module(pkgname)
     variant_dir = envs_dir / variant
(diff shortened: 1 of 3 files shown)


src/fromager/overrides.py
```diff
@@ -69,8 +70,9 @@ def extra_environ_for_pkg(
     Extra environment variables are stored in per-package .env files in the
     envs package, with a key=value per line.
 
+    By default all packages have access to the environment variables in the current process
     """
-    extra_environ = {}
+    extra_environ = dict(os.environ)
```
dhellmann, line 75: This looks like a reasonable approach, but it makes the name of this function seem wrong. The return value is no longer "extra" values, it's all of the values.

A similar change at https://github.com/python-wheel-build/fromager/blob/main/src/fromager/wheels.py#L114 would keep the intent of this function intact. The trade-off of that is that any plugin overriding the `build_wheel` function would have to do this step, too. Maybe that's reasonable, because perhaps one reason to override is to _not_ share some variable?

What do you think?

shubhbapna: That makes sense

dhellmann (approved):
Nice, thank you!

PR #130, opened 2024-07-03
trap fewer exceptions when resolving a dependency

dhellmann:
download_source() tries to resolve the source using multiple
servers. As part of the trial loop, it captures exceptions, logs them,
and continues with the next server. When there is a coding error in a
plugin, the capture behavior can mask the error. Instead of trapping
all exceptions, only catch the ones coming from resolvelib that are
also caught, and re-raised, in resolve_dist().


Gregory-Pereira (approved):
/lgtm

PR #133, opened 2024-07-04
Add an option to pass the number of max jobs that are available

### LINKED ISSUE #127: support parallel build controls
Some packages with compiled extensions support setting a variable to control how many compilers are run in parallel. There is no single standard, though, so it is difficult to control the builds. We should provide a way to control the value to use so that someone running builds can get a reasonable value from fromager instead of hard-coding it.

One way to do that is with a command line argument that sets a value in the work context that a plugin can read.

Another way to do it is to auto-detect some of the builders and set the variables for the user. That's much more error-prone.

Another way to do it is to let the user specify the variable name, and automatically set it based on the value.

shubhbapna:
fixes #127 

Implemented the first method where a max number of jobs is passed to fromager via CLI and is exposed to the plugins via context

### DIFF
--- a/src/fromager/__main__.py
+++ b/src/fromager/__main__.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import logging
+import os
 import pathlib
 
 import click
@@ -79,6 +80,12 @@ VERBOSE_LOG_FMT = "%(levelname)s:%(name)s:%(lineno)d: %(message)s"
     help="control removal of working files when a build completes successfully",
 )
 @click.option("--variant", default="cpu", help="the build variant name")
+@click.option(
+    "-j",
+    "--jobs",
+    type=int,
+    help="number of jobs available to run in parallel",
+)
 @click.pass_context
 def main(
     ctx,
@@ -93,6 +100,7 @@ def main(
     wheel_server_url: str,
     cleanup: bool,
     variant: str,
+    jobs: int,
 ):
     # Configure console and log output.
     stream_handler = logging.StreamHandler()
@@ -123,6 +131,7 @@ def main(
         wheel_server_url=wheel_server_url,
         cleanup=cleanup,
         variant=variant,
+        jobs=jobs if jobs is None or jobs > 0 else os.cpu_count(),
     )
     wkctx.setup()
     ctx.obj = wkctx
--- a/src/fromager/wheels.py
+++ b/src/fromager/wheels.py
@@ -84,6 +84,14 @@ def build_wheel(
     # TODO: refactor?
     # Build Rust without network access
     extra_environ["CARGO_NET_OFFLINE"] = "true"
+    # configure max jobs settings. should cover most of the cases, if not then the user can use ctx.jobs in their plugin
+    if ctx.jobs:
+        extra_environ["MAKEFLAGS"] = (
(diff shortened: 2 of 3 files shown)


src/fromager/__main__.py
```diff
@@ -79,6 +79,12 @@
     help="control removal of working files when a build completes successfully",
 )
 @click.option("--variant", default="cpu", help="the build variant name")
+@click.option(
+    "-n",
+    "--number-of-jobs-available",
```
tiran, line 84: I recommend a shorter name. `-j`, `--jobs`, and `--parallel`  are common options used by `make` and `cmake`. A long name also makes it harder to set the option with an env var. `FROMANGER_NUMBER_OF_JOBS_AVAILABLE=8` is a mouthful. :)

```suggestion
    "-j",
    "--jobs",
```

src/fromager/__main__.py
```diff
@@ -79,6 +79,12 @@
...
+    default=1,
```
tiran, line 85: I would prefer to make the default `None`, so it is easier to distinguish "user wants one job" from "user wants default". IIRC `cmake` and `ninja` automatically use all cores unless told otherwise.

```suggestion
    default=1,
```

tiran (commented):
Good job!

I suggest a shorter name, so we can set the option with `FROMAGER_JOBS=$(nproc)`.

If you add this snippet to the right place, then you have configured most build systems:

```
if self.jobs is not None:
    extra_environ["MAKEFLAGS"] = "-j{self.jobs}"
    extra_environ["CMAKE_BUILD_PARALLEL_LEVEL"] = "{self.jobs}"
```

src/fromager/wheels.py
```diff
@@ -84,6 +84,14 @@ def build_wheel(
     # TODO: refactor?
     # Build Rust without network access
     extra_environ["CARGO_NET_OFFLINE"] = "true"
+    # configure max jobs settings. should cover most of the cases, if not then the user can use ctx.jobs in their plugin
+    if ctx.jobs:
```
dhellmann, line 88: Do we need a check for `> 0` or do we assume if they give a negative number something they're building knows what to do with it?

tiran: Idea: check for `ctx.jobs is not None` and then map `ctx.jobs == 0` to `os.cpu_count()`.

src/fromager/__main__.py
```diff
@@ -79,6 +80,11 @@
     help="control removal of working files when a build completes successfully",
 )
 @click.option("--variant", default="cpu", help="the build variant name")
+@click.option(
+    "-j",
+    "--jobs",
+    help="number of jobs available to run in parallel",
+)
```
dhellmann, line 87: I think we need a `type` argument here to force the value to be converted to an integer.

```
$ tox -e cli -- -j nine bootstrap fromager
...
TypeError: '>' not supported between instances of 'str' and 'int'
```

dhellmann:
This looks good now. Could you squash everything into 1 commit for the new feature, please?

shubhbapna:
Rebased into 1 feature commit @dhellmann

dhellmann (approved):
Looks good!

dhellmann:
@mergifyio refresh

PR #134, opened 2024-07-05
Implement parameter expansion for env files

tiran:
Env files can now use simple parameter expansions like `$NAME` and `${NAME}`. The values are taken from previous lines, then proccess environment.

Subshell expressions `$()` and parameter expansions with defaults `${NAME:-default}` are not implemented. We can add them later if we need them.

### DIFF
--- a/src/fromager/overrides.py
+++ b/src/fromager/overrides.py
@@ -1,6 +1,8 @@
 import itertools
 import logging
+import os
 import pathlib
+import string
 import typing
 
 from packaging.utils import canonicalize_name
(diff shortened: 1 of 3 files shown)


dhellmann:
I'm a little torn about this. I could see it becoming a slippery slope. On the other hand, this minimal templating support is probably useful. I would definitely want to draw the line here, and not take anything more complicated (no conditionals, for example). If we agree on that, I think we can go ahead with this. Let me know what you think.

tiran:
No worries! I have no intention or motivation to add complex expressions like conditionals. The simply name lookups cover most use cases. For anything more complex, users have to write a plugin in Python.

src/fromager/overrides.py
```diff
@@ -82,7 +89,14 @@ def extra_environ_for_pkg(
         with open(env_file, "r") as f:
             for line in f:
                 key, _, value = line.strip().partition("=")
-                extra_environ[key.strip()] = value.strip()
+                key = key.strip()
+                value = value.strip()
+                if "$(" in value:
```
shubhbapna, line 94: Should we worry about the case where `"$("` might be escaped out and just part of the value as a string literal?

dhellmann: That's a good question.

I wonder how often that is likely to come up in practice? How hard would it be to detect that case?

dhellmann: Let's take this as it is, and consider the improvement as another PR. I'll open a ticket for it.

dhellmann (approved):

dhellmann:
@mergifyio refresh

Issue #129, opened 2024-07-03
document the override for creating a source distribution


dhellmann:
There are no docs for the override used in https://github.com/python-wheel-build/fromager/blob/main/src/fromager/sources.py#L269

dhellmann:
This was done in #141

PR #141, opened 2024-07-05
document override functions related to building sdist

dhellmann:
We were missing docs for both get_build_sdist_dependencies() and build_sdist().

Fixes document the override for creating a source distribution #129


rd4398:
I don't know how important it is but "The `build_sdist()` function is responsible forr creating a new source" there is an extra r in forr. (line  #340) Rest looks good to me +1:

dhellmann:
> I don't know how important it is but "The `build_sdist()` function is responsible forr creating a new source" there is an extra r in forr. (line #340) Rest looks good to me +1:

Good catch, thanks!

shubhbapna (approved):

PR #142, opened 2024-07-05
add list-overrides command

### LINKED ISSUE #135: add a command to list all packages with overrides
When setting up a build environment with overrides, it will be useful to be able to establish tests for each customized build. To do that, it would be useful to be able to know exactly which packages have overrides, since that would let someone create a "linter" for their CI config to ensure that as a new override is added, a new test job is also added.

We should create a command to list the available overrides, including all of the plugins that can be loaded, any packages that have patches, and any packages that have env vars. The list should not include packages listed as pre-built.

The output should be just the list of packages, without versions, one name per line written to the console. We could optionally include versions, but since plugins don't have versions that may be less useful, so let's consider that as an enhancement for later.

The function used to generate the list of names should be a stable API so that a linter can import it and invoke it to get an iterable of the same names that are printed by the new command.

dhellmann:
Provide a command to list all of the known
overrides of any sort in the current
configuration.

Fixes #135

### DIFF
--- a/src/fromager/commands/__init__.py
+++ b/src/fromager/commands/__init__.py
@@ -1,10 +1,19 @@
-from . import bootstrap, build, build_order, canonicalize, download_sequence, step
+from . import (
+    bootstrap,
+    build,
+    build_order,
+    canonicalize,
+    download_sequence,
+    list_overrides,
+    step,
+)
 
 commands = [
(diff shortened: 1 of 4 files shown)


dhellmann:
The way the patch filenames are being parsed is not correct.

Issue #102, opened 2024-06-24
accept constraints list as input


dhellmann:
There may be situations where fromager should constrain the versions of packages it builds beyond what the original requirements specifications say. For example, if someone wants to distribute a set of packages and exclude one version of one package with a known issue, it would be useful to be able to express that somehow to the resolver.

If we change a requirement specification based on a constraint, we should log that at INFO level.

We should think about whether we need to enforce a constraint fitting within the requested range. For example, if someone asks for `torch>=2.3.0,<2.4.0`, should it be valid to have a constraint `torch==2.5.0`? We could build that, but then it wouldn't be usable for the installation.

We may also need to support markers in constraints so that we can have different constraints for different platforms or python versions. We have [a function](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/dependencies.py#L217) to evaluate a marker expression.

Another case to consider is whether a constraint adds a cap or exclusion that is not present in the original requirement, but does not actually specify a particular version to use. If a requirement is `torch>=2.3.0` and the constraint adds `torch!=2.3.1,<2.4.0`, we could merge those settings into the existing requirement. We might consider that as a separate ticket, though, since the logic might be complex. Perhaps pip or resolvelib has something we can reuse, though?

Tasks:

- [x] Implement a simple constraints lists that only accepts `==` #144 
- [x] Add support for `<` #166

shubhbapna:
Should we add an `version_constraint` field in settings? So maybe something like:

```yaml
version_constraint:
  cuda:
     - torch==2.3.0
     - numpy<2.0.0
```

And then in the resolvers we can probably do the constraint check here:
- https://github.com/python-wheel-build/fromager/blob/main/src/fromager/resolver.py#L228
- https://github.com/python-wheel-build/fromager/blob/main/src/fromager/resolver.py#L297

dhellmann:
It would probably be most convenient to pass a constraints.txt file, instead of adding that info to the settings file. Lots of tools in the python ecosystem can produce constraints files.

I think we only need constraints for bootstrapping, since everything else already operates on an exact version of a package.

Fromager only works on one variant at a time, so I think we'd only need one constraints file at a time (the caller might have different constraints for different variants, but they can keep them in separate files).

dhellmann:
As far as what constraints we allow, we could start simple and just support exact versions using `==`. Later we could make it more complex, supporting ranges or exclusions. That would require modifying the resolver code, though, and I'm wary of making changes there right now.

[...]

PR #144, opened 2024-07-05
Add implementation for a basic constraints file

shubhbapna:
Part of #102 

Implements a simple constraints list where user is only allowed to use "==" constraints. Ensures that the given constraint version is not in conflict with the required version.

### DIFF
--- a/src/fromager/constraints.py
+++ b/src/fromager/constraints.py
@@ -21,18 +21,17 @@ class Constraints:
 
         # only allow one constraint per package
         if len(constraint.specifier) != 1:
-            logger.debug(
+            raise ValueError(
                 f"{constraint} is not allowed. Only one constraint per package is allowed"
             )
-            return new_req, None
 
         for spec in constraint.specifier:
             # only allow "=="
             if spec.operator != "==":
-                logger.debug(
+                raise ValueError(
                     f"{constraint} is not allowed. Only '<package>==<version>' constraints are allowed"
                 )
-                return new_req, None
+
             # ensure that constraint and req don't conflict
             if spec.version not in req.specifier:
                 raise ValueError(
@@ -56,10 +55,9 @@ def load(filename: pathlib.Path | None) -> Constraints:
         return Constraints({})
     filepath = pathlib.Path(filename)
     if not filepath.exists():
-        logger.debug(
-            "constraints file %s does not exist, ignoring", filepath.absolute()
+        raise FileNotFoundError(
+            f"constraints file {filepath.absolute()} does not exist, ignoring"
         )
-        return Constraints({})
     with open(filepath, "r") as f:
         logger.info("loading constraints from %s", filepath.absolute())
         return _parse(f)
--- a/src/fromager/constraints.py
+++ b/src/fromager/constraints.py
@@ -12,7 +12,7 @@ class Constraints:
     def __init__(self, data: dict[str, Requirement]):
         self._data = data
 
-    def get_new_requirement(self, req: Requirement):
+    def get_constrained_requirement(self, req: Requirement):
         constraint = self._data.get(req.name)
         new_req = copy.deepcopy(req)
 
--- a/src/fromager/constraints.py
+++ b/src/fromager/constraints.py
@@ -1,0 +1,54 @@
+import copy
+import logging
+import pathlib
+from io import TextIOWrapper
+
+from packaging.requirements import Requirement
+
+logger = logging.getLogger(__name__)
+
+
+class Constraints:
+    def __init__(self, data: dict[str, Requirement]):
+        self._data = data
+
+    def get_new_requirement(self, req: Requirement):
+        constraint = self._data.get(req.name)
+        new_req = copy.deepcopy(req)
+
+        if not constraint:
+            return new_req, None
+
+        # only allow "==" in constraints for now
+        if len(constraint.specifier) != 1 or not str(constraint.specifier).startswith(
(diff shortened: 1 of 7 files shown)


src/fromager/constraints.py
```diff
@@ -0,0 +1,54 @@
...
+        new_req = copy.deepcopy(req)
+
+        if not constraint:
+            return new_req, None
+
+        # only allow "==" in constraints for now
+        if len(constraint.specifier) != 1 or not str(constraint.specifier).startswith(
+            "=="
+        ):
+            logger.debug(
```
dhellmann, line 24: This should probably raise an exception instead of just logging. The exception will make it clear that the user is doing something unsupported, instead of hiding that and continuing the build operations.

src/fromager/constraints.py
```diff
@@ -0,0 +1,54 @@
...
+                f"{constraint} is not allowed. Only '<package>==<version>' constraints are allowed"
+            )
+            return new_req, None
+
+        new_req.specifier = constraint.specifier
+        return new_req, constraint
+
+
+def _parse(content: TextIOWrapper) -> Constraints:
```
dhellmann, line 46: There's some logic in https://github.com/python-wheel-build/fromager/blob/main/src/fromager/commands/bootstrap.py#L13 that handles comments and blank lines. We can go ahead with the version you have here, but it would be good to refactor both functions to share a common base loop for dealing with lines within the file.

src/fromager/constraints.py
```diff
@@ -0,0 +1,54 @@
...
+        constraints[req.name] = req
+    return Constraints(constraints)
+
+
+def load(filename: pathlib.Path | None) -> Constraints:
+    if not filename:
+        return Constraints({})
+    filepath = pathlib.Path(filename)
+    if not filepath.exists():
+        logger.debug(
```
dhellmann, line 59: A missing file is almost certainly a user error that should not be ignored.

src/fromager/sdist.py
```diff
@@ -75,95 +75,105 @@ def handle_requirement(
         f'{req.name}: {"*" * (len(why) + 1)} handling {req_type} requirement {req} {why}'
     )
 
+    new_req, constraint = ctx.constraints.get_new_requirement(req)
```
dhellmann, line 78:
```suggestion
    constrained_req, constraint = ctx.constraints.get_new_requirement(req)
```

`new_req` isn't particularly descriptive. Maybe `constrained_req`? Alternatively, you could save `req` as `original_req` and then replace its value with the constrained value. That has the appeal of changing fewer lines, and for cases where there is no constraint it's the same value anyway.

Same comment about the method name, "new" doesn't tell us how it's different. Something like `constrain_requirement(req)` or `get_constrained_requirement(req)` explains what sort of change is going to happen.

src/fromager/constraints.py
shubhbapna: I copied this part from `settings.py`. Will make a note to change it there as well in a separate patch

src/fromager/constraints.py
shubhbapna: Okay will do. I will make a separate patch for this

src/fromager/sdist.py
shubhbapna: This is so much better, yes!

src/fromager/constraints.py
dhellmann: Ah, in the case of the settings file there's a default and providing the file is optional. In this case, the user will be explicitly giving us a file and there is no default, so if they point us to something that doesn't exist it's more clearly a mistake.

src/fromager/sdist.py
```diff
@@ -60,21 +60,27 @@ def __init__(
...
         return ""
     logger.info(
-        f'{req.name}: {"*" * (len(why) + 1)} handling {req_type} requirement {req} {why}'
+        f'{original_req.name}: {"*" * (len(why) + 1)} handling {req_type} requirement {original_req} {why}'
     )
 
+    req, constraint = ctx.constraints.get_constrained_requirement(original_req)
+    if constraint:
+        logger.info(
+            f"incoming requirement {req} matches constraint {constraint}, the new requirement is {req}"
```
dhellmann, line 81:
```suggestion
            f"{req.name}: incoming requirement {original_req} matches constraint {constraint}, the new requirement is {req}"
```

Prefixing the log message with the requirement name helps keep the console output clear.  I think the other change, to include original_req, was missed in the naming changes.

src/fromager/sdist.py
```diff
@@ -233,7 +239,7 @@ def handle_requirement(
         if build_env:
             logger.debug(f"{req.name}: cleaning up build environment {build_env.path}")
             shutil.rmtree(build_env.path)
-            logger.debug("{req.name}: cleaned up build environment {build_env.path}")
+            logger.debug(f"{req.name}: cleaned up build environment {build_env.path}")
```
dhellmann, line 242: Nice clean-up, thank you for catching that.

dhellmann (approved):
Nice work. I spotted a couple of small issues that we should fix in a follow-up.

dhellmann:
@mergifyio refresh

src/fromager/sdist.py
dhellmann: #145

src/fromager/constraints.py
shubhbapna: Oh got it

PR #147, opened 2024-07-06
fix: only download sdist sources in download-sequence

dhellmann:
If the package has an override configured for download_source(), do
not try to download the sdist.


Gregory-Pereira (approved):
/lgtm

dhellmann:
@mergifyio refresh

PR #148, opened 2024-07-06
feature: post-build hooks

dhellmann:
Add a new extension point for plugging in after a
wheel is built. The hook should have enough
information to, for example, publish the newly
build sdist and wheel to a package index.


Gregory-Pereira (approved):
/lgtm

PR #154, opened 2024-07-08
Added checks for downloaded source and wheel file

rd4398:
Added the checks for downloaded source and wheel files by validating file extensions and matching requirement with downloaded tar

### DIFF
--- a/src/fromager/sdist.py
+++ b/src/fromager/sdist.py
@@ -8,8 +8,8 @@ import shutil
 import subprocess
 import sys
 import typing
+import zipfile
 from urllib.parse import urlparse
-from zipfile import ZipFile
 
 from packaging.requirements import Requirement
 
@@ -95,7 +95,7 @@ def handle_requirement(
         )
         if not wheel_filename.exists():
             logger.info(f"{req.name}: downloading pre-built wheel {wheel_url}")
-            wheel_filename = download_wheel_check(ctx.wheels_prebuilt, wheel_url)
+            wheel_filename = _download_wheel_check(ctx.wheels_prebuilt, wheel_url)
         else:
             logger.info(f"{req.name}: have pre-built wheel {wheel_filename}")
         # Add the wheel to the mirror so it is available to anything
@@ -233,15 +233,17 @@ def handle_requirement(
 
     return resolved_version
 
+
 # Helper method to check whether the .whl file is a zip file and has contents in it.
 # It will throw BadZipFile exception if any other file is encountered. Eg: index.html
-def download_wheel_check(wheels_prebuilt, wheel_url):
-    wheel_filename = sources.download_url(wheels_prebuilt, wheel_url)
-    wheel_directory_contents = ZipFile(wheel_filename).namelist()
-    if len(wheel_directory_contents) == 0:
-        raise TypeError("Empty zip file encountered")
+def _download_wheel_check(destination_dir, wheel_url):
+    wheel_filename = sources.download_url(destination_dir, wheel_url)
+    wheel_directory_contents = zipfile.ZipFile(wheel_filename).namelist()
+    if not wheel_directory_contents:
+        raise zipfile.BadZipFile("Empty zip file encountered")
 
     return wheel_filename
+
 
 def _sort_requirements(
     requirements: typing.Iterable[Requirement],
--- a/src/fromager/sdist.py
+++ b/src/fromager/sdist.py
@@ -9,6 +9,7 @@ import subprocess
 import sys
 import typing
 from urllib.parse import urlparse
+from zipfile import ZipFile
 
 from packaging.requirements import Requirement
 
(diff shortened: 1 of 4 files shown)


src/fromager/sdist.py
```diff
@@ -232,6 +233,15 @@ def handle_requirement(
 
     return resolved_version
 
+# Helper method to check whether the .whl file is a zip file and has contents in it.
+# It will throw BadZipFile exception if any other file is encountered. Eg: index.html
+def download_wheel_check(wheels_prebuilt, wheel_url):
```
dhellmann, line 238: Without the `ctx` prefix in the caller I don't know what `wheels_prebuilt` means. That's really the directory where the file is supposed to be downloaded, right? What is the similar argument called in `download_url()`?

src/fromager/sources.py
```diff
@@ -129,12 +129,29 @@ def default_download_source(
...
-    source_filename = download_url(ctx.sdists_downloads, url)
+    source_filename = download_source_check(ctx.sdists_downloads, url)
     logger.debug(
         f"{req.name}: have source for {req} version {version} in {source_filename}"
     )
     return (source_filename, version, url)
 
+# Helper method to check whether .zip /.tar / .tgz is able to extract and check its content.
+# It will throw exception if any other file is encountered. Eg: index.html
+def download_source_check(sdists_downloads, url):
```
dhellmann, line 140: Same comment on the name of the first argument here.

src/fromager/sdist.py
```diff
@@ -232,6 +233,15 @@ def handle_requirement(
...
+    wheel_filename = sources.download_url(wheels_prebuilt, wheel_url)
+    wheel_directory_contents = ZipFile(wheel_filename).namelist()
+    if len(wheel_directory_contents) == 0:
```
dhellmann, line 241:
```suggestion
    if not wheel_directory_contents:
```

src/fromager/sources.py
```diff
@@ -129,12 +129,29 @@ def default_download_source(
...
+    source_filename = download_url(sdists_downloads, url)
+    if source_filename.suffix == ".zip":
+        source_file_contents = ZipFile(source_filename).namelist()
+        if len(source_file_contents) == 0:
```
dhellmann, line 144:
```suggestion
        if not source_file_contents:
```

src/fromager/sources.py
```diff
@@ -129,12 +129,29 @@ def default_download_source(
...
+            raise TypeError("Bad Zip File encountered")
+    elif source_filename.suffix == ".tgz" or source_filename.suffix == ".gz":
+        tar = tarfile.open(source_filename)
+        contents = tar.getnames()
+        tar.close()
+        if len(contents) == 0:
```
dhellmann, line 150:
```suggestion
        if not contents:
```

src/fromager/sources.py
```diff
@@ -129,12 +129,29 @@ def default_download_source(
...
+# Helper method to check whether .zip /.tar / .tgz is able to extract and check its content.
+# It will throw exception if any other file is encountered. Eg: index.html
+def download_source_check(sdists_downloads, url):
+    source_filename = download_url(sdists_downloads, url)
+    if source_filename.suffix == ".zip":
+        source_file_contents = ZipFile(source_filename).namelist()
+        if len(source_file_contents) == 0:
+            raise TypeError("Bad Zip File encountered")
+    elif source_filename.suffix == ".tgz" or source_filename.suffix == ".gz":
+        tar = tarfile.open(source_filename)
```
dhellmann, line 147: If you use `with` you don't have to close the file explicitly.

```python
with tarfile.open(source_filename) as tar:
  contents = tar.getnames()
```

tests/test_sdist.py
```diff
@@ -30,3 +31,13 @@ def test_missing_dependency_format(resolve_dist, tmp_context):
     assert "setuptools>=40.8.0 -> 69.5.1" in s
     # Ensure we report what version we expect of all of the other dependencies
     assert "flit_core -> 3.9.0" in s
+
+@patch("fromager.sources.download_url")
+@patch("zipfile.ZipFile")
+def test_invalid_wheel_file_exception(mock_zipfile, mock_download_url):
+    mock_download_url.return_value = "fake_wheel.csv"
```
dhellmann, line 38: In order for the new check function to work, this file has to actually exist somewhere on the filesystem.

tests/test_sdist.py
```diff
@@ -30,3 +31,13 @@ def test_missing_dependency_format(resolve_dist, tmp_context):
...
+    mock_zip_instance = MagicMock()
+    mock_zip_instance.namelist.return_value = ["file1.txt", "file2.txt"]
```
dhellmann, line 40: Mocking out the `namelist()` method like this bypasses the checks in your test function. A more robust test would be to create the file on the filesystem as something other than a zip file, then let ZipFile try to read that file.

tests/test_sdist.py
```diff
@@ -30,3 +31,13 @@ def test_missing_dependency_format(resolve_dist, tmp_context):
...
+    mock_zipfile.return_value = mock_zip_instance
+    sdist.download_wheel_check(["prebuilt1", "prebuilt2"], "http://example.com/fake_wheel.csv")
+    assertRaises(exceptions.TypeError, sdist.download_wheel_check)
```
dhellmann, line 43: I think you want to be using `with pytest.raises()` here. There's an example in https://github.com/python-wheel-build/fromager/blob/main/tests/test_constraints.py#L25

dhellmann (changes requested):

src/fromager/sdist.py
rd4398: If I try to do `ctx.wheels_prebuilt`, it gives me syntax error for arguments. Should I pass `ctx` as parameter? or create a new `ctx` in helper function?

dhellmann: Sorry, I wasn't clear. The variable name doesn't make sense. Can we rename it to `destination_dir` like the argument to `download_url()`?

rd4398: oh, okay! sure I will do that!

PR #158, opened 2024-07-08
Refactor requirement file parsing

shubhbapna:
From: https://github.com/python-wheel-build/fromager/pull/144#discussion_r1667184891

Refactoring out the requirement file parsing done in bootstrap.py so that it can be used in constraints.py

### DIFF
--- a/src/fromager/requirements_file.py
+++ b/src/fromager/requirements_file.py
@@ -1,0 +1,20 @@
+import logging
+import pathlib
+import typing
+
+logger = logging.getLogger(__name__)
+
+
+def parse_requirements_file(
+    req_file: pathlib.Path,
+) -> typing.Iterable[str]:
+    lines = []
+    with open(req_file, "r") as f:
+        for line in f:
+            useful, _, _ = line.partition("#")
+            useful = useful.strip()
+            logger.debug("line %r useful %r", line, useful)
+            if not useful:
+                continue
+            lines.append(useful)
+    return lines
--- a/tests/test_constraints.py
+++ b/tests/test_constraints.py
@@ -69,7 +69,7 @@ def test_load_constraints_file(
 ):
     constraint_file = tmp_path / "constraint.txt"
     constraint_file.write_text("a\n")
-    parse_requirements_file.return_value = {str(constraint_file): ["torch==3.1.0"]}
+    parse_requirements_file.return_value = ["torch==3.1.0"]
     assert constraints.load(constraint_file)._data == {
         "torch": Requirement("torch==3.1.0")
     }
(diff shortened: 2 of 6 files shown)


src/fromager/requirements_file.py
```diff
@@ -0,0 +1,22 @@
...
+import logging
+import pathlib
+import typing
+
+logger = logging.getLogger(__name__)
+
+
+def parse_requirements_file(
+    req_files: typing.Iterable[pathlib.Path],
+) -> dict[str, typing.Iterable[str]]:
```
dhellmann, line 11: Why does this return a dict?

dhellmann: It feels like the fact that there might be multiple files and the requirements from each may need to be kept distinct is a problem for the caller to deal with. This function would be simpler if it only dealt with 1 file at a time. How many places do we have where it's called with multiple files, though? Is that the more common pattern?

shubhbapna: This function is used only twice: once in bootstrap.py to parse all the requirement files and once in constraint.py to parse a constraint file. In bootstrap.py it is called with multiple files.

Can't really tell from this if it will be a common pattern since it is used only twice.

dhellmann: OK, let's err on the side of keeping the parsing function simple and move the complexity to the caller. If we end up needing to deal with multiple files in more places, we can add another function that wraps this new one.

tests/test_constraints.py
```diff
@@ -47,3 +51,25 @@ def test_apply_constraint_to_req_with_extras():
...
+        constraints.load(non_existant_file)
+
+
+@patch("fromager.requirements_file.parse_requirements_file")
+def test_load_constraints_file(
+    parse_requirements_file: typing.Callable, tmp_path: pathlib.Path
+):
+    constraint_file = tmp_path / "constraint.txt"
+    constraint_file.write_text("a\n")
+    parse_requirements_file.return_value = {str(constraint_file): ["torch==3.1.0"]}
```
dhellmann, line 72: This no longer matches the API of the function.

Mocking fewer things and writing test content to real files will help expose any other cases like that.

shubhbapna: For unit tests, isn't better to mock `parse_requirements_file` so that if there is a test failure in `test_constraints.py` then we know it is because of the function we are testing and not cause some other function that it depends on?

dhellmann: Sometimes, depending on how complex the stack is. But this is a pretty shallow stack of relatively simple functions, so my inclination would be to mock little.

If you can mock it in a way that the API change would have been caught, that would be OK.

dhellmann (approved):

PR #160, opened 2024-07-09
Adding checks for downloaded source and tar files

rd4398:
Added the suggested changes from here: https://github.com/python-wheel-build/fromager/pull/154

Had to create a new branch since merge conflicts became a lot messy. I am really sorry about that.

### DIFF
--- a/src/fromager/sources.py
+++ b/src/fromager/sources.py
@@ -139,14 +139,14 @@ def _download_source_check(destination_dir, url):
     if source_filename.suffix == ".zip":
         source_file_contents = zipfile.ZipFile(source_filename).namelist()
         if not source_file_contents:
-            raise zipfile.BadZipFile("Bad Zip File encountered")
+            raise zipfile.BadZipFile("Empty zip file encountered")
     elif source_filename.suffix == ".tgz" or source_filename.suffix == ".gz":
         with tarfile.open(source_filename) as tar:
             contents = tar.getnames()
             if not contents:
-                raise TypeError("Bad Tar file encountered")
+                raise TypeError("Empty zip file encountered")
     else:
-        raise TypeError("Bad source file encountered")
+        raise TypeError("The source file encountered is not a zip or tar file")
     return source_filename
 
 
@@ -161,10 +161,7 @@ def download_url(destination_dir: pathlib.Path, url: str) -> pathlib.Path:
     # Open the URL first in case that fails, so we don't end up with an empty file.
     logger.debug(f"reading from {url}")
     with requests.get(url, stream=True) as r:
-        if r.status_code != 200:
-            raise requests.exceptions.HTTPError(
-                "Error in fetching url, status_code: ", r.status_code
-            )
+        r.raise_for_status()
         with open(outfile, "wb") as f:
             logger.debug(f"writing to {outfile}")
             for chunk in r.iter_content(chunk_size=1024 * 1024):
--- a/src/fromager/sources.py
+++ b/src/fromager/sources.py
@@ -125,11 +125,29 @@ def default_download_source(
     url, version = resolve_dist(
         ctx, req, sdist_server_url, include_sdists=True, include_wheels=False
     )
-    source_filename = download_url(ctx.sdists_downloads, url)
+    source_filename = _download_source_check(ctx.sdists_downloads, url)
     logger.debug(
         f"{req.name}: have source for {req} version {version} in {source_filename}"
     )
     return (source_filename, version, url)
+
+
+# Helper method to check whether .zip /.tar / .tgz is able to extract and check its content.
+# It will throw exception if any other file is encountered. Eg: index.html
+def _download_source_check(destination_dir, url):
+    source_filename = download_url(destination_dir, url)
+    if source_filename.suffix == ".zip":
+        source_file_contents = zipfile.ZipFile(source_filename).namelist()
+        if not source_file_contents:
+            raise zipfile.BadZipFile("Bad Zip File encountered")
+    elif source_filename.suffix == ".tgz" or source_filename.suffix == ".gz":
+        with tarfile.open(source_filename) as tar:
+            contents = tar.getnames()
+            if not contents:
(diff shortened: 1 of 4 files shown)


src/fromager/sources.py
```diff
@@ -143,6 +161,10 @@ def download_url(destination_dir: pathlib.Path, url: str) -> pathlib.Path:
     # Open the URL first in case that fails, so we don't end up with an empty file.
     logger.debug(f"reading from {url}")
     with requests.get(url, stream=True) as r:
+        if r.status_code != 200:
```
dhellmann, line 164: I think you can just call `raise_for_status()` here. https://requests.readthedocs.io/en/latest/user/quickstart/#response-status-codes

tests/test_sdist.py
```diff
@@ -45,3 +48,12 @@ def test_ignore_based_on_marker(tmp_context: WorkContext):
         why=[],
     )
     assert version == ""
+
+
+def test_invalid_wheel_file_exception(tmp_path: pathlib.Path):
+    fake_dir = tmp_path / "test"
+    fake_dir.mkdir()
+    test_url = "https://github.com/python-wheel-build/fromager/blob/main/README.md"
```
dhellmann, line 56: It's better if tests don't rely on being able to reach the internet to work. In an earlier version you were mocking out the call to download the file. If you do that, and then actually create a file, we won't be relying on GitHub.com access to run tests locally.

dhellmann (changes requested):

tests/test_sdist.py
rd4398: Okay, but the helper function I wrote takes URL as argument. How do I pass that to the function call in the test? I am a bit confused

dhellmann: You want to mock `sources.download_url()` so that it doesn't do anything. Then you would create a temporary file containing something that is not a zip file. Then you set up the mock so that it returns the filename to that temporary file. That way you bypass the actual download step and feed a file into the verification function.

Alternatively, you could separate the function you've already written so you can call the verification logic on a file. Then you could just test that function using temporary files.

rd4398: Okay, I mocked `sources.download_url()` by adding a `@patch("fromager.sources.download_url")`. I then created a temporary `txt` file that has something in it. I am now getting error as `FileNotFound: No such file fake_wheel.txt` 
But I am creating the file after mocking..

 ```@patch("fromager.sources.download_url")
def test_invalid_wheel_file_exception(mock_download_url, tmp_path: pathlib.Path):
    mock_download_url.return_value = "fake_wheel.txt"
    fake_dir = tmp_path / "test"
    fake_dir.mkdir()
    text_file = fake_dir / "fake_wheel.txt"
    text_file.write_text("This is a test file")
    with pytest.raises(zipfile.BadZipFile):
        sdist._download_wheel_check(fake_dir, mock_download_url)```

rd4398: Never mind, I was able to figure it out

rd4398:
I have pushed the changes as per your comments. Please let me know if any further changes are required.

tests/test_sdist.py
```diff
@@ -50,10 +50,12 @@ def test_ignore_based_on_marker(tmp_context: WorkContext):
...
+    mock_download_url.return_value = str(tmp_path / "test" / "fake_wheel.txt")
     fake_dir = tmp_path / "test"
     fake_dir.mkdir()
-    test_url = "https://github.com/python-wheel-build/fromager/blob/main/README.md"
-
+    text_file = fake_dir / "fake_wheel.txt"
+    text_file.write_text("This is a test file")
     with pytest.raises(zipfile.BadZipFile):
-        sdist._download_wheel_check(fake_dir, test_url)
+        sdist._download_wheel_check(fake_dir, mock_download_url)
```
dhellmann, line 61: This part doesn't look quite right. `mock_download_url` is a mock of a function, but this argument should be a string URL, right?

tests/test_sdist.py
```diff
@@ -50,10 +50,12 @@ def test_ignore_based_on_marker(tmp_context: WorkContext):
     assert version == ""
 
 
-def test_invalid_wheel_file_exception(tmp_path: pathlib.Path):
+@patch("fromager.sources.download_url")
+def test_invalid_wheel_file_exception(mock_download_url, tmp_path: pathlib.Path):
+    mock_download_url.return_value = str(tmp_path / "test" / "fake_wheel.txt")
```
dhellmann, line 55: I think this function is expected to return a `pathlib.Path()`, isn't it?

tests/test_sdist.py
rd4398: Hmm, I am unable to figure out what should go as parameter? Should I pass in the file name that I create? I am not sure

tests/test_sdist.py
rd4398: Yeah missed it, fixing now

rd4398:
I am now passing `text_file` which is a str instead of `mock_download_url`

tests/test_sdist.py
```diff
@@ -45,3 +48,14 @@ def test_ignore_based_on_marker(tmp_context: WorkContext):
...
+
+@patch("fromager.sources.download_url")
+def test_invalid_wheel_file_exception(mock_download_url, tmp_path: pathlib.Path):
+    mock_download_url.return_value = pathlib.Path(tmp_path / "test" / "fake_wheel.txt")
+    fake_dir = tmp_path / "test"
+    fake_dir.mkdir()
+    text_file = fake_dir / "fake_wheel.txt"
+    text_file.write_text("This is a test file")
+    with pytest.raises(zipfile.BadZipFile):
+        sdist._download_wheel_check(fake_dir, text_file)
```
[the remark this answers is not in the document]
rd4398: Okay, got it! So, should I create a fake url for this test then or a valid one like google.com?

rd4398:
I am now passing a fake url to the function as argument to avoid reliance on internet. Let me know if any more changes are required!
Thanks

Issue #151, opened 2024-07-07
highlight duplicates when writing the constraints file


dhellmann:
One of the outputs from bootstrapping is a constraints.txt file that can be used with pip to control the versions of packages that are installed, so they match the versions that will be built by the build order file.

Unfortunately, in some complex dependency trees there are duplicate requirements for the same package that end up being evaluated differently and producing conflicts in that constraints file.

We should ensure that any duplicates are listed at a the top of the file. We should pick a version for the real constraint, and include commented values for the other constraints.

rd4398:
Does duplicates mean requirement name duplicate with different version? Example: Will the following be considered duplicate?
flit-core==3.9.0
wheel==0.43.0
flit-core==3.11.0
wheel==0.44.0

dhellmann:
Yes, that's right.

rd4398:
Okay, I plan to solve this by processing list of requirements from constraints.txt and writing to a temp file. Then renaming the temp file to constraints.txt and deleting the old constraints.txt. Does that make sense? 

Also, is `We should pick a version for the real constraint, and include commented values for the other constraints.` a manual process or should code do it?

dhellmann:
The constraints file that needs to be updated is the one generated by fromager as it runs the bootstrap process. That code is [here](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/context.py#L121). The file is re-written every time that function is called, so you shouldn't need a temporary file. Just update the way it's output there.

dhellmann:
> Also, is `We should pick a version for the real constraint, and include commented values for the other constraints.` a manual process or should code do it?

That's a good question. Maybe we don't pick for now, and just list the ones with duplicates at the top of the file so it's easy for a person to pick them.

rd4398:
okay, will do!

PR #163, opened 2024-07-10
Find duplicates from constraints.txt

rd4398:
In this commit, I have created a helper to process duplicate entries in constraints.txt file and shift those to the start of the file as per discussion in https://github.com/python-wheel-build/fromager/issues/151

### DIFF
--- a/src/fromager/context.py
+++ b/src/fromager/context.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import pathlib
+from collections import Counter
 from urllib.parse import urlparse
 
 from packaging.requirements import Requirement
@@ -119,12 +120,22 @@ class WorkContext:
             # converted to JSON without help.
             json.dump(self._build_stack, f, indent=2, default=str)
         with open(self._constraints_filename, "w") as f:
-            for step in self._build_stack:
+            for step in self._process_constraints(self._build_stack):
                 comment = " ".join(
                     f"-{dep_type}-> {req.name}({version})"
                     for dep_type, req, version in step["why"]
                 )
                 f.write(f'{step["dist"]}=={step["version"]}  # {comment}\n')
+
+    # This is a helper method to find and write duplicates to start of the file from constraints.txt
+    # Input: list of objects (self._build_stack in this case)
+    # Returns: list of objects (processed_constraints in this case)
+    def _process_constraints(self, info):
+        count = Counter([item["dist"] for item in info])
+        processed_build_stack = sorted(info, key=lambda item: item["dist"])
+        return sorted(
+            processed_build_stack, key=lambda item: count[item["dist"]], reverse=True
+        )
 
     def setup(self):
         # The work dir must already exist, so don't try to create it.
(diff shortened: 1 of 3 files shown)


shubhbapna (commented):
You might get away with just sorting the `_build_stack` list actually so something like this:
```python
sorted(self._build_stack, lambda item: item["dist"])
```
This should end up grouping all the items which have the same dist together

src/fromager/context.py
```diff
@@ -119,13 +119,40 @@ def add_to_build_order(
             # converted to JSON without help.
             json.dump(self._build_stack, f, indent=2, default=str)
         with open(self._constraints_filename, "w") as f:
-            for step in self._build_stack:
+            for step in self._process_constraints(self._build_stack):
                 comment = " ".join(
                     f"-{dep_type}-> {req.name}({version})"
                     for dep_type, req, version in step["why"]
                 )
                 f.write(f'{step["dist"]}=={step["version"]}  # {comment}\n')
```
shubhbapna, line 127: After processing the build_stack wouldn't we have to comment out any duplicates?

rd4398: I discussed this  with Doug, and he said `Maybe we don't pick for now, and just list the ones with duplicates at the top of the file so it's easy for a person to pick them`

rd4398: I tried sorted() but we want all the duplicates at the top of file followed by non duplicates and sorted doesn't do that even though it groups items

shubhbapna: Oh so you want any pkg with duplicates to be on the top of the file not just grouped together in the file?

rd4398: Yeah!

shubhbapna: You could try this:
```python
# count the number of occurrences of a given dist
count = collections.Counter([item["dist"] for item in self._build_stack])
# first group all same dist items together as it will help with handling cases when 2 dists have the same count
grouped_build_stack = sorted(input, key=lambda item: item["dist"])
# sort based on the count
sorted(grouped_build_stack, key=lambda item: count[item["dist"]], reverse=True)
```

rd4398: Okay, let me try this

tests/test_context.py
```diff
@@ -166,3 +166,80 @@ def test_build_order_name_canonicalization(tmp_context):
         },
     ]
     assert expected == contents
+
+
+def test_processing_constraints_file(tmp_context):
```
shubhbapna, line 171: I would add more edge cases since we already found one that didn't work before

rd4398: Okay, I will do that!
Also, I am not able to understand why e2e tests are failing

shubhbapna (changes requested):
You can separate out the one big unit test into individual test cases that describes what kind of edge cases you are testing

rd4398:
Okay, I will do that

dhellmann:
It looks like @shubhbapna has the code review well in hand. My only feedback is to please squash the commits in your branch before the final review and merge.

shubhbapna (approved):
LGTM!

PR #164, opened 2024-07-10
Give the user an option to use predefined urls to download sdists from

### LINKED ISSUE #138: provide builtins for downloading GitHub release tarballs
Downstream we have several dependencies we're building from release tarballs on GitHub, instead of sdists. It would be convenient if there was an easy way to express that via first-class features of fromager, instead of having to hand-code each override.

[overrides.list_all()](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/overrides.py#L131) will need to be updated if we extend the settings schema.

shubhbapna:
Fixes #138 

User can provide a mapping of package to download source url in `settings.yaml`:

```yaml
download_source:
  torch:
      url: "https://github.com/pytorch/pytorch/releases/download/v${version}/pytorch-v${version}.tar.gz"
      rename_to: "torch-${version}"
```

User can define a predefined url for a package from which all its sources will be downloaded from. Optionally they can rename the downloaded sdist to whatever they want.

The only supported template variables are: `version`. `version` is replaced by the version returned by the resolver.

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -424,3 +424,17 @@ def post_build(
         f"{req.name}: running post build hook for {sdist_filename} and {wheel_filename}"
     )
 ```
+
+## Customizations using settings.yaml  
+
+To use predefined urls to download sources from, instead of overriding the entire `download_source` function, a mapping of package to download source url can be provided directly in settings.yaml:
+
+```yaml
+download_source:
+  torch:
+      url: "https://github.com/pytorch/pytorch/releases/download/v${version}/pytorch-v${version}.tar.gz"
+      rename_to: "torch-${version}"
+```
+
+User can define a predefined url for a package from which all its sources will be downloaded from. Optionally they can rename the downloaded sdist to whatever they want. The only supported template variable is `version` - it is replaced by the version returned by the resolver.  
+**Important**: Providing a version mapping doesn't set the version of whats being built. It simply maps a version to something that can be applied to get the sdist download url. If you want to download something with version X and have it built thinking it is version Y, you have to provide your own patch to do that.
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -433,8 +433,7 @@ To use predefined urls to download sources from, instead of overriding the entir
 download_source:
   torch:
       url: "https://github.com/pytorch/pytorch/releases/download/v${version}/pytorch-v${version}.tar.gz"
-      rename_to: "torch-${version}"
+      rename_to: "torch-${version}.tar.gz"
 ```
 
-User can define a predefined url for a package from which all its sources will be downloaded from. Optionally they can rename the downloaded sdist to whatever they want. The only supported template variable is `version` - it is replaced by the version returned by the resolver.  
-**Important**: Providing a version mapping doesn't set the version of whats being built. It simply maps a version to something that can be applied to get the sdist download url. If you want to download something with version X and have it built thinking it is version Y, you have to provide your own patch to do that.
+User can define a predefined url for a package from which all its sources will be downloaded from. Optionally they can rename the downloaded sdist to whatever they want. The only supported template variable is `version` - it is replaced by the version returned by the resolver.
--- a/src/fromager/overrides.py
+++ b/src/fromager/overrides.py
@@ -128,7 +128,12 @@ def find_override_method(distname: str, method: str) -> typing.Callable:
     return getattr(mod, method)
 
 
-def list_all(patches_dir: pathlib.Path, envs_dir: pathlib.Path, test: bool = False):
+def list_all(
+    patches_dir: pathlib.Path,
+    envs_dir: pathlib.Path,
+    download_source: dict[str, dict[str, str]],
+    test: bool = False,
+):
     exts = _get_extensions()
 
     def patched_projects():
@@ -162,6 +167,9 @@ def list_all(patches_dir: pathlib.Path, envs_dir: pathlib.Path, test: bool = Fal
         for item in envs_dir.glob("*/*.env"):
             yield item.stem
 
+    def projects_with_predefined_download_source():
+        yield from download_source
+
     # Use canonicalize_name() to ensure we can correctly remove duplicate
     # entries from the return list.
     return sorted(
@@ -172,6 +180,7 @@ def list_all(patches_dir: pathlib.Path, envs_dir: pathlib.Path, test: bool = Fal
                 patched_projects(),
                 patched_projects_legacy(),
                 env_projects(),
+                projects_with_predefined_download_source(),
             )
             if not test
             or n != "fromager_test"  # filter out test package except in test mode
--- a/src/fromager/settings.py
+++ b/src/fromager/settings.py
@@ -17,6 +17,19 @@ class Settings:
         names = p.get(variant) or []
         return set(overrides.pkgname_to_override_module(n) for n in names)
 
+    def download_source(self) -> dict[str, dict[str, str]]:
+        return self._data.get("download_source") or {}
+
+    def sdist_download_url(self, pkg: str) -> str | None:
+        p = self.download_source()
+        download_source = p.get(pkg) or {}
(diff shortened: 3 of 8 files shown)


src/fromager/sources.py
```diff
@@ -151,6 +166,24 @@ def download_url(destination_dir: pathlib.Path, url: str) -> pathlib.Path:
     return outfile
 
 
+def _apply_template(templates: list[str] | None, req: Requirement, version: str):
+    if not templates:
+        return None
+
+    template_env = {"version": version}
+    if req.url and req.url.startswith(f"git+{GITHUB_URL}"):
+        tag = req.url.split("@")[-1]
```
dhellmann, line 175: I think the above few lines are extracting a tag from a git URL, is that right? What if the URL only points to a branch and doesn't have a tag?

shubhbapna: Looking at the downstream plugins, i didn't find such a case. We can add it later if needed?

dhellmann: I was thinking you probably want a little error handling here for that case. Even if it just raises an exception with a more meaningful error, that would be good.

src/fromager/settings.py
```diff
@@ -17,6 +17,22 @@ def pre_built(self, variant: str) -> set[str]:
...
+        p = self._data.get("download_source") or {}
+        download_source = p.get(pkg) or {}
+        return download_source.get("url")
+
+    def rename_sdist_to(self, pkg: str) -> str | None:
+        p = self._data.get("download_source") or {}
+        download_source = p.get(pkg) or {}
+        return download_source.get("rename_to")
+
+    def map_version_to(self, pkg: str, version: str) -> str:
```
dhellmann, line 30: Same here, it's not clear what the version is being mapped to, or why. Maybe `version_override()`?

dhellmann (commented):
Could you update https://github.com/python-wheel-build/fromager/blob/main/src/fromager/overrides.py#L131 as well, please, to include these new types of overrides.

docs/customization.md
```diff
@@ -424,3 +424,26 @@ def post_build(
...
+  torch:
+      url: 
+           - "https://github.com/pytorch/pytorch/archive/refs/tags/${tag}.tar.gz"
+           - "https://github.com/pytorch/pytorch/releases/download/v${version}/pytorch-v${version}.tar.gz"
+      rename_to: "torch-${version}"
+      version_mapping:
+           "1.2.0": "2.2.0"
+```
+
+Although the user can provide multiple urls, fromager will only pick the first one in which the template was applied. If there was no url whose template fromager could resolve then it will raise an exception.
```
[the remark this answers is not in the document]
shubhbapna: If you would prefer to use `tags` from req.url if it is available. If it is not then you would want to drop back to using the url which uses `version`

src/fromager/sources.py
```diff
@@ -122,18 +124,40 @@ def default_download_source(
     sdist_server_url: str,
 ) -> tuple[pathlib.Path, str, str]:
     "Download the requirement and return the name of the output path."
-    url, version = resolve_dist(
-        ctx, req, sdist_server_url, include_sdists=True, include_wheels=False
+    custom_download_urls = ctx.settings.url_to_download_sdist_from(req.name)
+    rename_to_template = ctx.settings.rename_sdist_to(req.name)
+
+    # don't include sdists if the user wants to download source from a predefined url
```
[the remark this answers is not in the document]
shubhbapna: If the user is providing a url from which they want to download the sdist then the url that the resolver returns for the source dist won't be used anyways

shubhbapna:
> Could you update https://github.com/python-wheel-build/fromager/blob/main/src/fromager/overrides.py#L131 as well, please, to include these new types of overrides.

Sorry, I didn't quite understand what I should add here?

jwieleRH:
I was thinking of doing something like this myself for two reasons:
1. triton downloads the source from github in a file named by the version. This would be a reusable way to get the source and give the source archive a sensible name.
2. sacrebleu upstream builds its source distribution with setuptools, which leaves CHANGELOG.md out of the sdist which means you can't build the wheel from the sdist. Currently, we have a workaround that just creates an empty CHANGELOG.md file. Downloading the source from github instead of PyPI would allow us to build the wheel exactly the same way the upstream does, and of course the rename_to option would have the same application as with triton.

src/fromager/sources.py
shubhbapna: @tiran I was making the assumption that when a user defines a URL that uses tags for some package, then they know that the package will contain a URL which has tags (i.e. they have some background info on the package). Verifying whether it is a branch or tag might be quite a heavy operation.

shubhbapna: @tiran had a discussion with @dhellmann. We won't be implementing support for `tags` cause it will be removed from the downstream plugins. Plus like you said tags can't be reliable

src/fromager/overrides.py
```diff
@@ -128,7 +128,12 @@ def find_override_method(distname: str, method: str) -> typing.Callable:
     return getattr(mod, method)
 
 
-def list_all(patches_dir: pathlib.Path, envs_dir: pathlib.Path, test: bool = False):
+def list_all(
+    patches_dir: pathlib.Path,
+    envs_dir: pathlib.Path,
+    download_source: dict[str, dict[str, str]],
```
shubhbapna, line 134: @dhellmann couldn't pass the settings object directly here since that was resulting in a circular import

dhellmann: OK, maybe that's why all of those other arguments are separate, too.

For the linter in our downstream repo we'll need to figure out how to get that code to read the fromager settings and then pass the source, but that's not a big deal.

docs/customization.md
```diff
@@ -424,3 +424,17 @@ def post_build(
...
+
+## Customizations using settings.yaml  
+
+To use predefined urls to download sources from, instead of overriding the entire `download_source` function, a mapping of package to download source url can be provided directly in settings.yaml:
+
+```yaml
+download_source:
+  torch:
+      url: "https://github.com/pytorch/pytorch/releases/download/v${version}/pytorch-v${version}.tar.gz"
+      rename_to: "torch-${version}"
```
dhellmann, line 436: So this is purposefully only the base filename? What if the source is a zip file (we have some of those)?

docs/customization.md
```diff
@@ -424,3 +424,17 @@ def post_build(
...
+```
+
+User can define a predefined url for a package from which all its sources will be downloaded from. Optionally they can rename the downloaded sdist to whatever they want. The only supported template variable is `version` - it is replaced by the version returned by the resolver.  
+**Important**: Providing a version mapping doesn't set the version of whats being built. It simply maps a version to something that can be applied to get the sdist download url. If you want to download something with version X and have it built thinking it is version Y, you have to provide your own patch to do that.
```
dhellmann, line 440: I think this part is obsolete since that feature is removed, right?

src/fromager/sources.py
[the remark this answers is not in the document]
dhellmann: Including or excluding specific formats influences the resolver, too. It will only return a version the distribution type we've asked for exists for that version.

I guess in most cases where we download source from an alternative location there isn't an sdist to use, so this is probably OK. We may end up wanting to expose options to control which types of distributions to consider, though. Let's deal with that in a separate PR. In fact, we may want to split the step for resolving the version from the step for downloading the source for that version.

It would be really good to see a functioning version of this PR with a real use case based on our downstream builds.

docs/customization.md
shubhbapna: Right now it is up to the user to set the base name. If it is a zip file then they should name it `torch-${version}.zip`

dhellmann: Ah, OK, then the example should be updated to reflect that.

```suggestion
      rename_to: "torch-${version}.tar.gz"
```

dhellmann (approved):

PR #165, opened 2024-07-11
catch errors when trying to download wheels from cli

dhellmann:
When download-sequence tries to download pre-built
wheels, sometimes it finds packages that are only
available as source distributions. We can ignore
those errors because we will build the wheels
ourselves.

### DIFF
--- a/src/fromager/commands/download_sequence.py
+++ b/src/fromager/commands/download_sequence.py
@@ -55,9 +55,12 @@ def download_sequence(
                 )
 
             if include_wheels:
-                sdist.download_wheel(
-                    wkctx,
-                    req,
-                    wkctx.wheels_downloads,
-                    wheel_servers,
-                )
+                try:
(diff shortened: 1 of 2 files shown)


src/fromager/commands/download_sequence.py
```diff
@@ -55,9 +55,12 @@ def download_sequence(
...
-                )
+                try:
+                    sdist.download_wheel(
+                        wkctx,
+                        req,
+                        wkctx.wheels_downloads,
+                        wheel_servers,
+                    )
+                except Exception as err:
+                    logger.error(f"Failed to download {req}: {err}")
```
shubhbapna, line 66:
```suggestion
                    logger.error(f"Failed to download wheel for {req}: {err}")
```
just a nitpick to differentiate whether the failure was in downloading sources or wheels

shubhbapna (approved):

PR #169, opened 2024-07-11
Adding checks for downloaded source and tar files

rd4398:
Final version of changes as per the discussion : https://github.com/python-wheel-build/fromager/pull/160

(`sdist.py` had enough changes to cause merge conflicts and leading me to create another PR)

### DIFF
--- a/src/fromager/sdist.py
+++ b/src/fromager/sdist.py
@@ -8,6 +8,7 @@ import shutil
 import subprocess
 import sys
 import typing
+import zipfile
 from urllib.parse import urlparse
 
 from packaging.requirements import Requirement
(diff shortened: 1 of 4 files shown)


dhellmann (changes requested):
The logic looks good here. I want to make sure we're providing good feedback to users, so could you update all of the error messages so they include the filename, please? Also double-check the text, there's one where a tar file is called a zip file.

Otherwise this looks good.

src/fromager/sdist.py
```diff
@@ -256,14 +257,25 @@ def download_wheel(
...
 
+# Helper method to check whether the .whl file is a zip file and has contents in it.
+# It will throw BadZipFile exception if any other file is encountered. Eg: index.html
+def _download_wheel_check(destination_dir, wheel_url):
+    wheel_filename = sources.download_url(destination_dir, wheel_url)
+    wheel_directory_contents = zipfile.ZipFile(wheel_filename).namelist()
+    if not wheel_directory_contents:
+        raise zipfile.BadZipFile("Empty zip file encountered")
+
+    return wheel_filename
```
shubhbapna, line 276: Just a small suggestion to decouple `download_url` from `_download_wheel_check`

shubhbapna (commented):
Since @dhellmann was the original reviewer I will leave it for him for the final review

rd4398:
I have added filename to error messages and double checked whether they are correct (tar/zip)
Also, squashed the commits

dhellmann (approved):
Looks good, thanks!

PR #171, opened 2024-07-11
Allow all constraint operations

### LINKED ISSUE #166: add support for more expressions in constraints files
We have cases where we want to cap but not pin requirements. We need support for constraint expressions other than `==`.

shubhbapna:
fixes #166 

This adds support for all constraint operations. As discussed in #166, the implementation involves changes to the resolver. Since constraints are being directly passed to resolvers we don't need to construct new requirement like we were doing before to just support `==`. So the previous implementation has been removed


src/fromager/resolver.py
```diff
@@ -188,11 +189,13 @@ def __init__(
         include_sdists: bool = True,
         include_wheels: bool = True,
         sdist_server_url: str = "https://pypi.org/simple/",
+        constraints: Constraints | None = None,
     ):
         super().__init__()
         self.include_sdists = include_sdists
         self.include_wheels = include_wheels
         self.sdist_server_url = sdist_server_url
+        self.constraints = constraints if constraints else Constraints({})
```
dhellmann, line 198: You don't have to change it, but you can simplify this to 

```suggestion
        self.constraints = constraints or Constraints({})
```

src/fromager/constraints.py
```diff
@@ -14,34 +14,14 @@ class Constraints:
...
-            if spec.version not in req.specifier:
-                raise ValueError(
-                    f"Constraint {constraint} conflicts with requirement {req}"
-                )
-
-        new_req.specifier = constraint.specifier
-        return new_req, constraint
+    def is_satisfied_by(self, pkg_name: str, version: str | Version):
+        constraint = self._data.get(pkg_name)
+        if constraint:
```
dhellmann, line 22: Will the constraints with marker expressions that don't apply for the current system already have been filtered out by this point?

dhellmann (commented):
This looks nice and clean. I have a few questions inline, but I don't anticipate any of them holding this up.

src/fromager/constraints.py
shubhbapna: Oh thats right, I forgot about that. Adding

src/fromager/constraints.py
```diff
@@ -14,34 +14,14 @@ class Constraints:
...
-
-            # ensure that constraint and req don't conflict
-            if spec.version not in req.specifier:
-                raise ValueError(
-                    f"Constraint {constraint} conflicts with requirement {req}"
-                )
-
-        new_req.specifier = constraint.specifier
-        return new_req, constraint
+    def is_satisfied_by(self, pkg_name: str, version: str | Version):
```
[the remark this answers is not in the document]
shubhbapna: When fixing this here, I noticed that the type for `version` in `Candidate` is incorrect. It is not `str` and is in fact `Version`. Will fix that as well

dhellmann (approved):
Nice work!

PR #173, opened 2024-07-12
optimize sdist production during bootstrapping

rd4398:
I added to check whether sdist is already built before. If I get a `None`, call to `build_sdist()` 

1. Should there be anything in the `else` case? (Reference to my latest comment on issue)
2. Should I include a unit test for this?

### DIFF
--- a/e2e/test_optimize_build.sh
+++ b/e2e/test_optimize_build.sh
@@ -1,0 +1,63 @@
+#!/bin/bash
+# -*- indent-tabs-mode: nil; tab-width: 2; sh-indentation: 2; -*-
+
+set -x
+set -e
+set -o pipefail
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+OUTDIR="$(dirname "$SCRIPTDIR")/e2e-output"
+
+rm -rf "$OUTDIR"
+mkdir "$OUTDIR"
+
+tox -e e2e -n -r
+source .tox/e2e/bin/activate
+
+fromager \
+  --sdists-repo="$OUTDIR/sdists-repo" \
+  --wheels-repo="$OUTDIR/wheels-repo" \
+  --work-dir="$OUTDIR/work-dir" \
+  --log-file="$OUTDIR/logfile.txt" \
+  bootstrap 'stevedore==5.2.0'
+
+pass=true
+
+if  grep -q "have sdist version" "$OUTDIR/logfile.txt"; then
+    echo "Failed to build source distribution"
+    pass=false
+fi
+
+rm -rf "$OUTDIR/wheels-repo/downloads"
+rm -rf "$OUTDIR/logfile.txt"
+
+fromager \
+  --sdists-repo="$OUTDIR/sdists-repo" \
+  --wheels-repo="$OUTDIR/wheels-repo" \
+  --work-dir="$OUTDIR/work-dir" \
+  --log-file="$OUTDIR/logfile.txt" \
+  bootstrap 'stevedore==5.2.0'
+
+if ! grep -q "have sdist version" "$OUTDIR/logfile.txt"; then
+    echo "Building source distribution from scratch even after sdist exists"
+    pass=false
+fi
+
+rm -rf "$OUTDIR/sdists-repo/builds"
+rm -rf "$OUTDIR/wheels-repo/downloads"
+rm -rf "$OUTDIR/logfile.txt"
+
+fromager \
+  --sdists-repo="$OUTDIR/sdists-repo" \
+  --wheels-repo="$OUTDIR/wheels-repo" \
+  --work-dir="$OUTDIR/work-dir" \
+  --log-file="$OUTDIR/logfile.txt" \
+  bootstrap 'stevedore==5.2.0'
+
+if  grep -q "have sdist version" "$OUTDIR/logfile.txt"; then
+    echo "Failed to build source distribution"
+    pass=false
+fi
+
+rm -rf "$OUTDIR/logfile.txt"
+$pass
--- a/src/fromager/sdist.py
+++ b/src/fromager/sdist.py
@@ -195,7 +195,15 @@ def handle_requirement(
                 | build_sdist_dependencies,
             )
             try:
-                sources.build_sdist(ctx, req, sdist_root_dir)
+                find_sdist_result = finders.find_sdist(
+                    ctx.sdists_builds, req, resolved_version
+                )
+                if not find_sdist_result:
+                    sources.build_sdist(ctx, req, sdist_root_dir)
+                else:
+                    logger.info(
+                        f"{req.name} have sdist version {resolved_version}: {find_sdist_result}"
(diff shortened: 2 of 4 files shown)


dhellmann:
A test would definitely be good. I'm not sure how easy it would be to create a unit test, but see what you can come up with. If that looks really complicated, maybe an e2e test would make more sense in this case? There are some other e2e tests that do things like run a build, remove some of the artifacts, then build again that could serve as a starting point for an example. The test can also do things like look for specific log messages indicating that fromager is building the sdist.

rd4398:
Okay, I will try to come up with a unit test first else e2e test if things get complicated.
Thanks

rd4398:
Does this look like a good unit test?

tests/test_sdist.py
```diff
@@ -60,3 +60,44 @@ def test_invalid_wheel_file_exception(mock_download_url, tmp_path: pathlib.Path)
...
+    archive = builds / expected_base
+    archive.write_text("not-empty")
+
+    req = Requirement(dist_name)
+    ver = Version(version_string)
+
+    mock_build_sdist.return_value = None
+
+    # Give call to find_sdist and check if sdist is present
+    find_sdist_result = finders.find_sdist(builds, req, ver)
```
dhellmann, line 87: This is testing the finders package function, but we already have tests for those. 

I'm not sure how easy it's going to be to write a unit test for handle_requirement(), you might need to use an e2e test. For example, you could bootstrap something, then bootstrap the same thing again and look for a log message indicating that the existing sdist was used. Then you could remove the sdist and bootstrap again and look for a message indicating that it was created.

Keep in mind that the bootstrap log is appended, so you either need to provide a different file name each time or remove the file between bootstrap runs.

rd4398: Okay, will take a look

rd4398:
Fixed failing `pkglint `checks

e2e/test_optimize_build.sh
```diff
@@ -0,0 +1,60 @@
...
+fromager \
+  --sdists-repo="$OUTDIR/sdists-repo" \
+  --wheels-repo="$OUTDIR/wheels-repo" \
+  --work-dir="$OUTDIR/work-dir" \
+  bootstrap 'stevedore==5.2.0' 2>&1 | tee "$OUTDIR/logfile.txt"
+
+pass=true
+
+if ! grep -q "Building source distribution" "$OUTDIR/logfile.txt"; then
+    echo "grep not successful"
```
shubhbapna, line 26: Better test failure message will help us debug if there are any failures in the future

shubhbapna (commented):
Looks good! Will wait for @dhellmann approval as well

e2e/test_optimize_build.sh
```diff
@@ -0,0 +1,60 @@
...
+fi
+
+rm -rf "$OUTDIR/wheels-repo/downloads"
+rm -rf "$OUTDIR/logfile.txt"
+
+fromager \
+  --sdists-repo="$OUTDIR/sdists-repo" \
+  --wheels-repo="$OUTDIR/wheels-repo" \
+  --work-dir="$OUTDIR/work-dir" \
+  bootstrap 'stevedore==5.2.0' 2>&1 | tee "$OUTDIR/logfile.txt"
```
dhellmann, line 37: You can use `--log-file` to have fromager create a log file for you.

src/fromager/sdist.py
```diff
@@ -195,7 +195,14 @@ def handle_requirement(
                 | build_sdist_dependencies,
             )
             try:
-                sources.build_sdist(ctx, req, sdist_root_dir)
+                find_sdist_result = finders.find_sdist(
+                    ctx.sdists_builds, req, resolved_version
+                )
+                if not find_sdist_result:
+                    logger.info(f"{req.name}: Building source distribution")
```
dhellmann, line 202: `build_sdist()` already logs a message like this

src/fromager/sdist.py
```diff
@@ -195,7 +195,14 @@ def handle_requirement(
...
+                    sources.build_sdist(ctx, req, sdist_root_dir)
+                else:
+                    logger.info(f"{req.name} Already exists in the build directory")
```
dhellmann, line 205: Let's make this message consistent with the one for wheels on line 183.

dhellmann (commented):
The logic looks OK. I have a few comments on consistency.

rd4398:
Changed error messages as per suggestion and used --log-file to generate log file
Let me know if anymore changes are required!

rd4398:
Can you please approve and merge if everything looks fine?

src/fromager/sdist.py
```diff
@@ -195,7 +195,16 @@ def handle_requirement(
                 | build_sdist_dependencies,
             )
             try:
-                sources.build_sdist(ctx, req, sdist_root_dir)
+                find_sdist_result = finders.find_sdist(
+                    ctx.sdists_builds, req, resolved_version
+                )
+                if not find_sdist_result:
+                    logger.info(f"{req.name}: already resolved sdist version not found")
```
dhellmann, line 202: We don't need this much verbosity. Let's focus on logging things that are different from the expected default behavior. In this case, for example, it would be when we do find an sdist and so we are not going to rebuild it.

src/fromager/sdist.py
```diff
@@ -195,7 +195,16 @@ def handle_requirement(
...
+                    sources.build_sdist(ctx, req, sdist_root_dir)
+                else:
+                    logger.info(
+                        f"{req.name} has sdist version {resolved_version}: {find_sdist_result}"
```
dhellmann, line 206:
```suggestion
                        f"{req.name}: have sdist version {resolved_version}: {find_sdist_result}"
```

Minor tweak to the format of the message for consistency.

e2e/test_optimize_build.sh
```diff
@@ -0,0 +1,63 @@
...
+fromager \
+  --sdists-repo="$OUTDIR/sdists-repo" \
+  --wheels-repo="$OUTDIR/wheels-repo" \
+  --work-dir="$OUTDIR/work-dir" \
+  --log-file="$OUTDIR/logfile.txt" \
+  bootstrap 'stevedore==5.2.0'
+
+pass=true
+
+if ! grep -q "already resolved sdist version not found" "$OUTDIR/logfile.txt"; then
```
dhellmann, line 26: When you remove the log line that emits this message, this part of the test would have to change. You could look for the same "has sdist version" message that is checked on line 41, but expect to not find it (and so fail if you do).

dhellmann (commented):
I have some suggestions to adjust the logging. Once those are done, we'll look at merging this next week, after the deadline we have this week.

src/fromager/sdist.py
rd4398: Does `sdist not found, building sdist` sound good?

dhellmann: I think you want to remove this log message entirely.

rd4398: Okay

PR #175, opened 2024-07-12
Add caching to dependency resolution

### LINKED ISSUE #167: optimize dependency resolution
During bootstrap we write various requirements files as artifacts to examine different types of dependencies for packages. We could use those as a "cache" to avoid having to recompute the values, and save time when rerunning a bootstrap during override development. We already do something similar to avoid rebuilding a wheel if it already exists.

shubhbapna:
fixes #167 

Before resolving and getting requirements for dependency resolution, it checks first to see if the requirements have been produced before and if they have then they are loaded in from the file they were written to instead of resolving them again.

### DIFF
--- a/src/fromager/dependencies.py
+++ b/src/fromager/dependencies.py
@@ -152,21 +152,6 @@ def default_get_build_sdist_dependencies(
     return hook_caller.get_requires_for_build_wheel()
 
 
-def get_install_dependencies(
-    ctx: context.WorkContext,
-    req: Requirement,
-    sdist_root_dir: pathlib.Path,
-) -> set[Requirement]:
-    logger.info(
-        f"{req.name}: getting installation dependencies for {req} in {sdist_root_dir}"
-    )
-    dep_func = overrides.find_override_method(req.name, "get_install_dependencies")
-    if not dep_func:
-        dep_func = default_get_install_dependencies
-    deps = _filter_requirements(req, dep_func(ctx, req, sdist_root_dir))
-    return deps
-
-
 def get_install_dependencies_of_wheel(
     req: Requirement, wheel_filename: pathlib.Path, requirements_file_dir: pathlib.Path
 ) -> set[Requirement]:
--- a/src/fromager/dependencies.py
+++ b/src/fromager/dependencies.py
@@ -24,10 +24,23 @@ def get_build_system_dependencies(
     logger.info(
         f"{req.name}: getting build system dependencies for {req} in {sdist_root_dir}"
     )
+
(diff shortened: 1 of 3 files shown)


src/fromager/dependencies.py
```diff
@@ -24,10 +24,23 @@ def get_build_system_dependencies(
     logger.info(
         f"{req.name}: getting build system dependencies for {req} in {sdist_root_dir}"
     )
+
+    build_system_req_file = sdist_root_dir.parent / "build-system-requirements.txt"
+    if build_system_req_file.exists():
+        logger.info(
+            f"{req.name}: {build_system_req_file} already exists. Loading build system dependencies from {build_system_req_file.name}"
```
dhellmann, line 31: This log message will include the filename twice, right?

```suggestion
            f"{req.name}: loading build system dependencies from {build_system_req_file}"
```

src/fromager/dependencies.py
```diff
@@ -63,12 +76,25 @@ def get_build_backend_dependencies(
     logger.info(
         f"{req.name}: getting build backend dependencies for {req} in {sdist_root_dir}"
     )
+
+    build_backend_req_file = sdist_root_dir.parent / "build-backend-requirements.txt"
+    if build_backend_req_file.exists():
+        logger.info(
+            f"{req.name}: {build_backend_req_file} already exists. Loading build backend dependencies from {build_backend_req_file.name}"
```
dhellmann, line 83:
```suggestion
            f"{req.name}: loading build backend dependencies from {build_backend_req_file}"
```

src/fromager/dependencies.py
```diff
@@ -93,10 +119,23 @@ def get_build_sdist_dependencies(
     logger.info(
         f"{req.name}: getting build sdist dependencies for {req} in {sdist_root_dir}"
     )
+
+    build_sdist_req_file = sdist_root_dir.parent / "build-sdist-requirements.txt"
+    if build_sdist_req_file.exists():
+        logger.info(
+            f"{req.name}: {build_sdist_req_file} already exists. Loading build sdist dependencies from {build_sdist_req_file.name}"
```
dhellmann, line 126:
```suggestion
            f"{req.name}: loading build sdist dependencies from {build_sdist_req_file.name}"
```

dhellmann (changes requested):
The logic looks good. I have some comments on the log messages, and it would be good to have unit tests for the bypass behavior for the functions where you're reading the cache.

dhellmann (approved):
Looks good!

PR #186, opened 2024-07-15
try to make the output of build-order show-duplicates more readable

dhellmann:


src/fromager/commands/build_order.py
```diff
@@ -298,12 +298,22 @@ def show_duplicates(build_order_file: pathlib.Path, install_only: bool):
...
+            print(f"  {req}")
+            if len(why) > 1:
+                parents = " -- ".join(f"'{w[1]}'" for w in why[:-1])
+                print(
+                    f"    resolved to {step['version']} as {article} {req_type} dependency of {parents}"
+                )
+            else:
+                print(
+                    f"    resolved to {step['version']} as {article} {req_type} dependency"
+                )
```
shubhbapna, line 319:
```suggestion
            parents = " of " + " -- ".join(f"'{w[1]}'" for w in why[:-1]) if len(why) > 1 else ""
            print(f"\t resolved to {step['version']} as {article} {req_type} dependency{parents}")
```

shubhbapna (commented):
Instead of white space a tab character might be more consistent?

src/fromager/commands/build_order.py
dhellmann: Do you find that easier to read? It seems more complex to me, especially with all of the quotes.

shubhbapna: fair i was trying to reduce the duplicate print statements. How about something like this?

```python
show_duplicate_msg = f"\t resolved to {step['version']} as {article} {req_type} dependency"
if len(why) > 1:
    parents = " -- ".join(f"'{w[1]}'" for w in why[:-1])
    show_duplicate_msg = f"{show_duplicate_msg} of {parents}"
print(show_duplicate_msg)
```

dhellmann: That looks better, I'll make that update when I come back to this.

dhellmann: Updated

PR #187, opened 2024-07-16
Validating return value of version from download_source()

rd4398:
Let me know if changes are required

### DIFF
--- a/src/fromager/sources.py
+++ b/src/fromager/sources.py
@@ -14,7 +14,8 @@ from urllib.parse import urlparse
 import requests
 import resolvelib
 from packaging.requirements import Requirement
-from packaging.version import Version
+from packaging.version import InvalidVersion, Version
+from packaging.version import parse as validate_version
 
 from . import context, dependencies, overrides, resolver, tarballs, vendor_rust
 
@@ -52,13 +53,19 @@ def download_source(
                 raise ValueError(
                     f"do not know how to unpack {download_details}, expected 2 or 3 members"
                 )
+
+            # Validate version string by passing it to parse
+            validate_version(str(version))
+
         except (
             resolvelib.InconsistentCandidate,
             resolvelib.RequirementsConflicted,
             resolvelib.ResolutionImpossible,
+            InvalidVersion,
         ) as err:
             logger.debug(f"{req.name}: failed to resolve {req} using {url}: {err}")
             continue
+
         return (source_filename, version, source_url, source_type)
     servers = ", ".join(sdist_server_urls)
     raise ValueError(f"failed to find source for {req} at {servers}")
--- a/tests/test_sources.py
+++ b/tests/test_sources.py
@@ -84,3 +84,17 @@ def test_default_download_source_with_predefined_url(
     download_source_check.assert_called_with(
         tmp_context.sdists_downloads, "predefined_url-1.0", "foo-1.0"
     )
+
+
+@patch("fromager.sources.default_download_source")
+def test_invalid_version(mock_default_download):
+    ctx = context.WorkContext
+    req = Requirement("fake==1.0")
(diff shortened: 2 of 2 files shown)


src/fromager/sources.py
```diff
@@ -59,6 +59,13 @@ def download_source(
         ) as err:
             logger.debug(f"{req.name}: failed to resolve {req} using {url}: {err}")
             continue
+
+        # Validate version string by passing it to Version
+        try:
+            Version(str(version))
```
shubhbapna, line 65: Instead of initializing an object maybe using the `parse` function from `packaging.version` might be better since we are just trying to validate a version right now rather than create an object

shubhbapna, line 65: I think `version` is already a string so `str()` shouldn't be needed

tests/test_sources.py
```diff
@@ -84,3 +85,9 @@ def test_default_download_source_with_predefined_url(
     download_source_check.assert_called_with(
         tmp_context.sdists_downloads, "predefined_url-1.0", "foo-1.0"
     )
+
+
+def test_invalid_version():
+    version = "fake version 1.0"
+    with pytest.raises(InvalidVersion):
+        Version(version)
```
shubhbapna, line 93: This is testing the `Version` object from the `packaging` library which we have no control over anyways. It will make more sense to test if `download_source` throws an error if it receives an invalid version from the plugin or the default function

shubhbapna (changes requested):
I think we should move the validation of the version in the for loop and try the next url if we get an invalid version. IMO we shouldn't be returning an invalid version

tests/test_sources.py
rd4398: okay, will do

src/fromager/sources.py
rd4398: I checked, its not a `str()`. The e2e tests were failing because of that

rd4398: Okay, I will check out `parse`

shubhbapna: You are right. In fact `default_download_source` is returning a version object and is already doing this validation then. It gets the version from: https://github.com/python-wheel-build/fromager/blob/main/src/fromager/sources.py#L138

and in `resolve_dist`, the `candidate.version` field is being returned which is already a `Version`: https://github.com/python-wheel-build/fromager/blob/main/src/fromager/resolver.py#L44

We might want to fix this mismatch

rd4398: Hmm let me check. Does it mean we might not need validation again at all?

shubhbapna: The user defined override will a return a string. That's the problem

rd4398: Okay, got it!

rd4398:
I have made changes as per the suggestions in PR. Let me know if any more changes are required!
I believe the mismatch you pointed out is a separate conversation for the future

src/fromager/sources.py
```diff
@@ -14,7 +14,7 @@
 import requests
 import resolvelib
 from packaging.requirements import Requirement
-from packaging.version import Version
+from packaging.version import InvalidVersion, Version, parse
```
dhellmann, line 17: `parse` is a _very_ generic name to import like this. :-( 

I'll live with it, but if you have other changes to make, consider renaming it on import or importing the module and accessing the function using `version.parse`.

dhellmann (commented):
This looks OK. One suggestion for a cleanup, but I won't hold it up for that.

@shubhbapna had asked for changes, so I'll leave it to him to approve.

src/fromager/sources.py
rd4398: okay, I will change it!

shubhbapna (approved):
LGTM!

PR #190, opened 2024-07-17
build_sdist hook with extra_environ

tiran:
The `build_sdist` hook and `pep517_build_sdist` function now receive `extra_environ` parameter. This allows the caller to modify the environment for building an sdist.

PyTorch needs extra env vars to set the version and build number.

Breaking change, any existing `build_sdist` hook must be updated with the new argument.


shubhbapna (commented):
If `extra_environ` is needed in multiple places now, then shouldn't we add it to the context? It will be accessible without having to change the function signatures. Once it is available in the context you can define all the other env variables that depend on the resolved version by overriding `download_source` and won't have to pass `version` to other functions

Gregory-Pereira (approved):
Generally this looks great.
/lgtm

PR #192, opened 2024-07-17
This PR adds code to calculate time taken to build a wheel

rd4398:
I have used the time format as `hh:mm:ss` and ignored the milliseconds. Let me know if we want milliseconds as well.
Also, I have used the log message in `build_wheel()` to log this time required as `build_wheel()` is the function called by bootstrap, build and build-sequence.

Let me know if any changes are required!

### DIFF
--- a/src/fromager/wheels.py
+++ b/src/fromager/wheels.py
@@ -78,8 +78,6 @@ def build_wheel(
     logger.info(
         f"{req.name}: building wheel for {req} in {sdist_root_dir} writing to {ctx.wheels_build}"
     )
-    # Start the timer
-    start = datetime.now().replace(microsecond=0)
 
     builder = overrides.find_override_method(req.name, "build_wheel")
     if not builder:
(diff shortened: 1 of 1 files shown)


src/fromager/wheels.py
```diff
@@ -94,6 +98,11 @@ def build_wheel(
 
     builder(ctx, build_env, extra_environ, req, sdist_root_dir)
     wheels = list(ctx.wheels_build.glob("*.whl"))
+
+    # End the timer
+    end = datetime.now().replace(microsecond=0)
+    logger.info(f"Time taken to build wheel for {req.name}: {end - start}")
```
shubhbapna, line 104: To keep the logging format consistent start log statements with `req.name` like here - 
https://github.com/python-wheel-build/fromager/pull/192/files#diff-e7392f8e6edefea9b61e882f028ac83e81d7c40ca569de4fb02b380ffe2359f2L78

shubhbapna (changes requested):
We should only measure the time it took for `builder` to execute. Right now the timing is measuring the time to locate an override, executing builder and running glob

PR #193, opened 2024-07-17
This commit adds code to calculate time taken to build a wheel

rd4398:
Changes as per suggestion in https://github.com/python-wheel-build/fromager/pull/192#pullrequestreview-2183301298

### DIFF
--- a/src/fromager/wheels.py
+++ b/src/fromager/wheels.py
@@ -5,6 +5,7 @@ import platform
 import sys
 import tempfile
 import typing
+from datetime import datetime
 
 from packaging.requirements import Requirement
 from packaging.version import Version
@@ -94,6 +95,8 @@ def build_wheel(
         extra_environ["CMAKE_BUILD_PARALLEL_LEVEL"] = f"{ctx.jobs}"
         extra_environ["MAX_JOBS"] = f"{ctx.jobs}"
 
+    # Start the timer
+    start = datetime.now().replace(microsecond=0)
     builder(
         ctx=ctx,
         build_env=build_env,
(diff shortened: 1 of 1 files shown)


shubhbapna (commented):
Looks good! But will wait for @dhellmann approval for merge so as to not block this weeks builds

src/fromager/wheels.py
```diff
@@ -102,6 +105,9 @@ def build_wheel(
         sdist_root_dir=sdist_root_dir,
         version=version,
     )
+    # End the timer
+    end = datetime.now().replace(microsecond=0)
+    logger.info(f"{req.name}: time taken to build wheel {end - start}")
```
dhellmann, line 110: Does this output go to stdout?

What happens to the `fromager step build` command output with this info log message here? The command is supposed to print the filename it creates in a way that a shell script can grab it and do something with it. Will that still work with this change here?

rd4398: Yeah, I can see it on terminal and in file `*-prepare-build.log` file
I also ran `test_build_steps.sh` to see if it works.

Is there anything else / another way that I need to test?

dhellmann: What happens if you try to capture the output of a build command? Something like `wheel=$(fromager step build ...)`. Do you get just the filename, or does it include this new log message, too?

rd4398: I get just the filename
I ran the following command
```
OUTPUT="$(fromager \
        --log-file "$OUTDIR/build-logs/${dist}-prepare-build.log" \
        --work-dir "$OUTDIR/work-dir" \
        --sdists-repo "$OUTDIR/sdists-repo" \
        --wheels-repo "$OUTDIR/wheels-repo" \
        --wheel-server-url "${WHEEL_SERVER_URL}" \
        step build-wheel "$dist" "$version")"

echo "${OUTPUT}"
```
I get output as

`++ echo /home/rdevasth/fromager/e2e-output/wheels-repo/build/stevedore-5.2.0-py3-none-any.whl`

dhellmann: Good, that's what I was hoping.

dhellmann (approved):

PR #195, opened 2024-07-18
Feat: entry points with override methods

tiran:
The new entry point `fromager.override_methods` lists all fromager override methods by name and their default implementation. This allows consumers to get a list of known override methods and compare function signatures.

Also address some typing issues in override methods and helpers.


dhellmann:
I like this. We could document using [entry_point_inspector](https://pypi.org/project/entry-point-inspector/) for listing them.

This is relevant to #196, too, @shubhbapna

PR #197, opened 2024-07-18
feat: add implementation of invoke function

### LINKED ISSUE #196: make override invocations more backwards-compatible
We recently added new arguments to some of the override function definitions and that is making it more difficult to upgrade existing uses in the repo downstream. It would be good if those new arguments were considered optional, since they're obviously not needed by existing implementations.

We could add a function to the `overrides` module to invoke an override, passing only the arguments that the function actually accepts.  It could be called from within fromager like this:

```python
download_details = overrides.invoke(req.name, "download_source", default_download_source,
    ctx=ctx, req=req, url=url)
```

The first 3 arguments are the name of the package, the name of the override, and the default function to call if there is no override for that package. The remaining named arguments are all args to "download_source".

`invoke()` should use `inspect` to look at the function signature of the thing being called and pass any named arguments that it accepts. For each value received by `invoke` that the override doesn't take should produce a warning log message like "{req.name}: {override_name} override {callable} does not take argument {arg_name}".

All invocations of override functions in fromager should be updated to use the new `invoke()` function.

shubhbapna:
fixes #196

### DIFF
--- a/src/fromager/dependencies.py
+++ b/src/fromager/dependencies.py
@@ -32,10 +32,15 @@ def get_build_system_dependencies(
         )
         return _read_requirements_file(build_system_req_file)
 
-    dep_func = overrides.find_override_method(req.name, "get_build_system_dependencies")
-    if not dep_func:
-        dep_func = default_get_build_system_dependencies
-    deps = _filter_requirements(req, dep_func(ctx, req, sdist_root_dir))
+    orig_deps = overrides.find_and_invoke(
+        req.name,
+        "get_build_system_dependencies",
+        default_get_build_system_dependencies,
+        ctx=ctx,
+        req=req,
+        sdist_root_dir=sdist_root_dir,
+    )
+    deps = _filter_requirements(req, orig_deps)
 
     _write_requirements_file(
         deps,
@@ -84,12 +89,15 @@ def get_build_backend_dependencies(
         )
         return _read_requirements_file(build_backend_req_file)
 
-    dep_func = overrides.find_override_method(
-        req.name, "get_build_backend_dependencies"
+    orig_deps = overrides.find_and_invoke(
+        req.name,
+        "get_build_backend_dependencies",
+        default_get_build_backend_dependencies,
+        ctx=ctx,
+        req=req,
+        sdist_root_dir=sdist_root_dir,
     )
-    if not dep_func:
-        dep_func = default_get_build_backend_dependencies
-    deps = _filter_requirements(req, dep_func(ctx, req, sdist_root_dir))
+    deps = _filter_requirements(req, orig_deps)
 
     _write_requirements_file(
         deps,
@@ -127,10 +135,15 @@ def get_build_sdist_dependencies(
         )
         return _read_requirements_file(build_sdist_req_file)
 
-    dep_func = overrides.find_override_method(req.name, "get_build_sdist_dependencies")
-    if not dep_func:
-        dep_func = default_get_build_sdist_dependencies
-    deps = _filter_requirements(req, dep_func(ctx, req, sdist_root_dir))
+    orig_deps = overrides.find_and_invoke(
+        req.name,
+        "get_build_sdist_dependencies",
+        default_get_build_sdist_dependencies,
+        ctx=ctx,
+        req=req,
+        sdist_root_dir=sdist_root_dir,
+    )
+    deps = _filter_requirements(req, orig_deps)
 
     _write_requirements_file(
         deps,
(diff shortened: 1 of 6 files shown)


src/fromager/overrides.py
```diff
@@ -40,6 +41,23 @@ def _die_on_plugin_load_failure(
...
 
+def invoke(distname: str, method: str, default_fn: typing.Callable, **kwargs):
+    fn = find_override_method(distname, method)
+    if not fn:
+        fn = default_fn
+
+    sig = inspect.signature(fn)
+    # shallow copy because we do not want copies of dicts or objects passsed by the user as values to named args
+    fn_kwargs = kwargs.copy()
+    for arg_name in kwargs.keys():
```
tiran, line 52: - `**kwargs` is already a shallow copy. You can safely modify the dict without affecting the caller
- `keys()` is bad style, `for key in kwargs` works.

```suggestion
    for arg_name in kwargs:
```

shubhbapna: Yes that's not why I was using the shallow copy. The issue was popping the keys while iterating

tiran: ah, right!
The solution for that is `for arg_name in list(kwargs)`. This creates a copy of keys. The `dict.keys()` method returns a view (iterator), not a copy of keys.

shubhbapna: Ah nice thank you

src/fromager/overrides.py
```diff
@@ -40,6 +41,22 @@ def _die_on_plugin_load_failure(
...
+    if not fn:
+        fn = default_fn
+    return invoke(fn, **kwargs)
+
+
+def invoke(fn: typing.Callable, **kwargs):
+    sig = inspect.signature(fn)
+    for arg_name in list(kwargs):
+        if arg_name not in sig.parameters:
+            logger.warning(f"{fn.__name__} override does not take argument {arg_name}")
```
dhellmann, line 55: I can never remember how much info the `__name__` property of a function includes. Is it just the name, or does it include the module, etc.? Ideally we would log the full thing, like `package_overrides.torch.download_source` so someone looking at it can go from the warning message straight to the relevant code.

tests/test_overrides.py
```diff
@@ -147,3 +149,61 @@ def test_list_all(tmp_path: pathlib.Path):
...
+def test_invoke_override_with_exact_args(find_override_method: typing.Callable):
+    def default_foo():
+        return False
+
+    def foo(arg1, arg2):
+        return arg1 is not None and arg2 is not None
+
+    find_override_method.return_value = foo
+
+    assert overrides.invoke(foo, arg1="value1", arg2="value2")
```
dhellmann, line 164: You're just calling invoke, so mocking the find method shouldn't be needed? Maybe you meant to call through both code paths?

dhellmann (commented):
This looks very good. A few comments inline.

src/fromager/overrides.py
shubhbapna: It just includes the name. Let me try adding the full the thing

tests/test_overrides.py
shubhbapna: Ah yes forgot to remove that

dhellmann:
I'm happy with this version. I'd like to get @tiran's input before we merge it and cut a new release.

src/fromager/dependencies.py
```diff
@@ -32,10 +32,15 @@ def get_build_system_dependencies(
...
         return _read_requirements_file(build_system_req_file)
 
-    dep_func = overrides.find_override_method(req.name, "get_build_system_dependencies")
-    if not dep_func:
-        dep_func = default_get_build_system_dependencies
-    deps = _filter_requirements(req, dep_func(ctx, req, sdist_root_dir))
+    orig_deps = overrides.find_and_invoke(
+        req.name,
+        "get_build_system_dependencies",
+        default_get_build_system_dependencies,
```
tiran, line 38: This smells like a violation of DRY (don't repeat yourself). How about we enforce a consistent naming schema for default methods, then change `find_and_invoke` to get the name from the default function name?

```
assert default_fn.__name__.startswith("default_")
method = default_fn.__name__[8:]
```

shubhbapna: If we enforce this naming convention we will lose the opportunity to have common default functions or default functions that might have other use as well

dhellmann: I need this feature so I can update fromager in the builder repo today. Let's consider this improvement in a follow-up PR.

PR #205, opened 2024-07-19
optimizing sdist production during bootstrapping

rd4398:
As per changes and discussion here: https://github.com/python-wheel-build/fromager/pull/173

### DIFF
--- a/tests/test_sources.py
+++ b/tests/test_sources.py
@@ -96,5 +96,6 @@ def test_invalid_version(mock_default_download):
         "fake version 1.0",
         "fakesdisturl.com",
     )
+    mock_default_download.__name__ = "mock_default_download"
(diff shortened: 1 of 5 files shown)


tests/test_sources.py
```diff
@@ -96,5 +96,6 @@ def test_invalid_version(mock_default_download):
         "fake version 1.0",
         "fakesdisturl.com",
     )
+    mock_default_download.__name__ = "mock_default_download"
```
shubhbapna, line 99: Could you push this as a separate PR, need this fix in to stop the unit tests from failing

rd4398: okay, creating a separate PR now

dhellmann (approved):

PR #209, opened 2024-07-19
look up constraints using their canonical name

dhellmann:
Authors of requirements and constraints files are
not consistent in how they capitalize or punctuate
the package names. Convert the names to their
canonical forms in the dictionary and when we try
to look up a package so the names will match more
often. Change the API of
Constraints.get_constraint() to only take a name
so we can do the lookup more consistently.


shubhbapna (approved):

dhellmann:
@Mergifyio rebase

PR #212, opened 2024-07-20
ensure build commands pass version as class not str

### LINKED ISSUE #210: build commands pass version to builder as string not Version
builder plugins expect the version argument to be a Version but are receiving a string.

dhellmann:
Convert the input string to a Version instance
before calling the real build code.

Fixes #210


rd4398 (approved):
Looks good!

PR #214, opened 2024-07-21
add optional error log

dhellmann:
The debug log can be quite verbose, so provide an
option for logging just error messages to a
separate file for easier review.


tiran (commented):
LGTM

I wonder it it would be useful to log everything >= WARNING instead >= ERROR?

dhellmann:
> LGTM
> 
> I wonder it it would be useful to log everything >= WARNING instead >= ERROR?

That might be a good idea. We could come back and do that if we find warnings useful?

PR #215, opened 2024-07-21
fix return value for pep517_build_sdist

dhellmann:
Return the full path to the built file instead of
just the base name that is returned by the build
backend. Add an end-to-end test for building
sdists using pep517_build_sdist().


tiran (commented):
LGTM!

It's always the one function that is not covered by tests, isn't it?

Issue #202, opened 2024-07-19
warn when patches for old version of package exist


dhellmann:
As we update to new versions of dependencies, we sometimes have to update patches that let old versions of those same packages build. It's not always obvious that a new package is available, because it usually shows up when resolving something in the bootstrap process. It would be helpful to have clear warning-level messages when fromager prepares the source for a package and finds patches for old versions, but not the version being built at the time.

PR #219, opened 2024-07-22
This commit warns if patches for older versions exists

rd4398:
Fixes: https://github.com/python-wheel-build/fromager/issues/202

### DIFF
--- a/src/fromager/sources.py
+++ b/src/fromager/sources.py
@@ -270,8 +270,14 @@ def unpack_source(
     return (next(iter(unpack_dir.glob("*"))), True)
 
 
-def patch_source(ctx: context.WorkContext, source_root_dir: pathlib.Path) -> None:
+def patch_source(
+    ctx: context.WorkContext,
+    source_root_dir: pathlib.Path,
+    version: Version | None = None,
+) -> None:
     for p in overrides.patches_for_source_dir(ctx.patches_dir, source_root_dir.name):
+        if version:
+            _warn_for_old_patch(source_root_dir, p, version)
         logger.info("applying patch file %s to %s", p, source_root_dir)
         with open(p, "r") as f:
             subprocess.check_call(
@@ -279,6 +285,20 @@ def patch_source(ctx: context.WorkContext, source_root_dir: pathlib.Path) -> Non
                 stdin=f,
                 cwd=source_root_dir,
             )
+
+
+def _warn_for_old_patch(
+    source_root_dir: pathlib.Path, patch_filename: pathlib.Path, version: Version
+) -> None:
+    # Get the existing version and req name as per the source_root_dir naming conventions
+    req_name = source_root_dir.name.rsplit("-", 1)[0]
+    existing_version = Version(source_root_dir.name.rsplit("-", 1)[1])
+
+    # Check to see if patches for older version exists
+    if patch_filename.parent.name == source_root_dir.name:
+        logger.warning(
+            f"Patches for version {existing_version!s} of {req_name} exist but will not be applied"
+        )
 
 
 def write_build_meta(
--- a/tests/test_sources.py
+++ b/tests/test_sources.py
@@ -4,6 +4,7 @@ from unittest.mock import patch
 
 import pytest
 from packaging.requirements import Requirement
+from packaging.version import Version
 
 from fromager import context, settings, sources
 
@@ -99,3 +100,18 @@ def test_invalid_version(mock_default_download):
     mock_default_download.__name__ = "mock_default_download"
     with pytest.raises(ValueError):
         sources.download_source(ctx, req, sdist_server_urls)
+
+
+@patch("logging.Logger.warning")
+def test_warning_for_older_patch(mock, tmp_path: pathlib.Path):
+    source = "deepspeed-0.5.0"
+    patches_dir = tmp_path / source
+    patches_dir.mkdir()
+    patch_file = patches_dir / "deepspeed-0.5.0.patch"
+    patch_file.write_text("This is a test patch")
+
+    new_version = Version("0.6.0")
+    sources._warn_for_old_patch(patches_dir, patch_file, new_version)
+    mock.assert_called_with(
+        "Patches for version 0.5.0 of deepspeed exist but will not be applied"
+    )


shubhbapna (changes requested):
Minor edits

src/fromager/sources.py
```diff
@@ -270,7 +270,10 @@ def unpack_source(
     return (next(iter(unpack_dir.glob("*"))), True)
 
 
-def patch_source(ctx: context.WorkContext, source_root_dir: pathlib.Path) -> None:
+def patch_source(
+    ctx: context.WorkContext, source_root_dir: pathlib.Path, version
```
shubhbapna, line 274:
```suggestion
    ctx: context.WorkContext, source_root_dir: pathlib.Path, version: Version
```

src/fromager/sources.py
```diff
@@ -281,6 +284,18 @@ def patch_source(ctx: context.WorkContext, source_root_dir: pathlib.Path) -> Non
             )
 
 
+def _warn_for_old_patch(source_root_dir: pathlib.Path, version) -> None:
```
shubhbapna, line 287:
```suggestion
def _warn_for_old_patch(source_root_dir: pathlib.Path, version: Version) -> None:
```

src/fromager/sources.py
```diff
@@ -281,6 +284,18 @@ def patch_source(ctx: context.WorkContext, source_root_dir: pathlib.Path) -> Non
...
+    # Get the existing version and req name as per the source_root_dir naming conventions
+    req_name = source_root_dir.name.split("-")[0]
+    existing_version = Version(source_root_dir.name.split("-")[-1])
```
shubhbapna, line 290: Something that I learned from @tiran 
```suggestion
    existing_version = Version(source_root_dir.name.rsplit("-", 1)[1])
```

rd4398:
changes done as per suggestions!

src/fromager/commands/build.py
```diff
@@ -69,7 +69,7 @@ def build_sequence(
     with open(build_order_file, "r") as f:
         for entry in json.load(f):
             wheel_filename = _build(
-                wkctx, entry["dist"], entry["version"], sdist_server_url
+                wkctx, entry["dist"], Version(entry["version"]), sdist_server_url
```
dhellmann, line 72: Maybe we can merge https://github.com/python-wheel-build/fromager/pull/212 instead, and you can keep this commit focused on the warning changes.

src/fromager/sources.py
```diff
@@ -281,6 +284,18 @@ def patch_source(ctx: context.WorkContext, source_root_dir: pathlib.Path) -> Non
             )
 
 
+def _warn_for_old_patch(source_root_dir: pathlib.Path, version: Version) -> None:
+    # Get the existing version and req name as per the source_root_dir naming conventions
+    req_name = source_root_dir.name.rsplit("-", 1)[0]
+    existing_version = Version(source_root_dir.name.rsplit("-", 1)[1])
+
+    # Compare the version object to see if patches for older version exists
+    if version > existing_version:
```
dhellmann, line 293: We only want to emit these warnings if old patches exist and patches for the current version do not exist, because that might mean we need to update those patches to work with the newer version of the package.

src/fromager/sources.py
```diff
@@ -270,7 +270,10 @@ def unpack_source(
     return (next(iter(unpack_dir.glob("*"))), True)
 
 
-def patch_source(ctx: context.WorkContext, source_root_dir: pathlib.Path) -> None:
+def patch_source(
```
dhellmann, line 273: This function is invoked by some of our build overrides, and this change is backwards incompatible.

Could you make the version argument optional and only perform the check if a version is given? That would ensure we can update safely.

dhellmann (changes requested):

src/fromager/commands/build.py
rd4398: okay, sounds good

rd4398: Let me know when you merge https://github.com/python-wheel-build/fromager/pull/212 or should I merge it?

dhellmann: I think if you review and leave an approval it will merge.

src/fromager/sources.py
```diff
@@ -270,8 +270,12 @@ def unpack_source(
...
+    ctx: context.WorkContext, source_root_dir: pathlib.Path, version: Version = None
```
dhellmann, line 274:
```suggestion
    ctx: context.WorkContext, source_root_dir: pathlib.Path, version: Version | None = None
```

src/fromager/sources.py
```diff
@@ -281,6 +285,23 @@ def patch_source(ctx: context.WorkContext, source_root_dir: pathlib.Path) -> Non
             )
 
 
+def _warn_for_old_patch(
+    source_root_dir: pathlib.Path, path: str, version: Version
+) -> None:
+    if version:
```
dhellmann, line 291: This function will always be called with a valid version, right?

src/fromager/sources.py
```diff
@@ -281,6 +285,23 @@ def patch_source(ctx: context.WorkContext, source_root_dir: pathlib.Path) -> Non
             )
 
 
+def _warn_for_old_patch(
+    source_root_dir: pathlib.Path, path: str, version: Version
```
dhellmann, line 289: "path" doesn't really tell us what the argument is.

```suggestion
    source_root_dir: pathlib.Path, patch_filename: str, version: Version
```

src/fromager/sources.py
```diff
@@ -281,6 +285,23 @@ def patch_source(ctx: context.WorkContext, source_root_dir: pathlib.Path) -> Non
...
+) -> None:
+    if version:
+        # Get the existing version and req name as per the source_root_dir naming conventions
+        req_name = source_root_dir.name.rsplit("-", 1)[0]
+        existing_version = Version(source_root_dir.name.rsplit("-", 1)[1])
+        existing = req_name + "-" + str(existing_version)
+        new = req_name + "-" + str(version)
+
+        # Check to see if patches for older version exists
+        if version > existing_version and existing in path and new not in path:
```
dhellmann, line 299: Using "in" with paths is problematic, because the substrings don't honor the path separators so you can end up matching because one name is a subset of another name.

The patch files are expected to be in a subdirectory that matches the source directory where the patch will be applied. I think that means you don't need to do so much parsing. Something like this should do it.

```python
if patch_filename.parent.name != source_root_dir.name:
    logger.warning("ignoring patch for ...")
```

rd4398: Okay, will make changes!

src/fromager/sources.py
```diff
@@ -281,6 +287,20 @@ def patch_source(ctx: context.WorkContext, source_root_dir: pathlib.Path) -> Non
...
+    source_root_dir: pathlib.Path, patch_filename: pathlib.Path, version: Version
+) -> None:
+    # Get the existing version and req name as per the source_root_dir naming conventions
+    req_name = source_root_dir.name.rsplit("-", 1)[0]
+    existing_version = Version(source_root_dir.name.rsplit("-", 1)[1])
+
+    # Check to see if patches for older version exists
+    if patch_filename.parent.name == source_root_dir.name:
+        logger.warning(
+            f"Patch for older version {existing_version!s} of {req_name} exists"
```
dhellmann, line 300:
```suggestion
            f"Patches for version {existing_version!s} of {req_name} exist but will not be applied"
```

shubhbapna (commented):
lgtm!

shubhbapna (approved):

dhellmann (approved):

Issue #220, opened 2024-07-22
`resolver.find_matches` does not allow for resolution with `include_wheels=true` and `include_sdist=true`


Gregory-Pereira:
based on the way [the code](https://github.com/python-wheel-build/fromager/blob/a7333271a5f4ef1fda14e43c454dfee04036a02a/src/fromager/resolver.py#L237-L242) is written, it will check if `include_sdists` is true, and `continue` to passthrough, meaning that it will never check for both `include_sdists` and `include_wheels`.

PR #221, opened 2024-07-22
allow both include_sdists and include_wheels in resolution

Gregory-Pereira:
Resolves: #220 .

When working with bitsandbytes override, I found that our logic doesn't allow for resolving against both `include_sdists` and `include_wheels`.

/cc @dhellmann @tiran @shubhbapna

### DIFF
--- a/src/fromager/resolver.py
+++ b/src/fromager/resolver.py
@@ -234,10 +234,10 @@ class PyPIProvider(ExtrasProvider):
             if not self.constraints.is_satisfied_by(identifier, candidate.version):
                 continue
             # Only include sdists if we're asked to
-            if self.include_sdists and not candidate.is_sdist:
+            if candidate.is_sdist and not self.include_sdists:
                 continue
             # Only include wheels if we're asked to
-            if self.include_wheels and candidate.is_sdist:
+            if not candidate.is_sdist and not self.include_wheels:
                 continue
             candidates.append(candidate)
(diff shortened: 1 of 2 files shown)


src/fromager/resolver.py
```diff
@@ -234,10 +234,10 @@ def find_matches(self, identifier, requirements, incompatibilities):
             if not self.constraints.is_satisfied_by(identifier, candidate.version):
                 continue
             # Only include sdists if we're asked to
-            if self.include_sdists and not candidate.is_sdist:
+            if self.include_sdists and not candidate.is_sdist and not self.include_wheels:
```
dhellmann, line 237: I think we want to switch the order of the checks

```suggestion
            if candidate.is_sdist and not self.include_sdists:
```

tests/test_resolver.py
```diff
@@ -72,6 +75,35 @@ def test_provider_choose_sdist():
         assert str(candidate.version) == "1.3.2"
 
 
+def test_provider_choose_either_with_constraint():
```
dhellmann, line 78: We have some examples in this test file where we mock out the return value from the package server to avoid having to have network access to run the unit tests. That also makes the tests more reliable, since they don't rely on the remote service returning exactly the same answer each time.

dhellmann: Well, I see you're doing that below, just not in the method definition, so never mind.

dhellmann (approved):

PR #223, opened 2024-07-23
Do not vendor Rust test code for meson

tiran:
The directories `22 cargo subproject` and `25 cargo lock` contain broken `Cargo.toml`. Don't vendor Rust crates for test code. We ignore the code here because meson is both shipped stand-alone and vendored in NumPy 2.0.

### DIFF
--- a/src/fromager/vendor_rust.py
+++ b/src/fromager/vendor_rust.py
@@ -93,7 +93,15 @@ def vendor_rust(
     ``False``.
     """
     # check for Cargo.toml
-    manifests = list(project_dir.glob("**/Cargo.toml"))
+    manifests = []
+    for manifest in project_dir.glob("**/Cargo.toml"):
+        if any((d in manifest.parts) for d in ["22 cargo subproject", "25 cargo lock"]):
+            # HACK: meson 1.5.0 tests have broken Cargo.toml. Don't vendor
+            # test cases of meson and vendored copies of meson NumPy 2.0.1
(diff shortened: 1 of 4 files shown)


src/fromager/vendor_rust.py
```diff
@@ -93,7 +93,15 @@ def vendor_rust(
     ``False``.
     """
     # check for Cargo.toml
-    manifests = list(project_dir.glob("**/Cargo.toml"))
+    manifests = []
+    for manifest in project_dir.glob("**/Cargo.toml"):
+        if any((d in manifest.parts) for d in ["22 cargo subproject", "25 cargo lock"]):
```
dhellmann, line 98: Can we do something that's less tied to the specific package we're trying to build?

For example, patch these files in the upstream build? Or catch errors with the vendoring and ignore them by default in fromager? 

I opened #224 to track other ideas.

tiran: I have created https://github.com/python-wheel-build/fromager/pull/233 with a different approach. `vendor_rust` detects whether a project uses maturin or setuptools-rust. Meson and NumPy don't have a build dependency on any of the project and vendoring is skipped.

dhellmann: I like that better, it's less tightly coupled to specific packages and based more on general configuration.

dhellmann:
We merged #233 so I am going to close this PR.

Issue #213, opened 2024-07-20
include constraint info in comments in generated constraints.txt


dhellmann:
The constraints.txt file we produce includes comments about what depends on each package. It should also include constraint details that feed into fromager so a reader can understand if the original requirements were modified.

rd4398:
Can you please give an example of constraint info that should be included?

dhellmann:
One of the inputs to the bootstrap process is a constraints file that looks just like a requirements file. If we have one, the bootstrap process uses a [Constraints instance](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/constraints.py#L15) to look up whether a package has a constraint. That information is then included when the package is [added to the build order](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/context.py#L92).

The comment should include the contents of the "constraint" field in that build order entry.

PR #225, opened 2024-07-23
This commit adds constraints present in build order to constraints.txt

rd4398:
Fixes : https://github.com/python-wheel-build/fromager/issues/213

### DIFF
--- a/src/fromager/constraints.py
+++ b/src/fromager/constraints.py
@@ -66,4 +66,5 @@ def write_from_build_order(filename: pathlib.Path, build_stack: list[typing.Any]
                 f"-{dep_type}-> {req.name}({version})"
                 for dep_type, req, version in step["why"]
             )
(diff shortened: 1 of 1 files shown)


src/fromager/constraints.py
```diff
@@ -66,4 +66,6 @@ def write_from_build_order(filename: pathlib.Path, build_stack: list[typing.Any]
                 f"-{dep_type}-> {req.name}({version})"
                 for dep_type, req, version in step["why"]
             )
+            constraint = f' Constraints: {step["constraint"]}'
+            comment.join(constraint)
```
dhellmann, line 70: I don't think this is doing what you expect

```
>>> "a".join("b")
'b'
>>>
```

rd4398: Hmm, I have removed the `join()` call and added `constraints` to existing comment.

Issue #226, opened 2024-07-23
tracker: enforce typings and enable mypy checks in github actions


shubhbapna:
We have already been seeing type mismatch of between what is being passed to a function vs what the function parameter expects especially in the case of `version` where we have been passing strings as well as `Version` objects.

Enforcing types will help us sync everything and should prevent similar future issues. 

- [x] fix typings for src/fromager/clickext.py **(no errors detected)**
- [x] fix typings for src/fromager/constraints.py #227 
- [x] fix typings for src/fromager/context.py #229 
- [x] fix typings for src/fromager/dependencies.py #234 
- [x] fix typings for src/fromager/example_override.py **(no errors detected)**
- [x] fix typings for src/fromager/external_commands.py #234 
- [x] fix typings for src/fromager/extras_provider.py #273 
- [x] fix typings for src/fromager/finders.py **(callable function related errors only see tasks below)**
- [x] fix typings for src/fromager/hooks.py #245 
- [x] fix typings for src/fromager/\_\_main\_\_.py  #246 
- [x] fix typings for src/fromager/overrides.py #237 
- [x] fix typings for src/fromager/requirements_file.py #231 
- [x] fix typings for src/fromager/resolver.py #300
- [ ] fix typings for src/fromager/sdist.py 
- [x] fix typings for src/fromager/server.py #247
- [ ] fix typings for src/fromager/settings.py #257 
- [ ] fix typings for src/fromager/sources.py 
- [x] fix typings for src/fromager/tarballs.py #242
- [x] fix typings for src/fromager/vendor_rust.py #239 
- [x] fix typings for src/fromager/wheels.py #240
- [x] fix typings for src/fromager/commands/bootstrap.py #264 
- [x] fix typings for src/fromager/commands/build_order.py #265 
- [x] fix typings for src/fromager/commands/build.py #249
- [x] fix typings for src/fromager/commands/canonicalize.py #238 
- [x] fix typings for src/fromager/commands/download_sequence.py #238 
- [x] fix typings for src/fromager/commands/\_\_init\_\_.py **(no errors detected)**
- [x] fix typings for src/fromager/commands/list_overrides.py #238 
- [x] src/fromager/commands/server.py #238 
- [x] src/fromager/commands/step.py #248 
- [x] src/fromager/progress.py #297 

After all of the above are done:

[...]

PR #227, opened 2024-07-23
cleanup: fix typings for constraints.py

shubhbapna:
part of #226 

Fixes the following type errors:

```
src/fromager/constraints.py:19: error: Function is missing a return type annotation  [no-untyped-def]
src/fromager/constraints.py:22: error: Function is missing a return type annotation  [no-untyped-def]
src/fromager/constraints.py:54: error: Function is missing a return type annotation  [no-untyped-def]
src/fromager/constraints.py:54: error: Missing type parameters for generic type "dict"  [type-arg]
src/fromager/constraints.py:62: error: Function is missing a return type annotation  [no-untyped-def]
src/fromager/constraints.py:64: error: Argument 1 to "_organize_constraints" has incompatible type "list[Any]"; expected "dict[Any, Any]"  [arg-type]
```

### DIFF
--- a/src/fromager/constraints.py
+++ b/src/fromager/constraints.py
@@ -16,10 +16,10 @@ class Constraints:
     def __init__(self, data: dict[str, Requirement]):
         self._data = {canonicalize_name(n): v for n, v in data.items()}
 
-    def get_constraint(self, name: str):
+    def get_constraint(self, name: str) -> Requirement | None:
         return self._data.get(canonicalize_name(name))
 
-    def is_satisfied_by(self, pkg_name: str, version: Version):
+    def is_satisfied_by(self, pkg_name: str, version: Version) -> bool:
         constraint = self.get_constraint(pkg_name)
         if constraint:
             return version in constraint.specifier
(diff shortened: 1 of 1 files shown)


src/fromager/constraints.py
```diff
@@ -51,15 +51,17 @@ def load(filename: pathlib.Path | None) -> Constraints:
 # This is a helper function to find and write duplicates to start of the file
 # from constraints.txt Input: list of objects (self._build_stack in this case)
 # Returns: list of objects (processed_constraints in this case)
-def _organize_constraints(info: dict):
+def _organize_constraints(info: list[typing.Any]) -> list[typing.Any]:
```
tiran, line 54: ` list[typing.Any]` isn't a very useful type annotation. Is it a ` list[dict[str, typing.Any]]`?

shubhbapna: Yes that could describe it, but I was hoping to exact the exact type down later and use type alias for this cause this function accepts the `build_stack` which is a common data object across modules

tiran (approved):

Issue #222, opened 2024-07-23
Make fromager more resilient against network hickups


tiran:
Fromager does not recover when there is a temporary network hickup. Example:

```console
('Connection broken: IncompleteRead(16777216 bytes read, 86427880 more expected)', IncompleteRead(16777216 bytes read, 86427880 more expected))
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/urllib3/response.py", line 748, in _error_catcher
    yield
  File "/usr/local/lib/python3.11/site-packages/urllib3/response.py", line 894, in _raw_read
    raise IncompleteRead(self._fp_bytes_read, self.length_remaining)
urllib3.exceptions.IncompleteRead: IncompleteRead(16777216 bytes read, 86427880 more expected)
The above exception was the direct cause of the following exception:
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/requests/models.py", line 820, in generate
    yield from self.raw.stream(chunk_size, decode_content=True)
  File "/usr/local/lib/python3.11/site-packages/urllib3/response.py", line 1060, in stream
    data = self.read(amt=amt, decode_content=decode_content)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/urllib3/response.py", line 949, in read
    data = self._raw_read(amt)
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/urllib3/response.py", line 872, in _raw_read
    with self._error_catcher():
  File "/usr/lib64/python3.11/contextlib.py", line 158, in __exit__
    self.gen.throw(typ, value, traceback)
  File "/usr/local/lib/python3.11/site-packages/urllib3/response.py", line 772, in _error_catcher
    raise ProtocolError(arg, e) from e
urllib3.exceptions.ProtocolError: ('Connection broken: IncompleteRead(16777216 bytes read, 86427880 more expected)', IncompleteRead(16777216 bytes read, 86427880 more expected))
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/usr/local/lib/python3.11/site-packages/fromager/__main__.py", line 178, in invoke_main
    main(auto_envvar_prefix="FROMAGER")
  File "/usr/local/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
[...]

PR #228, opened 2024-07-23
Creating module for HTTP client with Retry approach to deal with network hiccups

rd4398:
Fixes: https://github.com/python-wheel-build/fromager/issues/222

Let me know if this looks ok. I will then proceed to make similar changes related to `requests.get()` that I encountered in `resolver.py`


tiran:
Nice work, @rd4398 

I would design this slightly different, so we have one central place to control, configure, and reuse a request session. A common session object also allows for session resumption and HTTP pipelining. The `WorkContext` object would be a logical place. It already holds configuration and state information.

```
class WorkContext:
    def __init__(...):
        ...
        self.requests = self._make_requests_session()
        ...

    def _make_requests_session(self) -> requests.Session:
        ...
```

then pass WorkContext to `download_url` and change `requests.get(...)`  to `ctx.requests.get`. This gives us one place to control download parameter.

How does `Retry` work? Does it resume a file download when it gets interrupted in the middle of an operation?

src/fromager/http_client.py
```diff
@@ -0,0 +1,36 @@
...
+It has been adopted from: https://www.raelldottin.com/2023/09/enhancing-resilience-in-your-python.html
+"""
+
+import requests
+from requests.adapters import HTTPAdapter, Retry
+
+
+class HTTPClient:
+    # Set up the retries, backoff_factor and status_forcelist
+    def __init__(self, base_url, retries=5, backoff_factor=0.1, status_forcelist=None):
```
dhellmann, line 12: If base_url isn't used until the call to get(), does it need to be an argument here?

src/fromager/http_client.py
```diff
@@ -0,0 +1,36 @@
...
+            status_forcelist=status_forcelist,
+        )
+
+        # Initialize the adapter object
+        adapter = HTTPAdapter(max_retries=retries)
+        self.session.mount("http://", adapter)
+        self.session.mount("https://", adapter)
+
+    # Get request function to make HTTP GET request using session and Retry
+    def get_request(self, endpoint="", **kwargs):
```
dhellmann, line 32: Maybe call this just get() so it mirrors the requests library API?

src/fromager/sources.py
```diff
@@ -199,12 +200,16 @@ def download_url(
...
-        r.raise_for_status()
-        with open(outfile, "wb") as f:
-            logger.debug(f"writing to {outfile}")
-            for chunk in r.iter_content(chunk_size=1024 * 1024):
-                f.write(chunk)
+
+    # Initialize the HTTPClient module
+    client = HTTPClient(url)
+    # Call HTTP GET for the URL
+    r = client.get_request()
```
dhellmann, line 207: We lost the `stream=True` argument there, which is important for making the loop below work. Without that argument requests downloads the whole file in one go, which means you have to be able to hold the whole thing in RAM. We're downloading very large wheels in some case, so we don't want to allocate that much RAM at one time.

rd4398:
> Nice work, @rd4398
> 
> I would design this slightly different, so we have one central place to control, configure, and reuse a request session. A common session object also allows for session resumption and HTTP pipelining. The `WorkContext` object would be a logical place. It already holds configuration and state information.
> 
> ```
> class WorkContext:
>     def __init__(...):
>         ...
>         self.requests = self._make_requests_session()
>         ...
> 
>     def _make_requests_session(self) -> requests.Session:
>         ...
> ```
> 
> then pass WorkContext to `download_url` and change `requests.get(...)` to `ctx.requests.get`. This gives us one place to control download parameter.
> 
> How does `Retry` work? Does it resume a file download when it gets interrupted in the middle of an operation?

No, it doesn't resume the download. It simply retries based on the `backoff_factor`. If the` backoff_factor = 0.1`, it will sleep for ` [0.05s, 0.1s, 0.2s, 0.4s, ...] between retries. ` depending on the `max_retries` parameter

src/fromager/sources.py
rd4398: Okay, I will make the changes and try to design this in a way @tiran suggested above

tiran:
> No, it doesn't resume the download. It simply retries based on the `backoff_factor`. If the` backoff_factor = 0.1`, it will sleep for `[0.05s, 0.1s, 0.2s, 0.4s, ...] between retries.` depending on the `max_retries` parameter

How is Retry going to help us to resume downloads? Either Retry helper must resume at exactly the same offset or we need to recognize resumption somehow and start download + write to file from scratch.

rd4398:
I made changes in the design as per suggestion above. Does it look ok? Should I proceed to make similar changes for other `request.get()` in `resolver.py`?

rd4398:
> > No, it doesn't resume the download. It simply retries based on the `backoff_factor`. If the` backoff_factor = 0.1`, it will sleep for `[0.05s, 0.1s, 0.2s, 0.4s, ...] between retries.` depending on the `max_retries` parameter
> 
> How is Retry going to help us to resume downloads? Either Retry helper must resume at exactly the same offset or we need to recognize resumption somehow and start download + write to file from scratch.

I researched about it but could not find anything. Even the documentation for [Retry](https://urllib3.readthedocs.io/en/latest/reference/urllib3.util.html#module-urllib3.util.retry) doesn't mention about it. 
Do you have any idea how we can implement this? 
(Also, not sure whether this should be included in the current PR or a new issue needs to be filed)

rd4398:
code has changed a lot since this PR which resulted in a messy rebase. Thus closing this PR.

PR #229, opened 2024-07-23
cleanup: fix typings for context.py

shubhbapna:
part of #226 

Fixes the following type errors:

```
src/fromager/context.py:54: error: Need type annotation for "_build_stack" (hint: "_build_stack: list[<type>] = ...")  [var-annotated]
src/fromager/context.py:55: error: Need type annotation for "_build_requirements" (hint: "_build_requirements: set[<type>] = ...")  [var-annotated]
src/fromager/context.py:62: error: Need type annotation for "_seen_requirements" (hint: "_seen_requirements: set[<type>] = ...")  [var-annotated]
src/fromager/context.py:69: error: List item 1 has incompatible type "str | None"; expected "str"  [list-item]
src/fromager/context.py:72: error: Missing type parameters for generic type "tuple"  [type-arg]
src/fromager/context.py:75: error: Function is missing a return type annotation  [no-untyped-def]
src/fromager/context.py:80: error: Missing type parameters for generic type "tuple"  [type-arg]
src/fromager/context.py:81: error: Incompatible return value type (got "bool", expected "tuple[Any, ...]")  [return-value]
src/fromager/context.py:83: error: Function is missing a return type annotation  [no-untyped-def]
src/fromager/context.py:88: error: Missing type parameters for generic type "list"  [type-arg]
src/fromager/context.py:125: error: Function is missing a return type annotation  [no-untyped-def]
src/fromager/context.py:125: note: Use "-> None" if function does not return a value
```

### DIFF
--- a/src/fromager/context.py
+++ b/src/fromager/context.py
@@ -1,10 +1,11 @@
 import json
 import logging
 import pathlib
+import typing
 from urllib.parse import urlparse
 
 from packaging.requirements import Requirement
-from packaging.utils import canonicalize_name
+from packaging.utils import NormalizedName, canonicalize_name
 from packaging.version import Version
 
 from . import constraints, settings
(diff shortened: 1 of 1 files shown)


src/fromager/context.py
```diff
@@ -51,46 +52,50 @@ def __init__(
...
+        self._build_stack: list[typing.Any] = []
+        self._build_requirements: set[tuple[NormalizedName, str]] = set()
 
         # Track requirements we've seen before so we don't resolve the
         # same dependencies over and over and so we can break cycles in
         # the dependency list. The key is the requirements spec, rather
         # than the package, in case we do have multiple rules for the same
         # package.
-        self._seen_requirements = set()
+        self._seen_requirements: set[tuple[NormalizedName, tuple[str, ...], str]] = (
```
tiran, line 63: It feels like `tuple[NormalizedName, tuple[str, ...], str]` should be a [type alias](https://docs.python.org/3/library/typing.html#type-aliases).

src/fromager/context.py
```diff
@@ -51,46 +52,50 @@ def __init__(
...
+    def has_been_seen(self, req: Requirement, version: Version) -> bool:
         return self._resolved_key(req, version) in self._seen_requirements
 
     def add_to_build_order(
         self,
         req_type: str,
         req: Requirement,
         version: Version,
-        why: list,
+        why: list[typing.Any],
```
tiran, line 93: `list[typing.Any]` isn't very useful. Does the list contain the same type of item all the time?

tiran (changes requested):

src/fromager/context.py
shubhbapna: I left the cleanup of long types as a separate task, in case if these type aliases are common across modules (the last one on the task list in the issue)

src/fromager/context.py
shubhbapna: I left the cleanup of `typing.Any` as a separate task cause these will most likely require type aliases (the last one on the task list in the issue)

src/fromager/context.py
tiran: Are you tracking these planned changes in the #226 ?

shubhbapna: Yep

tiran (approved):

PR #232, opened 2024-07-23
add command to produce rpm list from build-order

dhellmann:
Add `build-order rpms` to list the RPMs that look
like they are probably related to the dists in a
build-order file.

### DIFF
--- a/src/fromager/commands/build_order.py
+++ b/src/fromager/commands/build_order.py
@@ -8,7 +8,7 @@ import sys
 import click
 from packaging.requirements import Requirement
 
-from .. import clickext, constraints, overrides
+from .. import clickext, constraints, overrides, system_packages
 
 
 @click.group()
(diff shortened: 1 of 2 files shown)


tiran:
`dnf repoquery` uses the current list of enabled repos. We have several disabled repos that only get enabled for some steps, e.g. `EPEL`.

dhellmann:
> `dnf repoquery` uses the current list of enabled repos. We have several disabled repos that only get enabled for some steps, e.g. `EPEL`.

Is there a way to query those without enabling them?

tiran:
Do you want to query the available packages or the installed packages? There is always `rpm -qa`.

dhellmann:
> Do you want to query the available packages or the installed packages? There is always `rpm -qa`.

I'm interested in the installed and available packages. I think repoquery is showing both.

dhellmann:
I thought I'd written this code before, and I found it in the downstream tool where I put it, so we don't need this.

PR #233, opened 2024-07-24
Only vendor Rust when project uses setuptools-rust / maturin

tiran:
`vendor_rust` now inspects `pyproject.toml` and looks for `setuptools-rust` and `maturin` in `[build-system]requires`. When `pyproject.toml` does not exist or does not list a Rust build plugin as requirement, then Rust crates are not vendored.

The change indirectly addresses the issue with Meson's broken Rust crates in `test cases/rust`.

### DIFF
--- a/src/fromager/vendor_rust.py
+++ b/src/fromager/vendor_rust.py
@@ -10,7 +10,7 @@ import typing
 import toml
 from packaging.requirements import Requirement
 
-from . import external_commands
+from . import dependencies, external_commands
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


src/fromager/vendor_rust.py
```diff
@@ -84,24 +87,56 @@ def _cargo_config(project_dir: pathlib.Path):
...
+        return False
+
+    with pyproject_toml.open(encoding="utf-8") as f:
+        pyproject = toml.load(f)
+
+    try:
+        build_requires = pyproject["build-system"]["requires"]
+    except KeyError:
+        logger.debug(f"{req.name}: no [build-system]requires in pypyprojet.toml")
+        return False
```
dhellmann, line 108: A lot of the logic above reproduces https://github.com/python-wheel-build/fromager/blob/main/src/fromager/dependencies.py#L222. Maybe it makes sense to have it inline here anyway, to include the debug log messages?

dhellmann (commented):
Mostly good, one question about reuse.

src/fromager/vendor_rust.py
tiran: Valid point. I have rewritten the code to use the existing helpers.

dhellmann (approved):

PR #234, opened 2024-07-24
cleanup: fix typings for dependencies.py and external_commands.py

shubhbapna:
part of #226 

Fixes the following errors:

```
src/fromager/dependencies.py:33: error: Incompatible return value type (got "set[Requirement] | None", expected "set[Requirement]")  [return-value]
src/fromager/dependencies.py:73: error: Returning Any from function declared to return "Iterable[Requirement]"  [no-any-return]
src/fromager/dependencies.py:90: error: Incompatible return value type (got "set[Requirement] | None", expected "set[Requirement]")  [return-value]
src/fromager/dependencies.py:119: error: Returning Any from function declared to return "set[Requirement]"  [no-any-return]
src/fromager/dependencies.py:136: error: Incompatible return value type (got "set[Requirement] | None", expected "set[Requirement]")  [return-value]
src/fromager/dependencies.py:165: error: Returning Any from function declared to return "set[Requirement]"  [no-any-return]
src/fromager/dependencies.py:172: error: Argument 1 to "Wheel" has incompatible type "Path"; expected "str"  [arg-type]
src/fromager/dependencies.py:173: error: Argument 2 to "_filter_requirements" has incompatible type "Sequence[str]"; expected "Iterable[Requirement]"  [arg-type]
src/fromager/dependencies.py:207: error: Missing type parameters for generic type "dict"  [type-arg]
src/fromager/dependencies.py:222: error: Missing type parameters for generic type "dict"  [type-arg]
src/fromager/dependencies.py:239: error: Missing type parameters for generic type "dict"  [type-arg]
src/fromager/dependencies.py:243: error: Function is missing a type annotation  [no-untyped-def]
src/fromager/dependencies.py:262: error: Function is missing a return type annotation  [no-untyped-def]

src/fromager/external_commands.py:13: error: Missing type parameters for generic type "dict"  [type-arg]
```

fixes in dependencies.py needed fixes in external_commands.py.

Unfortunately we cannot upgrade pyproject_hooks to 1.1.0 to use their types since https://github.com/python-wheel-build/fromager/issues/226#issuecomment-2248693960
So the follow errors cannot be fixed: 
```
src/fromager/dependencies.py:119: error: Returning Any from function declared to return "Iterable[str]"  [no-any-return]
src/fromager/dependencies.py:165: error: Returning Any from function declared to return "Iterable[str]"  [no-any-return]
```

### DIFF
--- a/src/fromager/dependencies.py
+++ b/src/fromager/dependencies.py
@@ -51,7 +51,7 @@ def get_build_system_dependencies(
 
 def _filter_requirements(
     req: Requirement,
-    requirements: typing.Iterable[Requirement],
+    requirements: typing.Iterable[Requirement | str],
 ) -> set[Requirement]:
     requires = set()
     for r in requirements:
@@ -68,9 +68,9 @@ def default_get_build_system_dependencies(
     ctx: context.WorkContext,
     req: Requirement,
     sdist_root_dir: pathlib.Path,
-) -> typing.Iterable[Requirement]:
+) -> typing.Iterable[str]:
     pyproject_toml = get_pyproject_contents(sdist_root_dir)
-    return get_build_backend(pyproject_toml)["requires"]
+    return typing.cast(list[str], get_build_backend(pyproject_toml)["requires"])
 
 
 def get_build_backend_dependencies(
@@ -110,7 +110,7 @@ def default_get_build_backend_dependencies(
     ctx: context.WorkContext,
     req: Requirement,
     sdist_root_dir: pathlib.Path,
-) -> set[Requirement]:
+) -> typing.Iterable[str]:
     pyproject_toml = get_pyproject_contents(sdist_root_dir)
     extra_environ = overrides.extra_environ_for_pkg(ctx.envs_dir, req.name, ctx.variant)
     hook_caller = get_build_backend_hook_caller(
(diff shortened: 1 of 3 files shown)


src/fromager/dependencies.py
```diff
@@ -110,7 +110,7 @@ def default_get_build_backend_dependencies(
     ctx: context.WorkContext,
     req: Requirement,
     sdist_root_dir: pathlib.Path,
-) -> set[Requirement]:
+) -> typing.Iterable[str]:
```
tiran, line 113: Are you sure this is correct? `_filter_requirements` returns a set of Requirements.

shubhbapna: The return statement of this function is one of the hooks method: https://github.com/python-wheel-build/fromager/pull/234/files#diff-37a100b3c15c4653f7f126670e7e58c41fc41074761085dd21e375b905125b51L119

And the docs says that it returns a Sequence so I should probably change it to that: https://pyproject-hooks.readthedocs.io/en/latest/pyproject_hooks.html#pyproject_hooks.BuildBackendHookCaller.get_requires_for_build_editable

src/fromager/external_commands.py
```diff
@@ -2,15 +2,16 @@
...
 logger = logging.getLogger(__name__)
 
 
 # based on pyproject_hooks/_impl.py: quiet_subprocess_runner
 def run(
-    cmd: list[str],
+    cmd: typing.Sequence[str],
     cwd: str | None = None,
-    extra_environ: dict | None = None,
+    extra_environ: dict[str, typing.Any] | None = None,
```
tiran, line 14: That should be also `dict[str, str]`
```suggestion
    extra_environ: dict[str, str] | None = None,
```

tiran: I have added a TODO item to the ticket.

tiran (approved):

PR #239, opened 2024-07-24
cleanup: remove unused function from vendor_rust

shubhbapna:
part of #226 

fixes the following error:

```
src/fromager/vendor_rust.py:149: error: Function is missing a return type annotation  [no-untyped-def]
src/fromager/vendor_rust.py:149: note: Use "-> None" if function does not return a value
src/fromager/vendor_rust.py:176: error: Missing positional argument "project_dir" in call to "vendor_rust"  [call-arg]
src/fromager/vendor_rust.py:183: error: Call to untyped function "test" in typed context  [no-untyped-call]
```

I don't think we were using the `test` function anywhere and it was just put to quickly test this module manually

### DIFF
--- a/src/fromager/vendor_rust.py
+++ b/src/fromager/vendor_rust.py
@@ -1,4 +1,3 @@
-#!/usr/bin/python3
 """Vendor Rust crates into an sdist"""
 
 import json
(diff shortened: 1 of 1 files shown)


tiran (approved):
I used the test method while I was hacking on the module. I'm okay with removing it.

PR #240, opened 2024-07-24
cleanup: fix typings in wheels.py

shubhbapna:
part of #226 

fixes the following errors:

```
src/fromager/wheels.py:30: error: Call to untyped function "_createenv" in typed context  [no-untyped-call]
src/fromager/wheels.py:36: error: Function is missing a return type annotation  [no-untyped-def]
src/fromager/wheels.py:36: note: Use "-> None" if function does not return a value
src/fromager/wheels.py:42: error: List item 3 has incompatible type "Path"; expected "str"  [list-item]
src/fromager/wheels.py:54: error: Argument 1 to "run" has incompatible type "list[object]"; expected "list[str]"  [arg-type]
src/fromager/wheels.py:68: error: Argument "cwd" to "run" has incompatible type "Path"; expected "str | None"  [arg-type]
src/fromager/wheels.py:120: error: Missing type parameters for generic type "dict"  [type-arg]
```

### DIFF
--- a/src/fromager/wheels.py
+++ b/src/fromager/wheels.py
@@ -33,13 +33,13 @@ class BuildEnvironment:
     def python(self) -> pathlib.Path:
         return (self.path / "bin/python3").absolute()
 
-    def _createenv(self):
+    def _createenv(self) -> None:
         if self.path.exists():
             logger.info("reusing build environment in %s", self.path)
             return
 
         logger.debug("creating build environment in %s", self.path)
-        external_commands.run([sys.executable, "-m", "virtualenv", self.path])
+        external_commands.run([sys.executable, "-m", "virtualenv", str(self.path)])
         logger.info("created build environment in %s", self.path)
 
         req_filename = self.path / "requirements.txt"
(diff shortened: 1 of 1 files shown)


src/fromager/wheels.py
```diff
@@ -117,7 +117,7 @@ def build_wheel(
 def default_build_wheel(
     ctx: context.WorkContext,
     build_env: BuildEnvironment,
-    extra_environ: dict,
+    extra_environ: dict[str, typing.Any],
```
tiran, line 120: extra_environ is passed to functions that expect a `dict[str, str]` mapping.

```suggestion
    extra_environ: dict[str, str],
```

tiran (changes requested):
Looks ok except for `extra_environ` argument.

src/fromager/wheels.py
tiran: I have added a TODO item to the ticket.

shubhbapna: Yes I was planning to have type alias for all our environ maps. I wasn't sure if `str` was fine so just kept it any for now

PR #244, opened 2024-07-25
Fix mypy errors in settings.py

rd4398:
As part of: https://github.com/python-wheel-build/fromager/issues/226

### DIFF
--- a/src/fromager/settings.py
+++ b/src/fromager/settings.py
@@ -1,5 +1,6 @@
 import logging
 import pathlib
+import typing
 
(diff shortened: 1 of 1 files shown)


shubhbapna:
I would hold this for now for #207 cause a bunch of changes are coming in for settings.py

rd4398:
> I would hold this for now for #207 cause a bunch of changes are coming in for settings.py

Okay, I will convert it to draft PR

tiran (approved):

PR #245, opened 2024-07-25
cleanup: fix typings for hooks.py

shubhbapna:
part of #226 

fixes the following errors:

```
src/fromager/hooks.py:9: error: Need type annotation for "_mgrs" (hint: "_mgrs: dict[<type>, <type>] = ...")  [var-annotated]
src/fromager/hooks.py:28: error: Function is missing a return type annotation  [no-untyped-def]
src/fromager/hooks.py:36: error: Function is missing a return type annotation  [no-untyped-def]
```

### DIFF
--- a/src/fromager/hooks.py
+++ b/src/fromager/hooks.py
@@ -1,12 +1,13 @@
 import logging
 import pathlib
+import typing
 
 from packaging.requirements import Requirement
 from stevedore import extension, hook
 
 from fromager import context
 
-_mgrs = {}
+_mgrs: dict[str, hook.HookManager] = {}
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


src/fromager/hooks.py
```diff
@@ -29,7 +29,7 @@ def _die_on_plugin_load_failure(
     mgr: hook.HookManager,
     ep: extension.Extension,
     err: Exception,
-):
+) -> None:
```
tiran, line 32:
```suggestion
) -> typing.NoReturn:
```

shubhbapna: Ah is this to differentiate if we are actually returning None vs not doing any returns?

tiran: The function is raising an exception, so it never returns anything. https://docs.python.org/3/library/typing.html#typing.NoReturn

tiran (approved):

PR #248, opened 2024-07-25
cleanup: fix typings for command/step.py

shubhbapna:
part of #226 

fixes
```
src/fromager/commands/step.py:14: error: Function is missing a return type annotation  [no-untyped-def]
src/fromager/commands/step.py:14: note: Use "-> None" if function does not return a value
src/fromager/commands/step.py:24: error: Function is missing a return type annotation  [no-untyped-def]
src/fromager/commands/step.py:50: error: Function is missing a return type annotation  [no-untyped-def]
src/fromager/commands/step.py:64: error: Argument 3 to "find_sdist" has incompatible type "Version"; expected "str"  [arg-type]
src/fromager/commands/step.py:66: error: Need type annotation for "dir_contents" (hint: "dir_contents: list[<type>] = ...")  [var-annotated]
src/fromager/commands/step.py:81: error: Function is missing a return type annotation  [no-untyped-def]
src/fromager/commands/step.py:97: error: Argument 3 to "_find_source_root_dir" has incompatible type "Version"; expected "str"  [arg-type]
src/fromager/commands/step.py:98: error: Argument 3 to "BuildEnvironment" has incompatible type "None"; expected "Iterable[Requirement]"  [arg-type]
src/fromager/commands/step.py:127: error: Function is missing a return type annotation  [no-untyped-def]
src/fromager/commands/step.py:141: error: Argument 3 to "_find_source_root_dir" has incompatible type "Version"; expected "str"  [arg-type]
src/fromager/commands/step.py:149: error: Function is missing a return type annotation  [no-untyped-def]
src/fromager/commands/step.py:162: error: Argument 3 to "_find_source_root_dir" has incompatible type "Version"; expected "str"  [arg-type]
src/fromager/commands/step.py:163: error: Argument 3 to "BuildEnvironment" has incompatible type "None"; expected "Iterable[Requirement]"  [arg-type]
```

I didn't want to change the `finders.find_sdist` or `finders.find_source_dir` to accept `Version` instead of strings because I didn't want to break any downstream plugin. I will probably add that as a task in the issue


src/fromager/commands/step.py
```diff
@@ -109,9 +109,11 @@ def build_sdist(
 def _find_source_root_dir(
     work_dir: pathlib.Path,
     req: Requirement,
-    dist_version: str,
+    dist_version: Version,
 ) -> pathlib.Path:
-    source_root_dir = finders.find_source_dir(pathlib.Path(work_dir), req, dist_version)
+    source_root_dir = finders.find_source_dir(
+        pathlib.Path(work_dir), req, str(dist_version)
```
tiran, line 115: nit pick: work_dir is already a pathlib.Path:

```suggestion
        work_dir, req, str(dist_version)
```

tiran (changes requested):

tiran (approved):

PR #249, opened 2024-07-25
cleanup: fix typing in command/build.py

shubhbapna:
part of #226 

fixes
```
src/fromager/commands/build.py:19: error: Function is missing a return type annotation  [no-untyped-def]
src/fromager/commands/build.py:54: error: Function is missing a return type annotation  [no-untyped-def]
src/fromager/commands/build.py:110: error: Argument 3 to "BuildEnvironment" has incompatible type "None"; expected "Iterable[Requirement]"  [arg-type]
src/fromager/commands/build.py:134: error: Argument "dist_version" to "run_post_build_hooks" has incompatible type "Version"; expected "str"  [arg-type]
src/fromager/commands/build.py:136: error: Argument "wheel_filename" to "run_post_build_hooks" has incompatible type "Path | None"; expected "Path"  [arg-type]
src/fromager/commands/build.py:139: error: Incompatible return value type (got "Path | None", expected "Path")  [return-value]
```

### DIFF
--- a/src/fromager/commands/build.py
+++ b/src/fromager/commands/build.py
@@ -21,7 +21,7 @@ def build(
     dist_name: str,
     dist_version: Version,
     sdist_server_url: str,
-):
+) -> None:
     """Build a single version of a single wheel
 
     DIST_NAME is the name of a distribution
@@ -55,7 +55,7 @@ def build_sequence(
     wkctx: context.WorkContext,
     build_order_file: str,
     sdist_server_url: str,
-):
+) -> None:
     """Build a sequence of wheels in order
 
     BUILD_ORDER_FILE is the build-order.json files to build
@@ -131,7 +131,7 @@ def _build(
         ctx=wkctx,
         req=req,
         dist_name=dist_name,
-        dist_version=dist_version,
+        dist_version=str(dist_version),
         sdist_filename=sdist_filename,
         wheel_filename=wheel_filename,
     )


src/fromager/wheels.py
```diff
@@ -111,7 +111,7 @@ def build_wheel(
     wheels = list(ctx.wheels_build.glob("*.whl"))
     if wheels:
         return wheels[0]
-    return None
+    raise FileNotFoundError(f"Could not locate the built wheels for {req.name}")
```
shubhbapna, line 114: @dhellmann from what I noticed wherever this function is called, the caller already expects a non null return value. Do you think if we raise an error instead of returning null it would be fine? Or would you rather make this change later?

dhellmann: Let's do that separately so we can focus on test coverage for it.

shubhbapna: okay then I will fix the type errors from this separately

tiran, line 114: The signature of the exception class is a bit obscure. It takes an errno number, error string, and file or target name:

```suggestion
    raise FileNotFoundError(errno.ENOENT, "Could not locate the built wheels for", req.name)
```

```python
>>> import errno
>>> from packaging.requirements import Requirement
>>> req = Requirement("egg==0.1")
>>> raise FileNotFoundError(errno.ENOENT, "Could not locate the built wheels for", req.name)
FileNotFoundError: [Errno 2] Could not locate the built wheels for: 'egg'
```

tiran (changes requested):
Please change the exception call

shubhbapna:
> Please change the exception call

Will be adding the exception call in a separate PR. Have noted this down

tiran (approved):

PR #250, opened 2024-07-25
feat: add an option to skip building of wheel if it is exists on the wheel server

### LINKED ISSUE #235: add option to build-sequence to only build wheels that do not exist on the wheel server
Add an option to the build-sequence command to have it look up on the wheel server a wheel with the same name and version and skip the local build if the file exists. The default behavior should still be to build everything.

shubhbapna:
fixes #235 

It should be easy to toggle this option even via env variable by setting `FROMAGER_BUILD_SEQUENCE_SKIP_EXISTING=true` (see https://click.palletsprojects.com/en/8.1.x/options/#values-from-environment-variables:~:text=When%20using%20auto_envvar_prefix,is%20WEB_RUN_SERVER_HOST.)

I also moved `start_wheel_server` outside of the loop to avoid applying the `update_wheel_mirror` method twice per package

### DIFF
--- a/src/fromager/commands/build.py
+++ b/src/fromager/commands/build.py
@@ -43,6 +43,7 @@ def build(
     separately.
 
     """
+    server.start_wheel_server(wkctx)
     wheel_filename = _build(wkctx, dist_name, dist_version, sdist_server_url)
     print(wheel_filename)
 
@@ -50,11 +51,17 @@ def build(
 @click.command()
 @click.argument("build_order_file")
 @click.argument("sdist_server_url")
+@click.option(
+    "--skip-existing",
+    default=False,
+    is_flag=True,
+)
 @click.pass_obj
 def build_sequence(
     wkctx: context.WorkContext,
     build_order_file: str,
     sdist_server_url: str,
+    skip_existing: bool,
 ) -> None:
     """Build a sequence of wheels in order
 
@@ -66,12 +73,25 @@ def build_sequence(
     the build order file.
 
     """
+    server.start_wheel_server(wkctx)
+
     with open(build_order_file, "r") as f:
         for entry in progress.progress(json.load(f)):
             dist_name = entry["dist"]
             dist_version = Version(entry["version"])
+
+            if skip_existing and _is_wheel_built(wkctx, dist_name, dist_version):
+                logger.info(
+                    "%s: skipping building wheels for %s==%s since it already exists",
+                    dist_name,
+                    dist_name,
+                    dist_version,
+                )
+                continue
+
             logger.info("%s: building %s==%s", dist_name, dist_name, dist_version)
             wheel_filename = _build(wkctx, dist_name, dist_version, sdist_server_url)
+
             server.update_wheel_mirror(wkctx)
             # After we update the wheel mirror, the built file has
             # moved to a new directory.
@@ -85,8 +105,6 @@ def _build(
     dist_version: Version,
     sdist_server_url: str,
 ) -> pathlib.Path:
-    server.start_wheel_server(wkctx)
-
     req = Requirement(f"{dist_name}=={dist_version}")
 
     # Download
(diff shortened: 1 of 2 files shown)


Gregory-Pereira (commented):
Looks good, but going to leave formal approval to the experts
/lgtm

src/fromager/commands/build.py
```diff
@@ -137,3 +153,22 @@ def _build(
...
+) -> pathlib.Path | None:
+    req = Requirement(f"{dist_name}=={dist_version}")
+
+    try:
+        logger.info(f"{req.name}: checking if {req} was already built")
+        wheel_filename, _, _ = sdist.download_wheel(
+            wkctx,
+            req,
+            wkctx.wheels_build,
+            [wkctx.wheel_server_url],
```
dhellmann, line 169: I think we're going to need a different CLI flag for this case.

IIRC, when `--wheel-server-url` is set, then `start_wheel_server()` doesn't start a server. But we need the local server for most things because we have to assume we're building something that doesn't exist on the remote server. So for this case we're going to need to have 2 different wheel server URLs, for different purposes.

shubhbapna: Right now if we just pass a remote server to `--wheel-server-url` would it still build the package irrespective of whether it is there on the server or not?

dhellmann: Yes, today it always builds everything. The `--wheel-server-url` is used to find the build requirements. So if you're building a package that needs `setuptools` to make a wheel, fromager tells pip to download setuptools from whatever wheel server it knows, either the built-in one or the one provided as an argument on the command line.

shubhbapna: Right now for build sequence, we go through each item in the build order file (irrespective of the type of requirement it is), download the sources, prepare the sources and then build the wheel. The download and prepare source part should be applicable to something like `setuptools` as well. But when we come to building the actual wheel, will pip simply download `setuptools` from the wheel server instead of building it from the prepared sources because of this option: https://github.com/python-wheel-build/fromager/blob/main/src/fromager/wheels.py#L157 ?

dhellmann: Yes, that's right. When the build environment is set up, it [uses the context to get the wheel server](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/wheels.py#L63) and [that code](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/context.py#L68) uses the wheel server passed on the CLI or the one started by fromager if no value is given on the CLI.

dhellmann: The code you linked on line 157 should probably be updated to use the same approach as is used in creating the build environment.

shubhbapna: Doesn't look like we actually every execute [this command](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/wheels.py#L63) for build sequence since we initialize the build environment as [BuildEnvironment(wkctx, source_root_dir.parent, None)](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/commands/build.py#L111) and pass `build_requirements` as None. So in `_create_env` we return before we can execute pip install: https://github.com/python-wheel-build/fromager/blob/main/src/fromager/wheels.py#L51

dhellmann: The prepare_build_environment() call doesn't return the BuildEnvironment, so we have to re-instantiate it. We know it exists, so we just pass None as the requirements and it reuses the existing environment that was created [over here](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/sdist.py#L468).

The APIs of those functions should probably be changed so that prepare_build_environment returns the environment it creates, but I'd have to look at what other callers of that function do with the existing return value.

dhellmann: I was confused, we don't need a separate option because building only builds 1 thing at a time and uses the wheel server for its dependencies anyway.

shubhbapna:
As discussed, i have renamed the option to match what twine uses: https://twine.readthedocs.io/en/latest/#twine-upload

Also made the resolve_wheel function public

src/fromager/commands/build.py
```diff
@@ -66,12 +73,19 @@ def build_sequence(
...
 
     """
+    server.start_wheel_server(wkctx)
+
     with open(build_order_file, "r") as f:
         for entry in progress.progress(json.load(f)):
             dist_name = entry["dist"]
             dist_version = Version(entry["version"])
+
+            if skip_existing and _is_wheel_built(wkctx, dist_name, dist_version):
```
dhellmann, line 83: It would be good to have an INFO log message here saying that the build is being skipped because the wheel exists.

dhellmann (commented):
One small change

Issue #217, opened 2024-07-21
generate CLI documentation


dhellmann:
We should produce automatically generated documentation for all commands and CLI options.

rd4398:
Do we want to use something like https://pypi.org/project/md-click/?
Or it is to be done without using external libraries?

dhellmann:
If there are tools that will do it, that's fine.

We should consider how this would fit into an overall documentation build, too. I usually use Sphinx for python libs so I can publish on readthedocs.org, but that's not a hard requirement. If we don't publish there we need to figure out how we will publish.

rd4398:
Okay, I will research a bit more then

PR #253, opened 2024-07-26
Generating CLI documentation

rd4398:
Fixes: https://github.com/python-wheel-build/fromager/issues/217

References: 
https://sphinx-click.readthedocs.io/en/latest/examples/commands/
https://www.sphinx-doc.org/en/master/usage/quickstart.html
https://stackoverflow.com/questions/53668052/sphinx-cannot-find-my-python-files-says-no-module-named

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ __pycache__/
 /e2e/pyo3_test/target
 /e2e/pyo3_test/vendor
 /overrides/
+/docs/_build/
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,0 +1,30 @@
+from importlib import metadata
+
+# Configuration file for the Sphinx documentation builder.
+#
+# For the full list of built-in configuration values, see the documentation:
+# https://www.sphinx-doc.org/en/master/usage/configuration.html
+
+# -- Project information -----------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
+
+project = "Fromager"
+copyright = "2024, Doug Hellmann"
+author = "Doug Hellmann"
+release = metadata.version("fromager")
+
+# -- General configuration ---------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#general-configuration
+
+extensions = ["sphinx_click"]
+
+templates_path = ['_templates']
+exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
+
+language = 'English'
+
+# -- Options for HTML output -------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#options-for-html-output
+
+html_theme = 'alabaster'
+html_static_path = ['_static']
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -1,0 +1,18 @@
+.. Fromager documentation master file, created by
+   sphinx-quickstart on Fri Jul 26 11:06:15 2024.
+   You can adapt this file completely to your liking, but it should at least
+   contain the root `toctree` directive.
+
+Fromager documentation
+======================
+
+Add your content using ``reStructuredText`` syntax. See the
+`reStructuredText <https://www.sphinx-doc.org/en/master/usage/restructuredtext/index.html>`_
+documentation for details.
+
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Contents:
+
+   cli.rst
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,0 +1,2 @@
+sphinx
+sphinx-click
--- a/tox.ini
+++ b/tox.ini
@@ -59,3 +59,11 @@ commands=
     python -m build
     twine check dist/*.tar.gz dist/*.whl
     check-python-versions --only pyproject.toml,.github/workflows/test.yml
+
+[testenv:docs]
+description = sphinx docs
+basepython = python3.11
+deps =
+    -r docs/requirements.txt
+commands =
+    sphinx-build -M html docs docs/_build -j auto --keep-going {posargs:--fail-on-warning --fresh-env -n}
--- a/.github/workflows/check.yaml
+++ b/.github/workflows/check.yaml
@@ -85,3 +85,24 @@ jobs:
         #   files: 'docs/**/*.md'
         #   ignore: node_modules
         #   version: 0.28.1
+
+  docs:
+    name: readthedocs
+    runs-on: ubuntu-latest
+    if: ${{ !startsWith(github.ref, 'refs/tags') }}
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"  # minimum supported lang version
+
+      - name: Install dependencies
+        run: python -m pip install tox
+
+      - name: Run
+        run: tox -e docs
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,19 @@
+version: 2
+
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.11"
+
+sphinx:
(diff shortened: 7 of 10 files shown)


docs/index.rst
```diff
@@ -0,0 +1,20 @@
...
+Add your content using ``reStructuredText`` syntax. See the
+`reStructuredText <https://www.sphinx-doc.org/en/master/usage/restructuredtext/index.html>`_
+documentation for details.
+
+
+.. toctree::
+   :maxdepth: 2
+   :caption: Contents:
+
+.. click:: src.fomager.commands.bootstrap:bootstrap
```
tiran, line 18: The instruction expects a Python import name to the root click group. For Fromager it's `from fromager.__main__ import main`:

```suggestion
.. click:: fromager.__main__:main
```

tiran (changes requested):
It's a good idea and a good start. I have done the same for InstructLab a while ago,

Next steps:

- move the click doc to a separate page, https://github.com/instructlab/instructlab/blob/main/docs/ilab.rst
- configure the click extension, https://github.com/instructlab/instructlab/blob/main/docs/conf.py
- create a requirements.txt in the docs folder, https://github.com/instructlab/instructlab/blob/main/docs/requirements.txt
- create a tox target to build the docs: https://github.com/instructlab/instructlab/blob/main/tox.ini#L124-L134

docs/index.rst
rd4398: I changed this and now I get the following error:

```
/home/rdevasth/fromager/docs/cli.rst:1: ERROR: Failed to import "main" from "fromager.__main__". The following exception was raised:
  File "/home/rdevasth/fromager/venv/lib/python3.11/site-packages/sphinx_click/ext.py", line 444, in _load_module
    mod = __import__(module_name, globals(), locals(), [attr_name])
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'fromager'
```

rd4398:
> It's a good idea and a good start. I have done the same for InstructLab a while ago,
> 
> Next steps:
> 
> * move the click doc to a separate page, https://github.com/instructlab/instructlab/blob/main/docs/ilab.rst
> * configure the click extension, https://github.com/instructlab/instructlab/blob/main/docs/conf.py
> * create a requirements.txt in the docs folder, https://github.com/instructlab/instructlab/blob/main/docs/requirements.txt
> * create a tox target to build the docs: https://github.com/instructlab/instructlab/blob/main/tox.ini#L124-L134

I did first three steps but still I haven't been able to generate the documentation when I run `make html`
I also get a ModuleNotFound error that I mentioned above. I have pushed my latest attempt. Am I still missing something?

docs/index.rst
dhellmann: Do you have fromager installed? Sphinx is going to try to import the source code to generate those docs.

One convenient way to do that is to use tox instead of make. Tox lets you define a virtualenv with dependencies and then a set of commands to run. [This](https://github.com/sphinx-contrib/spelling/blob/main/tox.ini#L72) is what I use for generating documentation for another library that I maintain, for example.

rd4398:
cli
```
.. click:: fromager.__main__:main
  :prog: bootstrap
  :nested: full
```
Not sure whether `:prog: ` should be `bootstrap` since it looks like above. Any idea what it should be?

tiran:
> Not sure whether `:prog: ` should be `bootstrap` since it looks like above. Any idea what it should be?

`prog` should be the name of the script. In our case it is `fromager`.

rd4398:
Ah, okay! Changing that and adding `readthedocs.yaml` file as well

rd4398:
Let me know if any changes are required!

tox.ini
```diff
@@ -65,4 +65,12 @@ deps=
...
\ No newline at end of file
+    check-python-versions --only pyproject.toml,.github/workflows/test.yml
+
+[testenv:docs]
+description = sphinx docs
+basepython = python3.11
+deps =
+    -r docs/requirements.txt
+commands =
+    sphinx-build -M html docs docs/_build -j auto --keep-going {posargs:--fail-on-warning --fresh-env -n}
```
dhellmann, line 76: The `--fail-on-warning` flag makes this PR fail to build for me locally because there is no `docs/_static` directory. You could create that directory with an empty file named `.gitkeep` to resolve that.

dhellmann:
This looks good to me. @tiran had asked for a change, I'll wait for him to re-review before merging.

docs/conf.py
```diff
@@ -0,0 +1,28 @@
...
+# For the full list of built-in configuration values, see the documentation:
+# https://www.sphinx-doc.org/en/master/usage/configuration.html
+
+# -- Project information -----------------------------------------------------
+# https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
+
+project = 'Fromager'
+copyright = '2024, Doug Hellmann'
+author = 'Doug Hellmann'
+release = '0.21.3'
```
tiran, line 12: Let's use https://docs.python.org/3/library/importlib.metadata.html to get the version from the package metadata.

.gitignore
```diff
@@ -24,3 +24,4 @@ __pycache__/
 /e2e/pyo3_test/target
 /e2e/pyo3_test/vendor
 /overrides/
+/docs/_build/
```
tiran, line 27: Missing newline at EOF

docs/requirements.txt
```diff
@@ -0,0 +1,3 @@
+sphinx
+sphinx-click
+fromager
```
tiran, line 3: Missing newline at EOF
You also don't need to include the package as dependency. It's implied.

readthedocs.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+# See https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html
+python:
+  install:
+    - requirements: docs/requirements.txt
+    - method: pip
+      path: .
+
+submodules:
+  include: all
```
tiran, line 19: The file has to be named `.readthedocs.yaml` with a leading dot.

Missing newline at EOF

tox.ini
```diff
@@ -59,3 +59,11 @@ commands=
...
     twine check dist/*.tar.gz dist/*.whl
     check-python-versions --only pyproject.toml,.github/workflows/test.yml
+
+[testenv:docs]
+description = sphinx docs
+basepython = python3.11
+deps =
+    -r docs/requirements.txt
+commands =
+    sphinx-build -M html docs docs/_build -j auto --keep-going {posargs:--fail-on-warning --fresh-env -n}
```
tiran, line 69: Missing newline at EOF

tiran (changes requested):

docs/conf.py
rd4398: I tried using it but Sphinx is detecting it as function for some reason and I get the error:
`TypeError: expected string or bytes-like object, got 'function'`

Code:
```
from importlib.metadata import version
project = "Fromager"
copyright = "2024, Doug Hellmann"
author = "Doug Hellmann"
release = version("fromager")

```

tiran: I bet something is looking at `version` attribute of the module

do `from importlib import metadata`

rd4398:
I have made the suggested changes!
Let me know if any more changes are required.
Also, @tiran is there any reason to leave a blank line at EOF for some files you mentioned above? Just curious about why do we do that.

rd4398:
Not sure why one check is getting cancelled only after squishing commits :(

dhellmann:
I'm happy with this version. If we spot any other issues, let's plan to submit follow-ups after we have the doc publishing configured.

dhellmann:
> is there any reason to leave a blank line at EOF for some files you mentioned above? Just curious about why do we do that.

If we all configure our editors the same way it cuts down on spurious reformatting changes when we submit PRs. Having a blank line at the end of a file means if you `cat` it then it comes out with your command prompt on the next line instead of joined up with the last line of the file.

tiran (approved):

PR #255, opened 2024-07-26
feat: apply generic patches for all versions of packages

### LINKED ISSUE #230: Feature: Create overrides/patches/<package> directories
The overrides/patches directory contains patches that are applied to specific versions of packages.  For example, there exists a torch-2.3.0 and a torch-2.3.1.  This directory layout requires a new package-version directory be created every time the package is updated.  In many cases the patches may be simply updated and copied from one package-version directory to the next, and in some cases new patches are required.

In most cases the patches, can be simply copied from one directory to the next -- that is, there isn't any reason for a new package-version directory to be created.  We could simply create a general package directory, with patches, that could be applied to top-of-tree and only create a package-version directory if the patches changed.

This would minimize efforts when packages are updated.

shubhbapna:
fixes #230 

The `version` parameter was not being used in the `_warn_for_old_patch` function and so has been removed after talking to @rd4398. it was introduced to the `patch_source` function only for it to be passed to the warn function and so can be removed safely

### DIFF
--- a/src/fromager/sources.py
+++ b/src/fromager/sources.py
@@ -273,22 +273,35 @@ def unpack_source(
 def patch_source(
     ctx: context.WorkContext,
     source_root_dir: pathlib.Path,
-    version: Version | None = None,
+    req: Requirement,
 ) -> None:
-    for p in overrides.patches_for_source_dir(ctx.patches_dir, source_root_dir.name):
-        if version:
-            _warn_for_old_patch(source_root_dir, p, version)
-        logger.info("applying patch file %s to %s", p, source_root_dir)
-        with open(p, "r") as f:
-            subprocess.check_call(
-                ["patch", "-p1"],
-                stdin=f,
-                cwd=source_root_dir,
-            )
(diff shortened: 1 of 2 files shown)


rd4398 (commented):
Looks good to me! 
Will leave it to Doug for final review and approval

src/fromager/sources.py
```diff
@@ -273,22 +273,35 @@ def unpack_source(
...
+            ctx.patches_dir, source_root_dir.name
+        ):
+            _warn_for_old_patch(source_root_dir, p)
+            _apply_patch(p, source_root_dir)
+
+
+def _apply_patch(patch: pathlib.Path, source_root_dir: pathlib.Path):
+    logger.info("applying patch file %s to %s", patch, source_root_dir)
+    with open(patch, "r") as f:
+        subprocess.check_call(
```
dhellmann, line 296: We probably want to use external_command() for this. I'll file a separate ticket to fix that.

dhellmann (approved):

PR #257, opened 2024-07-27
Use Pydantic to validate and manage settings

tiran:
Pydantic provides a convenient way to validate data and transform it into type-safe Python objects with defaults.

### DIFF
--- a/src/fromager/settings_model.py
+++ b/src/fromager/settings_model.py
@@ -1,0 +1,136 @@
+import typing
+
+import yaml
+from packaging.utils import NormalizedName, canonicalize_name
+from pydantic import (
+    BaseModel,
+    BeforeValidator,
+    Field,
+    HttpUrl,
+    ValidationInfo,
+    field_validator,
+)
+
+from . import overrides
+
+# build variant
+Variant = typing.NewType("Variant", str)
+
+# Package validates and transforms name to canonicalized name
+Package = typing.Annotated[
+    NormalizedName,
+    BeforeValidator(lambda pkg: canonicalize_name(pkg, validate=True)),
+]
+
+# Set of packages, maps None to empty set
+PackageSet = typing.Annotated[
+    set[Package],
+    BeforeValidator(lambda v: v if v is not None else set()),
+]
+
+# URL with templating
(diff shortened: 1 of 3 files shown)


src/fromager/settings_model.py
```diff
@@ -0,0 +1,136 @@
...
+
+from . import overrides
+
+# build variant
+Variant = typing.NewType("Variant", str)
+
+# Package validates and transforms name to canonicalized name
+Package = typing.Annotated[
+    NormalizedName,
+    BeforeValidator(lambda pkg: canonicalize_name(pkg, validate=True)),
```
dhellmann, line 22: To be backwards compatible, we need to use `overrides.pkgname_to_override_module()`.

src/fromager/settings_model.py
```diff
@@ -0,0 +1,136 @@
...
+    pre_built: dict[Variant, PackageSet] = Field(default_factory=dict)
+    """Mark packages as pre-built"""
+
+    @field_validator("packages", "pre_built", mode="before")
+    @classmethod
+    def before_none_to_default(cls, v: typing.Any, info: ValidationInfo) -> typing.Any:
+        """Pre-hook to map None to default factory result"""
+        return _map_none_to_default(cls, v, info)
+
+    def get_package(
```
dhellmann, line 115: Maybe this should be "get_build_info"?

dhellmann (commented):
I like this direction! I had a couple of small comments, but I'll wait for a more complete review until you remove the draft flag.

src/fromager/settings_model.py
```diff
@@ -0,0 +1,136 @@
...
+PackageSet = typing.Annotated[
+    set[Package],
+    BeforeValidator(lambda v: v if v is not None else set()),
+]
+
+# URL with templating
+UrlTemplate = typing.NewType("UrlTemplate", HttpUrl)
+
+
+def _map_none_to_default(cls, v: typing.Any, info: ValidationInfo) -> typing.Any:
```
shubhbapna, line 35: instead of `typing.Any` could we use a generic type?

shubhbapna (commented):
Looks great! we can try porting the remaining methods and trigger the tests?

PR #258, opened 2024-07-27
only write installation requirements to constraints

dhellmann:
Keep a complete graph of all of the dependencies and their reasons so we
can use that to find only installation dependencies of the root items to
include in the constraints file.

Remove tests that required the generated constraints file to match the
build-order file, since that no longer applies.


dhellmann:
This version still does not eliminate duplicates, and our current requirements produces a duplicate constraint.

There will be some clean-up to remove the command for converting a build-order file to a constraints file. It should probably work on a graph file instead of a build-order file. I'd like to deal with that in a separate PR, since that command is not critical.

rd4398 (commented):
Looks good to me!
I will wait for one more set of eyes to review and then approve

src/fromager/context.py
```diff
@@ -54,6 +54,7 @@ def __init__(
         # be built in order.
         self._build_stack: list[typing.Any] = []
         self._build_requirements: set[tuple[NormalizedName, str]] = set()
+        self.all_edges: dict[NormalizedName, tuple(str, NormalizedName, Version)] = {}
```
shubhbapna, line 57:
```suggestion
        self.all_edges: dict[NormalizedName, list[tuple(str, NormalizedName, Version)]] = {}
```
Looking at `update_dependency_graph` method, the type of value seems to be a list

src/fromager/context.py
```diff
@@ -123,9 +124,39 @@ def add_to_build_order(
...
+        logger.debug(
+            "recording %s dependency %s -> %s %s",
+            req_type,
+            parent_req.name if parent_req else "(toplevel)",
+            req.name,
+            version,
         )
+        # If we have no parent requirement, we are processing an installation
+        # requirement even though it is called "toplevel" elsewhere. So, change
+        # the type. Also set the parent_ßkey to an empty string as a unique value
```
shubhbapna, line 147:
```suggestion
        # the type. Also set the parent_key to an empty string as a unique value
```

src/fromager/context.py
```diff
@@ -123,9 +124,39 @@ def add_to_build_order(
...
+        # that can't be mistaken for a real package.
+        if parent_req is None:
+            parent_key = ""
+            req_type = "install"
+        else:
+            parent_key = canonicalize_name(parent_req.name)
+        parent_edges = self.all_edges.setdefault(parent_key, [])
```
shubhbapna, line 154: should we move this to initialization and instead use `defaultdict(list)`?

src/fromager/context.py
```diff
@@ -123,9 +124,39 @@ def add_to_build_order(
...
+        parent_edges.append((req_type, canonicalize_name(req.name), str(version)))
+
+        graph_filename = self.work_dir / "graph.json"
+        with open(graph_filename, "w") as f:
+            json.dump(self.all_edges, f, indent=2)
```
shubhbapna, line 159: instead of writing to file each time we want to update the graph, can we do it seperately?

src/fromager/commands/bootstrap.py
```diff
@@ -72,3 +73,22 @@ def bootstrap(
...
+    logger.info(f"writing installation dependencies to {constraints_filename}")
+    with open(wkctx.work_dir / "constraints.txt", "w") as f:
+        for name, version in sorted(
+            set(_installation_dependencies(wkctx.all_edges, ""))
+        ):
+            f.write(f"{name}=={version}\n")
+
+
+def _installation_dependencies(
+    all_edges: dict[NormalizedName | str, tuple[str, NormalizedName, str]],
```
shubhbapna, line 87:
```suggestion
    all_edges: dict[NormalizedName, list[tuple[str, NormalizedName, Version]]],
```
Syncing up the type with `all_edges` defined in context.py. It might be helpful to define a reusable type alias for the edges/graph in this case

src/fromager/commands/bootstrap.py
```diff
@@ -72,3 +73,22 @@ def bootstrap(
...
+    name: NormalizedName | str,
+) -> typing.Iterable[tuple[NormalizedName, str]]:
```
shubhbapna, line 89:
```suggestion
    name: NormalizedName,
) -> typing.Iterable[tuple[NormalizedName, Version]]:
```

src/fromager/context.py
```diff
@@ -123,9 +124,39 @@ def add_to_build_order(
...
+            parent_req.name if parent_req else "(toplevel)",
+            req.name,
+            version,
         )
+        # If we have no parent requirement, we are processing an installation
+        # requirement even though it is called "toplevel" elsewhere. So, change
+        # the type. Also set the parent_ßkey to an empty string as a unique value
+        # that can't be mistaken for a real package.
+        if parent_req is None:
+            parent_key = ""
```
shubhbapna, line 150: We should define a constant something like `ROOT = ""` or `TOP_LEVEL = ""` since in `bootstrap.py` file the first call to `_installation_dependencies` is a bit confusing in terms of readability as we start with `""`

shubhbapna (changes requested):
This is great! I think in the long term moving to a graph structure will be quite helpful. We can run topological sort on these dependencies to figure out mutually exclusive dependencies that can be built in parallel and speed up the build process

src/fromager/commands/bootstrap.py
dhellmann: I kept "str" for the key type because one key is `""` and that's not really a NormalizedName. Maybe that's not necessary?

You're right about the list, I'll make that update.

src/fromager/context.py
dhellmann: Good catch, thanks!

src/fromager/context.py
dhellmann: Good idea, done.

src/fromager/context.py
dhellmann: I wasn't sure how to express that in the type. Maybe it's still just a dict?

src/fromager/context.py
dhellmann: We do the same for the build order. The idea is to keep track of what we've produced so far, so it can be analyzed if someone has to debug a build. Given the sorts of things involved in the build, writing the file isn't all that expensive.

src/fromager/commands/bootstrap.py
dhellmann: I changed the type of the value to Version

src/fromager/context.py
shubhbapna: yes I think it is still a dict

src/fromager/context.py
shubhbapna: make sense. maybe instead of opening it and closing it every time, is opening it once globally better? anyways as you said it is a micro-optimization compared to the things we run in the build

shubhbapna (changes requested):
Need an update here as well https://github.com/python-wheel-build/fromager/blob/main/src/fromager/commands/build_order.py#L340

src/fromager/context.py
tiran: It's a dict:
```python
>>> collections.defaultdict.__mro__
(<class 'collections.defaultdict'>, <class 'dict'>, <class 'object'>)
```

src/fromager/commands/bootstrap.py
```diff
@@ -72,3 +73,28 @@ def bootstrap(
...
+            f.write(f"{name}=={version}\n")
+
+
+def _installation_dependencies(
+    # The keys of all_edges will be either a validate package name or an empty
+    # string.
+    all_edges: context.BuildRequirements,
+    name: NormalizedName | str,
+) -> typing.Iterable[tuple[NormalizedName, str]]:
+    for edge in all_edges.get(name, []):
```
tiran, line 96: Are the values always tuples with three items? If yes, then I would find it more readable to write the loop as `for a, b, c in all_edges.get(name, [])` with proper names for a, b, and c.

src/fromager/context.py
dhellmann: Done

src/fromager/commands/bootstrap.py
dhellmann: Done

dhellmann:
> Need an update here as well https://github.com/python-wheel-build/fromager/blob/main/src/fromager/commands/build_order.py#L340

I'm going to remove the command, for now. We'll need a "graph to constraints" command to replace it, but that's not a high priority.

dhellmann:
I think I've addressed all review comments in the latest draft.

src/fromager/commands/bootstrap.py
```diff
@@ -72,3 +74,28 @@ def bootstrap(
...
+            )
+        ):
+            f.write(f"{name}=={version}\n")
+
+
+def _installation_dependencies(
+    # The keys of all_edges will be either a validate package name or an empty
+    # string.
+    all_edges: context.BuildRequirements,
+    name: NormalizedName | str,
```
shubhbapna, line 95: Sorry last nit pic since everything is NormalizedName now we can go ahead remove `str`
```suggestion
    name: NormalizedName
```

shubhbapna (commented):
Everything looks good! Just need to resolve conflicts

src/fromager/commands/bootstrap.py
dhellmann: Done

shubhbapna (approved):

PR #259, opened 2024-07-28
add GenericProvider for resolving versions

dhellmann:
Refactor existing resolvers to define base classes
to reduce duplication and make it easier to create
new providers.

Add a GenericProvider class that accepts a
callable as argument and picks versions from the
return values.


docs/customization.md
```diff
@@ -186,6 +186,27 @@ def get_resolver_provider(ctx, req, include_sdists, include_wheels, sdist_server
...
+from fromager import resolver
+
+VERSION_MAP = {'1.0': 'first-release', '2.0': 'second-release'}
+
+def _version_source(
+        identifier: str,
+        requirements: resolver.RequirementsMap,
+        incompatibilities: resolver.CandidatesMap,
+    ) -> typing.Iterable[Version]:
+    return VERSION_MAP.keys()
```
tiran, line 203: In this example `VERSION_MAP` keys are `str`, not `Version` objects.

dhellmann: The values can be str or Version objects. I'll clarify that.

docs/customization.md
```diff
@@ -186,6 +186,27 @@ def get_resolver_provider(ctx, req, include_sdists, include_wheels, sdist_server
     ...
 ```
 
+The `GenericProvider` is a convenient base class, or can be instantiated
+directly if given a `version_source` callable that returns an iterator of
+version values as `str` or `Version` objects.
```
shubhbapna, line 191: Why allow both? Since we want to start enforcing the use of `Version` and this is a new API we are introducing, we can enforce it without causing any breaking changes?

dhellmann: Ease of use from the caller's perspective, since most of those will be plugins. https://en.wikipedia.org/wiki/Robustness_principle

shubhbapna: ahh make sense

shubhbapna (approved):
LGTM!

PR #261, opened 2024-07-29
deal with setuptools 72.0.0

dhellmann:
Update our dependency and constrain dependency in CI.

In order to do that, apply a common startup script to each test script so we have a place to set up the constraints using values that should be common to each test.


tiran (approved):
LGTM

Issue #254, opened 2024-07-26
warning for old patches might not be working as intended


shubhbapna:
No warning is emitted if patches for some old version exists but no patch for the current version exists. I suspect adding the warn function in the [patch loop](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/sources.py#L278) might be incorrect since if we are in the loop then we did find a patch for the current version and we don't want to log in those cases:

The existing test for the warn function also seems to be incorrect since it is setting source directory to be the patches directory for that version: https://github.com/python-wheel-build/fromager/blob/main/tests/test_sources.py#L106
It fails the following test:

```python
@patch("logging.Logger.warning")
def test_warning_for_older_patch(mock, tmp_path: pathlib.Path):
     # create patches dir
    patches_dir = tmp_path / "patches_dir"
    patches_dir.mkdir()
    
    # create patches dir for old version of deepspeed
    deepspeed_old_patch = patches_dir / "deepspeed-0.5.0"
    deepspeed_old_patch.mkdir()
    patch_file = deepspeed_old_patch / "deepspeed-0.5.0.patch"
    patch_file.write_text("This is a test patch")

    # set current source to be a new version of deepspeed
    source_root_dir = tmp_path / "deepspeed-0.6.0"
    source_root_dir.mkdir()

    sources._warn_for_old_patch(source_root_dir, patch_file)
    mock.assert_called_with(
        "Patches for version 0.5.0 of deepspeed exist but will not be applied"
    )
```

rd4398:
I am trying to fix this bug for a while now but no luck so far. I also tried comparing versions along with names but it doesn't seem to work here:

```
if patch_filename.parent.name == source_root_dir.name:
        logger.warning(
            f"{req_name}: patches for version {existing_version!s} exist but will not be applied"
        )
```
Any ideas / suggestions?

Also, if I remove warn function from patch loop, how do I access patch_file which is returned by `overrides.patches_for_source_dir(ctx.patches_dir, source_root_dir.name)` ?

@dhellmann @shubhbapna

rd4398:
> > I am trying to fix this bug for a while now but no luck so far. I also tried comparing versions along with names but it doesn't seem to work here:
> > ```
> > if patch_filename.parent.name == source_root_dir.name:
> >         logger.warning(
> >             f"{req_name}: patches for version {existing_version!s} exist but will not be applied"
> >         )
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
[...]

PR #262, opened 2024-07-29
bug fix: warning for old patches

rd4398:
Fixes: https://github.com/python-wheel-build/fromager/issues/254

Also, I have used the test case suggested by @shubhbapna which was failing

### DIFF
--- a/src/fromager/overrides.py
+++ b/src/fromager/overrides.py
@@ -3,6 +3,7 @@ import itertools
 import logging
 import os
 import pathlib
+import re
 import string
 import typing
 
@@ -81,6 +82,32 @@ def patches_for_source_dir(
 
     """
     return sorted((patches_dir / source_dir_name).glob("*.patch"))
+
+
+def get_patch_directories(
+    patches_dir: pathlib.Path, source_root_dir: pathlib.Path
+) -> list[pathlib.Path]:
+    """
+    This function will return directories that may contain patches for a specific requirement.
+    It takes in patches directory and a source root directory as input.
+    The output will be a list of all directories containing patches for that requirement
+    """
+    # Get the req name as per the source_root_dir naming conventions
+    req_name = source_root_dir.name.rsplit("-", 1)[0]
+    patches = sorted((patches_dir).glob(f"{req_name}*"))
+    filtered_patches = _filter_patches_based_on_req(patches, req_name)
+    return filtered_patches
+
+
+# Helper method to filter the unwanted patches using a regex
+def _filter_patches_based_on_req(
+    patches: list[pathlib.Path], req_name: str
+) -> list[pathlib.Path]:
+    # Set up regex to filter out unwanted patches.
+    pattern = re.compile(rf"^{req_name}-v?(\d+\.)+\d+")
+    filtered_patches = [s for s in patches if pattern.match(s.name)]
+    # filtered_patches won't contain patches for current version of req
+    return filtered_patches
 
 
 def extra_environ_for_pkg(
--- a/src/fromager/sources.py
+++ b/src/fromager/sources.py
@@ -300,19 +300,26 @@ def patch_source(
     source_root_dir: pathlib.Path,
     req: Requirement,
 ) -> None:
+    # Flag to check whether patch has been applied
+    has_applied = False
     # apply any unversioned patch first
     for p in overrides.patches_for_source_dir(
         ctx.patches_dir, overrides.pkgname_to_override_module(req.name)
     ):
         _apply_patch(p, source_root_dir)
+        has_applied = True
 
     # make sure that we don't apply the generic unversioned patch again
     if source_root_dir.name != overrides.pkgname_to_override_module(req.name):
         for p in overrides.patches_for_source_dir(
             ctx.patches_dir, source_root_dir.name
         ):
-            _warn_for_old_patch(source_root_dir, p)
             _apply_patch(p, source_root_dir)
+            has_applied = True
+
+    # If no patch has been applied, call warn for old patch
+    if not has_applied:
+        _warn_for_old_patch(source_root_dir, ctx.patches_dir)
 
 
 def _apply_patch(patch: pathlib.Path, source_root_dir: pathlib.Path):
@@ -326,17 +333,18 @@ def _apply_patch(patch: pathlib.Path, source_root_dir: pathlib.Path):
 
 
 def _warn_for_old_patch(
-    source_root_dir: pathlib.Path, patch_filename: pathlib.Path
+    source_root_dir: pathlib.Path,
+    patches_dir: pathlib.Path,
 ) -> None:
-    # Get the existing version and req name as per the source_root_dir naming conventions
+    # Get the req name as per the source_root_dir naming conventions
     req_name = source_root_dir.name.rsplit("-", 1)[0]
-    existing_version = Version(source_root_dir.name.rsplit("-", 1)[1])
 
-    # Check to see if patches for older version exists
-    if patch_filename.parent.name == source_root_dir.name:
-        logger.warning(
-            f"{req_name}: patches for version {existing_version!s} exist but will not be applied"
-        )
+    # Filter the patch directories using regex
+    patch_directories = overrides.get_patch_directories(patches_dir, source_root_dir)
+
+    for dirs in patch_directories:
+        for p in dirs.iterdir():
+            logger.warning(f"{req_name}: patch {p} exists but will not be applied")
 
 
 def write_build_meta(
--- a/tests/test_sources.py
+++ b/tests/test_sources.py
@@ -102,20 +102,6 @@ def test_invalid_version(mock_default_download):
         sources.download_source(ctx, req, sdist_server_urls)
 
 
-@patch("logging.Logger.warning")
-def test_warning_for_older_patch(mock, tmp_path: pathlib.Path):
-    source = "deepspeed-0.5.0"
-    patches_dir = tmp_path / source
-    patches_dir.mkdir()
-    patch_file = patches_dir / "deepspeed-0.5.0.patch"
-    patch_file.write_text("This is a test patch")
-
-    sources._warn_for_old_patch(patches_dir, patch_file)
-    mock.assert_called_with(
-        "deepspeed: patches for version 0.5.0 exist but will not be applied"
-    )
-
-
 @patch("fromager.sources._apply_patch")
 def test_patch_sources_apply_unversioned_and_versioned(
     apply_patch: typing.Callable,
@@ -179,3 +165,43 @@ def test_patch_sources_apply_only_unversioned(
             call(unversioned_patch_file, source_root_dir),
         ]
     )
+
+
+@patch("logging.Logger.warning")
+def test_warning_for_older_patch(mock, tmp_path: pathlib.Path):
+    # create patches dir
+    patches_dir = tmp_path / "patches_dir"
+    patches_dir.mkdir()
+
+    # create patches dir for old version of deepspeed
+    deepspeed_old_patch = patches_dir / "deepspeed-0.5.0"
+    deepspeed_old_patch.mkdir()
+    patch_file = deepspeed_old_patch / "deepspeed-0.5.0.patch"
+    patch_file.write_text("This is a test patch")
+
+    # set current source to be a new version of deepspeed
+    source_root_dir = tmp_path / "deepspeed-0.6.0"
+    source_root_dir.mkdir()
+
+    sources._warn_for_old_patch(source_root_dir, patches_dir)
+    mock.assert_called()
+
+
+@patch("logging.Logger.warning")
+def test_warning_for_older_patch_different_req(mock, tmp_path: pathlib.Path):
+    # create patches dir
+    patches_dir = tmp_path / "patches_dir"
+    patches_dir.mkdir()
+
+    # create patches dir for old version of deepspeed
+    deepspeed_old_patch = patches_dir / "foo-0.5.0"
+    deepspeed_old_patch.mkdir()
+    patch_file = deepspeed_old_patch / "foo-0.5.0.patch"
+    patch_file.write_text("This is a test patch")
+
+    # set current source to be a new version of deepspeed
+    source_root_dir = tmp_path / "deepspeed-0.5.0"
+    source_root_dir.mkdir()
+
+    sources._warn_for_old_patch(source_root_dir, patches_dir)
+    mock.assert_not_called()
--- a/tests/test_overrides.py
+++ b/tests/test_overrides.py
@@ -199,3 +199,79 @@ def test_find_and_invoke(
     assert overrides.find_and_invoke(
         "pkg", "foo", default_foo, arg1="value1", arg2="value2"
     )
+
+
+def test_regex_dummy_package(tmp_path: pathlib.Path):
+    req_name = "foo"
+    patches_dir = tmp_path / "patches_dir"
+    patches_dir.mkdir()
+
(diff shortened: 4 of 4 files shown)


src/fromager/overrides.py
```diff
@@ -83,6 +83,18 @@ def patches_for_source_dir(
...
+def patches_for_req(
+    patches_dir: pathlib.Path, req: str
+) -> typing.Iterable[pathlib.Path]:
+    """
+    This function will return patches for a specific requirement.
+    It takes in patches directory and requirement as input.
+    The output will be a list of all directories containing patches for that requirement
+    """
+    glob_string = req + "*"
+    return sorted((patches_dir).glob(glob_string))
```
shubhbapna, line 95:
```suggestion
    return sorted((patches_dir).glob(f"{req}*"))
```

src/fromager/overrides.py
```diff
@@ -83,6 +83,18 @@ def patches_for_source_dir(
     return sorted((patches_dir / source_dir_name).glob("*.patch"))
 
 
+def patches_for_req(
+    patches_dir: pathlib.Path, req: str
```
shubhbapna, line 87:
```suggestion
    patches_dir: pathlib.Path, req_name: str
```
since you just pass only the req name so maybe this var name might be better

src/fromager/overrides.py
```diff
@@ -83,6 +83,18 @@ def patches_for_source_dir(
...
+) -> typing.Iterable[pathlib.Path]:
```
shubhbapna, line 88: I think a better return type would be a `list` since we are using `len` in the warn function and `len` is not necessarily part of an iterable. @tiran might be able to confirm?

src/fromager/sources.py
```diff
@@ -311,17 +309,19 @@ def patch_source(
...
-            f"{req_name}: patches for version {existing_version!s} exist but will not be applied"
-        )
+    # Create expected path for patch file
+    patch_path = pathlib.Path(patches_dir / source_root_dir)
+
+    # Get all the patches for that requirement
+    req_patches = overrides.patches_for_req(patches_dir, req_name)
+    # Check if old patches exist and current patch not in the list
+    if len(req_patches) > 0 and patch_path not in req_patches:
+        logger.warning(f"{req_name}: old patches exist but will not be applied")
```
shubhbapna, line 324: Probably want to log the version of the patches not being applied (logging the patches directory might be enough)

shubhbapna (changes requested):
Could you also add a test for a case where no warning should be applied

rd4398:
Okay, making changes!

src/fromager/overrides.py
rd4398: Done

src/fromager/overrides.py
rd4398: Done

src/fromager/overrides.py
rd4398: Okay, I have set the return type to list now. I will change it again if required.

src/fromager/sources.py
rd4398: Done

src/fromager/overrides.py
```diff
@@ -83,6 +83,15 @@ def patches_for_source_dir(
     return sorted((patches_dir / source_dir_name).glob("*.patch"))
 
 
+def patches_for_req(patches_dir: pathlib.Path, req_name: str) -> list[pathlib.Path]:
+    """
+    This function will return patches for a specific requirement.
+    It takes in patches directory and requirement as input.
+    The output will be a list of all directories containing patches for that requirement
+    """
+    return sorted((patches_dir).glob(f"{req_name}*"))
```
dhellmann, line 92: We have some packages that start with a common prefix, so this is going to return more than it should. For example, we have `instructlab` and `instructlab-sdg`. Any patches for `instructlab-sdg` will appear to be related to `instructlab`. The directory naming convention includes a `-` between the package name and version, so you could add that to the pattern to fix this.

There will also be patches in a directory just named for the requirement, without any version. For that case, you could just check for that directory to exist at all and include it if it is there without using a glob pattern.

The `patches_for_source_dir()` function takes the source directory, instead of a requirement, because the source directory doesn't have to match the requirement name. Requirements like `instructlab.sdg` and `instructlab-sdg` are equivalent. It looks like you're expecting the prefix of the source directory name, without any version information. This function might be clearer if it took the full source directory name and did its own parsing of that name to strip the version string, instead of having the caller do that step.

src/fromager/sources.py
```diff
@@ -311,17 +309,21 @@ def patch_source(
...
-        logger.warning(
-            f"{req_name}: patches for version {existing_version!s} exist but will not be applied"
-        )
+    # Get all the patches for that requirement
+    req_patches = overrides.patches_for_req(patches_dir, req_name)
+
+    # Check if old patches exist and current patch not in the list
+    if len(req_patches) > 0:
+        string_to_search = "\t".join(str(p) for p in req_patches)
+        if source_root_dir.name not in string_to_search:
```
dhellmann, line 323: I'm worried that relying on substrings is going to be brittle because of the name collisions I mentioned in the earlier comment.

If you build 2 sets, one containing all patches for all versions of a package and the other containing only patches for the current version of the package, you could use set.difference() to find the items that are available to be applied that won't be applied and then emit the warnings.

dhellmann (changes requested):

src/fromager/sources.py
rd4398: How can I build a set for patches for current version? Should I add a method in overrides doing something like:

```
def patches_directory_for_current_version(patches_dir: pathlib.Path, source_dir_name: str) -> pathlib.Path | None:
    """
    This method will return patches directory for current version
    """
    return patches_dir / source_dir_name if (patches_dir / source_dir_name).exists() else None
```
Or is there  a better way? I am not sure

rd4398:
After, doing a rebase and understanding Shubh's changes for generic patch, I have changed the logic for warning of old patches which now includes a regex to filter packages and avoid conflicts for packages like `instructlab `and `instructlab-sdg`
As per Doug's comment in https://github.com/python-wheel-build/fromager/issues/254:
`Remember, the goal is to report patches related to a package that will not be applied, but only when no patches will be applied.`
I think set difference is not needed and a simple flag in `patch_source` is enough. I have added test cases for different scenarios as well.
Does it make sense?
Let me know if any changes are required

src/fromager/sources.py
```diff
@@ -326,17 +336,30 @@ def _apply_patch(patch: pathlib.Path, source_root_dir: pathlib.Path):
...
-    if patch_filename.parent.name == source_root_dir.name:
-        logger.warning(
-            f"{req_name}: patches for version {existing_version!s} exist but will not be applied"
+    # Get all the patches for that requirement
+    req_patches = overrides.patches_for_req(patches_dir, source_root_dir.name)
+
+    if len(req_patches) > 0:
+        # Set up regex to filter out unwanted patches.
+        # Example: instructlab and instructlab-sdg
+        pattern = re.compile(rf"^{req_name}(-v?\d+\.\d+\.\d+)?$")
```
dhellmann, line 351: vllm will have 4 numerical parts to its version, and we have some packages that end with version `.post1`. If there are unversioned patches, in a directory that's the same as req_name, then those would have been applied and this function shouldn't be called, so I think you can ignore that case.

I think that's something like this:

```suggestion
        pattern = re.compile(rf"^{req_name}-v?(\d+\.)+\d+)")
```

It would be good to pull this regex out of the function and write some unit tests using known package names and versions that we have.

src/fromager/sources.py
```diff
@@ -326,17 +336,30 @@ def _apply_patch(patch: pathlib.Path, source_root_dir: pathlib.Path):
...
+        filtered_patches = [s for s in req_patches if pattern.match(s.name)]
+
+        current_path = source_root_dir.parent.joinpath(
+            "patches_dir", source_root_dir.name
         )
+        # If filtered patches contain patches for current version, remove the path because we dont want to warn
+        if current_path in filtered_patches:
+            filtered_patches.remove(current_path)
+        if len(filtered_patches) > 0:
+            logger.warning(f"{req_name}: old patches exist but will not be applied")
```
dhellmann, line 362: They might be patches for a newer version, so the message shouldn't assume which one is the newer/older. It would be good to log each patch name that is being ignored.

src/fromager/overrides.py
```diff
@@ -83,6 +83,17 @@ def patches_for_source_dir(
     return sorted((patches_dir / source_dir_name).glob("*.patch"))
 
 
+def patches_for_req(patches_dir: pathlib.Path, dir_name: str) -> list[pathlib.Path]:
+    """
+    This function will return patches for a specific requirement.
+    It takes in patches directory and requirement as input.
```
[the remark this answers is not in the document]
rd4398: Okay, changed it to `get_patches()`
Is it more generic? Should I change it?

src/fromager/sources.py
rd4398: Okay, I have added tests for regex and created a helper for it

src/fromager/sources.py
rd4398: Okay, I am now logging` filtered_patches`. 
Should I use `iterdir()` to log individual patch files?

rd4398:
I have made changes as per the suggestions. Let me know if any other changes are required

src/fromager/overrides.py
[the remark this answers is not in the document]
dhellmann: The docstring is out of date. What is `dir_name` expected to be?

This function is returning all patches for all versions of a dependency, starting from a source directory, right? If that's the case, please change `dir_name` to `source_root_dir` to match the value used elsewhere. Naming it consistently helps the caller understand what it is.

Since the function doesn't return patch filenames, it shouldn't be called `get_patches`. Maybe `get_patch_directories`?

src/fromager/sources.py
```diff
@@ -326,17 +336,36 @@ def _apply_patch(patch: pathlib.Path, source_root_dir: pathlib.Path):
...
+        # If filtered patches contain patches for current version, remove the path because we dont want to warn
+        if current_path in filtered_patches:
+            filtered_patches.remove(current_path)
+        if len(filtered_patches) > 0:
+            for p in filtered_patches:
+                logger.warning(f"{req_name}: patch {p} exists but will not be applied")
+
+
+# Helper method to filter the unwanted patches using a regex
+def _filter_patches(patches: list[pathlib.Path], req_name: str) -> list[pathlib.Path]:
```
dhellmann, line 363: This function needs a more descriptive name. 

Maybe this filtering should be done in the function in overrides?

src/fromager/sources.py
```diff
@@ -326,17 +336,36 @@ def _apply_patch(patch: pathlib.Path, source_root_dir: pathlib.Path):
...
+    if len(patches) > 0:
+        # Filter the patches using regex
+        filtered_patches = _filter_patches(patches, req_name)
+        current_path = source_root_dir.parent.joinpath(
+            "patches_dir", source_root_dir.name
         )
+        # If filtered patches contain patches for current version, remove the path because we dont want to warn
+        if current_path in filtered_patches:
+            filtered_patches.remove(current_path)
+        if len(filtered_patches) > 0:
```
dhellmann, line 357: Iterating over an empty list is OK, you don't need this test.

src/fromager/sources.py
```diff
@@ -326,17 +336,36 @@ def _apply_patch(patch: pathlib.Path, source_root_dir: pathlib.Path):
...
-    existing_version = Version(source_root_dir.name.rsplit("-", 1)[1])
 
-    # Check to see if patches for older version exists
-    if patch_filename.parent.name == source_root_dir.name:
-        logger.warning(
-            f"{req_name}: patches for version {existing_version!s} exist but will not be applied"
+    # Get all the patches for that requirement
+    patches = overrides.get_patches(patches_dir, source_root_dir.name)
+
+    if len(patches) > 0:
```
dhellmann, line 348: This test isn't needed. If the list is empty, no messages will be emitted.

src/fromager/sources.py
rd4398: Okay

src/fromager/sources.py
rd4398: Done. Also shifted it to overrides.py

src/fromager/overrides.py
```diff
@@ -83,6 +84,32 @@ def patches_for_source_dir(
     return sorted((patches_dir / source_dir_name).glob("*.patch"))
 
 
+def get_patch_directories(
+    patches_dir: pathlib.Path, source_root_dir: pathlib.Path
+) -> list[pathlib.Path]:
+    """
+    This function will return patches for a specific requirement.
```
dhellmann, line 91:
```suggestion
    This function will return directories that may contain patches for a specific requirement.
```

src/fromager/overrides.py
```diff
@@ -83,6 +84,32 @@ def patches_for_source_dir(
...
+    It takes in patches directory and requirement as input.
```
dhellmann, line 92:
```suggestion
    It takes in patches directory and a source root directory as input.
```

dhellmann (commented):
2 nits on that docstring, and then I think this is ready

src/fromager/overrides.py
rd4398: Okay, done!

src/fromager/overrides.py
rd4398: done!

dhellmann (commented):
This version looks good to me. @shubhbapna had comments, so let's get him to look again before we merge it.

tests/test_sources.py
```diff
@@ -179,3 +165,139 @@ def test_patch_sources_apply_only_unversioned(
...
+
+    expected = [
+        patches_dir / "vllm-1.1.0.9",
+        patches_dir / "vllm-v2.3.5.0.post1",
+        patches_dir / "vllm-v5.5.5.1",
+    ]
+
+    actual = overrides._filter_patches_based_on_req(lst, req_name)
+    assert len(expected) == len(actual)
+    assert expected == actual
```
shubhbapna, line 303: These tests should be in `test_overrides.py`

src/fromager/sources.py
```diff
@@ -326,17 +335,21 @@ def _apply_patch(patch: pathlib.Path, source_root_dir: pathlib.Path):
...
-    if patch_filename.parent.name == source_root_dir.name:
-        logger.warning(
-            f"{req_name}: patches for version {existing_version!s} exist but will not be applied"
-        )
+    # Filter the patches using regex
+    patch_directories = overrides.get_patch_directories(patches_dir, source_root_dir)
+    current_path = source_root_dir.parent.joinpath("patches_dir", source_root_dir.name)
+    # If filtered patches contain patches for current version, remove the path because we dont want to warn
+    if current_path in patch_directories:
+        patch_directories.remove(current_path)
```
shubhbapna, line 349: I don't think filtered patches will contain the current version since you are only calling this function if `has_applied` is false which implies that were no patches directory for the current version.

You can add a docstring for this helper to clarify this point so that any other caller can use it correctly

tests/test_sources.py
rd4398: okay, done!

src/fromager/sources.py
rd4398: Okay, I added it as docstring

rd4398:
The latest version is ready for review

dhellmann (approved):
This looks good, thanks for those updates!

shubhbapna (approved):
LGTM

PR #266, opened 2024-07-29
bug: fix installation of dependency by using the resolved_version instead of actual_version

shubhbapna:
We noticed that dependencies that were resolving to some different version were not actually being installed using that version. Here is my analysis of the bug:

[_maybe_install](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/sdist.py#L492) is called after handling each requirement. If the resolved version is not none, then it checks whether the actual version matches the resolved version and if it doesn't then it installs it using [safe_install](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/sdist.py#L517) but in`_maybe_install` it seems like we are passing the old requirement to `safe_install` instead of the resolved one: https://github.com/python-wheel-build/fromager/blob/main/src/fromager/sdist.py#L514

### DIFF
--- a/src/fromager/sdist.py
+++ b/src/fromager/sdist.py
@@ -507,6 +507,8 @@ def _maybe_install(
             logger.info(
                 f"{req.name}: found {req.name} {actual_version} installed, updating to {resolved_version}"
             )
+            safe_install(ctx, Requirement(f"{req.name}=={resolved_version}"), req_type)
+            return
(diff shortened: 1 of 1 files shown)


Gregory-Pereira (commented):
Overall /lgtm. This change makes sense, still want to wait for Doug to confirm on this one though.

dhellmann (approved):
Nice find!

Issue #263, opened 2024-07-29
Add option to build with network isolation


tiran:
Prarit and I just noticed that ROCm PyTorch builds are downloading projects from the internet during build. PyTorch's cmake files use [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html). We are worried that CUDA builds could behave the same, but we have no easy way to detect the problem.

Suggestion:
Add an option to run `sdist` and `wheel` build steps with `unshare -rn`. The unshare command creates a new network namespace without any network configuration. `-r` enables `--map-root-user`, which allows unshare as a regular user.

```
$ unshare -rn ping github.com
ping: connect: Network is unreachable
```

PR #269, opened 2024-07-30
Run sdist and wheel building with network isolation

tiran:
Prevent sdist and wheel build steps from accessing the internet. `externalcommand.run()` can use now run commands with unshare. `unshare --net --map-current-user` creates a new network namespace without access network.

Fixes: #263

### DIFF
--- a/src/fromager/__main__.py
+++ b/src/fromager/__main__.py
@@ -6,12 +6,29 @@ import pathlib
 
 import click
 
-from . import clickext, commands, constraints, context, overrides, settings
+from . import (
+    clickext,
+    commands,
+    constraints,
+    context,
+    external_commands,
+    overrides,
+    settings,
+)
 
 logger = logging.getLogger(__name__)
 
 TERSE_LOG_FMT = "%(message)s"
 VERBOSE_LOG_FMT = "%(levelname)s:%(name)s:%(lineno)d: %(message)s"
+
+try:
+    external_commands.detect_network_isolation()
+except Exception as e:
+    SUPPORTS_NETWORK_ISOLATION: bool = False
+    NETWORK_ISOLATION_ERROR: str | None = str(e)
+else:
+    SUPPORTS_NETWORK_ISOLATION = True
+    NETWORK_ISOLATION_ERROR = None
 
 
 @click.group()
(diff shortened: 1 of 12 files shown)


tiran:
fromager enables network isolation by default. I could make it opt-in.

src/fromager/external_commands.py
```diff
@@ -20,7 +22,12 @@ def run(
     env = os.environ.copy()
     env.update(extra_environ)
 
-    logger.debug(
+    if network_isolation:
+        # prevent network access by creating a new network namespace that
+        # has no routing configured.
+        cmd = ["unshare", "--net", "--map-current-user"] + cmd
+
+    logger.info(
```
dhellmann, line 30: Logging these commands at INFO is going to make the console output pretty verbose. Was this for your own debugging, or do you want to keep it this way?

src/fromager/external_commands.py
```diff
@@ -20,7 +22,12 @@ def run(
     env = os.environ.copy()
     env.update(extra_environ)
 
-    logger.debug(
+    if network_isolation:
+        # prevent network access by creating a new network namespace that
+        # has no routing configured.
+        cmd = ["unshare", "--net", "--map-current-user"] + cmd
```
dhellmann, line 28: Platforms other than linux may not have `unshare`. Even some linux hosts may not have it installed. We can deal with it in a follow-up, but we should decide how to handle that case.

We could look at the OS to set the default for the flag.

We could look for `unshare` in `$PATH` to set the default for the flag.

We could have a config file option to set the default.

We could just make the default `False`.

Other options?

dhellmann (commented):
One question about logging, and a follow-up for maintaining support for other OSes, but otherwise this looks good.

src/fromager/external_commands.py
tiran: I didn't meant to commit the change. I bumped the log level locally for testing.

src/fromager/external_commands.py
tiran: I have updated the PR to detect platform, `shutil.which('unshare')`, and if unshare command works. The default now depends on whether unshare works or not. Users can override the detection with `fromager --network-isolation` and `fromager --no-network-isolation`.

src/fromager/__main__.py
```diff
@@ -6,13 +6,30 @@
...
+    settings,
+)
 
 logger = logging.getLogger(__name__)
 
 TERSE_LOG_FMT = "%(message)s"
 VERBOSE_LOG_FMT = "%(levelname)s:%(name)s:%(lineno)d: %(message)s"
 
+try:
+    external_commands.supports_network_isolation()
```
dhellmann, line 25: I don't see a function with this name in this PR.

tiran: Another refactoring going wrong. Naming is hard.

dhellmann (approved):

PR #273, opened 2024-07-30
cleanup: fix typings in extras_provider.py

shubhbapna:
part of #226 

fixes:

```
src/fromager/extras_provider.py:24: error: Missing type parameters for generic type "AbstractProvider"  [type-arg]
src/fromager/extras_provider.py:27: error: Function is missing a type annotation  [no-untyped-def]
src/fromager/extras_provider.py:34: error: Function is missing a type annotation  [no-untyped-def]
src/fromager/extras_provider.py:41: error: Function is missing a type annotation  [no-untyped-def]
src/fromager/extras_provider.py:43: error: Call to untyped function "get_extras_for" in typed context  [no-untyped-call]
src/fromager/extras_provider.py:49: error: Function is missing a type annotation  [no-untyped-def]
src/fromager/extras_provider.py:52: error: Call to untyped function "get_base_requirement" in typed context  [no-untyped-call]
src/fromager/extras_provider.py:53: error: "Iterable[Any]" has no attribute "append"  [attr-defined]
```

fixes some of the resolver type errors that were coming from the Candidate class

### DIFF
--- a/src/fromager/candidate.py
+++ b/src/fromager/candidate.py
@@ -1,0 +1,85 @@
+import typing
+from email.message import EmailMessage, Message
+from email.parser import BytesParser
+from io import BytesIO
+from typing import TYPE_CHECKING
+from zipfile import ZipFile
+
+import requests
+from packaging.requirements import Requirement
+from packaging.utils import (
+    canonicalize_name,
+)
+from packaging.version import Version
+
+# fix for runtime errors caused by inheriting classes that are generic in stubs but not runtime
+# https://mypy.readthedocs.io/en/latest/runtime_troubles.html#using-classes-that-are-generic-in-stubs-but-not-at-runtime
+if TYPE_CHECKING:
+    Metadata = Message[str, str]
+else:
+    Metadata = Message
+
+
+class Candidate:
+    def __init__(
+        self,
+        name: str,
+        version: Version,
+        url: str,
+        extras: typing.Iterable[str] | None = None,
+        is_sdist: bool | None = None,
+    ):
+        self.name = canonicalize_name(name)
+        self.version = version
+        self.url = url
+        self.extras = extras
+        self.is_sdist = is_sdist
+
+        self._metadata: Metadata | None = None
+        self._dependencies: list[Requirement] | None = None
+
+    def __repr__(self) -> str:
+        if not self.extras:
+            return f"<{self.name}=={self.version}>"
+        return f"<{self.name}[{','.join(self.extras)}]=={self.version}>"
+
+    @property
+    def metadata(self) -> Metadata:
+        if self._metadata is None:
+            self._metadata = get_metadata_for_wheel(self.url)
+        return self._metadata
+
+    def _get_dependencies(self) -> typing.Iterable[Requirement]:
+        deps = self.metadata.get_all("Requires-Dist", [])
+        extras = self.extras if self.extras else [""]
+
+        for d in deps:
+            r = Requirement(d)
+            if r.marker is None:
+                yield r
+            else:
+                for e in extras:
+                    if r.marker.evaluate({"extra": e}):
+                        yield r
+
+    @property
+    def dependencies(self) -> list[Requirement]:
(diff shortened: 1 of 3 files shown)


src/fromager/resolver.py
```diff
@@ -37,59 +34,6 @@
...
-
-    @property
-    def metadata(self):
-        if self._metadata is None:
-            self._metadata = get_metadata_for_wheel(self.url)
-        return self._metadata
-
-    @property
-    def requires_python(self) -> bool:
-        return self.metadata.get("Requires-Python")
```
shubhbapna, line 71: Removed this function because IMO it is not being used anywhere even within the resolvelib:
fromager: https://github.com/search?q=repo%3Apython-wheel-build%2Ffromager%20requires_python&type=code
resolvelib: https://github.com/search?q=repo%3Asarugaku%2Fresolvelib%20requires_python&type=code

src/fromager/resolver.py
```diff
@@ -37,59 +34,6 @@
 SUPPORTED_TAGS = set(sys_tags())
 
 
-class Candidate:
```
shubhbapna, line 40: Had to shift this class to a separate module so that extras_provider can use it. Otherwise it was throwing a circular import error

src/fromager/extras_provider.py
```diff
@@ -18,37 +18,36 @@
...
         """
         raise NotImplementedError
-
-    def identify(self, requirement_or_candidate):
-        base = super().identify(requirement_or_candidate)
-        extras = self.get_extras_for(requirement_or_candidate)
-        if extras:
-            return (base, extras)
-        else:
-            return base
```
shubhbapna, line 47: This method calls `super().identify` but that will raise `NotImplementedError` since https://github.com/sarugaku/resolvelib/blob/main/src/resolvelib/providers.py#L25 

Since we have not seen this error I think we can safely remove it

src/fromager/extras_provider.py
```diff
@@ -18,37 +18,36 @@
...
-
-    def get_dependencies(self, candidate):
-        deps = super().get_dependencies(candidate)
-        if candidate.extras:
-            req = self.get_base_requirement(candidate)
-            deps.append(req)
-        return deps
```
shubhbapna, line 54: This method calls `super().get_dependencies` but that will raise `NotImplementedError` since https://github.com/sarugaku/resolvelib/blob/main/src/resolvelib/providers.py#L132

Since we have not seen this error I think we can safely remove it

src/fromager/candidate.py
```diff
@@ -0,0 +1,81 @@
...
+
+    @property
+    def dependencies(self) -> list[Requirement]:
+        if self._dependencies is None:
+            self._dependencies = list(self._get_dependencies())
+        return self._dependencies
+
+
+def get_metadata_for_wheel(url: str) -> Metadata:
+    data = requests.get(url).content
```
tiran, line 73: I just noticed that this is a problematic function. This line can download several GB of data and consume several GB of memory.

We should use https://peps.python.org/pep-0658/ if possible and use streaming IO if package metadata is not available.

src/fromager/extras_provider.py
tiran: SGTM

src/fromager/extras_provider.py
tiran: SGTM

src/fromager/resolver.py
tiran: Let's keep it in.

A resolver should take Requires-Python into account and fall back to an older version of a package.

tiran (changes requested):

src/fromager/candidate.py
shubhbapna: I will file a separate issue for this because the url being passed here is the download url and it doesn't have access to the tags that PEP 658 mentions. I will need to change some code in the resolver

tiran (approved):
Good work!

Issue #274, opened 2024-07-31
Automatically patch out cmake from pyproject.toml


tiran:
Several Python dist packages have a dependency on cmake  PyPI package. The PyPI package is a convenient provider for cmake, so users don't have to install the toot locally. Fromager assumes that the package is installed in build environments

Affected packages with cmake:
- torch
- triton
- vllm

It should be easy to automatically detect and remove the dependency. We have to use https://tomlkit.readthedocs.io/ to preserve style and comments in `pyproject.toml`.

dhellmann:
> Several Python dist packages have a dependency on cmake PyPI package. The PyPI package is a convenient provider for cmake, so users don't have to install the toot locally. Fromager assumes that the package is installed in build environments

We're making that assumption downstream. I don't know that I'd say all fromager users will want to make the same assumption.

I could see us making it easier to do this automatically, even to the point of having it expressed via the config file. I'm not sure we want to always enforce it.

dhellmann:
Maybe the way to approach this is to have a hook that is run as part of preparing the source tree, and then the logic for deciding whether to patch and how could be in the hook. We could publish a hook upstream, but it wouldn't have to be part of the core.

Issue #260, opened 2024-07-29
Auto-detect and auto-patch `[build-system] requires`


tiran:
Several packages in our dependency chain have incomplete or missing `[build-system] requires` in `pyproject.toml`. We should be able to automatically detect and fix most cases with a dash of AST magic. This will reduce the amount of patches in downstream.

1. detect absence of `pyproject.toml` or lack of `[build-system] requires`
2. use AST to parse and locate imports in `setup.py`
3. map known imports such as `packaging` and `torch` to requirements
4. detect requirement version from dist-info `METADATA` file, e.g. `torch >= 2.3.1`
5. write fixed `pyproject.toml` with `tomli`

AST detection of import statement is pretty straight forward:

```python
import ast

def detect_imports(codestring: str, filename: str = "<unknown>"):
    """Detect imported modules"""
    rootnode = ast.parse(codestring, filename)
    for node in ast.walk(rootnode):
        if isinstance(node, ast.Import):
            for subnode in node.names:
                yield subnode.name
        elif isinstance(node, ast.ImportFrom):
            yield node.module

# testing
print(
    list(
        detect_imports("""
from foo import bar
import foo.bar
import foo as bar
from foo import bar as baz
import egg, spam

def func():
    import camelot
""")
    )
)
```

Output:
```
['foo', 'foo.bar', 'foo', 'foo', 'egg', 'spam', 'camelot']
```

PR #275, opened 2024-07-31
WIP: Auto-fixer for pyproject.toml

tiran:
- add missing pyproject.toml
- add or update [build-system] map
- detect additional build requirements by parsing "setup.py"
- remove unwanted build requirements such as cmake

Fixes: #260
Fixes: #274

### DIFF
--- a/src/fromager/pyproject.py
+++ b/src/fromager/pyproject.py
@@ -1,0 +1,202 @@
+"""Tooling for pyproject.toml"""
+
+import ast
+import logging
+import pathlib
+import sys
+import typing
+
+import tomlkit
+from packaging.requirements import Requirement
+from packaging.utils import NormalizedName, canonicalize_name
+
+if typing.TYPE_CHECKING:
+    from . import context
+
+logger = logging.getLogger(__name__)
+
+TomlDict = dict[str, typing.Any]
+
+# section / key names
+BUILD_SYSTEM = "build-system"
+BUILD_BACKEND = "build-backend"
+BUILD_REQUIRES = "requires"
+
+
+class PyprojectOverrides(typing.NamedTuple):
+    """pyproject.toml overrides
+
+    import_build_requires: auto detect imports in setup.py
+    drop_build_requires: remove packages from [build-system]requires
+    replace_build_requires: replace a build requirement with another
+    """
+
+    enable: bool
+    auto_build_requires: dict[str, Requirement]
+    remove_build_requires: list[NormalizedName]
+    replace_build_requires: dict[NormalizedName, Requirement]
+
+    @classmethod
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -11,5 +11,6 @@ resolvelib
 setuptools<72.0.0
 stevedore
 toml
+tomlkit
```
dhellmann, line 14: Do we need both toml libraries? We don't have to remove the other one in this PR, but maybe a ticket to do that cleanup?

src/fromager/pyproject.py
```diff
@@ -0,0 +1,170 @@
...
+        self.req = req
+        self.root = sdist_root_dir
+        self.pyproject_toml = self.root / "pyproject.toml"
+        self.setup_py = self.root / "setup.py"
+
+    def update(self):
+        doc = self.load()
+        build_system = self.default_builds_system(doc)
+        self.update_build_requires(build_system)
+        self.save(doc)
```
dhellmann, line 64: Does tomlkit preserve comments? Maybe we don't care if we roundtrip the file and lose info?

dhellmann (commented):
Would you apply this to everything being built? Or is it meant to be an optional step only used for some packages?

src/fromager/pyproject.py
tiran: Yes, tomlkit preserves comments and even layout of the file. Fromager's `pyproject.toml` does roundtrip perfectly.

```python
>>> import pathlib, tomlkit, toml
>>> raw = pathlib.Path("pyproject.toml").read_text()
>>> tomlkit.dumps(tomlkit.loads(raw)) == raw
>>> print(tomlkit.dumps(tomlkit.loads(raw)))
>>> print(toml.dumps(toml.loads(raw)))
```
tomlkit keeps
```toml
[tool.setuptools.dynamic]
optional-dependencies.build = { file = ["requirements-build.txt"] }
optional-dependencies.mypy = { file = ["requirements-mypy.txt"] }
optional-dependencies.test = { file = ["requirements-test.txt"] }
dependencies = { file = ["requirements.txt"] }
```

while toml changes the lines to
```toml
[tool.setuptools.dynamic.optional-dependencies.build]
file = [ "requirements-build.txt",]

[tool.setuptools.dynamic.optional-dependencies.mypy]
file = [ "requirements-mypy.txt",]

[tool.setuptools.dynamic.optional-dependencies.test]
file = [ "requirements-test.txt",]
```

dhellmann: Nice, let's switch!

tiran: Let me make a separate PR for that.

tiran:
Replaced by #407

PR #282, opened 2024-08-01
Enforce MyPy type checks, ignore files with known issues

tiran:
Thanks to Shubh's relentless work, most files now pass MyPy type checker. It's time to ensure that new code does not introduce regressions.

- enforce successful MyPy runs in `tox.ini` and GitHub Actions
- ignore MyPy errors in files with known issues
- fix a bunch of low hanging fruits in tests

See: #226

### DIFF
--- a/src/fromager/external_commands.py
+++ b/src/fromager/external_commands.py
@@ -15,7 +15,7 @@ else:
     NETWORK_ISOLATION = None
 
 
-def network_isolation_cmd() -> tuple[str, ...]:
+def network_isolation_cmd() -> typing.Sequence[str]:
     """Detect network isolation wrapper
 
     Raises ValueError when network isolation is not supported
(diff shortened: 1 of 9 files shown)


shubhbapna:
Should we enable strict mode for mypy (at least for the source code)?

src/fromager/external_commands.py
```diff
@@ -37,7 +37,8 @@ def detect_network_isolation():
     """
     cmd = network_isolation_cmd()
     if os.name == "posix":
-        subprocess.check_call(cmd + ["true"], stderr=subprocess.STDOUT)
+        check = [*cmd, "true"]
+        subprocess.check_call(check, stderr=subprocess.STDOUT)
```
shubhbapna, line 41: I think this all needs to be called using `external_commands.run`. I will add it to the existing issue

tiran: The function runs before we have logging framework set up.

tiran:
> Should we enable strict mode for mypy (at least for the source code)?

Yes, let's fix all files first, then might MyPy tighter.

src/fromager/external_commands.py
shubhbapna: Ah got it

shubhbapna (approved):

PR #283, opened 2024-08-01
fix: set source type to override if a download url was specified in settings

shubhbapna:


shubhbapna:
@dhellmann do you want the sdist type to override even when just the `destination_filename` is specified in settings (and not the url)?

dhellmann (approved):

dhellmann:
> @dhellmann do you want the sdist type to override even when just the `destination_filename` is specified in settings (and not the url)?

I didn't see this before merging.

We only need to prevent the download command from trying to download an sdist that it won't find on pypi. I don't think there's a use case for downloading from pypi and renaming the file.

PR #284, opened 2024-08-01
feat: source finders should be aware of download config settings

### LINKED ISSUE #279: source finders should be aware of download config settings
If a user configures special downloader instructions using the settings file, especially if rename instructions are included, then it should not be necessary to create override functions for the finders to use to find those names.

shubhbapna:
fixes #279

### DIFF
--- a/src/fromager/settings.py
+++ b/src/fromager/settings.py
@@ -1,8 +1,12 @@
 import logging
 import pathlib
+import string
 import typing
 
 import yaml
+from packaging.requirements import Requirement
+from packaging.utils import canonicalize_name
+from packaging.version import Version
 
 from . import overrides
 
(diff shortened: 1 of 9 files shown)


src/fromager/settings.py
```diff
@@ -24,17 +28,35 @@ def packages(self) -> dict[str, dict[str, str]]:
...
         }
 
-    def download_source_url(self, pkg: str, default: str | None = None) -> str | None:
+    def download_source_url(
+        self,
+        pkg: str,
+        req: Requirement | None = None,
+        version: Version | str | None = None,
+        default: str | None = None,
+        resolve_template: bool = True,
```
dhellmann, line 37: Having the template rendering happen inside this module feels like a layering violation. It has to know what all the parameters allowed in the template might be, which means if we add parameters to the template we have to make changes in multiple places.

What if this method was `download_source_url_template(pkg: str)`? And the caller did the rendering? If the rendering logic is reusable, that could go into its own module.

shubhbapna: I put it in settings since templated values is something that settings provides to the user. I don't think the caller should be responsible or know how to do template resolution

dhellmann (approved):

PR #285, opened 2024-08-02
Fix progress bar recursion in bootstrap command

tiran:
The bootstrap command uses a recursive algorithm to fetch sources and to build sdists + wheels. The complete set of dependencies is unknown until all packages have been downloaded and built.

The bootstrap command now passes an optional `progressbar` object around. Each handler updates the total amount of items in the beginning of a loop, and notifies the progress bar when an item has finished.

### DIFF
--- a/src/fromager/commands/bootstrap.py
+++ b/src/fromager/commands/bootstrap.py
@@ -15,7 +15,7 @@ logger = logging.getLogger(__name__)
 def _get_requirements_from_args(
     toplevel: typing.Iterable[str],
     req_files: typing.Iterable[pathlib.Path],
-) -> typing.Iterable[tuple[str, str]]:
+) -> typing.Sequence[tuple[str, str]]:
     to_build: list[tuple[str, str]] = []
     to_build.extend(("toplevel", t) for t in toplevel)
     for filename in req_files:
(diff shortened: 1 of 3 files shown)


shubhbapna:
Question: Is it possible to define this progress bar in the context so that it can start immediately as soon as fromager starts running and be passed around to different functions? Then when each function is called it can update the progress bar with the total amount of items that are being handled? One added advantage would be that we get this progress immediately for all our commands (and any commands that we add in the future) without having to change anything

tiran:
Yes, we could do all sorts of complicated things with `tqdm`. I would prefer to keep simple things as simple as possible.

The download sequence and build sequence commands have a for loop that iterates over a sequence of known length. `tqdm` makes it very easy to add a progress bar to an iterator with `__len__`. Replace `for item in seq:`  with `for item in tqdm.tqdm(seq):` and you have a working progress bar.

The bootstrap command is different, because it does not know the total size. You start with one package: download, build, fetch sdist/build/install requirements. For each requirement, get its sdist/build/install requirements, and so on. Each package expands the total sum.

`tqdm` has more fancy features, e.g. nested progress bars, logger integration, or TUI features with rich. I tried the logger integration, but it's buggy and messes up log output in CI runs.

dhellmann (approved):

PR #289, opened 2024-08-03
show plugin package and version

dhellmann:
On startup, log the package and version from which
any override plugins are being loaded.

### DIFF
--- a/e2e/test_override.sh
+++ b/e2e/test_override.sh
@@ -13,6 +13,7 @@ source "$SCRIPTDIR/common.sh"
 pip install e2e/flit_core_override
 
 fromager \
+  --verbose \
   --log-file="$OUTDIR/bootstrap.log" \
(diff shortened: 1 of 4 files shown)


dhellmann:
@Gregory-Pereira this is an alternative way for us to get the SHA of the downstream repo into a build log, because the SHA abbreviation will be part of the package version string for the plugins.

e2e/test_override.sh
shubhbapna, line 1: Maybe add a simple check for one of these debug statements in this e2e test:
```
DEBUG:fromager.overrides:86: loaded override 'flit_core': from package_plugins.flit_core (flit-core-overrides 0.0.0)
DEBUG:fromager.overrides:86: loaded override 'fromager_test': from fromager.example_override (fromager 0.1.dev1+g95e2fab)
```

dhellmann: Done

shubhbapna (approved):

PR #290, opened 2024-08-04
fix filenames in tarballs for sdists

### LINKED ISSUE #287: Fromager creates bad sdist tar ball
```console
$ tar tf deepspeed-0.14.4.tar.gz | grep PKG-INFO
build-output/work-dir/deepspeed-0.14.4/deepspeed-0.14.4/PKG-INFO
```

The correct path is `deepspeed-0.14.4/PKG-INFO`. The leading segments `build-output/work-dir/deepspeed-0.14.4` are invalid.

dhellmann:
Remove the prefix from the members of the tarballs
created by default for sdists.

Fixes #287

### DIFF
--- a/src/fromager/tarballs.py
+++ b/src/fromager/tarballs.py
@@ -22,8 +22,17 @@ def _tar_reset(tarinfo: tarfile.TarInfo) -> tarfile.TarInfo:
     return tarinfo
 
 
-def tar_reproducible(tar: tarfile.TarFile, basedir: pathlib.Path) -> None:
-    """Create reproducible tar file"""
+def tar_reproducible(
+    tar: tarfile.TarFile,
+    basedir: pathlib.Path,
+    prefix: pathlib.Path | None = None,
+) -> None:
+    """Create reproducible tar file
+
+    Add content from basedir to already opened tar. If prefix is provided, use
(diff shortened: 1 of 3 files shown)


src/fromager/tarballs.py
```diff
@@ -33,5 +45,18 @@ def tar_reproducible(tar: tarfile.TarFile, basedir: pathlib.Path) -> None:
...
+        if cur_name[0] != os.sep:
+            cur_name = os.sep + cur_name
+        new_name = pathlib.Path(cur_name).relative_to(prefix)
+        tarinfo.name = str(new_name)
+        logger.debug("adding %s", tarinfo.name)
+        return tarinfo
+
     for fn in content:
-        tar.add(fn, filter=_tar_reset, recursive=False, arcname=fn)
+        tar.add(fn, filter=_tar_filter, recursive=False)
```
tiran, line 62: This code passes your tests and is easier to understand (at least for me):

```suggestion
        arcname = fn if prefix is None else os.path.relpath(fn, prefix)
        tar.add(fn, filter=_tar_reset, recursive=False, arcname=arcname)
```

tiran (approved):

PR #291, opened 2024-08-04
improve network isolation error handling

dhellmann:
Add an exception type for network isolation errors
and raise that type instead of a simple
CalledProcessError when we are running with
network isolation and see a message that looks
like that is the problem. This fixes a problem
with running the tests on different versions of
Linux, where the actual error message from the
program run by the test might be different.


tiran (approved):
That's a great idea!

PR #306, opened 2024-08-07
Fix: Network hiccups using Retry logic

rd4398:
Fixes: https://github.com/python-wheel-build/fromager/issues/222
As per discussion here: https://github.com/python-wheel-build/fromager/pull/228

### DIFF
--- a/src/fromager/sdist.py
+++ b/src/fromager/sdist.py
@@ -311,7 +311,7 @@ def download_wheel(
     wheel_filename = output_directory / os.path.basename(urlparse(wheel_url).path)
     if not wheel_filename.exists():
         logger.info(f"{req.name}: downloading pre-built wheel {wheel_url}")
-        wheel_filename = _download_wheel_check(output_directory, wheel_url)
+        wheel_filename = _download_wheel_check(ctx, output_directory, wheel_url)
         logger.info(f"{req.name}: saved wheel to {wheel_filename}")
     else:
         logger.info(f"{req.name}: have existing wheel {wheel_filename}")
@@ -321,8 +321,8 @@ def download_wheel(
 
 # Helper method to check whether the .whl file is a zip file and has contents in it.
 # It will throw BadZipFile exception if any other file is encountered. Eg: index.html
-def _download_wheel_check(destination_dir, wheel_url):
-    wheel_filename = sources.download_url(destination_dir, wheel_url)
+def _download_wheel_check(ctx, destination_dir, wheel_url):
+    wheel_filename = sources.download_url(ctx, destination_dir, wheel_url)
     wheel_directory_contents = zipfile.ZipFile(wheel_filename).namelist()
     if not wheel_directory_contents:
         raise zipfile.BadZipFile(f"Empty zip file encountered: {wheel_filename}")
--- a/src/fromager/sources.py
+++ b/src/fromager/sources.py
@@ -10,7 +10,6 @@ import typing
 import zipfile
 from urllib.parse import urlparse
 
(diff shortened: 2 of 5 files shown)


src/fromager/sdist.py
```diff
@@ -311,7 +311,7 @@ def download_wheel(
     wheel_filename = output_directory / os.path.basename(urlparse(wheel_url).path)
     if not wheel_filename.exists():
         logger.info(f"{req.name}: downloading pre-built wheel {wheel_url}")
-        wheel_filename = _download_wheel_check(output_directory, wheel_url)
+        wheel_filename = _download_wheel_check(output_directory, wheel_url, ctx)
```
dhellmann, line 314: We have the context as the first argument to functions everywhere else, could you follow that pattern here, too?

src/fromager/sources.py
```diff
@@ -165,7 +164,7 @@ def default_download_source(
     )
 
     source_filename = _download_source_check(
-        ctx.sdists_downloads, url, destination_filename
+        ctx.sdists_downloads, url, ctx, destination_filename
```
dhellmann, line 167: Same comment about argument ordering.

src/fromager/sources.py
```diff
@@ -213,7 +218,7 @@ def download_url(
         return outfile
     # Open the URL first in case that fails, so we don't end up with an empty file.
     logger.debug(f"reading from {url}")
-    with requests.get(url, stream=True) as r:
+    with ctx.requests.get(url, stream=True) as r:
```
dhellmann, line 221: How does retrying work with streaming? If it starts over, the caller needs to understand that so it doesn't just write duplicate copies of the data to the same file, right?

src/fromager/sdist.py
rd4398: okay, done!

src/fromager/sources.py
rd4398: okay, done!

src/fromager/sources.py
rd4398: Yeah, need to look at this

rd4398: Found this: https://stackoverflow.com/questions/25860105/python-requests-package-lost-connection-while-streaming but not sure whether it will help our cause

dhellmann: https://gist.github.com/tobiasraabe/58adee67de619ce621464c1a6511d7d9 looks potentially useful. The `Range` HTTP header tells the server where to start sending data, based on how much has already been downloaded. https://medium.com/@lope.ai/downloading-files-over-http-with-python-requests-e12e6b795e43 shows another example of tracking that.

rd4398: Agreed
I have tried to use Retry header in the latest version of this PR

src/fromager/sources.py
```diff
@@ -209,13 +214,26 @@ def download_url(
...
+            # Set the range header
+            headers = {"Range": f"bytes={outfile_size_on_disk}-"}
+    else:
+        outfile_size_on_disk = 0
+        headers = {}
+
     # Open the URL first in case that fails, so we don't end up with an empty file.
     logger.debug(f"reading from {url}")
-    with requests.get(url, stream=True) as r:
+    with ctx.requests.get(url, stream=True, headers=headers) as r:
```
dhellmann, line 234: I think we're going to need a loop here to handle the retry.

Let's set this aside for now and come back when we have more time to work on it together.

shubhbapna:
Converting to draft based on the last review comment

PR #307, opened 2024-08-07
feat: add new `build_dir` field in settings

### LINKED ISSUE #278: configuration option for build dir relative to source root
Downstream we have several packages that unpack a source archive and then need to build in a subdirectory of that. It would be convenient to have a configuration option for that.

shubhbapna:
fixes #278 

had to tell mypy to ignore `settings.py` until #257

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -455,7 +455,9 @@ To use predefined urls to download sources from, instead of overriding the entir
 - `version` - it is replaced by the version returned by the resolver  
 - `canonicalized_name` - it is replaced by the canonicalized name of the package specified in the requirement, specifically it applies `canonicalize_name(req.nam)`  
 
-Additionally, the source distribution index server used by the package resolver can be overriden for a particular package. The resolver can also be told to whether include wheels or sdist sources while trying to resolve the package. Templating is not supported here.
+The source distribution index server used by the package resolver can be overriden for a particular package. The resolver can also be told to whether include wheels or sdist sources while trying to resolve the package. Templating is not supported here.  
+
+A `build_dir` field can also be defined to indicate to fromager where the package should be build relative to the sdist root directory
 
 ```yaml
 packages:
@@ -467,4 +469,5 @@ packages:
             sdist_server_url: "https://pypi.org/simple"
             include_wheels: true
             include_sdists: false
+        build_dir: directory name relative to sdist directory, defaults to an empty string, which means to use the sdist directory
 ```
--- a/src/fromager/settings.py
+++ b/src/fromager/settings.py
@@ -82,6 +82,20 @@ class Settings:
             resolve_dist.get("include_sdists"), default
         )
 
+    def build_dir(self, pkg: str, sdist_root_dir: pathlib.Path) -> pathlib.Path:
+        p = self.get_package_settings(pkg)
+        if p.get("build_dir"):
+            input_build_dir = pathlib.Path(p.get("build_dir"))
+            # hack: make path absolute to ensure that any directory escaping is contained
+            if ".." in str(input_build_dir):
+                raise ValueError(
+                    f"{pkg}: build dir {input_build_dir} defined in settings is not relative to sdist root dir"
+                )
+            # ensure that absolute build_dir path from settings is converted to a relative path
+            relative_build_dir = input_build_dir.relative_to(input_build_dir.anchor)
+            return sdist_root_dir / relative_build_dir
+        return sdist_root_dir
+
     def get_package_settings(self, pkg: str) -> dict[str, dict[str, str]]:
         p = self.packages()
         return self._return_value_or_default(
--- a/src/fromager/sources.py
+++ b/src/fromager/sources.py
@@ -404,7 +404,9 @@ def build_sdist(
     sdist_root_dir: pathlib.Path,
     build_env: wheels.BuildEnvironment,
 ) -> pathlib.Path:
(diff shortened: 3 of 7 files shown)


docs/customization.md
```diff
@@ -467,4 +469,5 @@ packages:
             sdist_server_url: "https://pypi.org/simple"
             include_wheels: true
             include_sdists: false
+        build_dir: directory name relative to sdist directory
```
dhellmann, line 472:
```suggestion
        build_dir: directory name relative to sdist directory, defaults to an empty string, which means to use the sdist directory
```

src/fromager/settings.py
```diff
@@ -82,6 +82,12 @@ def resolver_include_sdists(
             resolve_dist.get("include_sdists"), default
         )
 
+    def build_dir(self, pkg: str, sdist_root_dir: pathlib.Path) -> pathlib.Path:
+        p = self.get_package_settings(pkg)
+        if p.get("build_dir"):
+            return sdist_root_dir / p.get("build_dir")
```
dhellmann, line 88: We should do something here to ensure that this relative path does not go outside of the sdist_root_dir. For example, if sdist_root_dir is `/foo/bar` and someone sets the build_dir to `/tmp/foo` we need to ensure that resolves to `/foo/bar/tmp/foo` and not `/tmp/foo`.

```
>>> import pathlib
>>> p = pathlib.Path('/foo/bar')
>>> t = pathlib.Path('/tmp/blah')
>>> p / t
PosixPath('/tmp/blah')
```

src/fromager/sources.py
```diff
@@ -414,7 +416,7 @@ def build_sdist(
         extra_environ=extra_environ,
         req=req,
         version=version,
-        sdist_root_dir=sdist_root_dir,
+        sdist_root_dir=ctx.settings.build_dir(req.name, sdist_root_dir),
```
dhellmann, line 419: I'm trying to think through if there are any cases where an override plugin would need to know both the root of the source tree and the directory where it should do the build. For example, the pyarrow build? Maybe we just don't use the configuration option in that case, though.

What do you think?

shubhbapna: Yeah, i think if the user needs both then they shouldn't use this option and use plugins instead

shubhbapna: or maybe for now i don't use this option for `build_wheel` which will allow us to use this feature for the pyarrow builds?

dhellmann: I think we'd want to be consistent. After prepare_source, all overrides should either get both paths or only one of them.

shubhbapna: i think then lets just return the build dir

src/fromager/settings.py
```diff
@@ -82,6 +82,15 @@ def resolver_include_sdists(
             resolve_dist.get("include_sdists"), default
         )
 
+    def build_dir(self, pkg: str, sdist_root_dir: pathlib.Path) -> pathlib.Path:
+        p = self.get_package_settings(pkg)
+        if p.get("build_dir"):
+            # this ensures that the build_dir path from settings is relative to sdist_root_dir
+            input_build_dir = pathlib.Path(p.get("build_dir"))
+            relative_build_dir = input_build_dir.relative_to(input_build_dir.anchor)
```
dhellmann, line 90: I think we're going to need some unit tests to verify the cases for this. I was able to get it to throw an exception:

```
>>> pathlib.Path('/var/tmp').relative_to(pathlib.Path('bah/humbug').anchor)
  File "/usr/lib64/python3.9/pathlib.py", line 939, in relative_to
    raise ValueError("{!r} is not in the subpath of {!r}"
ValueError: '/var/tmp' is not in the subpath of '' OR one path is relative and the other is absolute.
```

src/fromager/sources.py
dhellmann: Yeah, I think that works. For the pyarrow case, we'll need to compute the path relative to that directory to build the stuff outside of the python directory, but that's OK.

src/fromager/settings.py
shubhbapna: oh both the paths should be the same this example. I will still add some unit tests

shubhbapna:
@dhellmann now that i think about this, should i just apply this transformation: `ctx.settings.build_dir(req.name, sdist_root_dir)` in `prepare_source` and `find_source_dir`. What do you think?

shubhbapna:
Handling the case where user tries to escape the source root dir is a bit tough since ".." can appear anywhere in the path for example: "foo/../../../bar" will allow me to escape outside the source root dir. So I used a hack - if ".." appears anywhere in the path then resolve it to an absolute path and since I have code that can handle absolute path we are able to contain the directory escaping

src/fromager/settings.py
```diff
@@ -82,6 +82,18 @@ def resolver_include_sdists(
             resolve_dist.get("include_sdists"), default
         )
 
+    def build_dir(self, pkg: str, sdist_root_dir: pathlib.Path) -> pathlib.Path:
+        p = self.get_package_settings(pkg)
+        if p.get("build_dir"):
+            input_build_dir = pathlib.Path(p.get("build_dir"))
+            # hack: make path absolute to ensure that any directory escaping is contained
+            if ".." in str(input_build_dir):
+                input_build_dir = input_build_dir.resolve()
```
dhellmann, line 91: We could just raise an exception in this case instead of producing a value that's hard to predict.

dhellmann (approved):

PR #308, opened 2024-08-07
fix: use external_commands.run to apply patch

### LINKED ISSUE #280: use external_command.run() for applying patches
When we run the patch command we don't log the command line or its output because we call subprocess.check_run() instead of our external_process.run() wrapper.

_Originally posted by @dhellmann in https://github.com/python-wheel-build/fromager/pull/255#discussion_r1698736462_

shubhbapna:
fixes #280


dhellmann:
This looks good. I don't see an e2e test in the repo for building something with a patch, so it would be good to add one.

dhellmann (approved):
Good idea to reuse an existing test script for the new case.

PR #309, opened 2024-08-07
fix: expose `default_patch_source` and return `is_new`

### LINKED ISSUE #301: make _default_prepare_source function public
We have a few examples of overrides that want to extend, but not completely replace, the behavior of the default implementation for prepare_source. We should make the existing default public. 

We should also change the API to return the is_new flag so that those overrides can tell when a new source directory has been created so they can avoid repeating operations that may not be idempotent.

shubhbapna:
fixes #301

### DIFF
--- a/src/fromager/sources.py
+++ b/src/fromager/sources.py
@@ -369,32 +369,40 @@ def prepare_source(
     version: Version,
 ) -> pathlib.Path:
     logger.info(f"{req.name}: preparing source for {req} from {source_filename}")
-    source_root_dir = overrides.find_and_invoke(
+    prepare_source_details = overrides.find_and_invoke(
         req.name,
         "prepare_source",
-        _default_prepare_source,
+        default_prepare_source,
         ctx=ctx,
         req=req,
         source_filename=source_filename,
         version=version,
     )
+    if not isinstance(prepare_source_details, tuple):
+        source_root_dir = prepare_source_details
+    elif len(prepare_source_details) == 2:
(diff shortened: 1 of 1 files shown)


src/fromager/sources.py
```diff
@@ -369,10 +369,10 @@ def prepare_source(
     version: Version,
 ) -> pathlib.Path:
     logger.info(f"{req.name}: preparing source for {req} from {source_filename}")
-    source_root_dir = overrides.find_and_invoke(
+    source_root_dir, _ = overrides.find_and_invoke(
```
dhellmann, line 372: This will require changing the return value of all existing plugins to include the flag. We could have some more complicated code here to unpack the return value if it's a tuple or not if it's just a single string, and avoid breaking existing implementations.

dhellmann (changes requested):

src/fromager/sources.py
```diff
@@ -369,32 +369,40 @@ def prepare_source(
...
         req.name,
         "prepare_source",
-        _default_prepare_source,
+        default_prepare_source,
         ctx=ctx,
         req=req,
         source_filename=source_filename,
         version=version,
     )
+    if len(prepare_source_details) == 1:
```
dhellmann, line 381: If there's only 1 return value, it won't be a tuple, it'll just be that value. I think you'll need to use `isinstance()` in the checks.

shubhbapna: ahh yes

dhellmann (approved):

PR #310, opened 2024-08-08
cleanup: remove unused functions and overrides

### LINKED ISSUE #305: unused `default_get_install_dependencies` and not using the customization `get_install_dependencies`
I couldn't find any references of `default_get_install_dependencies`:  https://github.com/search?q=repo%3Apython-wheel-build%2Ffromager%20default_get_install_dependencies&type=code (apart from the docs)

Similarly for the custom override `get_install_dependencies` I can't find any reference to it:  https://github.com/search?q=repo%3Apython-wheel-build%2Ffromager+get_install_dependencies&type=code

It seems like we are using `get_install_dependencies_of_wheel` over these 2 functions: https://github.com/python-wheel-build/fromager/blob/main/src/fromager/sdist.py#L269

shubhbapna:
fixes #305


dhellmann (approved):
This makes sense. We want the wheel file to reflect the correct installation dependencies, so prepare_source and build_wheel should ensure that happens.

PR #314, opened 2024-08-11
add package-specific settings files

dhellmann:
Support settings for individual packages in separate files for easier
sharing and management. This also makes it easier to set up jobs to test
builds of specific packages when their settings change, without building
packages that have no changes.

### DIFF
--- a/src/fromager/settings.py
+++ b/src/fromager/settings.py
@@ -23,10 +23,7 @@ class Settings:
         return set(overrides.pkgname_to_override_module(n) for n in names)
 
     def packages(self) -> dict[str, dict[str, str]]:
-        p = self._return_value_or_default(self._data.get("packages"), {})
-        return {
-            overrides.pkgname_to_override_module(key): value for key, value in p.items()
-        }
+        return self._return_value_or_default(self._data.get("packages"), {})
 
     def download_source_url(
         self,
--- a/src/fromager/settings.py
+++ b/src/fromager/settings.py
@@ -148,16 +148,47 @@ def _resolve_template(
         raise
 
 
-def _parse(content: str) -> Settings:
-    data = yaml.safe_load(content)
-    return Settings(data)
+def load(settings_file: pathlib.Path, settings_dir: pathlib.Path) -> Settings:
+    settings_data = {}
 
-
-def load(filename: pathlib.Path) -> Settings:
-    filepath = pathlib.Path(filename)
+    filepath = pathlib.Path(settings_file)
     if not filepath.exists():
         logger.debug("settings file %s does not exist, ignoring", filepath.absolute())
-        return Settings({})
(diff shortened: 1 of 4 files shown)


src/fromager/settings.py
```diff
@@ -143,16 +143,46 @@ def _resolve_template(
...
+    if "packages" not in settings_data:
+        settings_data["packages"] = {}
+    else:
+        new_packages = {}
+        for name, value in settings_data["packages"].items():
+            package_name = overrides.pkgname_to_override_module(name)
+            new_packages[package_name] = value
+            package_settings_from[package_name] = settings_file
+
+    for package_file in sorted(settings_dir.glob("*.yaml")):
```
tiran, line 170: How do you feel about a separate directory for package settings like `settings.d`, `packages.d` or just `packages`?
```suggestion
    for package_file in sorted(settings_dir.glob("packages.d/*.yaml")):
```

dhellmann: I called it just `settings` to be consistent with our other names, like `env`.

I could see us eventually removing this global file and moving the pre-built flag into these files.

src/fromager/settings.py
```diff
@@ -143,16 +143,46 @@ def _resolve_template(
...
+    # will be used to look the value up. Transform any existing keys to that
+    # format so we can warn if there are overriding values.
+    package_settings_from = {}
+    if "packages" not in settings_data:
+        settings_data["packages"] = {}
+    else:
+        new_packages = {}
+        for name, value in settings_data["packages"].items():
+            package_name = overrides.pkgname_to_override_module(name)
+            new_packages[package_name] = value
```
shubhbapna, line 167: I think the `new_packages` variable is not being used anymore?

dhellmann: Ooo, that's a bug. I updated the test in the new version of the PR and made sure to replace the updated packages section.

shubhbapna (commented):
We can probably change this function now since are already doing the canonicalization while parsing
https://github.com/python-wheel-build/fromager/pull/314/files#diff-c369c9fdcaebf0093b83132356c90dff71d762e0ced16bee8fa37d0d96259ce4R25

```python
    def packages(self) -> dict[str, dict[str, str]]:
        return self._return_value_or_default(self._data.get("packages"), {})
```

dhellmann:
> We can probably change this function now since we are already doing the canonicalization while parsing https://github.com/python-wheel-build/fromager/pull/314/files#diff-c369c9fdcaebf0093b83132356c90dff71d762e0ced16bee8fa37d0d96259ce4R25
> 
> ```python
>     def packages(self) -> dict[str, dict[str, str]]:
>         return self._return_value_or_default(self._data.get("packages"), {})
> ```

I added a commit to make this update.

PR #320, opened 2024-08-12
feat: add changelog field to settings

### LINKED ISSUE #317: Add package-version-specific changelog to settings
- Add a new field `changelog` to the per package setting
- Include a way to specify versions
- We just care about the length of the changelog (the ordering of the entries doesn't matter to us. it is on the user to maintain the ordering)

Agreed upon template:
```yaml
torch:
  changelog:
    "2.3.1":
      - rebuild for xyz

```

shubhbapna:
fixes #317

### DIFF
--- a/src/fromager/settings.py
+++ b/src/fromager/settings.py
@@ -96,6 +96,11 @@ class Settings:
             return sdist_root_dir / relative_build_dir
         return sdist_root_dir
 
+    def build_tag(self, pkg: str, version: Version | str) -> int:
+        p = self.get_package_settings(pkg)
+        changelog = p.get("changelog", {}).get(str(version), [])
+        return len(changelog)
+
     def get_package_settings(self, pkg: str) -> dict[str, dict[str, str]]:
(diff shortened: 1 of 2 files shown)


src/fromager/settings.py
```diff
@@ -96,6 +96,11 @@ def build_dir(self, pkg: str, sdist_root_dir: pathlib.Path) -> pathlib.Path:
             return sdist_root_dir / relative_build_dir
         return sdist_root_dir
 
+    def build_tag(self, pkg: str, version: Version | str) -> int:
```
dhellmann, line 99: I think we said we wanted the variant to be part of this query, too.

shubhbapna: Oh i thought we were doing the per variant settings after pydantic lands in so that we could port all the fields

dhellmann: OK, we can do that. I didn't have that sequencing in mind.

dhellmann (approved):

PR #323, opened 2024-08-13
Define Pydantic model for package config

tiran:
A package config is a YAML file that contains all settings for a Python package. It combines a `packages` entry and variant `pre_built` flag from `settings.yaml` with env files for variants.


src/fromager/models.py
```diff
@@ -0,0 +1,244 @@
...
+                str, when_used="json"
+            ),
+        )
+
+
+# Changelog
+Changelog = dict[Version, list[str]]
+
+# environment variables map
+EnvVars = dict[str, str | int | float | bool]
```
dhellmann, line 65: Environment variables are always actually strings, right?

src/fromager/models.py
```diff
@@ -0,0 +1,244 @@
...
+
+    Download package sources from an alternative source, e.g. GitHub release.
+    """
+
+    model_config = MODEL_CONFIG
+
+    url: UrlTemplate | None = None
+    """Source download url (string template)"""
+
+    destination_filename: str | None = None
```
dhellmann, line 109: We want to make sure this is a valid filename, and has no path components. Maybe we can add that in a separate PR, once we have the basics in place.

src/fromager/models.py
```diff
@@ -0,0 +1,244 @@
...
+    @classmethod
+    def from_file(cls, filename: pathlib.Path) -> "PackageSettings":
+        """Load from file
+
+        Raises :exc:`FileNotFound` when the file is not found.
+        The package name is taken from the stem of the file name.
+        """
+        filename = filename.absolute()
+        raw_yaml = filename.read_text(encoding="utf-8")
+        return cls.from_string(filename.stem, raw_yaml)
```
dhellmann, line 210: If this fails, the error message won't include a reference to the filename.
```suggestion
        try:
          return cls.from_string(filename.stem, raw_yaml)
        except Exception as err:
          raise RuntimeError(f"failed to load settings from {filename}: {err}") from err
```

src/fromager/models.py
```diff
@@ -0,0 +1,244 @@
...
+
+        extra_environ: dict[str, str] = {}
+        for key, value in entries:
+            if isinstance(value, bool):
+                value = "1" if value else "0"
+            elif isinstance(value, int | float):
+                value = str(value)
+            value = value.strip()
+            if "$(" in value:
+                raise ValueError(f"'{value}': subshell '$()' is not supported.")
```
dhellmann, line 235: In another PR, we should move this validation to when the file is read. Otherwise it might not be triggered for quite a while into a build process.

src/fromager/models.py
tiran: Yes, but the idea is that we accept int, float, and bool values in YAML files and convert them to strings. I have now switch to a custom str type with a before validator. The validator converts int, str, bool to str. It also checks for `$(` in str.

src/fromager/models.py
tiran: I'm adding a simple validator that checks for `os.pathsep not in value`

src/fromager/models.py
tiran: Good idea!

src/fromager/models.py
dhellmann: OK, that makes sense.

src/fromager/models.py
```diff
@@ -0,0 +1,262 @@
...
+
+    url: UrlTemplate | None = None
+    """Source download url (string template)"""
+
+    destination_filename: str | None = None
+    """Rename file (filename without path)"""
+
+    @pydantic.field_validator("destination_filename")
+    @classmethod
+    def name_must_contain_space(cls, v):
```
dhellmann, line 125: I'm confused by this function name.

dhellmann (commented):
This looks good.

What's the next step? Updating settings.py to use it?

shubhbapna (commented):
Could we also add a method to resolve templates? https://github.com/python-wheel-build/fromager/blob/main/src/fromager/settings.py#L55

IMO, the caller doesn't knows what template variables are supported by the settings and so it should be on the settings to provide an API to resolve the template

src/fromager/models.py
tiran: Copy and paste...

src/fromager/models.py
tiran: The input is now validated when the config file is loaded.

src/fromager/settings.py
shubhbapna, line 1: Do we still need this settings class? Can we use the packagesettings directly?

tiran: Eventually we can get rid of the settings object entirely. Or we may keep it for global settings.

dhellmann (commented):
This is really close. Other recent conversations mean I have questions about a global config file after the migration, but otherwise I think we're ready to merge it.

shubhbapna:
I was thinking that since a variant once passed to fromager doesn't change and we have access to the variant before we load the settings, we can update settings to only give back data that is applicable to that variant? The variant variable can be passed here to settings: https://github.com/python-wheel-build/fromager/blob/main/src/fromager/__main__.py#L187

dhellmann:
> I was thinking that since a variant once passed to fromager doesn't change and we have access to the variant before we load the settings, we can update settings to only give back data that is applicable to that variant? The variant variable can be passed here to settings: https://github.com/python-wheel-build/fromager/blob/main/src/fromager/__main__.py#L187

I like that. Maybe we can add that in a follow-up?

tiran:
> I was thinking that since a variant once passed to fromager doesn't change and we have access to the variant before we load the settings, we can update settings to only give back data that is applicable to that variant? The variant variable can be passed here to settings: https://github.com/python-wheel-build/fromager/blob/main/src/fromager/__main__.py#L187

That's already implemented. The Pydantic model `PackageSettings` is an internal implementation details. The rest of Fromager uses `PackageBuildInfo`, which wraps `PackageSettings` and variant from `context.WorkContext`.

docs/customization.md
```diff
@@ -21,60 +21,99 @@ Set the variant using the `--variant` command line option or the
...
+    destination_filename: "${canonicalized_name}-${version}.tar.gz"
+resolver_dist:
+    sdist_server_url: "https://pypi.org/simple"
+    include_wheels: true
+    include_sdists: false
+build_dir: directory name relative to sdist directory, defaults to an empty string, which means to use the sdist directory
+env:
+    USE_FFMPEG: "0"
+    USE_LEVELDB: "0"
+    USE_LMDB: "1"
```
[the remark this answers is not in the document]
tiran: Good point! I have tests in place that ensure the YAML mappings are loaded and processed in order.

docs/customization.md
```diff
@@ -21,60 +21,99 @@ Set the variant using the `--variant` command line option or the
...
+
+Package settings are read from their own file in the directory specified with
+`--settings-dir`. Files should be named using the canonicalized form of the
+package name with the suffix `.yaml`, e.g. `torch.yaml` or
+`llama_cpp_python.yaml`. Files are read in lexicographical order.
+
+### Example config
+
+```yaml
+# settings/torch.yaml
```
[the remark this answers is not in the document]
tiran: Right now I'm using `--settings-dir` for package `yaml` files. We could rename the option or standardize on `settings/packages`. It's a simple search and replace operation.

I have removed `packages` block from global settings. I can add it back if you like.

tiran:
PR is almost ready. I have to port two tests to the new API.

docs/customization.md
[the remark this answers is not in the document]
dhellmann: I forgot we had separate flags for the settings file and settings dir. As long as we keep both, I think we're OK.

At some point, I'd like to simplify the CLI options to just point to the overrides directory and have everything else use a standard layout under that path. That doesn't have to be part of this change, of course.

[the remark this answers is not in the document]
tiran: I agree! A `--settings-dir` option to a directory with `settings.yaml`, `settings/*.yaml`, and `patches`?

dhellmann (commented):
This looks really close to me. I had a couple of comments about cleanups, and one question about what I think is a change in behavior but might just be a subtlety of the new API.

src/fromager/commands/bootstrap.py
```diff
@@ -55,9 +55,9 @@ def bootstrap(
         )
     logger.info("bootstrapping %r variant of %s", wkctx.variant, to_build)
 
-    pre_built = wkctx.settings.pre_built(wkctx.variant)
+    pre_built = wkctx.settings.list_pre_built()
```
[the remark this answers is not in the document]
tiran: `Settings` has a reference to the context's `variant` and filters by variant internally.

src/fromager/packagesettings.py
```diff
@@ -626,27 +634,42 @@ def package_build_info(self, package: str | Package) -> PackageBuildInfo:
...
             Package(canonicalize_name(name, validate=True)) for name in exts.names()
         )
+
         # patches
         for patchfile in self.patches_dir.glob("*/*.patch"):
             # parent directory has format "package-version"
             name = patchfile.parent.name.rsplit("-", 1)[0]
             packages.add(Package(canonicalize_name(name, validate=True)))
 
+        if not test:
```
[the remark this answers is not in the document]
tiran: I have removed the code

src/fromager/wheels.py
```diff
@@ -210,7 +211,11 @@ def add_extra_metadata_to_wheels(
                 )
                 data_to_add = {}
 
-        settings = ctx.settings.get_package_settings(req.name)
+        build_file = dist_info_dir / "fromager-build-settings"
```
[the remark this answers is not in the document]
tiran: It was a merge conflict resolution gone wrong. I have addressed the issue.

src/fromager/commands/bootstrap.py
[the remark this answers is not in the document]
dhellmann: OK, good, that covers my concern.

dhellmann (commented):
I'm happy with this version

I see a test failure, and I want to give @shubhbapna a chance to review, too, so I'll leave it open for now.

shubhbapna (approved):

PR #325, opened 2024-08-13
feat: add build tags to built wheels

### LINKED ISSUE #318: Change filename of wheel with build number based on length of changelog
- Bump the build tag of wheel name based on the length of changelog: https://packaging.python.org/en/latest/specifications/binary-distribution-format/#file-name-convention
- Start all of the actual builds with 0 to differentiate them from upstream packages (need to test that pip and gitlab pypi can handle this)
- Do not change wheel version metadata
- Make any changes to resolver code and finder code to account for this new wheel file name
  - Resolver in fromager should not look at the build number because it can’t use build number to pick a specific artifact. The caller should look at the wheel filename returned by the resolver

shubhbapna:
fixes #318


e2e/test_build_order.sh
shubhbapna, line 1: I have updated this e2e test to mimic our downstream build scenario where we are using our private index. Here the "private" index is the local pypi mirror

src/fromager/commands/build.py
```diff
@@ -165,13 +175,25 @@ def _build(
...
+        url, _ = sdist.resolve_prebuilt_wheel(wkctx, req, [wkctx.wheel_server_url])
+        build_tag = wkctx.settings.build_tag(req.name, dist_version)
+        wheel_filename = urlparse(url).path.rpartition("/")[-1]
+        _, _, build_tag_from_name, _ = parse_wheel_filename(wheel_filename)
+        if build_tag_from_name:
+            increment_by = max(build_tag_from_name[0] - build_tag, 0)
+            if increment_by > 0:
+                increment_by += 1  # make the new build tag greater than the latest one
+                logger.warn(
+                    f"{dist_name}: changelog for version {dist_version} is inconsistent. Found build tag {build_tag_from_name[0]} but expected {build_tag}. Will tag the wheel with build number {build_tag + increment_by}"
```
shubhbapna, line 192: This is to solve this problem: https://github.com/python-wheel-build/fromager/issues/318#issuecomment-2287160456

We will warn about the changelog and then use the latest build tag + 1 to build the wheel

src/fromager/finders.py
```diff
@@ -90,29 +90,33 @@ def find_sdist(
...
 
 def find_wheel(
+    ctx: context.WorkContext,
     downloads_dir: pathlib.Path,
     req: Requirement,
     dist_version: str,
 ) -> pathlib.Path | None:
     filename_prefix = _dist_name_to_filename(req.name)
     canonical_name = canonicalize_name(req.name)
+    build_tag = ctx.settings.build_tag(req.name, dist_version)
```
dhellmann, line 100: I wonder if the caller should pass this in, instead of us computing it inside? It feels a bit like the version. Maybe that means changing more places?

e2e/test_bootstrap.sh
```diff
@@ -10,12 +10,14 @@
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 source "$SCRIPTDIR/common.sh"
 
+# passing settings to bootstrap but should have 0 effect on it
```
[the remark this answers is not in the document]
shubhbapna: I think in our meeting we decided that the default build tag would be 0 for all the built wheels. Currently this applies to bootstrap as well

e2e/test_bootstrap_build_tags.sh
```diff
@@ -0,0 +1,91 @@
...
+
+# check if tmp dir was created
+if [[ ! "$TEMP_DIR" || ! -d "$TEMP_DIR" ]]; then
+  echo "Could not create temp dir"
+  exit 1
+fi
+
+SETTINGS_FILE_WITH_ADDED_CHANGELOG=$TEMP_DIR/changelog_settings_$RANDOM.yaml
+cp $SCRIPTDIR/changelog_settings.yaml $SETTINGS_FILE_WITH_ADDED_CHANGELOG
+echo -e "- test2" >> $SETTINGS_FILE_WITH_ADDED_CHANGELOG
```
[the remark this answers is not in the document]
shubhbapna: It is being created in a temp directory. It is not changing any file in the repository itself

src/fromager/commands/build.py
```diff
@@ -165,13 +175,25 @@ def _build(
...
     try:
         logger.info(f"{req.name}: checking if {req} was already built")
-        sdist.resolve_prebuilt_wheel(wkctx, req, [wkctx.wheel_server_url])
-        return True
+        url, _ = sdist.resolve_prebuilt_wheel(wkctx, req, [wkctx.wheel_server_url])
+        build_tag = wkctx.settings.build_tag(req.name, dist_version)
+        wheel_filename = urlparse(url).path.rpartition("/")[-1]
+        _, _, build_tag_from_name, _ = parse_wheel_filename(wheel_filename)
+        if build_tag_from_name:
+            increment_by = max(build_tag_from_name[0] - build_tag, 0)
```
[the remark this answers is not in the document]
shubhbapna: The issue was that `build_wheel` was getting the build tag from `settings.build_tag`. I am passing the `increment_by` as an amount that needs to be added to `settings.build_tag`. I guess I should maybe change the `build_wheel` command to take a `build_tag` as an argument instead of getting it directly from the settings

src/fromager/finders.py
shubhbapna: I could do that, optionally pass `build_tag`

e2e/test_bootstrap.sh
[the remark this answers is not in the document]
shubhbapna: I think bootstrap should run with whatever settings are there, including adding any build tags if needed

src/fromager/commands/build.py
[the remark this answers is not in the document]
shubhbapna: now that i think of it all the time the `build_tag` should only come from settings. Only in this edge case we need a different `build_tag`. I think it might make more sense to add an optional `override_build_tag` parameter to `build_wheel` instead so that i don't have to change a lot of code

e2e/test_bootstrap.sh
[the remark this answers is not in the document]
dhellmann: That will keep the code simpler. It also means if settings change so that a new build tag is needed, that wheel build will be testable with the bootstrap.

shubhbapna:
@dhellmann as discusse fromager will now raise an error if the changelog value is lesser than the latest build tag on the index

src/fromager/commands/build.py
```diff
@@ -170,8 +172,19 @@ def _is_wheel_built(
...
+        _, _, build_tag_from_name, _ = parse_wheel_filename(wheel_filename)
+        if build_tag_from_name:
+            if build_tag_from_name[0] > build_tag:
+                raise ValueError(
+                    f"{dist_name}: changelog for version {dist_version} is inconsistent. Found build tag {build_tag_from_name[0]} but expected {build_tag}"
+                )
+            return build_tag_from_name[0] == build_tag
+
+        # build tag == 0 is equivalent to having no build tag
+        return build_tag == 0
```
dhellmann, line 187: This logic is correct, but not as clear as it could be. See what you think of these changes.

```suggestion
        _, _, build_tag_from_name, _ = parse_wheel_filename(wheel_filename)
        if build_tag_from_name:
          existing_build_tag = build_tag_from_name[0]
        else:
          existing_build_tag = 0  # default to 0, even if it is not in the name

        if existing_build_tag > build_tag:
            raise ValueError(
                    f"{dist_name}: changelog for version {dist_version} is inconsistent. Found build tag {existing_build_tag} but expected {build_tag}"
                )

        return existing_build_tag == build_tag
```

shubhbapna: I like it, will update

dhellmann (commented):
I'm happy with this version. It's going to be a big change to introduce downstream, so I'd like to hold off for a day or two. That will give @tiran time to review it, too.

src/fromager/candidate.py
```diff
@@ -28,12 +28,14 @@ def __init__(
         url: str,
         extras: typing.Iterable[str] | None = None,
         is_sdist: bool | None = None,
+        build_tag: int | None = None,
```
tiran, line 31: I recommend to use a new, user-defined type for build tag. Build tags are not limited to numbers. They can be any string. pip sorts build tags by converting the leading digits to an integer and then sorting by tuple `(int(digits), remainder)`. In the future we can use the build tag to annotate the build flavor, e.g. `0.cpu_ubi9`.

src/fromager/commands/build.py
```diff
@@ -170,8 +172,19 @@ def _is_wheel_built(
 
     try:
         logger.info(f"{req.name}: checking if {req} was already built")
-        sdist.resolve_prebuilt_wheel(wkctx, req, [wkctx.wheel_server_url])
-        return True
+        url, _ = sdist.resolve_prebuilt_wheel(wkctx, req, [wkctx.wheel_server_url])
+        build_tag = wkctx.settings.build_tag(req.name, dist_version)
+        wheel_filename = urlparse(url).path.rpartition("/")[-1]
```
tiran, line 177:
```suggestion
        wheel_filename = urlparse(url).path.rsplit("/", 1)[-1]
```

src/fromager/wheels.py
```diff
@@ -308,3 +309,23 @@ def default_build_wheel(
...
+    build_tag: int = 0,
+) -> pathlib.Path:
+    cmd = [
+        "wheel",
+        "tags",
+        "--build",
+        str(build_tag),
+        "--remove",
+        str(wheel_filename),
+    ]
```
tiran, line 326: `wheel pack` can also add a build tag. For efficiency, I recommend to fold this feature into `add_extra_metadata_to_wheel`. `wheel tag` unpacks the wheel file and repacks a new wheel file with a build tag. Every unpack and pack operation also validates or updates the SHA256 digest of every file. It's slow for large wheels like PyTorch.

src/fromager/commands/build.py
shubhbapna: Took this from the resolver. Will update it there as well

src/fromager/candidate.py
shubhbapna: Should settings also return the `BuildTag` type then for now it could just be `(int(digits), "")`?

e2e/test_bootstrap_build_tags.sh
```diff
@@ -0,0 +1,83 @@
...
+fromager \
+  --log-file="$OUTDIR/bootstrap.log" \
+  --error-log-file="$OUTDIR/fromager-errors.log" \
+  --sdists-repo="$OUTDIR/sdists-repo" \
+  --wheels-repo="$OUTDIR/wheels-repo" \
+  --work-dir="$OUTDIR/work-dir" \
+  --settings-file="$SCRIPTDIR/bootstrap_settings.yaml" \
+  bootstrap 'stevedore==5.1.0'
+
+# Check for log message that the override is loaded
```
dhellmann, line 43: This comment isn't accurate. You could probably just delete it, since the error messages make clear what's expected.

e2e/changelog_settings.yaml
```diff
@@ -0,0 +1,6 @@
+packages:
+  stevedore:
+    changelog:
+      "5.1.0":
+        - test 1
+
```
dhellmann, line 6: Relying on this file to retain trailing whitespace on this line is going to be brittle. My editor is configured to automatically strip whitespace. Please update the line that edits the file to include the needed whitespace to indent the new entry properly.

e2e/test_build_order.sh
```diff
@@ -1,42 +1,55 @@
...
-# not available when setting up to build a package.
+# Test build-sequence (with and without skipping already built wheel)
 
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 source "$SCRIPTDIR/common.sh"
 
 # What are we building?
 DIST="stevedore"
-VERSION="5.2.0"
+VERSION="5.1.0"
```
dhellmann, line 11: Why change the version?

src/fromager/wheels.py
```diff
@@ -189,13 +193,32 @@ def add_extra_metadata_to_wheels(
         for req_file in req_files:
             shutil.copy(req_file, dist_info_dir / f"fromager-{req_file.name}")
 
-        cmd = ["wheel", "pack", str(dist_info_dir.parent), "-d", str(wheel_file.parent)]
+        build_tag = ctx.settings.build_tag(req.name, version)
+
+        cmd = [
+            "wheel",
+            "pack",
+            str(dist_info_dir.parent),
```
dhellmann, line 201: I think we want `os.fspath()` here instead of just `str()`.

dhellmann (changes requested):

e2e/test_build_order.sh
shubhbapna: Yep, don't need this now

e2e/changelog_settings.yaml
shubhbapna: I think it already does that: https://github.com/python-wheel-build/fromager/pull/325/files#diff-1cb0faa8083db060cd12ad0b5cbc65c23ecd1fc853f3bf7022cc8737d3c48ca5R53

https://github.com/python-wheel-build/fromager/pull/325/files#diff-4a06375158e17eabf9367b420b56d9e7f3e510c8d043938d6cf4b0f7c827ad2fR125

src/fromager/wheels.py
shubhbapna: I am not sure would we need os.fspath? dist_info_dir is pathlib.Path object which has implemented the `__str__` method to return the string representation of the file system path

dhellmann: Huh, we're using fspath in a bunch of other places, and I thought the reason was to ensure the right encoding. It doesn't look like that's what it does, though. @tiran why have we been using os.fspath?

e2e/test_bootstrap_build_tags.sh
```diff
@@ -0,0 +1,83 @@
...
+if ! grep -q "stevedore: have wheel version 5.1.0: $OUTDIR/wheels-repo/downloads/stevedore-5.1.0-1" "$OUTDIR/bootstrap.log"; then
+  echo "FAIL: Did not find log message have wheel version in $OUTDIR/bootstrap.log" 1>&2
+  pass=false
+fi
+
+$pass
+
+SETTINGS_FILE_WITH_ADDED_CHANGELOG=$OUTDIR/changelog_settings_$RANDOM.yaml
+cp $SCRIPTDIR/changelog_settings.yaml $SETTINGS_FILE_WITH_ADDED_CHANGELOG
+echo -e "- test2\n" >> $SETTINGS_FILE_WITH_ADDED_CHANGELOG
```
dhellmann, line 53:
```suggestion
echo -e "        - test2\n" >> $SETTINGS_FILE_WITH_ADDED_CHANGELOG
```

e2e/changelog_settings.yaml
dhellmann, line 6:
```suggestion

```

dhellmann (commented):
I'm happy with this version. I'll leave it open because @tiran had comments, too.

src/fromager/wheels.py
shubhbapna: I think it is strictly pathlib.Path object

tiran (approved):
LGTM, excellent work!

src/fromager/commands/build.py
```diff
@@ -170,8 +172,20 @@ def _is_wheel_built(
...
-        return True
+        url, _ = sdist.resolve_prebuilt_wheel(wkctx, req, [wkctx.wheel_server_url])
+        build_tag_from_settings = wkctx.settings.build_tag(req.name, dist_version)
+        build_tag = build_tag_from_settings if build_tag_from_settings else (0, "")
+        wheel_filename = urlparse(url).path.rsplit("/", 1)[-1]
+        _, _, build_tag_from_name, _ = parse_wheel_filename(wheel_filename)
+        existing_build_tag = build_tag_from_name if build_tag_from_name else (0, "")
+        if (
+            existing_build_tag[0] > build_tag[0]
+            and existing_build_tag[1] == existing_build_tag[1]
```
dhellmann, line 183: This condition looks weird. Isn't the second part comparing the same values?

I remember we wanted the `existing_build_tag[0] > build_tag[0]` part. I'm not sure what the second part is meant to be checking.

src/fromager/settings.py
```diff
@@ -93,10 +93,12 @@ def build_dir(self, pkg: str, sdist_root_dir: pathlib.Path) -> pathlib.Path:
             return sdist_root_dir / relative_build_dir
         return sdist_root_dir
 
-    def build_tag(self, pkg: str, version: Version | str) -> int:
+    def build_tag(self, pkg: str, version: Version | str) -> BuildTag:
         p = self.get_package_settings(pkg)
         changelog = p.get("changelog", {}).get(str(version), [])
-        return len(changelog)
+        if not changelog:
+            return ()
```
dhellmann, line 100: Maybe the caller would want this to be `(0, "")` to make unpacking the tuple consistent? Or is an empty tuple recognized by everything that wants the BuildTag?

src/fromager/wheels.py
```diff
@@ -200,13 +198,33 @@ def add_extra_metadata_to_wheels(
...
 
-        cmd = ["wheel", "pack", str(dist_info_dir.parent), "-d", str(wheel_file.parent)]
+        build_tag_from_settings = ctx.settings.build_tag(req.name, version)
+        build_tag = build_tag_from_settings if build_tag_from_settings else (0, "")
+
+        cmd = [
+            "wheel",
+            "pack",
+            str(dist_info_dir.parent),
+            "-d",
```
dhellmann, line 208: What does `-d` do? Is there a more verbose version of that flag name?

src/fromager/commands/build.py
shubhbapna: I think at some point @tiran wants to add variants to the build tag so it would end up looking like: `0.cpu`. So it only makes sense to check the build number if the variant matches. Right now it is a no-op but is immediately usable if we whenever we decide to introduce variants to build tags

src/fromager/settings.py
shubhbapna: yep empty tuple is recognized by everything that wants the BuildTag. This keeps the types consistent and we anyways to perform a check to ensure the tuple is not empty anywhere we use this method. If we don't then mypy complains about it

src/fromager/wheels.py
shubhbapna: It is the destination dir: https://wheel.readthedocs.io/en/stable/reference/wheel_pack.html#cmdoption-d

Should I use the more verbose name?

src/fromager/commands/build.py
dhellmann: OK, but it's literally comparing a variable to itself so it's always going to be true. Is it meant to be comparing the same part of 2 different variables?

src/fromager/settings.py
dhellmann: Makes sense.

src/fromager/wheels.py
dhellmann: Yes, please. I like long names in places like this because it means I spend less time looking at docs for commands I'm not familiar with.

src/fromager/commands/build.py
shubhbapna: ahhh i completely missed that. Thank you!

dhellmann (approved):
Good! Let's merge this and set up for a release next week so we can test consuming it.

PR #327, opened 2024-08-14
cleanup: remove example_override.py

### LINKED ISSUE #324: remove `example_override.py`
Can we remove `example_override.py` and https://github.com/python-wheel-build/fromager/blob/main/pyproject.toml#L46 ?

I don't think we are testing overrides on fromager itself. The e2e tests seem to be using flit core

shubhbapna:
fixes #324


dhellmann (approved):
At some point we'll want an example like this in the docs, but I agree it's confusing tests and production deployments.

PR #329, opened 2024-08-14
feat: add extra metadata files while building wheels

### LINKED ISSUE #319: Add extra build-log metadata to wheel content
Add extra build-log metadata to wheel content to include:
- changelog
- plugin that allows to add more stuff such as build-env-specific settings:
  - Git SHA
  - Env vars

TOML or JSON format, because parsers for those formats are in the stdlib for py 3.11 and later

shubhbapna:
fixes #319 

We will create a `BUILD` file in the `.dist-info` directory of the wheel. This file is of TOML format and contains the entire settings of the package by default. There is an option for the user to include more data in this file by defining a plugin.

Additionally, I added the build system, build backend and build sdist requirements file (hope is that we can use this later to speedup bootstrapping)

### DIFF
--- a/src/fromager/wheels.py
+++ b/src/fromager/wheels.py
@@ -3,6 +3,7 @@ import logging
 import os
 import pathlib
 import platform
+import shutil
 import sys
 import tempfile
 import typing
@@ -10,6 +11,7 @@ import zipfile
 from datetime import datetime
 
 import elfdeps
+import tomlkit
 from packaging.requirements import Requirement
 from packaging.utils import parse_wheel_filename
 from packaging.version import Version
@@ -129,6 +131,73 @@ def analyze_wheel_elfdeps(
     return requires, provides
 
 
+def add_extra_metadata_to_wheels(
+    ctx: context.WorkContext,
+    req: Requirement,
+    version: Version,
+    extra_environ: dict[str, str],
+    sdist_root_dir: pathlib.Path,
+    wheel_file: pathlib.Path,
+) -> None:
+    extra_data_plugin = overrides.find_override_method(
+        req.name, "add_extra_metadata_to_wheels"
+    )
+    data_to_add = {}
+    if extra_data_plugin:
+        data_to_add = overrides.invoke(
+            extra_data_plugin,
+            ctx=ctx,
+            req=req,
+            version=version,
+            extra_environ=extra_environ,
+            sdist_root_dir=sdist_root_dir,
+        )
+        if not isinstance(data_to_add, dict):
+            logger.warning(
+                f"{req.name}: unexpected return type from plugin add_extra_metadata_to_wheels. Expected dictionary. Will ignore"
+            )
+            data_to_add = {}
+
+    with tempfile.TemporaryDirectory() as dir_name:
+        cmd = ["wheel", "unpack", str(wheel_file), "-d", dir_name]
+        external_commands.run(
+            cmd,
+            cwd=dir_name,
+            network_isolation=ctx.network_isolation,
+        )
+
+        wheel_file_name_parts = wheel_file.name.split("-")
+        distribution_name = f"{wheel_file_name_parts[0]}-{wheel_file_name_parts[1]}"
+        dist_info_dir = (
+            pathlib.Path(dir_name)
+            / distribution_name
+            / f"{distribution_name}.dist-info"
+        )
+        if not dist_info_dir.exists():
+            logger.debug(
+                f"{req.name}: could not add additional metadata. {dist_info_dir} does not exist"
+            )
+            return
+
+        build_file = dist_info_dir / "fromager-build-settings"
+        settings = ctx.settings.get_package_settings(req.name)
+        if data_to_add:
+            settings["metadata-from-plugin"] = data_to_add
+
+        build_file.write_text(tomlkit.dumps(settings))
+
+        req_files = sdist_root_dir.parent.glob("*-requirements.txt")
+        for req_file in req_files:
+            shutil.copy(req_file, dist_info_dir / f"fromager-{req_file.name}")
+
+        cmd = ["wheel", "pack", str(dist_info_dir.parent), "-d", str(wheel_file.parent)]
+        external_commands.run(
+            cmd,
+            cwd=dir_name,
+            network_isolation=ctx.network_isolation,
+        )
+
+
 def build_wheel(
     ctx: context.WorkContext,
     req: Requirement,
@@ -172,6 +241,14 @@ def build_wheel(
         # TODO: raise error?
         return None
     wheel = wheels[0]
+    add_extra_metadata_to_wheels(
+        ctx=ctx,
+        req=req,
+        version=version,
+        extra_environ=extra_environ,
+        sdist_root_dir=sdist_root_dir,
+        wheel_file=wheel,
+    )
     logger.info(f"{req.name}: built wheel '{wheel}' in {end - start}")
     analyze_wheel_elfdeps(ctx, req, wheel)
     return wheel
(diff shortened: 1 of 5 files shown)


src/fromager/wheels.py
```diff
@@ -129,6 +131,73 @@ def analyze_wheel_elfdeps(
...
+            extra_environ=extra_environ,
+            sdist_root_dir=sdist_root_dir,
+        )
+        if not isinstance(data_to_add, dict):
+            logger.warning(
+                f"{req.name}: unexpected return type from plugin add_extra_metadata_to_wheels. Expected dictionary. Will ignore"
+            )
+            data_to_add = {}
+
+    with tempfile.TemporaryDirectory() as dir_name:
```
dhellmann, line 161: Have a look at #104, it shows how to use a class in auditwheel to edit wheel contents and automatically repack the file.

shubhbapna: It doesn't seem like auditwheel was made to be used as an API? It also has a non python dependency: https://github.com/pypa/auditwheel?tab=readme-ov-file#requirements

I would prefer to stick to `wheel` for now until the wheel API PR lands https://github.com/pypa/packaging/pull/805

src/fromager/wheels.py
```diff
@@ -129,6 +131,73 @@ def analyze_wheel_elfdeps(
...
+            return
+
+        build_file = dist_info_dir / "BUILD"
+        build_file.write_text(
+            tomlkit.dumps(ctx.settings.get_package_settings(req.name))
+        )
+        if data_to_add:
+            build_file.write_text(tomlkit.dumps(data_to_add))
+
+        req_files = sdist_root_dir.parent.glob("*-requirements.txt")
```
dhellmann, line 189: We talked about renaming these files to include "fromager" as a prefix, to avoid potentially colliding with standard files.

shubhbapna: Should add the fromager prefix to the `BUILD` file as well?

src/fromager/wheels.py
```diff
@@ -129,6 +131,73 @@ def analyze_wheel_elfdeps(
...
+            / distribution_name
+            / f"{distribution_name}.dist-info"
+        )
+        if not dist_info_dir.exists():
+            logger.debug(
+                f"{req.name}: could not add additional metadata. {dist_info_dir} does not exist"
+            )
+            return
+
+        build_file = dist_info_dir / "BUILD"
```
dhellmann, line 182: Let's call this "fromager-build-settings" maybe?

src/fromager/wheels.py
dhellmann: Yes, I think everything we add that's unique to our build tool should be marked clearly.

src/fromager/wheels.py
```diff
@@ -129,6 +131,73 @@ def analyze_wheel_elfdeps(
...
 
+def add_extra_metadata_to_wheels(
+    ctx: context.WorkContext,
+    req: Requirement,
+    version: Version,
+    extra_environ: dict[str, str],
+    sdist_root_dir: pathlib.Path,
+    wheel_file: pathlib.Path,
+) -> None:
+    extra_data_plugin = overrides.find_override_method(
```
dhellmann, line 142: Why not find_and_invoke()?

src/fromager/wheels.py
```diff
@@ -129,6 +131,73 @@ def analyze_wheel_elfdeps(
...
+                f"{req.name}: could not add additional metadata. {dist_info_dir} does not exist"
+            )
+            return
+
+        build_file = dist_info_dir / "BUILD"
+        build_file.write_text(
+            tomlkit.dumps(ctx.settings.get_package_settings(req.name))
+        )
+        if data_to_add:
+            build_file.write_text(tomlkit.dumps(data_to_add))
```
dhellmann, line 187: Does this over-write the file, or append to it? Maybe what we really want to do is add the extra data from the plugin to the other settings as a key in the other dict. Something like `custom-metadata-extensions`? It's a long name, so something shorter that captures the meaning might be better.

Does TOML impose any restrictions on the syntax for keys? (no spaces, case-insensitive, etc.)

shubhbapna: okay I can either add the custom metadata in some key or another option is that i make the settings data as the default function. If a plugin is specified then all the data is added only from there?

TOML does have some restrictions: https://github.com/toml-lang/toml/blob/main/toml.md#keys

dhellmann: How likely is it that data we are pulling from YAML or other places won't be compatible with TOML?

src/fromager/wheels.py
dhellmann: We could have a default function that returns an empty dict, but maybe it's not worth it.

src/fromager/wheels.py
shubhbapna: i think the tomlkit handles it for us. It was able to successfully convert this yaml without an error:

```yaml
packages:
  stevedore:
    download_source:
      url: https://files.pythonhosted.org/packages/e7/c1/b210bf1071c96ecfcd24c2eeb4c828a2a24bf74b38af13896d02203b1eec/stevedore-${version}.tar.gz
      destination_filename: stevedore-custom-${version}.tar.gz
    resolver_dist:
      sdist_server_url: "https://pypi.org/simple"
      include_sdist: false
      include_wheels: true
    changelog:
      "1.0.0 hahaha":
        - test hello world
```

src/fromager/wheels.py
```diff
@@ -129,6 +131,73 @@ def analyze_wheel_elfdeps(
...
+                f"{req.name}: could not add additional metadata. {dist_info_dir} does not exist"
+            )
+            return
+
+        build_file = dist_info_dir / "fromager-build-settings"
+        build_file.write_text(
+            tomlkit.dumps(ctx.settings.get_package_settings(req.name))
+        )
+        if data_to_add:
+            build_file.write_text(tomlkit.dumps({"metadata-from-plugin": data_to_add}))
```
dhellmann, line 187: I verified that calling write_text 2 times overwrites the contents of the file.

```
>>> import pathlib
>>> p = pathlib.Path('tmp.txt')
>>> p.write_text('foo')
3
>>> p.read_text()
'foo'
>>> p.write_text('bar')
3
>>> p.read_text()
'bar'
```

I think you want to combine the data structures before writing. Something like this:

```suggestion
        settings = ctx.settings.get_package_settings(req.name)
        if data_to_add:
            settings['metadata-from-plugin'] = data_to_add
        else:
            settings['metadata-from-plugin'] = {}
        build_file.write_text(
            tomlkit.dumps(settings)
        )
```

shubhbapna: Do we want an empty `metadata-from-plugin` key even if there is no data from the plugin?

dhellmann: Normally I'd assume so because it makes it more consistent for something reading the file. Does that make weird looking TOML, though?

shubhbapna: Yes it looks like this:

```toml
[download_source]
url = "https://files.pythonhosted.org/packages/e7/c1/b210bf1071c96ecfcd24c2eeb4c828a2a24bf74b38af13896d02203b1eec/stevedore-${version}.tar.gz"
destination_filename = "stevedore-custom-${version}.tar.gz"

[resolver_dist]
sdist_server_url = "https://pypi.org/simple"
include_sdist = false
include_wheels = true

[metadata-from-plugin]
```

dhellmann: So a little weird but not terrible.

I guess in other files the keys are left out if there are no settings, so maybe we should follow that pattern.

dhellmann (approved):

PR #330, opened 2024-08-15
pass constraints file when installing build dependencies

dhellmann:
If we're given a constraints file to decide which build dependencies to
use, apply those same constraints during installation. This ensures that if
the private wheel server has multiple versions of a package, we use the
right one to do the build.

### DIFF
--- a/src/fromager/sdist.py
+++ b/src/fromager/sdist.py
@@ -28,10 +28,18 @@ from . import (
 
 logger = logging.getLogger(__name__)
 
-# Pip has no API, so parse its output looking for what it couldn't
-# install.
+# Pip has no API, so parse its output looking for what it couldn't install.
+# Verbose regular expressions ignore blank spaces, so we have to escape those to
+# have them recognized as being part of the pattern.
 _pip_missing_dependency_pattern = re.compile(
-    r"Could not find a version that satisfies the requirement (\w+)",
+    r"""(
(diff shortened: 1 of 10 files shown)


src/fromager/sdist.py
```diff
@@ -579,6 +589,7 @@ def safe_install(
             ":all:",
         ]
         + ctx.pip_wheel_server_args
+        + ctx.pip_constraint_args
```
shubhbapna, line 592: Is this needed for `safe_install`? We already fix the version when calling it from `_maybe_install`: https://github.com/python-wheel-build/fromager/pull/330/files#diff-3f261cb4d56f2826fc58f44967dd7d1b2a92b5e4e9da0133b832d0e9027ff011R564
(and i don't think `resolved_version` is every `None` in that function - we should probably remove that check)

dhellmann: OK, I included it here because I did see that conditional check in the caller.

I would prefer to be safe and always use the constraints file. Those rules should be leading us to produce the same version, so it's redundant but still safe.

shubhbapna: okay fair enough, the e2e tests are failing though not sure why

shubhbapna (approved):
Looks good but there is still one e2e test failing

dhellmann (commented):
This is ready for re-review

PR #335, opened 2024-08-16
download multiple packages in parallel in download-sequence

dhellmann:
We use a build-order file to know what to
download, but we don't have to wait for one
download to complete to start the next one. Use
threads to download several packages in parallel.
Set a limit, but let the user control the limit
with a command line option.

### DIFF
--- a/src/fromager/commands/download_sequence.py
+++ b/src/fromager/commands/download_sequence.py
@@ -1,5 +1,8 @@
 import json
 import logging
+import os
+import typing
+from concurrent import futures
 
 import click
 from packaging.requirements import Requirement
@@ -23,6 +26,11 @@ logger = logging.getLogger(__name__)
     default=False,
     is_flag=True,
 )
+@click.option(
+    "--num-threads",
+    default=os.cpu_count(),
+    type=int,
+)
 @click.pass_obj
 def download_sequence(
     wkctx: context.WorkContext,
@@ -30,6 +38,7 @@ def download_sequence(
     sdist_server_url: str,
     include_wheels: bool,
     ignore_missing_sdists: bool,
+    num_threads: int,
 ) -> None:
     """Download a sequence of source distributions in order.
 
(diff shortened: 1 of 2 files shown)


src/fromager/commands/download_sequence.py
```diff
@@ -47,32 +54,39 @@ def download_sequence(
         wheel_servers = [sdist_server_url]
 
     with open(build_order_file, "r") as f:
-        for entry in progress.progress(json.load(f)):
-            if entry["prebuilt"]:
-                logger.info(f"{entry['dist']}: uses a pre-built wheel, skipping")
-                continue
+        build_order = json.load(f)
+
+    def download_one(entry):
```
shubhbapna, line 59: Could you add the types here

src/fromager/commands/download_sequence.py
```diff
@@ -23,13 +24,19 @@
     default=False,
     is_flag=True,
 )
+@click.option(
+    "--num-threads",
+    default=3,
```
shubhbapna, line 29: Why not `os.cpu_count`?

shubhbapna (commented):
This will be a nice improvement! Just a couple of small comments

src/fromager/commands/download_sequence.py
dhellmann: I'm worried about throttling on the server side causing errors. Maybe that's an overblown concern for modern services?

src/fromager/commands/download_sequence.py
dhellmann: Done, although it's a vague one. Do we have an alias for the entry in a build-order file?

src/fromager/commands/download_sequence.py
shubhbapna: Oh yeah, we might get rate limited as well. make sense

src/fromager/commands/download_sequence.py
shubhbapna: Its on my todo list for the types cleanup

dhellmann: OK. I've added at least the `dict` part here.

src/fromager/commands/download_sequence.py
dhellmann: I went ahead and set it to the CPU count. Let's see what that does in the build environment.

src/fromager/commands/download_sequence.py
```diff
@@ -23,13 +26,19 @@
     default=False,
     is_flag=True,
 )
+@click.option(
+    "--num-threads",
+    default=os.cpu_count(),
```
dhellmann, line 31: mypy didn't like it when sched_getaffinity() did not exist so I'm dropping that for now so we can land this.

shubhbapna (approved):

PR #341, opened 2024-08-19
upload logs on e2e failures to help debug

shubhbapna:
I am having trouble debugging macos failures and the only way I can do that is via github actions since I don't have access to a mac machine

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -99,3 +99,9 @@ jobs:
 
       - name: Run tests
         run: ./e2e/test_${{ matrix.test-script }}.sh
+      
+      - name: Upload logs for debugging
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v4
+        with:
+          path: e2e-output


.github/workflows/test.yaml
```diff
@@ -99,3 +99,8 @@ jobs:
 
       - name: Run tests
         run: ./e2e/test_${{ matrix.test-script }}.sh
+      
+      - name: Upload logs for debugging
+        if: ${{ failure() }}
+        uses: actions/upload-artifact@v4
+        path: e2e-output
```
dhellmann, line 106: Do we have any data limits we need to be aware of? There might be some big wheels or sdists in that directory.

dhellmann (commented):
LGTM, just one question before approving.

.github/workflows/test.yaml
shubhbapna: I am not sure i think there is a total limit of 100 GB. Someone experimented and said that greater than 5 GB uploads might start breaking (but i think that is the size of the zip archive): https://github.com/actions/upload-artifact/issues/9

shubhbapna: Could you also the default retention period of artifacts on this repository to be 7 days. It should be somewhere in the repo settings

dhellmann (approved):

dhellmann:
The retention period is set to 90 days right now. Let's see how that works out, we can adjust if we need to.

Issue #338, opened 2024-08-19
Include ELF requirements and provides in dist-info


tiran:
The new feature `add_extra_metadata_to_wheels` allows us to add additional metadata files to wheel's dist-info. Let's add ELF provides and ELF requires information. The functionality of `analyze_wheel_elfdeps` should be moved into the `add_extra_metadata_to_wheels`  function, too. `analyze_wheel_elfdeps` needs to unpack the wheel zip file.

Should we store the information in a single TOML file or two simple text files?

dhellmann:
I like separate text files.

PR #345, opened 2024-08-21
Include ELF requirements and provides in dist-info

tiran:
The `add_extra_metadata_to_wheels()` step now writes
`fromager-elf-provides.txt` and `fromager-elf-requires.txt` files with
provided and required shared object names.

I have also fixed several minor issues with the extraction code. It now
analyzes any file with a Python extension suffix and any files that
matches `lib*.so*` pattern.

Fixes: #338


tiran:
Depends on #344

shubhbapna:
I think it needs a rebase? #344 is showing up as changes in the diff

tiran:
The changes in this PR conflicted with changes from the build tag PR. I have addresses the conflict.

src/fromager/wheels.py
```diff
@@ -81,31 +85,65 @@ def _createenv(self) -> None:
...
+    settings: elfdeps.ELFAnalyzeSettings,
+) -> typing.Generator[elfdeps.ELFInfo, None, None]:
+    """Locate shared libraries and yield ELF information"""
+    for root, _, files in os.walk(wheel_root_dir):
+        rootpath = pathlib.Path(root)
+        for filename in files:
+            # any Python extension module or 'lib*.so*'
+            if filename.endswith(tuple(EXTENSION_SUFFIXES)) or (
+                filename.startswith("lib") and ".so" in filename
+            ):
```
shubhbapna, line 100: will we need this after https://github.com/python-wheel-build/elfdeps/pull/6 ?

tiran: No, we don't need this any more. I have changed the code to use the new `analyze_dirtree` function.

shubhbapna (approved):

Issue #339, opened 2024-08-19
Pass dist_info_dir to `add_extra_metadata_to_wheels` override


tiran:
The `add_extra_metadata_to_wheels` override method should have access to the unpacked wheel files. Let's run the method after unpacking the wheel and pass the `dist_info_dir` to the method, too. The caller can assume that the parent directory of `dist_info_dir` is the root directory of the unpacked wheel.

dhellmann:
I think that method is designed with the idea that it returns data, and something else writes it to the wheel. @shubhbapna can confirm.

shubhbapna:
Yes, currently it is designed to only return data which we will write into the `fromager-build-settings` file. Will giving access to `dist_info_dir` to the plugin cause the following issues:
- user decides to overwrite `fromager-build-settings` file. We can lose any data that fromager put their
- adds a file with the name `fromager-build-backend-requirements.txt` with their custom data but that will be overwritten by fromager and the user will lose that data

dhellmann:
We give plugins write information to lots of other aspects of the build, so if they choose to do something that undoes fromager's work I think I'd just say that's up to them and they get what they get as a result. We could also deal with that by writing the core metadata after the plugin is called, so we would be overwriting any files they create with our names.

The bigger question for me is, do we want plugins to write files or do we want them to give us data and let us write files? Letting them write files is more flexible, since the plugin can create whatever types of files it wants/needs. Having them return data to us means we control the format and location of the data, which is more consistent.

shubhbapna:
I prefer to have control over the format and location of data:
- fromager itself might use the data it creates : eg using the *-requirement files directly from the wheels while bootstrapping. if the user overwrites that then the changes from `build-sequence` will propogate to `bootstrap` which will make things difficult to debug
- having control over the format will allow us to switch to a different format in the future if needed without breaking anything

tiran:
I don't have any intent to **write** any files from `add_extra_metadata_to_wheels`. I would like to give the plugin method access to the unpacked files, so it can read and analyze the content of the wheel file.

shubhbapna:
Make sense we can add it

PR #346, opened 2024-08-21
Adding `dist_info_dir` to `add_extra_metadata_to_wheels`

rd4398:
Fixes: https://github.com/python-wheel-build/fromager/issues/339

### DIFF
--- a/src/fromager/wheels.py
+++ b/src/fromager/wheels.py
@@ -158,20 +158,6 @@ def add_extra_metadata_to_wheels(
         req.name, "add_extra_metadata_to_wheels"
     )
     data_to_add = {}
-    if extra_data_plugin:
-        data_to_add = overrides.invoke(
-            extra_data_plugin,
-            ctx=ctx,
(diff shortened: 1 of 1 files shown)


src/fromager/wheels.py
```diff
@@ -189,6 +175,22 @@ def add_extra_metadata_to_wheels(
                 f"{req.name}: {wheel_file} does not contain {dist_info_dir.name}"
             )
 
+        if extra_data_plugin:
+            data_to_add = overrides.invoke(
```
dhellmann, line 179: A while back, @tiran added plugin references pointing to the default functions for all of our overrides so that consumers downstream can find those functions and check their function signatures against the customizations in the downstream build. I missed that when we added this new override without a default. We don't have to hold up this PR over it, but we should have an issue ticket to make sure we add that.

shubhbapna (approved):

dhellmann (approved):

PR #349, opened 2024-08-26
macOS: Run unit tests and less e2e tests

tiran:
Configure unit test runner to run tests on macOS

Only run Python 3.12 tests on macOS. The macOS runner is slower and congested.


dhellmann (approved):

tiran:
Manual merge because we changed mergify settings.

PR #351, opened 2024-08-26
separate macos and linux e2e jobs

dhellmann:
Separate the macos jobs and only run them for the latest version of
python. Update the linter for github actions and mergify settings.


dhellmann:
Let's take #349 instead

PR #352, opened 2024-08-26
feat: add global changelog to trigger rebuilds of all packages with new build tag

### LINKED ISSUE #347: Add global changelog to settings to trigger a rebuild of every package with a new build tag
It is difficult to identify whether a version bump is significant enough to cause a rebuild of any package that depends on it. So we will leave it to the user to update the changelog. There might be times where we need a to trigger a build of every package (eg: maybe when setuptools get upgraded to some version with a lot of changes). In this case we need a way for the user to tell fromager to rebuild everything with a new build tag. Adding a global changelog will be helpful for this

shubhbapna:
fixes #347

### DIFF
--- a/tests/test_settings.py
+++ b/tests/test_settings.py
@@ -159,10 +159,30 @@ def test_changelog():
             - "rebuild xyz"
     """)
     )
-    assert s.build_tag("foo", Version("2.1.0")) == (2, "")
-    assert s.build_tag("foo", "2.1.0") == (2, "")
-    assert s.build_tag("foo", "3.1.0") == ()
-    assert s.build_tag("bar", "2.1.0") == ()
+    assert s.build_tag("foo", Version("2.1.0"), "variant") == (2, "")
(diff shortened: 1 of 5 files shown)


tests/test_settings.py
```diff
@@ -165,6 +165,24 @@ def test_changelog():
     assert s.build_tag("bar", "2.1.0") == ()
 
 
+def test_global_changelog():
+    s = _parse(
+        textwrap.dedent("""
+    changelog:
```
dhellmann, line 171: We'll want the changelog to have a variant component so that a change to one variant doesn't trigger a complete rebuild of all of the other variants.

We should probably make the per-package changelog variant-specific, too.

shubhbapna: Should we address this once #323 lands cause it defines a per-package variant specific settings and it might be easier to build on top of that?

dhellmann: Yeah, we can put off the per-package change until after 323, but let's make the global list variant-specific here.

PR #354, opened 2024-08-27
Pydantic model for global and package settings

tiran:
Introduce Pydantic to validate and manage `settings.yaml` and package-specific `package.yaml` config files. Pydantic handles validation and default management of settings files.

The package settings files now contains environment variables overrides for package and package variants as well as pre-built flag for each variant. The `env` files and most fields in from the old `settings.yaml` will be replaced eventually.

A new command `fromager migrate-config` handles migration of config and env files from the old format to the new format.

The new code is not used, yet. Code will be switched to the new API in a subsequent PR.


tiran:
I have split up PR #323 and moved the core changes into this PR. I'll update the other PR after we have agreed on a migration path.

src/fromager/commands/migrate_config.py
shubhbapna, line 1: need to add the new command here https://github.com/python-wheel-build/fromager/blob/main/src/fromager/commands/__init__.py#L13

shubhbapna (commented):
Looks good, just one note

dhellmann:
Given the pain of rebasing, I'd be OK with merging this as-is and following up to add the command integration that @shubhbapna pointed out.

tiran:
@shubhbapna @dhellmann I have added the missing import and added functional tests for the new migration system.

Issue #340, opened 2024-08-19
per-package constraints for build requirements


dhellmann:
A stable release branch may want to minimize the changes that go into the environment. We can already constrain globally, but some packages need different versions of the same build dependency. We could address that with per-package constraints files _only_ for build requirements.

shubhbapna:
Summarizing my discussion with @dhellmann and the next steps:

- installation constraints are strict and typically won't be allowed to have multiple conflicting versions of the same package. this is how pip also works
- we already produce a constraints file for all installation dependencies so don't have to worry about that
- we need a constraint file for all build dependencies. The problem with this one is that packages are allowed to have "conflicting" build dependencies since they are built in isolation i.e. if package `A` needs `numpy>=2` as a build dependency and package `B` needs `numpy<2` as a build dependency, then we will need 2 versions of numpy which is valid dependency resolution in the context of build requirements
- our goal is to essentially to create `pip freeze` for build requirements 

Next steps:
- Take advantage of the dependency graph and produce a global build requirement constraint file for whatever packages we can. if there are packages that require 2 different versions then produce a per package build requirement constraint file

dhellmann:
Some other requirements to consider: 

1. We want the user to be able to change the constraints.
2. We want to minimize the burden of managing those constraints. It may be simple to think about a per-package constraint file, but for a large set of packages that would mean managing a _lot_ of files.

shubhbapna:
To address making edits to the graph, we can introduce a new fromager command `bump`.  It will follow these rules while updating the graph:
-  the new version is only valid is it satisfies the req specifier
- for all `req_type = install` the new version must satisfy the req specifier for all the nodes of the package we are updating
- for all `req_type != install` it is fine if there are nodes with req specifier that don't satisfy this new version. it won't update those nodes and only update the ones that satisfy it.

Examples:
`fromager bump foo` -> bumps foo to the latest version for all install and build dependencies
`fromager bump foo <version>` -> bumps foo to the <version> for all install and build dependencies
`fromager bump foo --no-installation` -> only bumps the build dependencies
`fromager bump foo --no-build` -> only bumps the installation dependencies

dhellmann:
> To address making edits to the graph, we can introduce a new fromager command `bump`. It will follow these rules while updating the graph:
> 
> * the new version is only valid is it satisfies the req specifier
> * for all `req_type = install` the new version must satisfy the req specifier for all the nodes of the package we are updating

[...]

PR #358, opened 2024-08-27
feat: add a cache to resolver

### LINKED ISSUE #359: add caching in resolver
This will allow us to reuse already seen candidates if they match the current conditions of the package we are trying to resolve. 

Advantages:
- clean data: reduce the number of different package versions we see in graph.json and build-order.json. This will help in #340 
- micro-optimization: we might not need to make a network connection for package resolution

shubhbapna:
fixes #359

### DIFF
--- a/src/fromager/candidate.py
+++ b/src/fromager/candidate.py
@@ -27,6 +27,7 @@ class Candidate:
         extras: typing.Iterable[str] | None = None,
         is_sdist: bool | None = None,
         build_tag: BuildTag = (),
+        identifier: str | None = None,
     ):
         self.name = canonicalize_name(name)
         self.version = version
@@ -34,6 +35,7 @@ class Candidate:
         self.extras = extras
         self.is_sdist = is_sdist
         self.build_tag = build_tag
+        self.identifier = identifier or name
 
         self._metadata: Metadata | None = None
         self._dependencies: list[Requirement] | None = None
--- a/src/fromager/context.py
+++ b/src/fromager/context.py
@@ -10,7 +10,7 @@ from packaging.requirements import Requirement
 from packaging.utils import NormalizedName, canonicalize_name
 from packaging.version import Version
 
-from . import constraints, settings
+from . import candidate, constraints, settings
 
 logger = logging.getLogger(__name__)
 
@@ -57,6 +57,9 @@ class WorkContext:
         self.variant = variant
         self.jobs = jobs
         self.network_isolation = network_isolation
+        self.resolver_cache: dict[str, list[candidate.Candidate]] = (
+            collections.defaultdict(list[candidate.Candidate])
+        )
 
         self._build_order_filename = self.work_dir / "build-order.json"
         self._constraints_filename = self.work_dir / "constraints.txt"
--- a/src/fromager/resolver.py
+++ b/src/fromager/resolver.py
@@ -6,6 +6,7 @@
 import logging
 import os
 import typing
+from collections import defaultdict
 from operator import attrgetter
 from platform import python_version
 from urllib.parse import urljoin, urlparse
@@ -117,6 +118,7 @@ def get_project_from_pypi(
             extras=extras,
             is_sdist=is_sdist,
             build_tag=build_tag,
+            identifier=project,
         )
         if DEBUG_RESOLVER:
             logger.debug(
@@ -140,12 +142,14 @@ class BaseProvider(ExtrasProvider):
         include_wheels: bool = True,
         sdist_server_url: str = "https://pypi.org/simple/",
         constraints: Constraints | None = None,
+        cache: dict[str, list[Candidate]] | None = None,
     ):
         super().__init__()
         self.include_sdists = include_sdists
         self.include_wheels = include_wheels
         self.sdist_server_url = sdist_server_url
         self.constraints = constraints or Constraints({})
+        self.cache = cache or defaultdict(list[Candidate])
 
     def identify(self, requirement_or_candidate: Requirement | Candidate) -> str:
         return canonicalize_name(requirement_or_candidate.name)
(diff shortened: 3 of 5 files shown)


src/fromager/resolver.py
```diff
@@ -220,50 +297,23 @@ def find_matches(
...
-                    logger.debug(
-                        f"{identifier}: skipping {candidate} because it is a wheel"
-                    )
-                continue
-            candidates.append(candidate)
+            if not self.is_in_cache(identifier, candidate) and self.is_valid_candidate(
+                candidate, identifier, bad_versions, identifier_reqs
+            ):
+                candidates.append(candidate)
+                self.add_to_cache(identifier, candidate)
```
dhellmann, line 316: How would we get a candidate that's already in the cache here?

Can we hide that check by having `add_to_cache()` perform it? That would make it work more like a `set()`.

This also feels like it is going to cache some candidates, but not all, and not the actual resolved result. 

We could build a new class that takes as input a resolver instance and provides the same interface for looking up versions that the resolver does, but that saves the results before returning them. That would be a relatively small class, I think?

shubhbapna: We get a candidate that is in cache here: https://github.com/python-wheel-build/fromager/pull/358/files#diff-7dc38361431a0d4763f2cae35a298b8b64a0885f49485d4911096dca9d6e9a4bR300

And right now we will are only storing candidates in cache that were actually potential candidates for some package condition. Do you want me store all seen candidates in the cache?

dhellmann: I was thinking we would only cache the resolved _results_, not the candidates.

shubhbapna: I can do that

shubhbapna: I just realized that we initialize a new provider object each time we do resolution. So we will need to define a global cache and pass that to the provider or initialize the provider only once

src/fromager/candidate.py
```diff
@@ -43,6 +43,20 @@ def __repr__(self) -> str:
...
+    def __eq__(self, other: object) -> bool:
+        if isinstance(other, Candidate):
+            return (
+                self.name == other.name
+                and self.version == other.version
+                and self.build_tag == other.build_tag
+                and self.extras == other.extras
+                and self.is_sdist == other.is_sdist
+            )
+        return False
```
tiran, line 58: You have to return NotImplemented of you don't know how to compare an object.
```suggestion
        return NotImplemented
```

src/fromager/candidate.py
```diff
@@ -43,6 +43,20 @@ def __repr__(self) -> str:
             return f"<{self.name}=={self.version}>"
         return f"<{self.name}[{','.join(self.extras)}]=={self.version}>"
 
+    def __hash__(self) -> int:
+        return hash(self.__repr__())
```
tiran, line 47: This is not a good implementation of `__hash__`. The hash and eq methods should operate on the same set of fields. It might be better idea to replace the `Candidate` object with a frozen [dataclass](https://docs.python.org/3/library/dataclasses.html). A frozen dataclass implements eq and hash correctly. You may have to redefine `extras` as `tuple[str, ...] | None` to make it work.

tiran (changes requested):

src/fromager/resolver.py
dhellmann: That feels like another reason to keep the cache on the outside of the provider.

We have an [override method](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/sources.py#L101) for getting per-package resolver providers, so we can't have a singleton resolver.

However, we would expect those unique providers to give us the same results, and a cache hit could mean we avoid invoking any provider at all.

We're only looking at using the cache for bootstrap, right? Because for the build we are given an exact version and there's no need to resolve it. So maybe we should look for a place where we could put a cache that would only show up in the code path used for bootstrapping?

shubhbapna: We use it here as well: https://github.com/python-wheel-build/fromager/blob/main/src/fromager/commands/build.py#L175

shubhbapna: Nvm caching won't make a difference here so yeah just bootstrap

shubhbapna: I think we will still need to access the cache inside the provider because only the provider knows how to check if a candidate is can be used or not

dhellmann: Why do we need to re-evaluate each of the previously seen candidates?

dhellmann: You're thinking of the evaluation criteria other than the version?

shubhbapna: yep all the different if checks we do apart from just checking if the version matches the requirement

src/fromager/candidate.py
shubhbapna: I don't think I will need a set in my cache. So won't need to implement this anymore

src/fromager/candidate.py
shubhbapna: Won't need this anymore

shubhbapna:
The cache is now stored in context to make it easy to pass it to each instance of the provider. This also makes it easy to add only the candidate we choose to work with to the cache. It is on the provider to decide if a cached candidate is valid or not

src/fromager/context.py
```diff
@@ -57,6 +57,9 @@ def __init__(
         self.variant = variant
         self.jobs = jobs
         self.network_isolation = network_isolation
+        self.resolver_cache: dict[str, list[candidate.Candidate]] = (
```
dhellmann, line 60: It's going to be a dict of normalized names to candidates, right? Do we have a special type for normalized names?

src/fromager/resolver.py
```diff
@@ -220,50 +298,22 @@ def find_matches(
         identifier_reqs = list(requirements[identifier])
         bad_versions = {c.version for c in incompatibilities[identifier]}
 
+        candidates = self.get_from_cache(identifier, bad_versions, identifier_reqs)
+        if candidates:
+            return sorted(
```
dhellmann, line 303: A log message here about using the cached results would be good.

src/fromager/resolver.py
```diff
@@ -286,10 +367,15 @@ def find_matches(
...
```
dhellmann, line 372: Same comment about logging.

src/fromager/resolver.py
```diff
@@ -220,50 +298,22 @@ def find_matches(
...
-                continue
-            # Only include wheels if we're asked to
-            if not candidate.is_sdist and not self.include_wheels:
-                if DEBUG_RESOLVER:
-                    logger.debug(
-                        f"{identifier}: skipping {candidate} because it is a wheel"
-                    )
-                continue
-            candidates.append(candidate)
+            if not self.is_in_cache(identifier, candidate) and self.is_valid_candidate(
```
dhellmann, line 313: Why do we only want candidates that are not in the cache here?

src/fromager/sources.py
```diff
@@ -123,8 +124,10 @@ def resolve_dist(
         logger.debug(f"{req.name}: could not resolve {req} with {constraint}: {err}")
         raise
 
-    candidate: resolver.Candidate
+    candidate: Candidate
     for candidate in result.mapping.values():
+        if candidate not in ctx.resolver_cache[candidate.identifier]:
```
dhellmann, line 129: Having to remember to check this from the outside of the cache before adding something makes me think we do want a cache class so we can have an `add()` method that hides the check.

shubhbapna:
Marking this as draft, because this cleaning of data will break repeatability of builds downstream. We can come back to this after #340

PR #373, opened 2024-08-30
feat: use prebuilt wheel server url override while bootstrapping

### LINKED ISSUE #302: Variant overrides for package source
I'm maintaining a variant that requires several pre-built wheels. The wheels are not hosted on PyPI. I need a way to configure Fromager to pull the pre-built wheels from the costum package index and to use the custom package index for resolving versions. The options must not influence other variants in the settings file.

shubhbapna:
fixes #302 

With the pydantic stuff in, #302 was essentially done (thanks @tiran!). This PR is to just start using the override while bootstrapping

### DIFF
--- a/src/fromager/sdist.py
+++ b/src/fromager/sdist.py
@@ -123,9 +123,13 @@ def handle_requirement(
 
     else:
         logger.info(f"{req.name}: {req_type} requirement {req} uses a pre-built wheel")
-        servers = [sources.PYPI_SERVER_URL]
-        if ctx.wheel_server_url:
-            servers.insert(0, ctx.wheel_server_url)
+        if pbi.wheel_server_url:
+            # use only the wheel server from settings if it is defined. Do not fallback to other URLs
+            servers = [pbi.wheel_server_url]
+        else:
+            servers = [sources.PYPI_SERVER_URL]
+            if ctx.wheel_server_url:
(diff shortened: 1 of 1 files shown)


src/fromager/sdist.py
```diff
@@ -126,6 +126,9 @@ def handle_requirement(
         servers = [sources.PYPI_SERVER_URL]
         if ctx.wheel_server_url:
             servers.insert(0, ctx.wheel_server_url)
+        if pbi.wheel_server_url:
+            # use only the wheel server from settings if it is defined. Do not fallback to other URLs
+            servers = [pbi.wheel_server_url]
```
dhellmann, line 131: It's a little subtle to notice that line 131 replaces the variable and that's the intent. Something like this emphasizes that more:

```suggestion
        if pbi.wheel_server_url:
            # use only the wheel server from settings if it is defined. Do not fallback to other URLs
            servers = [pbi.wheel_server_url]
        else:
            servers = [sources.PYPI_SERVER_URL]
            if ctx.wheel_server_url:
                servers.insert(0, ctx.wheel_server_url)
```

dhellmann (changes requested):
The change looks functionally correct, but I worry about maintaining it in this form. I've proposed an alternative that emphasizes the special case.

dhellmann (approved):

PR #378, opened 2024-09-03
fix: allow prereleases of packages using constraints

### LINKED ISSUE #377: should we allow prereleases when resolving packages
Downstream certain prebuilt wheels have prerelease versions and by default the resolver does not consider prerelease candidates: https://packaging.pypa.io/en/stable/specifiers.html#packaging.specifiers.Specifier.contains

Should we allow prerelease candidates by default? Or should we only allow it for prebuilt wheels?

shubhbapna:
fixes #377


src/fromager/resolver.py
```diff
@@ -179,10 +179,10 @@ def get_preference(
     def is_satisfied_by(self, requirement: Requirement, candidate: Candidate) -> bool:
         if canonicalize_name(requirement.name) != candidate.name:
             return False
-        return (
-            candidate.version in requirement.specifier
-            and self.constraints.is_satisfied_by(requirement.name, candidate.version)
-        )
+        allow_prerelease = self.constraints.allow_prerelease(requirement.name)
```
dhellmann, line 182: So this means we would normally not allow pre-release versions, unless a constraint is in place that mentions a pre-release version?

dhellmann (commented):
This looks good, just one question about how it actually works.

src/fromager/resolver.py
shubhbapna: yep

dhellmann (approved):

Issue #292, opened 2024-08-04
document formats of input and output files


dhellmann:
- [ ] constraints/requriements
- [ ] build-order.json
- [ ] graph.json

dhellmann:
> * What all should the documentation contain? Example of what's included in each file? or just formats like `.txt, .json, etc` ?

A list of the files with descriptions and short examples.

> * Where should this documentation go? in separate `.md` file?

A separate file does make sense.

PR #382, opened 2024-09-03
RHELAI-1056: Document formats of input and output files

rd4398:
JIRA: https://issues.redhat.com/browse/RHELAI-1056

Fixes: https://github.com/python-wheel-build/fromager/issues/292

### DIFF
--- a/docs/files.md
+++ b/docs/files.md
@@ -1,0 +1,95 @@
+# Input and Output Files
+
+## constraints.txt
+
+The `constraints.txt` is the input file used to control the resolution of python packages that are needed to build a wheel. This file is similar to `requirements.txt` but instead of listing packages that must be installed, it sets constraints on versions of packages, ensuring that certain versions are either used or avoided during the wheel building process.
+
+Each line in the `constraints.txt` file should contain the name of a package followed by a version specifier. The syntax for version specifier can be found [here](https://pip.pypa.io/en/stable/reference/requirements-file-format/)
+
+### Example constraints.txt
+
+```
+torch==2.3.1
+pandas<2.2.2
+setuptools!=72.0.0
+```
+
+## build-order.json
+
+The `build-order.json` file is an output file that contains the bottom-up order in which the dependencies need to be built for a specific wheel.
+
+This file contains an array of objects where each object represents a dependency and necessary information related to it. The order of these objects determines the build sequence, starting with the top-level dependency at the bottom of the array.
+
+### Example build-order.json
+
+The following example shows `build-order.json` file for the top-level dependency `wheel`
+
+```json
+[
+  {
+    "type": "build-system",
+    "req": "flit_core<4,>=3.8",
+    "constraint": "",
+    "dist": "flit-core",
+    "version": "3.9.0",
+    "prebuilt": false,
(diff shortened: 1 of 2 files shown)


docs/files.md
```diff
@@ -0,0 +1,97 @@
+# Input and Output Files
+
+## constraints.txt
+
+The `constraints.txt` is the input file used to specify the exact version constraints of python packages that are needed to build a wheel. This file is similar to `requirements.txt` but instead of listing packages that must be installed, it sets constraints on versions of packages, ensuring that certain versions are either used or avoided during the wheel building process.
```
dhellmann, line 5:
```suggestion
The `constraints.txt` is the input file used to control the resolution of python packages that are needed to build a wheel. This file is similar to `requirements.txt` but instead of listing packages that must be installed, it sets constraints on versions of packages, ensuring that certain versions are either used or avoided during the wheel building process.
```

We do tend to use it to set specific versions, but it's also possible to just use `!=` rules to avoid packages.

docs/files.md
```diff
@@ -0,0 +1,97 @@
...
+
+Each line in the `constraints.txt` file should contain the name of a package followed by a version specifier. The version specifier can be one of the following:
+
+* `==`: Specifies an exact version of the package.
+* `<`, `<=`, `>`, `>=`: Specifies a version range.
+* `!=`: Excludes a specific version
```
dhellmann, line 11: We could point to the upstream documentation for the syntax, too.

docs/files.md
```diff
@@ -0,0 +1,97 @@
...
+    "prebuilt": false,
+    "source_url": "https://files.pythonhosted.org/packages/b7/a0/95e9e962c5fd9da11c1e28aa4c0d8210ab277b1ada951d2aee336b505813/wheel-0.44.0.tar.gz#sha256=a29c3f2817e95ab89aa4660681ad547c0e9547f20e75b0562fe7723c9a2a9d49",
+    "source_url_type": "sdist"
+  }
+]
+```
+
+## graph.json
+
+The `graph.json` is an output file that contains all the paths fromager can take to resolve a dependency during building the wheel.
```
dhellmann, line 58: The graph represents all of the dependencies encountered during a build of a wheel collection ("collection" is a new term we're introducing). It shows the parent/child relationships of all dependencies, including the type of dependency.

The format is a mapping from resolved versions expressed as a requirement specification (package==version) to a mapping with properties of the package and the list of dependencies of that package.

docs/files.md
```diff
@@ -0,0 +1,97 @@
...
+
+The following example shows `graph.json` file for the top-level dependency `wheel`
+
+```json
+{
+  "": {
+    "download_url": "",
+    "version": "0",
+    "canonicalized_name": "",
+    "edges": [
```
dhellmann, line 70: @shubhbapna I didn't catch this earlier. Maybe instead of "edges" we want to say "dependencies" here?

dhellmann (commented):
This is a great start. I tweaked the wording in one place and gave you some more details to add in the graph section.

docs/files.md
rd4398: Okay, changed it!

docs/files.md
rd4398: Umm, which upstream documentation? The customization.md or README.md or general fromager documentation?

docs/files.md
rd4398: Okay, added it!

docs/files.md
dhellmann: That file format is described as part of the python packaging standards or pip documentation, I'm not sure which. Have a look through https://www.pypa.io/en/latest/specifications/ and look for the official pip docs to find a good reference.

rd4398: Okay, done!

dhellmann (approved):

PR #384, opened 2024-09-03
fix: decouple downloading of sources from resolution of package source

shubhbapna:
part of #368 

I will update the fields in settings in a separate PR

PR #386, opened 2024-09-04
fix: rename `resolver_dist` to `resolve_source` and pass output dir to `download_source`

### LINKED ISSUE #385: update fields in settings due to the decoupling of downloading and resolving

shubhbapna:
fixes #385 and also adds the nit picks from #384 

I know we discussed renaming fields in settings such that they are decoupled from their corresponding override but i think in this case `resolve_source` is a good name :)

### DIFF
--- a/src/fromager/sources.py
+++ b/src/fromager/sources.py
@@ -51,6 +51,7 @@ def download_source(
         req=req,
         version=version,
         download_url=download_url,
+        sdists_downloads_dir=ctx.sdists_downloads,
     )
 
     if not isinstance(source_path, pathlib.Path):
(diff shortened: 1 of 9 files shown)


src/fromager/sources.py
```diff
@@ -107,28 +108,30 @@ def default_resolve_source(
...
     return url, version
 
 
 def default_download_source(
-    ctx: context.WorkContext, req: Requirement, version: Version, download_url: str
+    ctx: context.WorkContext,
+    req: Requirement,
+    version: Version,
+    download_url: str,
+    output_dir: pathlib.Path,
```
dhellmann, line 128: Let's make clear that plugins should only put sdists here.
```suggestion
    sdists_downloads_dir: pathlib.Path,
```

dhellmann (approved):

PR #387, opened 2024-09-04
fix: use download links from build order file in download sequence and build sequence instead of resolving again

### LINKED ISSUE #92: overrides for downloading source need a way to tell if they are bootstrapping or building
In bootstrap mode, an override should pull source from an upstream location. The download-source command should, too. When building, the override needs to (possibly) pull from an alternative location. We need to pass enough information through the API so the override can make that decision.

### LINKED ISSUE #374: use the download url from build order file in all the build commands

shubhbapna:
fixes #374 
fixes #92 

Breaking change in the `build-sequence` command: it doesn't need the sdist_server_url arg anymore

### DIFF
--- a/src/fromager/commands/build.py
+++ b/src/fromager/commands/build.py
@@ -5,7 +5,7 @@ from urllib.parse import urlparse
 
 import click
 from packaging.requirements import Requirement
-from packaging.utils import parse_wheel_filename
+from packaging.utils import canonicalize_name, parse_wheel_filename
 from packaging.version import Version
 
 from fromager import clickext, context, hooks, progress, sdist, server, sources, wheels
@@ -46,13 +46,16 @@ def build(
 
     """
     server.start_wheel_server(wkctx)
-    wheel_filename = _build(wkctx, dist_name, dist_version, sdist_server_url)
+    req = Requirement(f"{dist_name}=={dist_version}")
+    source_url, version = sources.resolve_source(
+        ctx=wkctx, req=req, sdist_server_url=sdist_server_url
+    )
+    wheel_filename = _build(wkctx, version, req, source_url)
     print(wheel_filename)
 
 
 @click.command()
 @click.argument("build_order_file")
-@click.argument("sdist_server_url")
 @click.option(
     "--skip-existing",
     default=False,
@@ -62,7 +65,6 @@ def build(
 def build_sequence(
     wkctx: context.WorkContext,
     build_order_file: str,
-    sdist_server_url: str,
     skip_existing: bool,
 ) -> None:
     """Build a sequence of wheels in order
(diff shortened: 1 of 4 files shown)


src/fromager/commands/download_sequence.py
```diff
@@ -59,31 +60,26 @@ def download_sequence(
...
-                    version=version,
-                    download_url=source_url,
-                )
-            except Exception as err:
-                logger.error(f"failed to download sdist for {req}: {err}")
-                if not ignore_missing_sdists:
-                    raise
-        else:
-            logger.info(
-                f"{entry['dist']}: uses a {entry['source_url_type']} downloader, skipping"
```
shubhbapna, line 85: I think the logic here should be the same if it is not a prebuilt wheel. Apart from "sdist" the only other type we have is "override" in which case we expect that there is some override that will invoked in `download_source` anyways that handles that particular package

src/fromager/commands/build.py
```diff
@@ -85,19 +87,28 @@ def build_sequence(
...
                     dist_name,
                     dist_name,
-                    dist_version,
+                    resolved_version,
                 )
                 continue
 
-            logger.info("%s: building %s==%s", dist_name, dist_name, dist_version)
-            wheel_filename = _build(wkctx, dist_name, dist_version, sdist_server_url)
+            logger.info("%s: building %s==%s", dist_name, dist_name, resolved_version)
```
dhellmann, line 103: Since sometimes we'll be downloading and not building, maybe change this to

```suggestion
            logger.info("%s: do not have wheel for %s==%s", dist_name, dist_name, resolved_version)
```

and then ensure that both `download_wheel()` and `_build()` log their action, too.

src/fromager/commands/download_sequence.py
```diff
@@ -59,31 +60,26 @@ def download_sequence(
         build_order = json.load(f)
 
     def download_one(entry: dict[str, typing.Any]):
+        req = Requirement(f"{entry['dist']}=={entry['version']}")
+
         if entry["prebuilt"]:
-            logger.info(f"{entry['dist']}: uses a pre-built wheel, skipping")
+            if include_wheels:
+                wheels.download_wheel(req, entry["source_url"], wkctx.wheels_downloads)
```
dhellmann, line 67: Will the URL in the entry of a build-order file point to a wheel?

src/fromager/commands/download_sequence.py
dhellmann: It was only handling sdists because we expect overrides for download source to be downloading something that is not an sdist file (either a release tarball or cloning a git repository).

src/fromager/commands/download_sequence.py
shubhbapna: yes it should https://github.com/python-wheel-build/fromager/blob/main/src/fromager/sdist.py#L140

src/fromager/commands/download_sequence.py
shubhbapna: I see should I revert this then? Although the description of the command says that `Performs the equivalent of the 'step download-source-archive' command for each item in the build order file` and the `step download-source-archive` command would have downloaded non-sdist types as well

src/fromager/commands/build.py
shubhbapna: I think all the other functions in `_build` log something related to their action so I don't think we need anything else there

src/fromager/commands/download_sequence.py
dhellmann: nice, ok

src/fromager/commands/download_sequence.py
shubhbapna: ahhh I see. Yeah it puts them in the same place. In that case I will add that if check again

shubhbapna: done

dhellmann (approved):

PR #392, opened 2024-09-06
Redesign parallel build configuration

tiran:
The configuration for parallel build jobs can take available memory into account. Scaling factors for memory and CPU cores can be set per package.

- `--jobs` now sets the maximum number of cores to use (default: all cores)

Per package override is configured in `settings/mypackage.yaml`

```yaml
build_options:
    cpu_scaling: 4
    memory_scaling: 4
```

### DIFF
--- a/src/fromager/packagesettings.py
+++ b/src/fromager/packagesettings.py
@@ -6,6 +6,7 @@ import types
 import typing
 from collections.abc import Mapping
 
+import psutil
 import pydantic
 import yaml
 from packaging.utils import BuildTag, NormalizedName, canonicalize_name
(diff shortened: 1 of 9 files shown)


src/fromager/packagesettings.py
```diff
@@ -142,6 +143,25 @@ def validate_destination_filename(cls, v):
         return v
 
 
+class BuildOptions(pydantic.BaseModel):
+    """Build system options"""
+
+    model_config = MODEL_CONFIG
+
+    cpu_scaling: int = 1
```
dhellmann, line 151: It sounds like from the description that this is the number of CPU cores to be given to each job, is that right? If so, a name like "cpu_cores_per_job" would make that clearer.

src/fromager/packagesettings.py
```diff
@@ -142,6 +143,25 @@ def validate_destination_filename(cls, v):
...
+    """scale parallel jobs by available CPU cores
+
+    1: as many parallel jobs as CPU cores
+    2: allocate 2 cores per job
+    """
+
+    memory_scaling: int = 2
```
dhellmann, line 158: Similarly here, expressing this as a unit, rather than just a scaling factor, would make it easier to understand how to adjust the value. Is it counting 512MB at a time?

dhellmann (commented):
A couple of nits about names, but the implementation looks good.

src/fromager/packagesettings.py
tiran: I have changed the name to `cpu_cores_per_job` and added a constraint to require a value `>= 1`.

src/fromager/packagesettings.py
tiran: I have renamed the option to ` memory_per_job_gb`. It's now a floating point number `>=0.1`. `0.5` is 512 MB.

dhellmann:
It looks like some of the unit tests need updating, but otherwise this looks good to me.

dhellmann (approved):
We'll want some good examples of this in the documentation. There are a bunch of other doc tickets already, so maybe file an issue so we don't forget to include that topic?

PR #393, opened 2024-09-06
skip prebuilt wheels in build sequence

shubhbapna:


src/fromager/commands/build.py
```diff
@@ -89,8 +89,18 @@ def build_sequence(
...
 
-            logger.info("%s: building %s==%s", dist_name, dist_name, dist_version)
-            wheel_filename = _build(wkctx, dist_name, dist_version, sdist_server_url)
+            if entry["prebuilt"]:
+                logger.info(
+                    "%s: skipping %s==%s since it is a prebuilt",
+                    dist_name,
+                    dist_name,
+                    dist_version,
+                )
```
dhellmann, line 98: I think you want a `continue` here, too, because otherwise line 108 is going to fail with an undefined variable, right?

dhellmann (approved):

tiran (approved):

PR #394, opened 2024-09-06
fix url in version source for generic provider

shubhbapna:
part of #389 
Also adds a function to resolve when given a provider.


dhellmann (approved):
I like these small PRs!

PR #399, opened 2024-09-06
feat: allow dependency resolution using the result of a previous bootstrap

### LINKED ISSUE #389: use dependency graph as a resolver cache for bootstrapping

shubhbapna:
fixes #389

### DIFF
--- a/src/fromager/commands/bootstrap.py
+++ b/src/fromager/commands/bootstrap.py
@@ -11,8 +11,11 @@ from .. import (
     dependency_graph,
     progress,
     requirements_file,
+    resolver,
     sdist,
     server,
+    sources,
+    wheels,
 )
 
 # Map child_name==child_version to list of (parent_name==parent_version, Requirement)
@@ -57,11 +60,19 @@ def _get_requirements_from_args(
     type=clickext.ClickPath(),
     help="pip requirements file",
 )
+@click.option(
+    "-p",
+    "--previous-bootstrap",
+    "prev_graph_file",
+    type=clickext.ClickPath(),
+    help="graph file produced from a previous bootstrap",
+)
 @click.argument("toplevel", nargs=-1)
 @click.pass_obj
 def bootstrap(
     wkctx: context.WorkContext,
     requirements_files: list[pathlib.Path],
+    prev_graph_file: pathlib.Path | None,
     toplevel: list[str],
 ) -> None:
     """Compute and build the dependencies of a set of requirements recursively
@@ -77,11 +88,39 @@ def bootstrap(
         )
     logger.info("bootstrapping %r variant of %s", wkctx.variant, to_build)
 
+    if prev_graph_file:
+        prev_graph = dependency_graph.DependencyGraph.from_file(prev_graph_file)
+    else:
+        prev_graph = None
+
     pre_built = wkctx.settings.list_pre_built()
     if pre_built:
         logger.info("treating %s as pre-built wheels", sorted(pre_built))
 
     server.start_wheel_server(wkctx)
+
+    # we need to resolve all the top level dependencies before we start bootstrapping.
+    # this is to ensure that if we are using an older bootstrap to resolve packages
+    # we are able to upgrade a package anywhere in the dependency tree if it is mentioned
+    # in the toplevel without having to fall back to history
+    for req in to_build:
+        pbi = wkctx.package_build_info(req)
+        if pbi.pre_built:
+            servers = wheels.get_wheel_server_urls(wkctx, req)
+            source_url, version = wheels.resolve_prebuilt_wheel(wkctx, req, servers)
+        else:
+            source_url, version = sources.resolve_source(
+                wkctx, req, resolver.PYPI_SERVER_URL
+            )
+        wkctx.dependency_graph.add_dependency(
+            parent_name=None,
+            parent_version=None,
+            req_type=requirements_file.RequirementType.TOP_LEVEL,
+            req=req,
+            req_version=version,
+            download_url=source_url,
+            pre_built=pbi.pre_built,
+        )
 
     with progress.progress_context(total=len(to_build)) as progressbar:
         for req in to_build:
@@ -90,6 +129,7 @@ def bootstrap(
                 req,
                 req_type=requirements_file.RequirementType.TOP_LEVEL,
                 progressbar=progressbar,
+                prev_graph=prev_graph,
             )
             progressbar.update()
 
--- a/src/fromager/sdist.py
+++ b/src/fromager/sdist.py
@@ -19,6 +19,7 @@ from . import (
     sources,
     wheels,
 )
+from .dependency_graph import DependencyGraph
 from .requirements_file import RequirementType
 
 logger = logging.getLogger(__name__)
@@ -28,8 +29,9 @@ def handle_requirement(
     ctx: context.WorkContext,
     req: Requirement,
     req_type: RequirementType = RequirementType.TOP_LEVEL,
-    why: list | None = None,
+    why: list[tuple[RequirementType, Requirement, Version]] | None = None,
     progressbar: progress.Progressbar | None = None,
+    prev_graph: DependencyGraph | None = None,
 ) -> str:
     if why is None:
         why = []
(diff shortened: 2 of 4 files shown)


src/fromager/commands/bootstrap.py
```diff
@@ -77,19 +87,41 @@ def bootstrap(
...
     if pre_built:
         logger.info("treating %s as pre-built wheels", sorted(pre_built))
 
     server.start_wheel_server(wkctx)
 
+    # we need to resolve all the top level dependencies before we start bootstrapping.
+    # this is to ensure that if we are using an older bootstrap to resolve packages
+    # we are able to upgrade a package anywhere in the dependency tree if it is mentioned
+    # in the toplevel without having to fall back to history
+    for req in to_build:
```
dhellmann, line 104: What happens if a prebuilt  appears in this list?

src/fromager/sdist.py
```diff
@@ -93,16 +101,17 @@ def handle_requirement(
...
-        parent_name=canonicalize_name(why[-1][1].name) if why else None,
-        parent_version=why[-1][2] if why else None,
-        req_type=req_type,
-        req=req,
-        req_version=Version(str(resolved_version)),
-        download_url=source_url,
-        pre_built=pre_built,
-    )
-    ctx.write_to_graph_to_file()
+    if req_type != RequirementType.TOP_LEVEL:
```
dhellmann, line 104: Maybe `add_dependency()` should be idempotent instead of asking the caller to check?

src/fromager/sdist.py
```diff
@@ -259,6 +278,150 @@ def handle_requirement(
...
+    if req_type == RequirementType.TOP_LEVEL:
+        for edge in ctx.dependency_graph.get_root_node().get_outgoing_edges(
+            req.name, RequirementType.TOP_LEVEL
+        ):
+            if edge.req == req:
+                return (
+                    edge.destination_node.download_url,
+                    edge.destination_node.version,
+                )
+        # this should never happen
```
dhellmann, line 300: Maybe say why, because I had to sit for a minute to think about it. And maybe raise an exception in that case?

src/fromager/sdist.py
```diff
@@ -259,6 +278,150 @@ def handle_requirement(
...
+    ):
+        possible_versions_in_top_level.append(
+            (
+                top_level_edge.destination_node.download_url,
+                top_level_edge.destination_node.version,
+            )
+        )
+        seen_version.add(str(top_level_edge.destination_node.version))
+
+    resolver_result = _resolver_for_graph(ctx, possible_versions_in_top_level, req)
```
dhellmann, line 321: This function is named like it's going to return a resolver, but it's used like it's returning results.

src/fromager/sdist.py
```diff
@@ -259,6 +278,150 @@ def handle_requirement(
...
+    if resolver_result:
+        return resolver_result
+
+    # only if there is nothing in top level reqs, resolve using history
+    possible_versions_from_graph = []
+    # check all nodes which have the same parent name irrespective of the parent's version
+    for parent_node in prev_graph.get_nodes_by_name(
+        parent_req.name if parent_req else None
+    ):
+        # if the edge matchesthe current req and type then it is a possible candidate
```
dhellmann, line 331: We talked through this, but it would be good to record the reason for filtering on type here in a comment so we have a good record if we end up needing to come back and  change it.

src/fromager/sdist.py
```diff
@@ -259,6 +278,150 @@ def handle_requirement(
...
+            version_source=lambda x, y, z: version_source, constraints=ctx.constraints
+        )
+        return resolver.resolve_from_provider(provider, req)
+    except Exception as err:
+        logger.debug(
+            f"{req.name}: could not resolve {req} from {version_source}: {err}"
+        )
+
+
+def _resolve_source(
```
dhellmann, line 364: This name is very similar to the function in the sources module. It would be good to differentiate more. You could call it something like `_resolve_source_with_history` to emphasize that it looks at that graph for past resolutions.

src/fromager/sdist.py
```diff
@@ -280,7 +444,12 @@ def _handle_build_system_requirements(
     for dep in _sort_requirements(build_system_dependencies):
         try:
             resolved = handle_requirement(
-                ctx, dep, RequirementType.BUILD_SYSTEM, why, progressbar=progressbar
+                ctx,
```
dhellmann, line 447: It would be good to be consistent about calling by name here.

src/fromager/sdist.py
shubhbapna: I could add checks to make sure there are no duplicate edges, but in our current downstream we allow duplicate edges to show up

src/fromager/commands/bootstrap.py
shubhbapna: ah nice catch this is why the e2e test was failing. Thank you!

src/fromager/commands/bootstrap.py
```diff
@@ -77,19 +88,47 @@ def bootstrap(
...
+        pbi = wkctx.package_build_info(req)
+        if pbi.pre_built:
+            servers = wheels.get_wheel_server_urls(wkctx, req)
+            source_url, version = wheels.resolve_prebuilt_wheel(wkctx, req, servers)
+        else:
+            source_url, version = sources.resolve_source(
+                wkctx, req, resolver.PYPI_SERVER_URL
+            )
```
dhellmann, line 113: What other cases do we deal with in other places? Does this single `if` cover everything now? Your comments last week about reusing the resolver have me wondering about how much benefit we would get from sharing this logic in other places by creating a function?

src/fromager/sdist.py
dhellmann: If we only need to prevent the top level duplicates, it's OK to just do the check here. Maybe expand the comment to mention that we're expecting top level edges to already exist because they were all resolved before starting to bootstrap, like you have on line 289.

src/fromager/commands/bootstrap.py
shubhbapna: the only other place we do this `if` based resolution is in sdist.py: https://github.com/python-wheel-build/fromager/pull/399/files/810230839452154a7b3b284a3a7d8a6ab57c51b0#diff-3f261cb4d56f2826fc58f44967dd7d1b2a92b5e4e9da0133b832d0e9027ff011R57

dhellmann: Ok,it felt like it might have been a pattern but it sounds like not.

dhellmann (commented):
This looks good to me. I think @tiran should also take a look.

src/fromager/sdist.py
```diff
@@ -259,6 +279,152 @@ def handle_requirement(
...
+            f"{req.name}: {req} appears as a toplevel requirement but it's resolution does not exist in the root node of the graph"
+        )
+
+    if not prev_graph:
+        return None
+
+    seen_version = set()
+
+    # first perform resolution using the top level reqs before looking at history
+    possible_versions_in_top_level = []
```
tiran, line 313:
```suggestion
    possible_versions_in_top_level: list[tuple[str, Version]] = []
```

src/fromager/sdist.py
```diff
@@ -259,6 +279,152 @@ def handle_requirement(
...
+        seen_version.add(str(top_level_edge.destination_node.version))
+
+    resolver_result = _resolve_from_version_source(
+        ctx, possible_versions_in_top_level, req
+    )
+    if resolver_result:
+        return resolver_result
+
+    # only if there is nothing in top level reqs, resolve using history
+    possible_versions_from_graph = []
```
tiran, line 332:
```suggestion
    possible_versions_from_graph: list[tuple[str, Version]] = []
```

src/fromager/sdist.py
```diff
@@ -259,6 +279,152 @@ def handle_requirement(
...
+                    )
+                )
+                seen_version.add(str(edge.destination_node.version))
+
+    return _resolve_from_version_source(ctx, possible_versions_from_graph, req)
+
+
+def _resolve_from_version_source(
+    ctx: context.WorkContext, version_source: list[Requirement], req: Requirement
+):
```
tiran, line 360: Missing return type annotation

src/fromager/sdist.py
```diff
@@ -259,6 +279,152 @@ def handle_requirement(
...
+        )
+
+
+def _resolve_source_with_history(
+    ctx: context.WorkContext,
+    prev_graph: DependencyGraph | None,
+    req: Requirement,
+    req_type: RequirementType,
+    parent_req: Requirement | None,
+):
```
tiran, line 380: Missing return type annotation

src/fromager/sdist.py
```diff
@@ -259,6 +279,152 @@ def handle_requirement(
...
+    return (source_url, resolved_version)
+
+
+def _resolve_prebuilt_with_history(
+    ctx: context.WorkContext,
+    prev_graph: DependencyGraph | None,
+    req: Requirement,
+    req_type: RequirementType,
+    parent_req: Requirement | None,
+):
```
tiran, line 407: Missing return type annotation

tiran (commented):
Some nit picks and missing type annotations.

I wonder why mypy hasn't been complaining about the missing annotations. Are we ignoring the file?

tiran (approved):
LGTM

Let's land the PR. I'm going to do another test run on the big server tomorrow.

PR #400, opened 2024-09-07
Fix resolver_dist and config migration

tiran:
The option is called `resolver_dist`, not `resolve_source`. Rename helper properties and function to match the names of Fromager 0.27.

Fix config migration when a variant block is None instead of empty mapping.

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -61,7 +61,7 @@ $ fromager --settings-dir=overrides/settings ...
 download_source:
     url: "https://github.com/pytorch/pytorch/releases/download/v${version}/pytorch-v${version}.tar.gz"
     destination_filename: "${canonicalized_name}-${version}.tar.gz"
-resolve_source:
+resolver_dist:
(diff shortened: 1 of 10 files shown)


shubhbapna:
In 0.28 we are already breaking the settings config so I thought I could go ahead and change `resolver_dist` to `resolve_source` since we don't use `resolver_dist` anymore (and `resolver_source` is a more accurate name). I guess the migration command doesn't pick that up?

shubhbapna:
Does #401 help instead of reverting resolve_source to resolver_dist? Since we have to run the migrate-config command anyways?

tiran:
I didn't mean to change the names of the config elements. I misunderstood the code and then copy 'n pasted the wrong names into the new code. Now I understand what the code is suppose to do. Your original names make more sense and explain the purpose of the settings.

I would prefer to go back to your naming convention instead. We could yank the 0.28.0 release and declare it a bad release.

PR #402, opened 2024-09-08
fix migrate-config command

shubhbapna:
Some settings weren't being correctly migrated and were in fact being ignored. This PR fixes that as well as adds changing "resolver_dist" field to "resolve_source" and fixes config migration when a variant block is None instead of empty mapping


dhellmann (commented):
It sounds like the migration code needs more tests.

tiran:
The migration problem in our repo is caused by a malformed file name. `cuda-ubi9/vllm-flash-attn.env` should be `cuda-ubi9/vllm_flash_attn.env`. The combination of the malformed file name and `vllm_flash_attn` in `settings.yaml` breaks the migration system. I have pulled @shubhbapna's fix into PR #400.

shubhbapna:
closing and using #400 instead

PR #403, opened 2024-09-09
Improve parallel job configuration

tiran:
Use `DIST_EXTRA_CONFIG` env var to configure `setuptools` to use parallel builds in `build_ext`. This speeds up compilation of packages such as Cython. The feature is disabled by default, because it breaks some packages like Numba 0.60.0.

Apply package env variables last. Package configuration can now override variables like `MAX_JOBS`.

### DIFF
--- a/src/fromager/packagesettings.py
+++ b/src/fromager/packagesettings.py
@@ -148,6 +148,13 @@ class BuildOptions(pydantic.BaseModel):
 
     model_config = MODEL_CONFIG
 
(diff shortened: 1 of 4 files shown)


tiran:
On my test system, the change speeds up `Cython-3.0.11-0-cp311-cp311-linux_x86_64.whl` build from `2:42` to `0:57`.

dhellmann (approved):
Nice!

PR #405, opened 2024-09-09
fix: download-sequence store prebuilts in the prebuilt directory

### LINKED ISSUE #390: download-sequence should handle prebuilt wheels as prebuilt
We need to put prebuilt wheels in the `prebuilt` directory as well as the `downloads` directory so that the bootstrap process will recognize them properly.

shubhbapna:
fixes #390

### DIFF
--- a/src/fromager/commands/download_sequence.py
+++ b/src/fromager/commands/download_sequence.py
@@ -64,7 +64,7 @@ def download_sequence(
 
         if entry["prebuilt"]:
             if include_wheels:
-                wheels.download_wheel(req, entry["source_url"], wkctx.wheels_downloads)
+                wheels.download_wheel(req, entry["source_url"], wkctx.wheels_prebuilt)
             else:
                 logger.info(f"{entry['dist']}: uses a pre-built wheel, skipping")
             return


src/fromager/commands/download_sequence.py
```diff
@@ -64,7 +64,7 @@ def download_one(entry: dict[str, typing.Any]):
 
         if entry["prebuilt"]:
             if include_wheels:
-                wheels.download_wheel(req, entry["source_url"], wkctx.wheels_downloads)
+                wheels.download_wheel(req, entry["source_url"], wkctx.wheels_prebuilt)
```
dhellmann, line 67: I think we need it in both places, and we need to trigger rebuilding the local wheel index if we're running one. That way a prebuilt dependency of something else will be available when it's needed to build that thing.

shubhbapna: won't all of that happen on its own like you pointed out here: https://github.com/python-wheel-build/fromager/issues/390#issuecomment-2332694340

dhellmann: Ah, right, OK.

dhellmann (approved):

PR #417, opened 2024-09-13
fix: create all directories on setup

### LINKED ISSUE #416: some directories needed by fromager don't exist
Sometimes the user passes path to directories that don't exist, in which case fromager should create them instead of failing

shubhbapna:
fixes #416


dhellmann (approved):

dhellmann (approved):

e2e/test_elfdeps.sh
```diff
@@ -14,7 +14,7 @@ VERSION="2.1.5"
 OS=$(uname)
 if [ "$OS" = "Darwin" ]; then
     NETWORK_ISOLATION=""
-    WHEEL_PLATFORM_TAG="macosx_10_9_universal2"
+    WHEEL_PLATFORM_TAG="macosx_10_13_universal2"
```
dhellmann, line 17: I have a different fix for this in #420 that should be future proof and allow the tests to work on my dev box, too.

tiran (approved):

e2e/test_elfdeps.sh
tiran: I have merged main into the branch and resolved the merge conflict.

PR #424, opened 2024-09-16
standardize patch directories using override module names

### LINKED ISSUE #421: Patches are not applied by override name
https://github.com/python-wheel-build/fromager/pull/420 added a `lint` subcommand to Fromager. The linter assumes that patch directories use the override module name (canonical name, `-` -> `_`). However this does not work for `tqdm-multiprocess`. We either need to ensure that patches use the override name or change the linter again.

dhellmann:
Require patch directories be named using the override module form of the
name of the package being patched. In addition to being more predictable,
this allows us to simplify the logic for detecting patches for other
versions of a project that are not being applied.

Standardize the naming of unpacked source distributions to make the name
predictable.

We log the individual filenames when patches are applied, but if we do not
have a patch there is no indication that we tried to patch the source at
all.

Fixes #421


rd4398 (approved):
This looks good! I will wait for others to review and approve

shubhbapna (approved):

PR #429, opened 2024-09-17
Feat: Per wheel log

rd4398:
This commit is related to per wheel log feature for the `build-sequence` command

JIRA Epic: https://issues.redhat.com/browse/RHELAI-1323
JIRA ticket: https://issues.redhat.com/browse/RHELAI-1487

References:
- https://docs.python.org/3/library/logging.html#logging.Logger.addHandler

### DIFF
--- a/e2e/test_build_order.sh
+++ b/e2e/test_build_order.sh
@@ -50,6 +50,7 @@ if grep -q "skipping building wheels for stevedore" "$log"; then
   pass=false
 fi
 
+
 EXPECTED_FILES="
 $OUTDIR/wheels-repo/downloads/setuptools-*.whl
 $OUTDIR/wheels-repo/downloads/pbr-*.whl
@@ -58,6 +59,10 @@ $OUTDIR/wheels-repo/downloads/stevedore-*.whl
 $OUTDIR/sdists-repo/downloads/stevedore-*.tar.gz
 $OUTDIR/sdists-repo/downloads/setuptools-*.tar.gz
 $OUTDIR/sdists-repo/downloads/pbr-*.tar.gz
+
+$OUTDIR/work-dir/logs/stevedore-*.log
+$OUTDIR/work-dir/logs/setuptools-*.log
+$OUTDIR/work-dir/logs/pbr-*.log
 "
 
 pass=true
--- a/src/fromager/__main__.py
+++ b/src/fromager/__main__.py
@@ -66,7 +66,7 @@ else:
     "--work-dir",
     default=pathlib.Path("work-dir"),
     type=clickext.ClickPath(),
-    help="location to manage working files, including builds",
+    help="location to manage working files, including builds and logs",
 )
 @click.option(
     "-p",
--- a/src/fromager/commands/build.py
+++ b/src/fromager/commands/build.py
@@ -13,6 +13,7 @@ from fromager import (
     clickext,
     context,
     hooks,
+    overrides,
     progress,
     server,
     sources,
@@ -141,6 +142,16 @@ def _build(
     req: Requirement,
     source_download_url: str,
 ) -> pathlib.Path:
+    per_wheel_logger = logging.getLogger("")
+
+    module_name = overrides.pkgname_to_override_module(req.name)
+    log_filename = module_name + "_current.log"
+
+    wheel_log = wkctx.logs_dir / log_filename
+
+    file_handler = logging.FileHandler(str(wheel_log), mode="w")
+    per_wheel_logger.addHandler(file_handler)
+
     source_filename = sources.download_source(
         ctx=wkctx,
         req=req,
@@ -190,6 +201,12 @@ def _build(
         sdist_filename=sdist_filename,
         wheel_filename=wheel_filename,
     )
+
+    per_wheel_logger.removeHandler(file_handler)
+    file_handler.close()
+
+    new_filename = wheel_log.with_name(wheel_filename.stem + ".log")
+    wheel_log.rename(new_filename)
 
     return wheel_filename
 
--- a/src/fromager/context.py
+++ b/src/fromager/context.py
@@ -65,6 +65,7 @@ class WorkContext:
         self.wheel_server_dir = self.wheels_repo / "simple"
         self.work_dir = pathlib.Path(work_dir).absolute()
         self.wheel_server_url = wheel_server_url
+        self.logs_dir = self.work_dir / "logs"
         self.cleanup = cleanup
         self.variant = variant
         self.network_isolation = network_isolation
@@ -178,6 +179,7 @@ class WorkContext:
             self.wheels_downloads,
             self.wheels_prebuilt,
             self.wheels_build,
+            self.logs_dir,
         ]:
             if not p.exists():
                 logger.debug("creating %s", p)


src/fromager/commands/build.py
```diff
@@ -19,7 +19,7 @@
     wheels,
 )
 
-logger = logging.getLogger(__name__)
+logger = logging.getLogger('')
```
dhellmann, line 22: Let's not rename this one, and just get the root logger in the function where we need to use it.

src/fromager/commands/build.py
```diff
@@ -101,13 +101,20 @@ def build_sequence(
             source_download_url = entry["source_url"]
             req = Requirement(f"{dist_name}=={resolved_version}")
 
+            log_filename = str(resolved_version) + "-" + dist_name + ".log"
+            wheel_log = wkctx.work_dir / log_filename
```
dhellmann, line 105: We should include the build tag in the filename. @shubhbapna do we have enough information to compute the build tag here before we build the wheel?

src/fromager/commands/build.py
```diff
@@ -101,13 +101,20 @@ def build_sequence(
...
+
+            file_handler = logging.FileHandler(str(wheel_log), mode="w")
+            logger.addHandler(file_handler)
```
dhellmann, line 108: Doing the logging here is going to cause us to produce a log with the skipping messages on line 112. We don't want to upload those and overwrite other log files that have useful information. Can we do the logging down in the `_build()` function, where we know we're going to try to build the wheel?

src/fromager/commands/build.py
rd4398: Okay, will do

src/fromager/commands/build.py
rd4398: I agree

rd4398:
I have pushed new changes

src/fromager/commands/build.py
```diff
@@ -141,6 +141,19 @@ def _build(
     req: Requirement,
     source_download_url: str,
 ) -> pathlib.Path:
+    pbi = wkctx.package_build_info(req)
+    build_tag_from_settings = pbi.build_tag(resolved_version)
+    build_tag = build_tag_from_settings if build_tag_from_settings else (0, "")
+    build_tag_as_string = " ".join(map(str, build_tag))
```
dhellmann, line 147: This will put a space in the filename, right? That tends to make shell scripting harder, so let's use another separator. Could we use the same standard that's used to add a build tag to a version in the wheel filename?

src/fromager/commands/build.py
```diff
@@ -141,6 +141,19 @@ def _build(
...
+    per_wheel_logger = logging.getLogger("")
+    log_filename = (
+        req.name + " " + str(resolved_version) + "-" + build_tag_as_string + ".log"
```
dhellmann, line 150: Another space in the filename?

Instead of req.name, we should use our function that converts the name to the override module format for consistency. https://github.com/python-wheel-build/fromager/blob/main/src/fromager/overrides.py#L128

src/fromager/commands/build.py
rd4398: oh yes okay. I replaced it with the same standard used to add a build tag to a version in the wheel filename

src/fromager/commands/build.py
rd4398: okay, done in latest push!

rd4398:
@dhellmann @shubhbapna I need your inputs on whether we should test this and how?

src/fromager/commands/build.py
```diff
@@ -141,6 +142,23 @@ def _build(
...
+    pbi = wkctx.package_build_info(req)
+    build_tag_from_settings = pbi.build_tag(resolved_version)
+    build_tag = build_tag_from_settings if build_tag_from_settings else (0, "")
+    build_tag_as_string = f"{build_tag[0]}{build_tag[1]}"
+
+    per_wheel_logger = logging.getLogger("")
+
+    module_name = overrides.pkgname_to_override_module(req.name)
+    log_filename = (
+        module_name + "-" + str(resolved_version) + "-" + build_tag_as_string + ".log"
```
dhellmann, line 154: We could probably use a less complicated name, since it's only temporary. Something like `_current.log` even, to make it stand out.

src/fromager/commands/build.py
```diff
@@ -191,6 +209,11 @@ def _build(
         wheel_filename=wheel_filename,
     )
 
+    new_filename = wheel_log.with_name(wheel_filename.stem + ".log")
+    wheel_log.rename(new_filename)
+
+    per_wheel_logger.removeHandler(file_handler)
```
dhellmann, line 215: We probably want to remove the handler and make sure it has closed the file before we rename it.

We should also think about the right thing to do if there's a build error. We won't upload the file in the production pipeline, but a developer debugging that issue locally is going to want to know how to find the log file and to know that it has everything in it.

dhellmann (commented):
For testing, you could add some filenames to the lists that we expect to see in some of the existing jobs.

src/fromager/commands/build.py
rd4398: Hmm, okay. Any suggestions about what is the right thing to do? Add documentation for it?

src/fromager/commands/build.py
rd4398: okay, done!

rd4398:
> For testing, you could add some filenames to the lists that we expect to see in some of the existing jobs.

Okay, I will add this

src/fromager/commands/build.py
rd4398: > We probably want to remove the handler and make sure it has closed the file before we rename it.

okay, done!

e2e/test_build_order.sh
```diff
@@ -50,6 +50,8 @@ if grep -q "skipping building wheels for stevedore" "$log"; then
   pass=false
 fi
 
+APP_ROOT="$(dirname "$(dirname "$(readlink -fm "$0")")")"
```
shubhbapna, line 53: The logs should be created in the $OUTDIR not the APP_ROOT

src/fromager/__main__.py
```diff
@@ -68,6 +68,13 @@
     type=clickext.ClickPath(),
     help="location to manage working files, including builds",
 )
+@click.option(
+    "-t",
```
shubhbapna, line 72: `-t` is already being used by `--word-dir`. Use another shorthand like `-l` maybe?

shubhbapna (changes requested):

e2e/test_build_order.sh
rd4398: Hmm okay, but we are trying to create `logs` at root level in fromager just like how `work-dir` and `sdists-repo` are created. How do we test this at root level since $OUTDIR is pointing to `e2e-output`. Am I missing something?

src/fromager/__main__.py
tiran: Do we need a separate log dir? We could store the logs in `ctx.work_dir / "logs"`.

tiran: Did Doug say why he wanted a different directory? I don't see a comment about the log directory from him.

The work directory also contains a bunch of temporary stuff like unpacked sdists and wheels. It makes sense to have a separate directory with state that we want to archive and retain for future use. What do you think about a more general name (state, output, ...)? We could move the graph, build-order, and constraints file into the directory and the new logs in a "logs" subdirectory.

rd4398: I guess @shubhbapna is working on a similar feature for archiving `sdists` and I remember Doug mentioning this so that we both are on same track. Also, he mentioned that this format will help us use the upload plugin which we have in builder to upload this dir.

I can move it to a sub-directory inside` work-dir` if that's a better approach :)

rd4398: I have updated the code to make `logs` as sub-directory of `work-dir`

shubhbapna (approved):
Nice, LGTM!

Issue #431, opened 2024-09-18
Support building for different Python version


tiran:
In Fromager 0.31 and earlier, we assume that packages are built for the same Python version and interpreter as Fromager's Python. For example if Fromager is installed in a Python 3.11 virtual env, then all packages are built for Python 3.11.

Fromager should be able to build a package for another Python interpreter, e.g.

```shell
fromager --python=/usr/bin/python3.12 ...
```

I think all it takes is to replace `sys.executable` with a command line parameter and get `platform.python_version()` from that interpreter. Then Fromager will create virtual envs for a different Python version.

dhellmann:
The resolver is also going to need to be updated to apply the python version when evaluating dependencies.

I think this is more complex than the benefits we could reasonable expect to get from it.

PR #432, opened 2024-09-18
Draft: Build wheels for different Python interpreter

tiran:
Fromager now support building wheels for a different Python interpreter:

```shell
fromager --python-interpreter=/path/to/python ...
```

Fixes: #431


tiran:
I need to add some e2e tests before the PR is ready for review.

dhellmann:
The resolver is also going to need to be updated to apply the python version when evaluating dependencies.

I think this is more complex than the benefits we could reasonable expect to get from it.

Issue #201, opened 2024-07-18
Attempts to install build tools to system Python environment


jaraco:
Using [pip-run](https://pypi.org/project/pip-run) to invoke fromager, it gets a lot done before attempting to install setuptools to the system Python environment and then fails because pip rejects the installation:

```
 fromager 🐚 py -3.12 -m pip-run fromager -- -m fromager bootstrap requests
bootstrapping 'cpu' variant of [('toplevel', 'requests')]
requests: * handling toplevel requirement requests []
saved /Users/jaraco/fromager/sdists-repo/downloads/requests-2.32.3.tar.gz
requests: new toplevel dependency requests resolves to 2.32.3
requests: preparing source for requests from /Users/jaraco/fromager/sdists-repo/downloads/requests-2.32.3.tar.gz
requests: prepared source for requests at /Users/jaraco/fromager/work-dir/requests-2.32.3/requests-2.32.3
requests: getting build system dependencies for requests in /Users/jaraco/fromager/work-dir/requests-2.32.3/requests-2.32.3
setuptools: ** handling build-system requirement setuptools>=40.8.0 [('toplevel', <Requirement('requests')>, <Version('2.32.3')>)]
saved /Users/jaraco/fromager/sdists-repo/downloads/setuptools-71.0.3.tar.gz
setuptools: new build-system dependency setuptools>=40.8.0 resolves to 71.0.3
setuptools: preparing source for setuptools>=40.8.0 from /Users/jaraco/fromager/sdists-repo/downloads/setuptools-71.0.3.tar.gz
setuptools: prepared source for setuptools>=40.8.0 at /Users/jaraco/fromager/work-dir/setuptools-71.0.3/setuptools-71.0.3
setuptools: getting build system dependencies for setuptools>=40.8.0 in /Users/jaraco/fromager/work-dir/setuptools-71.0.3/setuptools-71.0.3
setuptools: getting build backend dependencies for setuptools>=40.8.0 in /Users/jaraco/fromager/work-dir/setuptools-71.0.3/setuptools-71.0.3
setuptools: getting build sdist dependencies for setuptools>=40.8.0 in /Users/jaraco/fromager/work-dir/setuptools-71.0.3/setuptools-71.0.3
setuptools: adding ('setuptools', '71.0.3') to build order
setuptools: preparing to build wheel for version 71.0.3
created build environment in /Users/jaraco/fromager/work-dir/setuptools-71.0.3/build-3.12.4
[...]

Issue #126, opened 2024-07-03
use `build` instead of `pip wheel`?


dhellmann:
Revisit the decision to use `pip wheel` to build wheels. The build command supports a `--no-isolation` flag, which should prevent installing anything and give the same protection that flag does when running pip.

What was it that led us to switch from build to pip?

tiran:
We have to switch to `build` so we can build `matplotlib` with system packages. A bug in `pip` prevents us from passing multiple arguments to `pip wheel`

https://matplotlib.org/stable/install/dependencies.html#use-system-libraries

```shell
python -m pip install \
  --config-settings=setup-args="-Dsystem-freetype=true" \
  --config-settings=setup-args="-Dsystem-qhull=true" \
  .
```

https://meson-python.readthedocs.io/en/stable/how-to-guides/config-settings.html

> Please note that, while pypa/build concatenates arguments for the same key passed to the -C option, pip up to version 23.0.1 does not offer any way to set a build config setting to a list of strings: later values for the same key passed to --config-settings override earlier ones.
>
> This effectively limits the number of options that can be passed to each command invoked in the build process to one. This limitation is tracked in [pip issue #11681](https://github.com/pypa/pip/issues/11681). This limitation should be removed in the next version of pip.

dhellmann:
> We have to switch to `build` so we can build `matplotlib` with system packages. A bug in `pip` prevents us from passing multiple arguments to `pip wheel`

Good to know. matplotlib isn't likely to be a priority for a while, but this definitely motivates a move.

shubhbapna:
> The `build` API lets us run the hooks with a custom interpreter and runner instead: `build.BuildEnv(python_executable="$VIRTUAL_ENV/bin/python")`. https://build.pypa.io/en/stable/api.html#build.ProjectBuilder

I like this idea!

PR #434, opened 2024-09-19
Add run method to BuildEnvironment

tiran:
The logic to run a command in a virtual environment has been moved from `build_wheel()` function into `BuildEnvironment.run()` method. It allows users to run any program in the context of a virtual environment.

The `PackageBuildInfo.get_extra_environ()` method applies all 
package-related env vars and optionally `BuildEnvironment` virtual env 
vars `PATH` and `VIRTUAL_ENV`. Before, parallel job settings were only 
applied for wheel building. Package env did not have access to modified
`PATH` and `VIRTUAL_ENV`.

### DIFF
--- a/docs/hooks.rst
+++ b/docs/hooks.rst
@@ -242,8 +242,16 @@ Additional types
 ----------------
 
 .. autoclass:: fromager.build_environment.BuildEnvironment
+   :members: run
+
+   .. autoproperty:: python
+
 .. autoclass:: fromager.context.WorkContext
+
 .. autoclass:: fromager.resolver.PyPIProvider
+
 .. autoclass:: fromager.resolver.GenericProvider
+
 .. autoclass:: fromager.resolver.GitHubTagProvider
+
 .. autofunction:: fromager.sources.prepare_new_source
--- a/src/fromager/build_environment.py
+++ b/src/fromager/build_environment.py
@@ -1,11 +1,13 @@
 import importlib.metadata
 import logging
+import os
 import pathlib
 import platform
 import re
 import subprocess
 import sys
 import typing
+from io import TextIOWrapper
 
 from packaging.requirements import Requirement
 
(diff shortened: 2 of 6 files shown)


tiran:
The change prepares Fromager for changes related to #126 and #201.

src/fromager/build_environment.py
```diff
@@ -70,18 +73,114 @@ class BuildEnvironment:
...
+
+        If `extra_environ` is set, then the value must come from
+        :meth:`get_extra_environ`. Fromage hooks are called with expected
+        environment.
+
+        `network_isolation` defaults to context setting.
+        """
+        # Gotcha: If user provides extra_environ, then we cannot apply
+        # our `get_extra_environ` method. The extra environ may already
+        # have the values. We must not apply extra environments twice.
```
dhellmann, line 163: This feels like a logic error in the callers that we should fix.

src/fromager/build_environment.py
```diff
@@ -70,18 +73,114 @@ class BuildEnvironment:
...
+        if extra_environ is None:
+            extra_environ = self.get_extra_environ(extra_environ)
+        else:
```
dhellmann, line 166: What happens here if extra_environ is a dict with content?

tiran: Argh, I got the logic wrong. Fixing ...

docs/hooks.rst
```diff
@@ -242,8 +242,16 @@ Additional types
 ----------------
 
 .. autoclass:: fromager.build_environment.BuildEnvironment
+   :members: get_extra_environ, run
```
dhellmann, line 245:
```suggestion
   :members: add_venv_environ, run
```

tiran: D'oh

src/fromager/wheels.py
```diff
@@ -268,11 +266,14 @@ def build_wheel(
...
 
-    # add package and variant env vars last
+    # add package and variant env vars
     template_env = os.environ.copy()
     template_env.update(extra_environ)
     extra_environ.update(pbi.get_extra_environ(template_env=template_env))
 
+    # activate virtual env
+    build_env.add_venv_environ(extra_environ)
+
```
dhellmann, line 276: This section is confusing. It seems to be updating multiple dicts with similar information, iteratively? 

I think I understand that `pbi.get_extra_environ()` needs the template environment in order to evaluate the template for some of the values. Are we sure that will never need to include `$PATH`, `$VIRTUAL_ENV`, or any of the values defined in `extra_environ` above? In particular, I could see `MAX_JOBS` being useful if the way to specify that for a build tool is using a different variable name.

If `add_venv_environ()` just returned those settings, then could we put them into `template_env` before calling `get_extra_environ()`? Or would we have to worry about the ordering?

tiran: You are right, the logic was confusing. I wasn't happy with the previous implementation, either. It's inconsistent, too. For example the parallel job settings were only applied in the build wheel path, not in other paths. The package env templates did not have access to correct `VIRTUAL_ENV` either.

Now `get_extra_environ()` applies the parallel job settings and takes an optional `build_env` arg to apply virtual env settings from the build environment.

shubhbapna (approved):

PR #438, opened 2024-09-20
feat: add a summary of wheels built in build-sequence

### LINKED ISSUE #272: [Feat] Transaction summary
`fromager` should have some way of reporting the transaction summary. This need comes from running bootstraps with large variations in time differences. Bumping to `instructlab[cuda]==0.18.0b3` for instance only took about 21 minutes, where as some of our earlier builds took up to a few hours. Knowing which packages were re-bootstrapped and or re-built would be nice, and mesh well with Shubh's chances to allow only rebuilding if the flag is passed and a wheel exists in the given `WHEEL_SERVER_URL` . This could look like a directive we pass to build-sequence and bootstrap full.

shubhbapna:
fixes #272

### DIFF
--- a/src/fromager/commands/build.py
+++ b/src/fromager/commands/build.py
@@ -92,7 +92,7 @@ def build_sequence(
             "skipping builds for versions of packages available at %s",
             wkctx.wheel_server_url,
         )
-
+    packages_built = []
     logger.info("reading build order from %s", build_order_file)
     with open(build_order_file, "r") as f:
         for entry in progress.progress(json.load(f)):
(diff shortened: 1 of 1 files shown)


src/fromager/commands/build.py
```diff
@@ -127,13 +127,24 @@ def build_sequence(
...
                 )
+            packages_built.append(entry)
 
             server.update_wheel_mirror(wkctx)
             # After we update the wheel mirror, the built file has
             # moved to a new directory.
             wheel_filename = wkctx.wheels_downloads / wheel_filename.name
             print(wheel_filename)
 
+    logger.info("\nTransaction summary:")
```
dhellmann, line 138: How hard would it be to get this into a separate file?

src/fromager/commands/build.py
```diff
@@ -127,13 +127,24 @@ def build_sequence(
                 wheel_filename = _build(
                     wkctx, resolved_version, req, source_download_url
                 )
+            packages_built.append(entry)
```
dhellmann, line 130: If we're using the logger, and writing to a separate file, we could just log here instead of waiting until the end. That way even if the build fails, we'll have a nice summary of what we were able to do.

src/fromager/commands/build.py
shubhbapna: Probably just adding a file handler

src/fromager/commands/build.py
shubhbapna: I wonder how it will work with #429? Should we try to combine this with #429?

shubhbapna:
closing in favor of #443

Issue #218, opened 2024-07-21
user guide


dhellmann:
We need a user guide. Some topics to consider, in no particular order

- [ ] approaches for building in containers
- [ ] approaches for testing plugins for building specific packages using CI jobs
- [ ] development workflows vs. production build workflows
- [ ] example production pipelines (we have a nice diagram in Miro, maybe we can convert that to something we can publish)
- [ ] concepts guide (bootstrapping, build one vs. sequence, step commands)
- [ ] scripting using step commands
- [ ] uploading builds (at the end of build sequence vs. one by one)

PR #441, opened 2024-09-20
add a "getting started" guide

dhellmann:
Add an example of how to bootstrap a set of wheels and fix one that does not build immediately.

Related to #218


rd4398 (approved):
This looks good! 
I will wait for others to review and approve

PR #444, opened 2024-09-25
fix sdist type to be override if provider plugin is defined

shubhbapna:
When a resolver provider is defined the sdist type should still be override since ` download-sequence` currently does not support such plugins


rd4398 (approved):
Looks good to me!

PR #445, opened 2024-09-25
feat: measure time for different steps in bootstrapping and build-sequence

### LINKED ISSUE #408: add time measurements for all steps in bootstrapping and build-sequence
Right now we only measure the amount of time taken to build a wheel, we should also measure time for other steps during bootstrapping and build-sequence to help us detect performance bottlenecks. Maybe we should also measure other performance counters like memory usage?

shubhbapna:
fixes #408 

I couldn't find a nice way to measure memory usage for child processes. even using cgroups won't be as straightforward. Anyways measuring time is probably a higher priority than memory so we can discuss how to do memory usage later

For measuring time, instead of wrapping each function call we want to measure individually, I decided it might be better to create a decorator which we can call for whichever function we want to measure. This will make it easy to store metrics in a single store and print it at the end as well. Plus, in the future if we want to measure more functions we simply have to add this decorator


src/fromager/metrics.py
```diff
@@ -0,0 +1,82 @@
...
+_time_store: dict[str, dict[str, float]] = defaultdict(dict[str, float])
+_time_description_store: dict[str, str] = {}
+
+
+def timeit(description: str):
+    def timeit_decorator(func: typing.Callable):
+        _time_description_store[func.__name__] = description
+
+        @functools.wraps(func)
+        def wrapper_timeit(*args: typing.Any, **kwargs: typing.Any):
```
tiran, line 20: You are missing return type annotations.

tiran (changes requested):
The ` _extract_req_and_version_from_args` involves a lot of hackery and black magic. Let's make it easier and more sustainable for us:

1. Change every function with a `timeit()` decorator to accept only keyword arguments, e.g. `def download_source(*, ctx: context.WorkContext, req: Requirement, version: Version, download_url: str)`. The lonely `*` indicates keyword-only for remaining arguments.
2. Change the internal wrapper to `def wrapper_timeit(*, ctx: context.WorkContext, req: Requirement, version: Version, **kwargs: typing.Any) -> typing.Any`.

Now you have convenient access to the variables without hackery. It makes the timeit function less generic, but that's okay for our purpose. You can also store the timing information on the context instead of a global variable.

shubhbapna:
> 1. Change every function with a `timeit()` decorator to accept only keyword arguments, e.g. `def download_source(*, ctx: context.WorkContext, req: Requirement, version: Version, download_url: str)`. The lonely `*` indicates keyword-only for remaining arguments.

Thats a breaking change for a lot of our APIs and everytime we want to measure a new function we will have to change its API to ensure that it is compatible with our metrics decorators. Plus we won't be able to measure functions like `prepare_build_environment`

tiran:
> Thats a breaking change for a lot of our APIs and everytime we want to measure a new function we will have to change its API to ensure that it is compatible with our metrics decorators. Plus we won't be able to measure functions like `prepare_build_environment` . Right now `timeit` functions allows to measure any function and prints a debug log immediately.

keyword-only argument cause some breakage, but it's not as bad as you think. I would argue that they also improve code for the better. I have implemented keyword-only args for several functions in PR #459.

Do we need to measure any function? I don't think so. We want to measure how long build steps of a requirement are taking. The API functions all take a `ctx` and a `req`, most take a `version`, too. To handle `prepare_build_environment`, you can make the version argument optional:

```python
def wrapper_timeit(*, ctx: context.WorkContext, req: Requirement, **kwargs: typing.Any) -> typing.Any:
    version = kwargs.get("version", kwargs.get("dist_version"))
    ...
```

Another approach is to use `inspect.signature` to create a Signature object for a callable, then bind the signature to `*args` and `**kwargs`:

```python
>>> import inspect
>>> from fromager import build_environment
>>> from packaging.requirements import Requirement
>>> import pathlib

>>> sig = inspect.signature(build_environment.prepare_build_environment)
>>> args = ("ctx", Requirement("egg"), pathlib.Path("source"))
>>> kwargs = {}
>>> bound = sig.bind(*args, **kwargs)
>>> bound.apply_defaults()
>>> bound
<BoundArguments (ctx='ctx', req=<Requirement('egg')>, sdist_root_dir=PosixPath('source'))>
>>> bound.arguments
{'ctx': 'ctx', 'req': <Requirement('egg')>, 'sdist_root_dir': PosixPath('source')}
```

shubhbapna:
That make sense I like it. What about the resolver functions? They don't take in `version` but return a version. Should I use the binding the signature approach and then check the return value of the function to extract the version?

shubhbapna:
> How much granularity do we need for timing?

Since we are trying to time some heavy weight functions (almost all of them requiring some sort of IO), I think we can get away with not being as fine grained in measuring them. We might not gain much by optimizing these functions at the micro-second level anyways. So just a blanket measurement for each step might be enough for now?

dhellmann:
> > How much granularity do we need for timing?
> 
> Since we are trying to time some heavy weight functions (almost all of them requiring some sort of IO), I think we can get away with not being as fine grained in measuring them. We might not gain much by optimizing these functions at the micro-second level anyways. So just a blanket measurement for each step might be enough for now?

So we would time how long it takes to resolve, download, and build?

Should we separate the time spent in plugins vs. the time spent in the core of fromager?

shubhbapna:
> So we would time how long it takes to resolve, download, and build?

And prepare as well

> Should we separate the time spent in plugins vs. the time spent in the core of fromager?

I think we can just think of the plugin as some IO fromager is performing that we can't control much

dhellmann:
> > Should we separate the time spent in plugins vs. the time spent in the core of fromager?
> 
> I think we can just think of the plugin as some IO fromager is performing that we can't control much

That makes sense.

src/fromager/metrics.py
```diff
@@ -0,0 +1,74 @@
...
+    return timeit_decorator
+
+
+def summarize(prefix: str) -> None:
+    logger = logging.getLogger(__name__)
+    for req in sorted(_time_store.keys()):
+        total_time = sum(_time_store[req].values())
+        log = f"{prefix} {req} took {total_time:.2f} seconds total"
+        for fn_name, time_taken in _time_store[req].items():
+            log += f", {time_taken:.2f} seconds to {_time_description_store[fn_name]}"
```
dhellmann, line 62: Some of the longer builds take on the order of an hour, so it would be good to show the duration in a format that accounts for that. [datetime.timedelta](https://docs.python.org/3/library/datetime.html#timedelta-objects) does that automatically when converted to a string.

dhellmann (commented):
I have one formatting suggestion that I think we should address now.

We may also find we want a machine-readable output file in the future. Let's do that later, though, when we have some experience using the log output so we can decide what format to use.

dhellmann (approved):

src/fromager/metrics.py
```diff
@@ -0,0 +1,77 @@
...
+import time
+import typing
+from collections import defaultdict
+from datetime import timedelta
+
+from packaging.requirements import Requirement
+from packaging.version import Version
+
+_time_store: dict[str, dict[str, float]] = defaultdict(dict[str, float])
+_time_description_store: dict[str, str] = {}
```
tiran, line 12: I would prefer to store this on the context. All functions should have a `ctx` argument.

src/fromager/metrics.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+def timeit(description: str):
```
tiran, line 15:
```suggestion
def timeit(description: str) -> typing.Callable:
```

src/fromager/metrics.py
```diff
@@ -0,0 +1,77 @@
...
+            end = time.perf_counter()
+            # get the logger for the module from which this function was called
+            logger = logging.getLogger(func.__module__)
+            version = (
+                kwargs.get("version")
+                or kwargs.get("dist_version")
+                or kwargs.get("resolved_version")
+            )
+            if not version:
+                version = _extract_version_from_return(ret)
```
tiran, line 34: Which function needs this hack?

src/fromager/metrics.py
```diff
@@ -0,0 +1,77 @@
...
+
+            runtime = end - start
+
+            if req:
+                logger.debug(
+                    f"{req.name}: took {timedelta(seconds=runtime)} to {description}"
```
tiran, line 40:
```suggestion
                    f"{req.name}: {func.__name__} took {timedelta(seconds=runtime)} to {description}"
```

src/fromager/metrics.py
```diff
@@ -0,0 +1,77 @@
...
+                )
+            else:
+                logger.debug(f"took {timedelta(seconds=runtime)} to {description}")
```
tiran, line 43:
```suggestion
                logger.debug(f"{func.__name__} took {timedelta(seconds=runtime)} to {description}")
```

src/fromager/metrics.py
shubhbapna: any of the version resolvers like this one: https://github.com/python-wheel-build/fromager/blob/main/src/fromager/sources.py#L73

src/fromager/metrics.py
```diff
@@ -0,0 +1,80 @@
...
+        @functools.wraps(func)
+        def wrapper_timeit(
+            *,
+            ctx: context.WorkContext,
+            req: Requirement | None = None,
+            **kwargs: typing.Any,
+        ) -> typing.Any:
+            ctx.time_description_store[func.__name__] = description
+
+            start = time.perf_counter()
```
rd4398, line 24: I am using `datetime.now(`) here https://github.com/python-wheel-build/fromager/blob/80b8563d9cb77e342a142ef7ebd9836994a45c6a/src/fromager/wheels.py#L274 and was wondering whether we should be consistent in using the timing libraries. What do you think?

shubhbapna: `perf_counter` is mostly used for measuring performance:
- https://pymotw.com/3/time/ (Doug's blog)
- https://stackoverflow.com/questions/66036844/time-time-or-time-perf-counter-which-is-faster/66037261#66037261

rd4398: Ah okay!

tiran: There are multiple clock sources. `datetime` uses wall clock as clock source. A wall clock can jump backwards and forwards in time, e.g. during DST switch, when NTP syncs the wall clock with a remote server, or when a computer is put to sleep and wakes up again.

`perf_counter` is a monotonic clock (does not go back) with a high resolution. There is also `process_time`, which has different properties than `perf_counter`. See https://docs.python.org/3/library/time.html#time.get_clock_info for all clock types. 

`perf_counter` is a good choice to measure deltas between two time stamps.

rd4398 (approved):
LGTM!

shubhbapna:
@tiran are anymore changes required for this PR?

tiran (approved):

Issue #409, opened 2024-09-11
Log return value of override methods


tiran:
`find_and_invoke` should log the return value of override methods. In Fromager 0.29.0, we log that we are calling an override method, but not the arguments and return value.

PR #451, opened 2024-09-26
 Log return value of override methods

rd4398:
Fixes: https://github.com/python-wheel-build/fromager/issues/409

This commit adds log statements to the `find_and_invoke` method to log arguments and return value of the override method

### DIFF
--- a/src/fromager/overrides.py
+++ b/src/fromager/overrides.py
@@ -50,7 +50,10 @@ def find_and_invoke(
     fn = find_override_method(distname, method)
     if not fn:
         fn = default_fn
-    return invoke(fn, **kwargs)
+
+    result = invoke(fn, **kwargs)
+    logger.info(f"{distname}: override method {fn.__name__} returned {result}")
+    return result
 
 
 def invoke(fn: typing.Callable, **kwargs: typing.Any) -> typing.Any:
@@ -61,6 +64,7 @@ def invoke(fn: typing.Callable, **kwargs: typing.Any) -> typing.Any:
                 f"{fn.__module__}.{fn.__name__} override does not take argument {arg_name}"
             )
             kwargs.pop(arg_name)
+        logger.debug(f"{fn.__name__} takes argument: {arg_name}")
     return fn(**kwargs)
 
 


src/fromager/overrides.py
```diff
@@ -50,7 +50,11 @@ def find_and_invoke(
     fn = find_override_method(distname, method)
     if not fn:
         fn = default_fn
-    return invoke(fn, **kwargs)
+
+    logger.debug(f"The arguments for override methods are: {kwargs}")
```
shubhbapna, line 54: I don't think we want to log kwargs here cause not all of them might be used. Take a look at the `invoke` function

src/fromager/overrides.py
```diff
@@ -50,7 +50,11 @@ def find_and_invoke(
...
+    result = invoke(fn, **kwargs)
+    logger.info(f"The override method {fn!s} returned: {result}")
```
shubhbapna, line 56:
```suggestion
    logger.info(f"{distname}: override method {fn.__name__} returned {result}")
```

shubhbapna (changes requested):

src/fromager/overrides.py
rd4398: hmm, okay do we want to log them in `invoke` after ` kwargs.pop(arg_name)` outside for loop?
Or there is no need at all to log kwargs anywhere?

shubhbapna: I think @tiran wanted the arguments to be logged (#409). So probably log the ones we actually use outside of the if statement

rd4398: okay, will do

src/fromager/overrides.py
rd4398: okay, done!

dhellmann:
Let's not link upstream work to downstream tickets that no one else can see in the private jira. Please remove the jira ID from the PR title and any commit messages.

rd4398:
> Let's not link upstream work to downstream tickets that no one else can see in the private jira. Please remove the jira ID from the PR title and any commit messages.

Oh okay, sorry about that! I will keep this in mind

shubhbapna (approved):

Issue #343, opened 2024-08-19
tracker: speedup bootstrap


shubhbapna:
We can speed up bootstrap by:
- [x] #450
- [x] #449
- [x] #507 
- [x] #510 
- [ ] A prefetcher thread that gets the "future" wheel/sdists

dhellmann:
Why would the index for pre-populating the bootstrap directory be different from the one passed to `--wheel-server`?

dhellmann:
The prefetch thread idea is interesting, but may be unnecessary. I'm going to close this as done and we can reconsider the prefetch idea if it proves useful.

PR #453, opened 2024-09-26
modularize `handle_requirement`

### LINKED ISSUE #452: modularize `handle_requirement` in the sdist module
The `handle_requirement` function handles the entire bootstrap process and is getting bigger and more difficult to maintain. Plus adding more features like #343 will make it more difficult maintain. So lets modularize this function before we start adding more functionality

shubhbapna:
fixes #452 

This is essentially a one-to-one refactor of sdist module and breaks down `handle_requirement` to multiple functions (no logic change). This is to help with #343


src/fromager/bootstrapper.py
```diff
@@ -0,0 +1,489 @@
...
+                wheel_filename = found_wheel_filename
+
+        self.ctx.add_to_build_order(
+            req=req,
+            version=resolved_version,
+            source_url=source_url,
+            source_url_type=source_url_type,
+            prebuilt=pbi.pre_built,
+            constraint=constraint,
+        )
```
shubhbapna, line 108: I moved this to after building of wheels because:
- We have graph.json that as already recorded this dependency
- if the build fails then anyways bootstrapping stops and we don't end up using that build order file

src/fromager/bootstrapper.py
```diff
@@ -0,0 +1,489 @@
...
+            )
+
+        pbi = self.ctx.package_build_info(req)
+        if pbi.pre_built:
+            resolved_version, wheel_url, wheel_filename, unpack_dir = (
+                self._handle_prebuilt_sources(req, req_type)
+            )
+            # Remember that this is a prebuilt wheel, and where we got it.
+            source_url = wheel_url
+            source_url_type = "prebuilt"
```
tiran, line 58: Suggestion: add "prebuilt" to RequirementType enum

src/fromager/bootstrapper.py
```diff
@@ -0,0 +1,489 @@
...
+    def _prepare_source(
+        self, req: Requirement, source_filename: pathlib.Path, resolved_version: Version
+    ) -> tuple[pathlib.Path, pathlib.Path, set[Requirement]]:
+        sdist_root_dir = sources.prepare_source(
+            self.ctx, req, source_filename, resolved_version
+        )
+        unpack_dir = sdist_root_dir.parent
+
+        build_system_dependencies = dependencies.get_build_system_dependencies(
+            self.ctx, req, sdist_root_dir
```
tiran, line 205:
```suggestion
            ctx=self.ctx, req=req, sdist_root_dir=sdist_root_dir
```

src/fromager/bootstrapper.py
```diff
@@ -0,0 +1,489 @@
...
+        )
+        self._handle_build_requirements(
+            RequirementType.BUILD_SYSTEM,
+            build_system_dependencies,
+        )
+
+        build_backend_dependencies = dependencies.get_build_backend_dependencies(
+            self.ctx, req, sdist_root_dir
```
tiran, line 213:
```suggestion
            ctx=self.ctx, req=req, sdist_root_dir=sdist_root_dir
```

src/fromager/bootstrapper.py
```diff
@@ -0,0 +1,489 @@
...
+        )
+        self._handle_build_requirements(
+            RequirementType.BUILD_BACKEND,
+            build_backend_dependencies,
+        )
+
+        build_sdist_dependencies = dependencies.get_build_sdist_dependencies(
+            self.ctx, req, sdist_root_dir
```
tiran, line 221:
```suggestion
            ctx=self.ctx, req=req, sdist_root_dir=sdist_root_dir
```

src/fromager/bootstrapper.py
```diff
@@ -0,0 +1,489 @@
...
+        logger.info(
+            f"{req.name}: built wheel for version {resolved_version}: {wheel_filename}"
+        )
+        return wheel_filename, build_env
+
+    def _prepare_source(
+        self, req: Requirement, source_filename: pathlib.Path, resolved_version: Version
+    ) -> tuple[pathlib.Path, pathlib.Path, set[Requirement]]:
+        sdist_root_dir = sources.prepare_source(
+            self.ctx, req, source_filename, resolved_version
```
tiran, line 200:
```suggestion
            ctx=self.ctx, req=req, source_filename=source_filename, version=resolved_version
```

src/fromager/bootstrapper.py
shubhbapna: This is more for sdist types which are: "sdist", "override" and "prebuilt". I can create a new enum for that

rd4398 (approved):
Looks good to me !
Nice work with the Boostrapper class!

src/fromager/sdist.py
dhellmann, line 1: It might be unfortunate to lose the git history of this file. I don't know if git is smart enough to recognize that the new bootstrapper.py is mostly the same content, in a different format. One way to deal with that would be to include a commit in this PR that renames sdist.py to bootstrapper.py, and then make all of the edits in a second commit. I wouldn't expect the tests to pass on that rename commit.

I could also be overthinking it, so let's see what other reviewers think.

src/fromager/bootstrapper.py
```diff
@@ -0,0 +1,492 @@
...
+            | build_backend_dependencies
+            | build_sdist_dependencies,
+        )
+
+    def _handle_build_requirements(
+        self, build_type: RequirementType, build_dependencies: set[Requirement]
+    ) -> None:
+        self.progressbar.update_total(len(build_dependencies))
+
+        for dep in self._sort_requirements(build_dependencies):
```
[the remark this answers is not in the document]
shubhbapna: Yep

src/fromager/bootstrapper.py
```diff
@@ -0,0 +1,492 @@
...
+from .dependency_graph import DependencyGraph
+from .requirements_file import RequirementType, SourceType
+
+if typing.TYPE_CHECKING:
+    from . import context
+
+logger = logging.getLogger(__name__)
+
+
+class Bootstrapper:
```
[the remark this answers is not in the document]
shubhbapna: I think the graph is accessed in other commands as well. Maybe I can try to move build order stuff, `mark_as_seen` and `has_been_seen`

src/fromager/sdist.py
shubhbapna: Why would we like to preserve the history of this file?

src/fromager/bootstrapper.py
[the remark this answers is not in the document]
shubhbapna: okay moved these to the bootstrapper class

src/fromager/bootstrapper.py
```diff
@@ -0,0 +1,492 @@
...
+            build_backend_dependencies,
+        )
+
+        build_sdist_dependencies = dependencies.get_build_sdist_dependencies(
+            ctx=self.ctx, req=req, sdist_root_dir=sdist_root_dir
+        )
+        self._handle_build_requirements(
+            RequirementType.BUILD_SDIST,
+            build_sdist_dependencies,
+        )
```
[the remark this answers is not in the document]
shubhbapna: Refactored it let me know if the current version is better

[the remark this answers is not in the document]
shubhbapna: nevermind the refactor caused some e2e tests to fail so maybe it is better suited for a seperate PR

src/fromager/sdist.py
dhellmann: Maybe I'm a packrat.

I like to be able to see the evolution of code, even if that means spanning files sometimes. It's not the end of the world to not have it.

Issue #411, opened 2024-09-12
Doc: Document output directories created by fromager 


rd4398:
We need to document the output directories created by fromager

rd4398:
@dhellmann I have a few questions:

- Do we create a separate .`md ` file for this like we did for input/output files?
- Just to confirm, we want to write about: `sdists-repo`, 'wheels-repo` and `work-dir` right?
- How much detailed information do we want? Is short description okay? Do we want to include sub-directory structure as well?

dhellmann:
> @dhellmann I have a few questions:
> 
> * Do we create a separate .`md ` file for this like we did for input/output files?

The `files.md` page has the title "Input and Output files", so maybe the new info can go there?

> * Just to confirm, we want to write about: `sdists-repo`, 'wheels-repo`and`work-dir` right?

All of the output directories and their subdirectories. Look at the output from a simple bootstrap to see what you get.

> * How much detailed information do we want? Is short description okay? Do we want to include sub-directory structure as well?

Yes, subdirectories. It would be good to know what sorts of files are written to each directory, what files we write, etc. Everything the user needs to understand what we're doing and what information we're generating.

PR #454, opened 2024-09-27
Document output directories generated by fromager

rd4398:
This commit adds documentation for the output directories generated by fromager during the wheel build process.

Fixes: https://github.com/python-wheel-build/fromager/issues/411


docs/files.md
```diff
@@ -93,3 +93,132 @@ The following example shows `graph.json` file for the top-level dependency `whee
...
+
+This directory contains the source distributions for the package and its dependencies that we are building. The directory structure of `sdists-repo` is as follows
+
+```
+sdists-repo
+├── builds
+└── downloads
+```
+
+The `builds` and `downloads` sub-directories contain the source distributions that are built and downloaded respectively in `tar.gz` format. For example, the `sdists-repo` for `stevedore` package looks as follows:
```
dhellmann, line 111: You could clarify here that the `downloads` directory includes the original sdist from upstream and the `builds` directory contains the sdist created by fromager after any patches are applied.

docs/files.md
```diff
@@ -93,3 +93,132 @@ The following example shows `graph.json` file for the top-level dependency `whee
...
+
+```
+wheels-repo
+├── build
+├── downloads
+├── prebuilt
+└── simple
+```
+
+* The `build` sub-directoy contains the configuration information related to the wheel build
```
dhellmann, line 140: The build directory holds temporary builds. We use it as the output directory when building the wheel because we can't predict the filename, and so using an empty directory with a name we know gives us a way to find the file and move it into the `downloads` directory after it's built.

Explaining that is highlighting for me that we may want to change some of these names.

docs/files.md
```diff
@@ -93,3 +93,132 @@ The following example shows `graph.json` file for the top-level dependency `whee
...
+* The `downloads` sub-directory contains the wheels in `.whl` format that were downloaded from an index
```
dhellmann, line 141: This isn't quite correct. We combine the wheels we build with the pre-built wheels here so we can create a local package index in `simple`. Again, this name isn't ideal.

docs/files.md
```diff
@@ -93,3 +93,132 @@ The following example shows `graph.json` file for the top-level dependency `whee
...
+* The `prebuilt` sub-directory contains wheels that are being used as prebuilt
+* The `simple` sub-directory consists of sub-directories that include wheels built by fromager in `whl` format along with an `index.html` file which contains the link to download the wheel. The `index.html` file present at the root level of this directory includes links to download wheels for the package and its dependencies
```
dhellmann, line 143: The `simple` directory is managed by [pypi-mirror](https://pypi.org/project/pypi-mirror/) to create a local wheel index.

docs/files.md
```diff
@@ -93,3 +93,132 @@ The following example shows `graph.json` file for the top-level dependency `whee
...
+    │   └── setuptools-75.1.0-0-py3-none-any.whl
+    └── stevedore
+        ├── index.html
+        └── stevedore-5.3.0-0-py3-none-any.whl
+
+```
+
+### work-dir
+
+This directory contains information that is required during the wheel build process. This information includes logs, constraints, requirements, graph for dependency resolution and the order in which the package and its dependencies will be built. The directory structure of `work-dir` is as follows:
```
dhellmann, line 171: One kind of item not shown here is the unpacked source code of each wheel that is used for the build. If you use the `--no-cleanup` flag those directories will be retained.

docs/files.md
```diff
@@ -93,3 +93,132 @@ The following example shows `graph.json` file for the top-level dependency `whee
...
+   ├── build-backend-requirements.txt
+   ├── build.log
+   ├── build-meta.json
+   ├── build-sdist-requirements.txt
+   ├── build-system-requirements.txt
+   └── requirements.txt
+
+```
+
+* The `build-order.json` file is an output file that contains the bottom-up order in which the dependencies need to be built for a specific wheel
```
dhellmann, line 189: You could link to the existing description of this file for more details.

docs/files.md
```diff
@@ -93,3 +93,132 @@ The following example shows `graph.json` file for the top-level dependency `whee
...
+* The `constraints.txt` is the input file used to control the resolution of python packages that are needed to build a wheel
```
dhellmann, line 190: There are 2 constraints files. This one is the _output_ file, produced by fromager, showing all of the versions of the packages that are install-time dependencies of the top-level items.

docs/files.md
```diff
@@ -93,3 +93,132 @@ The following example shows `graph.json` file for the top-level dependency `whee
...
+* The `graph.json` is an output file that contains all the paths fromager can take to resolve a dependency during building the wheel
```
dhellmann, line 191: You could link to the section that explains this file.

docs/files.md
```diff
@@ -93,3 +93,132 @@ The following example shows `graph.json` file for the top-level dependency `whee
...
+* The `logs` sub-directory contains detailed logs for fromager's `build-sequence` command including various settings and overrides for each individual package and its dependencies whose wheel was built by fromager
```
dhellmann, line 192: It would be good to clarify what each file contains. It's just the build of the given wheel, right? There are files for the dependencies, if they are built, but the log for a wheel does not include the info about building its dependencies?

docs/files.md
```diff
@@ -93,3 +93,132 @@ The following example shows `graph.json` file for the top-level dependency `whee
...
+* The `work-dir` also includes sub-directories for the package and its dependencies. These sub-directories include various types of requirements files including `build-backend-requirements.txt`, `build-sdists-requirements.txt`, `build-system-requirements.txt` and the general `requirements.txt`. Files like `build.log` which store the logs generated by PyPI and `build-meta.json` that stores the metadata for the build are also located in `work-dir`. These sub-directories also include all the other relevant information for a particular package
```
dhellmann, line 193: The build.log is generated by `pip`, not PyPI.

docs/files.md
```diff
@@ -93,3 +93,132 @@ The following example shows `graph.json` file for the top-level dependency `whee
...
+
+For example, the `work-dir` for `stevedore` package after `bootstrap` command looks as follows:
+
+```
+.
+├── build-order.json
+├── constraints.txt
+├── graph.json
+├── logs
```
dhellmann, line 202: Why is this directory empty?

docs/files.md
rd4398: I have tried to add more details but not sure whether that is what is required.

rd4398:
I have made the changes based on above feedback

docs/files.md
```diff
@@ -93,3 +93,133 @@ The following example shows `graph.json` file for the top-level dependency `whee
...
+
+This directory contains the source distributions for the package and its dependencies that we are building. The directory structure of `sdists-repo` is as follows
+
+```
+sdists-repo
+├── builds
+└── downloads
+```
+
+The `builds` and `downloads` sub-directories contain the source distributions that are built and downloaded respectively in `tar.gz` format. The `downloads` directory includes the original sdist from upstream and the `builds` directory contains the sdist created by fromager after any patches are applied. For example, the `sdists-repo` for `stevedore` package looks as follows:
```
dhellmann, line 111: Some sdists are zip files, so let's not specify anything about the format.

```suggestion
The `builds` and `downloads` sub-directories contain the source distributions that are built and downloaded respectively. The `downloads` directory includes the original sdist from upstream and the `builds` directory contains the sdist created by fromager after any patches are applied. For example, the `sdists-repo` for `stevedore` package looks as follows:
```

rd4398: okay, done!

dhellmann (approved):

PR #458, opened 2024-10-02
shlex.quote environment values

tiran:
`external_commands.run()` now quotes environment values, too. This enables copy and pasting when an environment variable contains spaces.

See: #455


dhellmann (approved):
Nice improvement, thank you!

PR #459, opened 2024-10-02
Use keyword-only args for API functions

tiran:
Import API functions no longer accept positional arguments and only support keyword arguments. This makes the code more reliable, readable, and prepares the functions for the `timeit` decorator.

Related: #408
Related: #445


shubhbapna (commented):
LGTM, but will wait for @dhellmann review since this is a breaking change

dhellmann (commented):
This will make upgrades downstream harder again, but long-term it's a good change.

When is the best time to merge this?

dhellmann (approved):

PR #466, opened 2024-10-02
throw an error there is no version that works for all dependencies when generating the constraints file

### LINKED ISSUE #465: throw an error if fromager cannot decide which dependency to pick when generating the constraints file
Right now when fromager cannot pick one version that works for all dependencies when generating the constraints file, it just adds that error as a comment in the file. Lets complete the generation of the constraints file but also throw an error at the end to bring this to the attention of the user immediately

shubhbapna:
fixes #465


src/fromager/commands/bootstrap.py
```diff
@@ -156,18 +156,21 @@ def bootstrap(
 
     constraints_filename = wkctx.work_dir / "constraints.txt"
     logger.info(f"writing installation dependencies to {constraints_filename}")
-    with open(wkctx.work_dir / "constraints.txt", "w") as f:
-        write_constraints_file(graph=wkctx.dependency_graph, output=f)
+    with open(constraints_filename, "w") as f:
+        if not write_constraints_file(graph=wkctx.dependency_graph, output=f):
+            raise ValueError(
+                f"Could not produce a pip compatible constraints file. Please review {constraints_filename} for more details"
+            )
```
dhellmann, line 163: If we throw the exception inside the `with` block, is the file still written properly?

shubhbapna: the writing happens inside the `write_constraints_file` function so it should be fine

shubhbapna: not sure, know any small enough packages that have conflicting installation time deps?

dhellmann: What if we construct our own fake package?

shubhbapna: I realized we can just pass 2 conflicting toplevel dependencies

PR #470, opened 2024-10-07
Extended template variables for extra_environ

tiran:
Default values for extra environ templates can now be set in the global settings file. The new feature can be used to set defaults like `CFLAGS: ''`, so `CFLAGS: '-mavx2 ${CFLAGS}'` always works.


docs/customization.md
```diff
@@ -154,6 +154,45 @@ variants:
             PATH: "/cpu/bin:$PATH"
 ```
 
+#### template envs
+
+.. versionadded: 0.31
+
+Template variables for extra environment variables can be configured in the
```
dhellmann, line 161: So these are only included in the logic for expanding template values, but are not "defaults" for these variables because they aren't passed unless the variant-specific configuration sets the variable, too?

tiran: That's correct.

dhellmann: That seems likely to introduce some confusion. What's the downside to just treating them as global defaults?

tiran:
PR #479 added support for `${NAME:-somedefault}`. The new features solves the problem in a easy-to-understand way.

PR #471, opened 2024-10-10
feat: add caching to resolver

### LINKED ISSUE #359: add caching in resolver

shubhbapna:
fixes #359 

Summarizing the needs from #358 (and from our previous calls)

- we shouldn't use caching for install type requirements because we always want the latest version that matches the reqs and constraints and we cannot guarantee that the latest version will exist in the cache. For example:
  - first we resolve `numpy<2` which ends up populating the cache with say `numpy==1`. Then we want to resolve `numpy>=1` as an install type req. if we end up using the cache it would return `numpy==1` even though there is a `numpy==2` that we can use.
  - we want install type reqs to be the latest possible version so that they have the latest possible cve fixes in them

- the cache is something internal to the resolver and should not be influenced directly by the caller of the resolver

### DIFF
--- a/src/fromager/resolver.py
+++ b/src/fromager/resolver.py
@@ -8,6 +8,7 @@ from __future__ import annotations
 import logging
 import os
 import typing
+from collections import defaultdict
 from operator import attrgetter
 from platform import python_version
 from urllib.parse import urljoin, urlparse
@@ -31,6 +32,7 @@ from .candidate import Candidate
 from .constraints import Constraints
 from .extras_provider import ExtrasProvider
 from .request_session import session
+from .requirements_file import RequirementType
 
 if typing.TYPE_CHECKING:
     from . import context
@@ -51,6 +53,7 @@ def resolve(
     sdist_server_url: str,
     include_sdists: bool = True,
     include_wheels: bool = True,
+    req_type: RequirementType | None = None,
 ) -> tuple[str, Version]:
     # Create the (reusable) resolver.
     provider = overrides.find_and_invoke(
@@ -62,6 +65,7 @@ def resolve(
         include_sdists=include_sdists,
         include_wheels=include_wheels,
         sdist_server_url=sdist_server_url,
+        req_type=req_type,
     )
     return resolve_from_provider(provider, req)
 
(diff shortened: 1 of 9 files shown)


src/fromager/resolver.py
```diff
@@ -200,12 +206,14 @@ def __init__(
         include_wheels: bool = True,
         sdist_server_url: str = "https://pypi.org/simple/",
         constraints: Constraints | None = None,
+        req_type: RequirementType | None = None,
```
dhellmann, line 209: It feels a little odd to have the requirement type passed to the constructor instead of validate_candidate(). That implies the resolver will only resolve this type of dependencies. What led to that choice?

How much more code has to change if we add the argument to validate_candidate()?

src/fromager/resolver.py
```diff
@@ -262,151 +305,198 @@ def find_matches(
...
+            constraints=constraints,
+            req_type=req_type,
+        )
 
-    def find_matches(
+    @classmethod
+    def get_cache(cls) -> dict[str, list[Candidate]]:
+        return cls.pypi_resolver_cache
+
+    def validate_candidate(
```
dhellmann, line 330: How different are the 2 implementations of this method?

src/fromager/resolver.py
shubhbapna: validate_candidate is called by find_matches which is internally called by the resolvelib library functions. I am not sure how we will pass req_type to validate_candidate

src/fromager/resolver.py
shubhbapna: we don't do the sdist checks in generic provider but I see your point. Let me refactor it

src/fromager/resolver.py
dhellmann: Ah, ok, it sounds like we can't.

shubhbapna:
@dhellmann anymore changes required?

dhellmann (approved):

Issue #472, opened 2024-10-10
Use bubblewrap (bwrap) instead of unshare for network isolation


tiran:
Fromager implements network isolation with `unshare -rn`. The command creates a new network namespace without any interface and connection to the rest of the world. This works fine in most cases. However OpenMPI 4.x's compiler `mpicc` does not like our approach. It prints warnings to standard streams, which break's CMake's `FindMPI`.

One simple solution for the problem is [`bubblewrap`](https://github.com/containers/bubblewrap). The tool creates namespaces like `unshare` but with extra bells and whistles. `bwrap --unshare-net --dev-bind / /` unshares the network and creates a loopback interface in the new namespace. The `--dev-bind` is necessary to bind parent mounts into the new namespace and allow `/dev` access.

The tool is tiny (~ 100 kB) and available on all our target platforms.

pnasrat:
I''m defintely in favor of bubblewrap will take a look at the PR

shubhbapna:
`bwrap` didn't work for us as documented here https://github.com/python-wheel-build/fromager/pull/473#issuecomment-2412940902

we ended up using `unshare` with `ip link` see #477

PR #473, opened 2024-10-10
Prefer bubblewrap for network isolation

tiran:
Bubblewrap is another tool for unsharing namespaces. It sets up a network namespace with a disconnected loopback.

Fixes: #472

### DIFF
--- a/src/fromager/external_commands.py
+++ b/src/fromager/external_commands.py
@@ -9,9 +9,12 @@ from io import TextIOWrapper
 
 logger = logging.getLogger(__name__)
 
-NETWORK_ISOLATION: list[str] | None
+NETWORK_ISOLATION: list[list[str]] | None
 if sys.platform == "linux":
-    NETWORK_ISOLATION = ["unshare", "--net", "--map-current-user"]
+    NETWORK_ISOLATION = [
+        ["bwrap", "--unshare-network", "--dev-bind", "/", "/", "--"],
+        ["unshare", "--net", "--map-current-user"],
+    ]
 else:
     NETWORK_ISOLATION = None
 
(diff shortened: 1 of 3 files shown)


dhellmann (commented):
This looks OK for now. We should consider if this should be configurable by the user somehow, but this at least lets us move forward.

Edit: nvm.  I tested it and it seems to work -- see comments below about adding --privileged to docker and fixing the --unshare-net option.

prarit:
Let's try this and see if it works better.  OOC do you test if it solved the MPI issue in a container?

tiran:
> Let's try this and see if it works better. OOC do you test if it solved the MPI issue in a container?

```shell
$ bwrap --unshare-net --dev-bind / / -- /usr/lib64/openmpi/bin/mpicc -showme:compile
-I/usr/include/openmpi-x86_64

$ unshare -rn -- /usr/lib64/openmpi/bin/mpicc -showme:compile
[hostname:1498104] opal_ifinit: unable to find network interfaces.
-I/usr/include/openmpi-x86_64
```

src/fromager/external_commands.py
```diff
@@ -9,9 +9,12 @@
 
 logger = logging.getLogger(__name__)
 
-NETWORK_ISOLATION: list[str] | None
+NETWORK_ISOLATION: list[list[str]] | None
 if sys.platform == "linux":
-    NETWORK_ISOLATION = ["unshare", "--net", "--map-current-user"]
+    NETWORK_ISOLATION = [
+        ["bwrap", "--unshare-network", "--dev-bind", "/", "/", "--"],
```
prarit, line 15: --unshare-network doesn't appear to be valid?  Is this supposed to be --unshare-net ?

prarit: FYI: this needs docker to specify --privileged in builder's bin/boostrap.sh

tiran:
`bwrap` does not work for us. It requires additional permissions and adjustments to work in an unprivileged container. We want a system that works in `podman` out of the box and in `docker` with minimal adjustments.

PR #474, opened 2024-10-11
throw an error when there is no version that works for all dependencies when generating the constraints file

shubhbapna:
backport of #466


dhellmann (approved):
Thanks!

PR #476, opened 2024-10-11
Implement unshare as preexec function

tiran:
Roll our own `unshare` implementation that sets up a loopback device inside the new namespace. Some software like OpenMPI's `mpicc` do not work correctly without a loopback device.

### DIFF
--- a/e2e/test_build.sh
+++ b/e2e/test_build.sh
@@ -14,13 +14,6 @@ VERSION="5.2.0"
 # Install hook for test
 pip install e2e/post_build_hook
 
-OS=$(uname)
-if [ "$OS" = "Darwin" ]; then
-    NETWORK_ISOLATION=""
-else
-    NETWORK_ISOLATION="--network-isolation"
-fi
-
 # Bootstrap the test project
 fromager \
     $NETWORK_ISOLATION \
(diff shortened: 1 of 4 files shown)


dhellmann:
This implementation feels pretty low level to live in the core code. I'd be much more comfortable with a configuration option for a "build command wrapper" with some examples of how to do this sort of thing as shell scripts. We should document them, or deliver them, or both.

shubhbapna:
I am not sure how much of a concern this is but the documentation says that we should avoid using `preexec_fn` in the presence of threads. I do plan to use threads as part of performance improvements #343

prarit:
I'm curious what this looks like when fromager logs a command.  Today I would see something like "unshare --net --map-current-user $env $command" which is easy to execute on a command line in a container.  Do we lose that ease of execution?

tiran:
@prarit Yes, you would no longer see any indication in the command line of the process. The `preexec_fn` runs between `fork` and `exec` syscalls and modifies the execution environment of the child process.

@shubhbapna Threading deadlocks can occur when the `preexec_fn` callback is executed in a subthread and the callback interacts with a lock from another thread. `fork` and threads don't mix well. Simply speaking, `fork` only clones the current thread and the child becomes single-threaded.

- https://en.wikipedia.org/wiki/Fork%E2%80%93exec
- https://man7.org/linux/man-pages/man2/fork.2.html
- https://man7.org/linux/man-pages/man7/signal-safety.7.html

Because this approach has some issues and is harder to maintain, I'm abandoning it in favor of a shell wrapper. It is not as elegant and fancy, but it's a more reasonable and easier to understand implementation.

PR #477, opened 2024-10-15
Fix: enable loopback device in network isolation

tiran:
Some software like OpenMPI's `mpicc` do not work correctly without a loopback device. Replace the `unshare` call with a wrapper script that combines `unshare -rn` and `ip link set lo up` before `exec`-ing the final command.

`unshare` is a `util-linux-core` command and therefore available pretty much in all Linux distributions.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -38,6 +38,14 @@ jobs:
           default: true
           override: true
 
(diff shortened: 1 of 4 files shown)


dhellmann (approved):
I like this approach. It's much easier to document the environment in which an individual wheel build runs, and leaves the possibility that we make it user-configurable in the future.

PR #482, opened 2024-10-22
expose extra_environ argument in get build dependency hooks

### LINKED ISSUE #481: add `extra_environ`  as an argument to dependencies hooks
Packages like pypdfium2 need certain environment variable set dynamically before calling the `get_build_backend_dependencies` and `get_build_sdist_dependencies`. So we need to expose the `extra_environ` as an argument so that plugins can modify it and call the default functions

shubhbapna:
fixes #481 

Passes `extra_environ` as an argument to get build dependency hooks instead of getting them from settings directly. This will allow plugins to dynamically add more env variables

### DIFF
--- a/src/fromager/dependencies.py
+++ b/src/fromager/dependencies.py
@@ -102,6 +102,7 @@ def get_build_backend_dependencies(
         )
         return _read_requirements_file(build_backend_req_file)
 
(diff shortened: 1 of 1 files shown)


tiran (approved):
I like the idea!

PR #485, opened 2024-10-25
Support variant specific patches

### LINKED ISSUE #486: Support variant specific patches
When building wheels for different variants, the code base might be different from one variant to the other, for example when the code comes from a fork with variant specific features. And with a different code base, the patches may fail to be applied.

fabiendupont:
When building wheels for different variants, the code base might be
different from one variant to the other, for example when the code comes
from a fork with variant specific features. And with a different code
base, the patches may fail to be applied.

This changes proposes to support patch files stored in a variant
specific path under the existing unversioned and versioned patch
directories. Below is an example of a patch tree:

```
.../patches
.../patches/deepspeed
.../pacthes/deepspeed/0001-unversioned.patch
.../pacthes/deepspeed/0002-unversioned.patch
.../patches/deepspeed/brie/0001-unversioned.patch
.../patches/deepspeed/feta/0002-unversioned.patch
.../patches/deepspeed-0.5.0/0001-versioned.patch
.../patches/deepspeed-0.5.0/brie/0003-versioned.patch
```

When applying the patches for the `brie` variant, the following patches
would be applied:

```
.../pacthes/deepspeed/0001-unversioned.patch
.../patches/deepspeed/brie/0001-unversioned.patch
.../pacthes/deepspeed/0002-unversioned.patch
.../patches/deepspeed-0.5.0/0001-versioned.patch
.../patches/deepspeed-0.5.0/brie/0003-versioned.patch
```

When applying the patches for the `feta` variant, the following patches
would be applied:

```
.../pacthes/deepspeed/0001-unversioned.patch
.../pacthes/deepspeed/0002-unversioned.patch
.../patches/deepspeed/feta/0002-unversioned.patch
.../patches/deepspeed-0.5.0/0001-versioned.patch
```

Fixes #486

### DIFF
--- a/src/fromager/overrides.py
+++ b/src/fromager/overrides.py
@@ -98,6 +98,7 @@ def patches_for_requirement(
     patches_dir: pathlib.Path,
     req: Requirement,
     version: Version,
+    variant: str = "",
 ) -> typing.Iterable[pathlib.Path]:
     """Iterator producing patches to apply to the source for a given version of a requirement.
 
@@ -108,12 +109,29 @@ def patches_for_requirement(
     override_name = pkgname_to_override_module(req.name)
     unversioned_patch_dir = patches_dir / override_name
     versioned_patch_dir = patches_dir / f"{override_name}-{version}"
+
+    unversioned_patch_files = list(unversioned_patch_dir.glob("*.patch"))
+    versioned_patch_files = list(versioned_patch_dir.glob("*.patch"))
+
+    # The list of files must exist to be joined to the global patch files
+    if variant:
+        unversioned_variant_patch_dir = unversioned_patch_dir / variant
+        if unversioned_variant_patch_dir.exists():
+            unversioned_patch_files.extend(
+                list(unversioned_variant_patch_dir.glob("*.patch"))
+            )
+        versioned_variant_patch_dir = versioned_patch_dir / variant
+        if versioned_variant_patch_dir.exists():
+            versioned_patch_files.extend(
+                list(versioned_variant_patch_dir.glob("*.patch"))
+            )
+
     return itertools.chain(
         # Apply all of the unversioned patches first, in order based on
         # filename.
-        sorted(unversioned_patch_dir.glob("*.patch")),
+        sorted(unversioned_patch_files, key=lambda f: f.name),
         # Then apply any for this specific version, in order based on filename.
-        sorted(versioned_patch_dir.glob("*.patch")),
+        sorted(versioned_patch_files, key=lambda f: f.name),
     )
 
 
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -174,6 +174,10 @@ name and use the suffix `.patch`. The filenames are sorted lexicographically, so
 any text between the prefix and suffix can be used to ensure the patches are
 applied in a specific order.
 
+Patch files can also be placed in a variant specific subdirectory, in order
+to allow variant specific patches, e.g. when the code base is a variant specific
+fork of the package and the global patches don't apply.
+
 Patches are applied by running `patch -p1 filename` while inside the root of the
 source tree.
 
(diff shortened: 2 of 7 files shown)


src/fromager/overrides.py
```diff
@@ -108,12 +109,23 @@ def patches_for_requirement(
     override_name = pkgname_to_override_module(req.name)
     unversioned_patch_dir = patches_dir / override_name
     versioned_patch_dir = patches_dir / f"{override_name}-{version}"
+
+    unversioned_patch_files = list(unversioned_patch_dir.glob("*.patch"))
+    versioned_patch_files = list(versioned_patch_dir.glob("*.patch"))
+
+    # The list of files must exist to be joined to the global patch files
```
dhellmann, line 116: For versioned vs. unversioned we use different directories. We follow that pattern elsewhere downstream with directories for the variant-specific parts. What's the benefit of introducing a new approach for these files?

fabiendupont: Not sure I get the question. I am only moving this `glob` here to allow adding files to the list before sorting it. The versioned and unversioned directories don't change, only the suffix we're looking for.

src/fromager/overrides.py
```diff
@@ -108,12 +109,23 @@ def patches_for_requirement(
...
+    if variant:
+        unversioned_patch_files += list(
+            unversioned_patch_dir.glob(f"*.patch.{variant}")
+        )
+        versioned_patch_files += list(versioned_patch_dir.glob(f"*.patch.{variant}"))
```
tiran, line 121: - I don't like the `.variant` suffix. How about a subdirectory?
- use list's `extend()` method to extend the lists.

src/fromager/overrides.py
```diff
@@ -98,6 +98,7 @@ def patches_for_requirement(
     patches_dir: pathlib.Path,
     req: Requirement,
     version: Version,
+    variant: str = "",
```
tiran, line 101: The variant name should be passed by passing a `Context` object as first argument.

src/fromager/overrides.py
```diff
@@ -108,12 +109,23 @@ def patches_for_requirement(
...
+        unversioned_patch_files += list(
+            unversioned_patch_dir.glob(f"*.patch.{variant}")
+        )
+        versioned_patch_files += list(versioned_patch_dir.glob(f"*.patch.{variant}"))
+
     return itertools.chain(
         # Apply all of the unversioned patches first, in order based on
         # filename.
-        sorted(unversioned_patch_dir.glob("*.patch")),
+        sorted(unversioned_patch_files),
```
tiran, line 126: I suggest to sort by the file's name, not the full path.

tiran (changes requested):
I would prefer `patch_dir / ctx.variant / "*.patch"`.

There is also code in `packagesettings` that deals with patches. You have to extend the code and its test it, too.

Please add documentation.

src/fromager/overrides.py
tiran: I have suggested `patches_dir / override_name / variant / "*.patch"` and `patches_dir / f"{override_name}-{version}" / variant / "*.patch"`. This keeps the variant patches in a subfolder. IMHO that is less clutter and easier to understand.

dhellmann: I like that. It also makes it less likely we would have some sort of issue parsing the path name.

src/fromager/overrides.py
tiran: Sounds complicated. I won't be able to recall the correct order in a week from now. How about we throw everything in a single list, then sort all files by their base name?

dhellmann: We already give precedence to the files in the directory without the version. I think we don't have any of those downstream, so we probably wouldn't break anything making the change you propose. I guess it lets users manage precedence themselves using numerical prefixes, too, so maybe that's a good thing?

fabiendupont: Yes, that was the idea. The prefix is what decides the order. The user can easily insert a variant specific patch by using the same number as the global patch. `variant/0001-xx.patch` will have precedence over `0002-aa.patch`.

src/fromager/overrides.py
fabiendupont: That would be the only method with a `Context` object in this package. This is mostly helpers. And an extra variable with a default value reduces the size of the change.

Not saying I'm against the idea. Just giving more context about why I did it this way.

fabiendupont:
Seems that https://github.com/python-wheel-build/fromager/issues/486 is also implementing this change. @tiran, @dhellmann, I guess this PR can be closed, right?

dhellmann:
> Seems that #488 is also implementing this change. @tiran, @dhellmann, I guess this PR can be closed, right?

I think so, yes.

fabiendupont:
Closing in favor of #488

PR #488, opened 2024-10-28
Implement variant-specific patches

tiran:
Fromager now supports variant-specific patches in additional to versioned and unversioned patches. Variant-specific patches are in subfolders, e.g. `atches/test_pkg-1.0.2/cpu/005-cpuver.patch`.

Patch files are sorted alphanumerically by their *base* name, not their full path. This allows users to sort versioned/unversioned and plain/variant patches as they like.

Internal changes: `sources.patch_source()` now uses `PackageBuildInfo` to get patches. `patches_for_requirement()` has been removed.

Fixes: #486

### DIFF
--- a/src/fromager/packagesettings.py
+++ b/src/fromager/packagesettings.py
@@ -88,7 +88,7 @@ Package = typing.Annotated[
 ]
 
 # patch mapping
-PatchMap = dict[Version, typing.Iterable[pathlib.Path]]
+PatchMap = dict[Version | None, list[pathlib.Path]]
 
 # URL or filename with templating
 Template = typing.NewType("Template", str)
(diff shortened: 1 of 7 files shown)


rd4398 (approved):
LGTM!
I will leave it to @dhellmann for final approval and merging

src/fromager/packagesettings.py
```diff
@@ -493,18 +493,38 @@ def plugin(self) -> types.ModuleType | None:
...
 
-    def get_patches(self) -> PatchMap:
+    def get_all_patches(self) -> PatchMap:
         """Get a mapping of version to list of patches"""
         if self._patches is None:
             patches: PatchMap = {}
-            pattern = f"{self.override_module_name}-*"
-            prefix_len = len(pattern) - 1
+            version: Version | None
+            pattern = f"{self.override_module_name}*"
```
dhellmann, line 501: This pattern is going to find patches for `foo-bar` when it the module is `foo`.

tiran: Patch directories use module override names, so `foo-bar` is normalized to `foo_bar`.

dhellmann: I added a test and fix in #493

PR #491, opened 2024-11-01
only show build settings for build commands

### LINKED ISSUE #440: the canonicalize command dumps a bunch of configuration settings in addition to the expected output
```
$ fromager canonicalize pydantic-core
primary settings file: overrides/settings.yaml
per-package settings dir: overrides/settings
variant: cpu
patches dir: overrides/patches
maximum concurrent jobs: None
constraints file: None
wheel server url:
network isolation: True
pydantic_core
```

dhellmann:
Fixes #440

### DIFF
--- a/src/fromager/__main__.py
+++ b/src/fromager/__main__.py
@@ -179,15 +179,20 @@ def main(
     if error_log_file:
         logger.info("logging errors to %s", error_log_file)
 
-    logger.info(f"primary settings file: {settings_file}")
-    logger.info(f"per-package settings dir: {settings_dir}")
-    logger.info(f"variant: {variant}")
-    logger.info(f"patches dir: {patches_dir}")
-    logger.info(f"maximum concurrent jobs: {jobs}")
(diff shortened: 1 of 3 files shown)


rd4398 (commented):
The changes look good! Just a couple of things:

- Looks like the `mypy` checks are failing
- The linked issue mentions similar behavior for the `graph` based commands as well. However, I did not see any changes related to this. Can the similar changes be applied there? or did I miss anything?

dhellmann:
> The changes look good! Just a couple of things:
> 
> * Looks like the `mypy` checks are failing

I fixed that in a separate PR, so CI should be green now.

> * The linked issue mentions similar behavior for the `graph` based commands as well. However, I did not see any changes related to this. Can the similar changes be applied there? or did I miss anything?

I implemented the check by looking for a flag to be set, and defaulting to False if the flag is not found. That way I only had to enable the flag on a few commands where we do want the output. I'll add a comment to that effect where the flag is being checked.

rd4398 (approved):

PR #493, opened 2024-11-03
ensure packages with common prefixes do not mix patches

dhellmann:
Fix a problem introduced in #488 that could cause packages with a prefix
common to another package's full name to have their patches included
when scanning for patches of the package with the shorter name. For
example, with patches for both 'foo' and 'foo-bar' when patches for
'foo' are loaded they would have included patches for 'foo-bar'.


tiran (approved):
Ouch, now I understand the problem. Good catch!

PR #497, opened 2024-11-07
Bug Fix: terminate when error is encountered

rd4398:
This commit fixes a bug related to error message handling. The `main` will now exit using `sys.exit(1)` when there is an error

### DIFF
--- a/src/fromager/__main__.py
+++ b/src/fromager/__main__.py
@@ -2,6 +2,7 @@
 
 import logging
 import pathlib
+import sys
 
 import click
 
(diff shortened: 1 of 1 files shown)


src/fromager/__main__.py
```diff
@@ -249,7 +250,7 @@ def invoke_main() -> None:
         )  # log the full traceback details to the debug log file, if any
         logger.error(f"ERROR: {err}")
         if _DEBUG:
-            raise
+            sys.exit(1)
```
dhellmann, line 253: This isn't quite what we want.

We always want the exit with an error if there's an exception. When `_DEBUG` is set, then we want to raise the exception again to get the traceback. So, I think we want something like this

```suggestion
            raise
        sys.exit(1)
```

rd4398: Ah okay. Pushed changes again

dhellmann (approved):

PR #499, opened 2024-11-07
improve error and debugging info during bootstrapping

dhellmann:


src/fromager/overrides.py
```diff
@@ -50,7 +50,12 @@ def find_and_invoke(
         fn = default_fn
 
     result = invoke(fn, **kwargs)
-    logger.info(f"{distname}: override method {fn.__name__} returned {result}")
+    if fn is default_fn:
+        log_method = logger.debug
+    else:
+        log_method = logger.info
+    log_method(f"{distname}: override method {fn.__name__} returned {result}")
+
```
rd4398, line 58: Any specific reason why we are using both ` logger.debug` and` logger.info` depending on log_method?
(Just wanted to know)

dhellmann: Good question. I explained in the commit message, but should have included a comment here.

I'm trying to make the info level output, that goes to the console, useful without being overly verbose. So, when there is a default function, we want to see the return values but we can be pretty sure those are the right "type" of value, so I just log them at debug. When the function is an actual plugin, it is more likely that the value returned will be of interest and so I log it at info level (so it goes to the console and to the log file).

rd4398: Ah, okay got it!
Thank you

rd4398 (approved):
Looks good to me!

PR #502, opened 2024-11-12
use a wheel server as a cache for bootstrapping

### LINKED ISSUE #449: Use the `*-requirements.txt` file put into the wheels by fromager instead of recomputing them

### LINKED ISSUE #450: Add a flag that downloads all the wheels from the given index instead of using `download-sequence`
Advantages over `download-sequence`:
- we only download things as and when needed
- don't need the user to be worries about whether `download-sequence` is downloading the wheels at the right place
- we can download the wheel before we even download the source to enable #449

shubhbapna:
fixes #449 fixes #450 

By using a cache wheel server url, fromager can download already built wheels as and when needed to speed up the bootstrapping process by avoiding to actually build those wheels. Plus, if those wheels were built by fromager itself, then it can extract the build requirements files from it instead of recomputing them again.


src/fromager/bootstrapper.py
```diff
@@ -101,19 +104,33 @@ def bootstrap(self, req: Requirement, req_type: RequirementType) -> Version:
...
-                sdist_root_dir,
-                unpack_dir,
-                build_dependencies,
-            ) = self._prepare_source(req, source_filename, resolved_version)
-
-            found_wheel_filename = self._is_wheel_built(req, resolved_version)
-            if not found_wheel_filename:
-                wheel_filename, build_env = self._build(
-                    req, resolved_version, sdist_root_dir, build_dependencies
+            unpacked_cached_wheel, cached_wheel_filename = (
```
dhellmann, line 107: This section looks correct, but might actually be clearer if the stages were handled one after the other instead of having 2 branches.

```
if not unpacked_cached_wheel:
  # prepare source

# prepare build dependencies (always, so no "if" statement)

if not unpacked_cached_wheel and not cached_wheel_filename:
  # build
```

That sort of organization shows that it's a linear flow through a set of steps but that we skip some steps. If we add a step later, we can add it to the sequence at the right point with the right conditions.

It would also let us move the logic into a separate function (`_handle_building_wheel`?) that just returns early if it can download the wheel from the cache.

src/fromager/bootstrapper.py
```diff
@@ -316,11 +303,81 @@ def _handle_prebuilt_sources(
...
-            unpack_dir.mkdir()
+        unpack_dir = self._create_unpack_dir(req, resolved_version)
         return (resolved_version, wheel_url, wheel_filename, unpack_dir)
 
+    def _download_wheel_from_cache(
+        self, req: Requirement, resolved_version: Version
+    ) -> tuple[pathlib.Path | None, pathlib.Path | None]:
+        if not self.cache_wheel_server_url:
+            return None, None
+        logger.info(f"{req.name}: checking if wheel was already built")
```
dhellmann, line 314:
```suggestion
        logger.info(f"{req.name}: checking if wheel was already uploaded to {self.cache_wheel_server_url}")
```

src/fromager/bootstrapper.py
```diff
@@ -316,11 +303,81 @@ def _handle_prebuilt_sources(
...
+                include_wheels=True,
+            )
+            cached_wheel = wheels.download_wheel(
+                req=req, wheel_url=wheel_url, output_directory=self.ctx.wheels_downloads
+            )
+            server.update_wheel_mirror(self.ctx)
+
+            pbi = self.ctx.package_build_info(req)
+            expected_build_tag = pbi.build_tag(resolved_version)
+            dist_name, dist_version, build_tag, _ = wheels.extract_info_from_wheel_file(
```
dhellmann, line 330: Do we have to download the wheel because we want to call this function and it needs the file? Or can we parse the URL to get the build tag and avoid doing the download if it's wrong?

src/fromager/bootstrapper.py
```diff
@@ -316,11 +303,81 @@ def _handle_prebuilt_sources(
...
+                    unpack_dir / dependencies.BUILD_SDIST_REQ_FILE_NAME,
+                )
+                logger.info(f"{req.name}: extracted build requirements from wheel")
+                return unpack_dir / dist_filename, cached_wheel
+            except Exception:
+                # implies that the wheel server hosted non-fromager built wheels
+                logger.info(
+                    f"{req.name}: could not extract build requirements from wheel"
+                )
+                return None, cached_wheel
```
dhellmann, line 374: If we can't extract all of the requirements info, should we delete whatever we could extract so that we recompute it? Or does this return value trigger that behavior?

src/fromager/commands/bootstrap.py
```diff
@@ -67,12 +67,19 @@ def _get_requirements_from_args(
...
+    "cache_wheel_server_url",
+    help="url to a wheel server from where fromager can download the wheels that it has built before",
+)
 @click.argument("toplevel", nargs=-1)
 @click.pass_obj
 def bootstrap(
     wkctx: context.WorkContext,
     requirements_files: list[pathlib.Path],
     previous_bootstrap_file: pathlib.Path | None,
+    cache_wheel_server_url: str | None,
```
dhellmann, line 82: We should log this configuration info on startup.

src/fromager/bootstrapper.py
shubhbapna: The return value triggers that behavior. This handles the case where the cache wheel server url is hosting non-fromager built wheels. This will allow the user to still skip building the wheel itself but will trigger a recomputation of the build requirements

src/fromager/bootstrapper.py
shubhbapna: Ah yes I think there is a way, let me try

shubhbapna: done

src/fromager/bootstrapper.py
```diff
@@ -316,11 +303,83 @@ def _handle_prebuilt_sources(
...
+        unpack_dir = self._create_unpack_dir(req, resolved_version)
         return (resolved_version, wheel_url, wheel_filename, unpack_dir)
 
+    def _download_wheel_from_cache(
+        self, req: Requirement, resolved_version: Version
+    ) -> tuple[pathlib.Path | None, pathlib.Path | None]:
+        if not self.cache_wheel_server_url:
+            return None, None
+        logger.info(
+            f"{req.name}: checking if wheel was already checking if wheel was already uploaded to {self.cache_wheel_server_url}"
```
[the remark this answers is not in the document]
shubhbapna: ahh oops

src/fromager/bootstrapper.py
shubhbapna: ah actually you are right we do need to do some cleanup

dhellmann (approved):

PR #503, opened 2024-11-12
improve error message when generated constraints file is invalid

### LINKED ISSUE #500: better error message when the generated constraints file is invalid
After bootstrap, fromager generates a constraints file to freeze all the install time dependencies. Sometimes it is not possible to generate this file because fromager is not able to pick one single version for a dependency. When this happens lets show why those versions of those dependencies were being pulled in by displaying the output of the `fromager graph why` command or `fromager graph explain-duplicates` command

shubhbapna:
fixes #500 

print why the conflicting dependencies were being pulled in at the end of bootstrap

### DIFF
--- a/src/fromager/commands/bootstrap.py
+++ b/src/fromager/commands/bootstrap.py
@@ -17,6 +17,7 @@ from .. import (
     sources,
     wheels,
 )
+from .graph import find_why
 
 # Map child_name==child_version to list of (parent_name==parent_version, Requirement)
 ReverseRequirements = dict[str, list[tuple[str, Requirement]]]
@@ -167,6 +168,7 @@ def write_constraints_file(
     # each package are needed.
     conflicts = graph.get_install_dependency_versions()
     ret = True
+    conflicting_deps = set()
     for dep_name, nodes in sorted(conflicts.items()):
         versions = [node.version for node in nodes]
         if len(versions) == 0:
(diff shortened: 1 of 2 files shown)


src/fromager/commands/bootstrap.py
```diff
@@ -226,9 +228,14 @@ def write_constraints_file(
...
             )
-            logging.error("%s: no single version meets all requirements", dep_name)
+            logger.error("%s: no single version meets all requirements", dep_name)
+            conflicting_deps.add(dep_name)
             for dv in sorted(versions):
                 output.write(f"{dep_name}=={dv}\n")
+    for dep_name in conflicting_deps:
+        logger.error("finding why %s was being used", dep_name)
+        for node in graph.get_nodes_by_name(dep_name):
+            find_why(graph, node, -1, 1, [])
```
shubhbapna, line 238: using `-1` here will cause `find_why` to print up till root level (i.e. for each dependency it will print parents, grandparents, great grandparents till the toplevel). If this is too much, I can change to only print the immediate parents.

dhellmann: The `explain-duplicates` output gives all of the packages that depend on the duplicate, and the version ranges allowed in each case. Will this output give us that information clearly?

shubhbapna: `explain-duplicates` will only give the packages that directly depend on the duplicate. `find_why` gives this information and will also tell us where the package that depends on the duplicate comes from

dhellmann: So we get the full dependency tree, including the actual requirements specifications at each step?

shubhbapna: yep including the req all the way upto the toplevel

dhellmann: OK, sounds good

dhellmann (approved):

PR #504, opened 2024-11-12
allow use of url for req, constraint and graph files

### LINKED ISSUE #501: allow use of urls to read information from file
Allow the use of urls for things like requirements file, constraints file and graph file when passing them as an argument to fromager

shubhbapna:
fixes #501 

users can now pass a url instead of a path to a local file to commands that need req files, constraints file or graph files. pip allows using url for req and constraints file so this is a pip compatible change

### DIFF
--- a/src/fromager/context.py
+++ b/src/fromager/context.py
@@ -7,11 +7,7 @@ from packaging.requirements import Requirement
 from packaging.utils import NormalizedName, canonicalize_name
 from packaging.version import Version
 
-from . import (
-    constraints,
-    dependency_graph,
-    packagesettings,
-)
+from . import constraints, dependency_graph, packagesettings, request_session
 
 logger = logging.getLogger(__name__)
 
@@ -24,7 +20,7 @@ class WorkContext:
     def __init__(
         self,
         active_settings: packagesettings.Settings | None,
-        constraints_file: pathlib.Path | None,
+        constraints_file: str | None,
         patches_dir: pathlib.Path,
         sdists_repo: pathlib.Path,
         wheels_repo: pathlib.Path,
@@ -44,12 +40,12 @@ class WorkContext:
                 max_jobs=max_jobs,
             )
         self.settings = active_settings
-        self.input_constraints_file: pathlib.Path | None
+        self.input_constraints_uri: str | None
         if constraints_file is not None:
-            self.input_constraints_file = constraints_file.absolute()
+            self.input_constraints_uri = constraints_file
             self.constraints = constraints.load(constraints_file)
         else:
-            self.input_constraints_file = None
+            self.input_constraints_uri = None
             self.constraints = constraints.Constraints({})
         self.sdists_repo = pathlib.Path(sdists_repo).absolute()
         self.sdists_downloads = self.sdists_repo / "downloads"
@@ -81,9 +77,19 @@ class WorkContext:
 
     @property
     def pip_constraint_args(self) -> list[str]:
-        if not self.input_constraints_file:
+        if not self.input_constraints_uri:
             return []
-        return ["--constraint", os.fspath(self.input_constraints_file)]
+
+        if self.input_constraints_uri.startswith(("https://", "http://", "file://")):
+            path_to_constraints_file = self.work_dir / "input-constraints.txt"
+            if not path_to_constraints_file.exists():
+                response = request_session.session.get(self.input_constraints_uri)
+                path_to_constraints_file.write_text(response.text)
+        else:
+            path_to_constraints_file = pathlib.Path(self.input_constraints_uri)
+
+        path_to_constraints_file = path_to_constraints_file.absolute()
+        return ["--constraint", os.fspath(path_to_constraints_file)]
 
     def write_to_graph_to_file(self):
         with open(self.work_dir / "graph.json", "w") as f:
--- a/src/fromager/read.py
+++ b/src/fromager/read.py
@@ -1,0 +1,26 @@
+import io
+import pathlib
+import typing
+from contextlib import contextmanager
+from urllib.parse import urlparse
+
+from .request_session import session
+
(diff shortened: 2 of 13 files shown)


src/fromager/context.py
```diff
@@ -84,7 +91,7 @@ def pip_wheel_server_args(self) -> list[str]:
     def pip_constraint_args(self) -> list[str]:
         if not self.input_constraints_file:
             return []
-        return ["--constraint", os.fspath(self.input_constraints_file)]
+        return ["--constraint", self.input_constraints_file]
```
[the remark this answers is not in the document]
tiran: What Doug says

Also please rename the attribute to `input_constraints_uri` to reflect that fact that it can be an URI.

src/fromager/read.py
```diff
@@ -0,0 +1,22 @@
...
+
+from .request_session import session
+
+
+@contextmanager
+def read_file_or_url(
+    path_or_url: str | pathlib.Path,
+) -> typing.Generator[io.TextIOWrapper, typing.Any, None]:
+    location = str(path_or_url)
+    if location.startswith("http://") or location.startswith("https://"):
```
tiran, line 14: Let's also support `file://`. 
```suggestion
    if location.startswith(("https://", "http://", "file://")):
```

Several `str` methods support `str | tuple[str, ...]`.

```
Help on method_descriptor:

startswith(...) unbound builtins.str method
    S.startswith(prefix[, start[, end]]) -> bool

    Return True if S starts with the specified prefix, False otherwise.
    With optional start, test S beginning at that position.
    With optional end, stop comparing S at that position.
    prefix can also be a tuple of strings to try.
```

src/fromager/context.py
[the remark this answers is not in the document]
shubhbapna: Ah yes that make sense, I will download the file and save it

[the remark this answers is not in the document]
shubhbapna: Done

src/fromager/context.py
```diff
@@ -81,9 +77,19 @@ def pip_wheel_server_args(self) -> list[str]:
 
     @property
     def pip_constraint_args(self) -> list[str]:
-        if not self.input_constraints_file:
+        if not self.input_constraints_uri:
             return []
-        return ["--constraint", os.fspath(self.input_constraints_file)]
+
+        if self.input_constraints_uri.startswith(("https://", "http://", "file://")):
```
dhellmann, line 83: This is going to cause `file://` URLs to be copied into the local temporary file. Those could be handled like regular files if we parsed the URL and extracted the path.

OTOH, if requests can already deal with that type of URL, then maybe it's OK.

src/fromager/read.py
```diff
@@ -0,0 +1,22 @@
...
+from .request_session import session
+
+
+@contextmanager
+def open_file_or_url(
+    path_or_url: str | pathlib.Path,
+) -> typing.Generator[io.TextIOWrapper, typing.Any, None]:
+    location = str(path_or_url)
+    if location.startswith(("https://", "http://", "file://")):
+        response = session.get(location)
```
dhellmann, line 15: session.get() and open() will produce different types of exceptions, so every caller using this function would have to handle both exception types. That leaks details of the implementation outside of the function.

We can keep it like this in this MR, but it would be better to trap the exceptions and produce a common one that is advertised as the API for this function.

src/fromager/context.py
shubhbapna: Actually it looks like requests can't handle `file://` url. If that is the case then should we even allow such URLs? @tiran 

```
>>> requests.get("file:///path/to/file")
  File "/home/projects/fromager/venv/lib/python3.11/site-packages/requests/api.py", line 73, in get
    return request("get", url, params=params, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/projects/fromager/venv/lib/python3.11/site-packages/requests/api.py", line 59, in request
    return session.request(method=method, url=url, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/projects/fromager/venv/lib/python3.11/site-packages/requests/sessions.py", line 589, in request
    resp = self.send(prep, **send_kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/projects/fromager/venv/lib/python3.11/site-packages/requests/sessions.py", line 697, in send
    adapter = self.get_adapter(url=request.url)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/projects/fromager/venv/lib/python3.11/site-packages/requests/sessions.py", line 792, in get_adapter
    raise InvalidSchema(f"No connection adapters were found for {url!r}")
requests.exceptions.InvalidSchema: No connection adapters were found for 'file:///path/to/file'
```

src/fromager/read.py
shubhbapna: Ah that makes sense

shubhbapna: Actually won't it help during debugging if the exceptions were different?

src/fromager/context.py
shubhbapna: It wasn't too bad to implement so lets keep it now :shrug:

src/fromager/read.py
dhellmann: If you do `raise IOError('local message') from err` then the caller only has to catch `IOError`, and the traceback shows all of the details of the original exception.

src/fromager/context.py
dhellmann: OK, what you have looks fine.

dhellmann (approved):
We can deal with the exception handling in another PR.

PR #516, opened 2024-11-25
add `--cache-wheel-server-url` and `--force` options to build-sequence

### LINKED ISSUE #505: add cache wheel server url argument to build-sequence
Just like in bootstrap, add the `--cache-wheel-server-url` argument to `build-sequence` so that it checks if the wheel exists against that url

shubhbapna:
fixes #505 

build-sequence will behave like make now where by default it doesn't build wheels if it has already been built. We can change this by passing the `--force` flag which will cause fromager to rebuild all wheels.
To check if a wheel has already been built, fromager looks only in the local pypi server it spins up. we can check an additional external server by passing the `--cache-wheel-server-url` option.

### DIFF
--- a/e2e/test_build_order.sh
+++ b/e2e/test_build_order.sh
@@ -21,7 +21,7 @@ fromager \
 # Save the build order file but remove everything else.
 cp "$OUTDIR/work-dir/build-order.json" "$OUTDIR/"
 
-# Rebuild everything
+# Rebuild everything even if it already exists
 log="$OUTDIR/build-logs/${DIST}-build.log"
 fromager \
     --log-file "$log" \
@@ -29,7 +29,7 @@ fromager \
     --sdists-repo "$OUTDIR/sdists-repo" \
     --wheels-repo "$OUTDIR/wheels-repo" \
     --settings-dir="$SCRIPTDIR/changelog_settings" \
-    build-sequence "$OUTDIR/build-order.json"
+    build-sequence --force "$OUTDIR/build-order.json"
 
 find "$OUTDIR/wheels-repo/"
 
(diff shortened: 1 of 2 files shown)


shubhbapna:
I am not sure what would be the best way to implement the cache wheel server url argument. I didn't think it was best UX to have 2 options to actually enable skip existing behavior `--skip-existing` and `--cache-wheel-server-url`. 

I went with the option of using only `--skip-existing` and passing a url instead of a boolean to implement this. The user can pass "localhost" if they want to use the wheel server that fromager spins up (this is consistent with the existing behavior)

thoughts @dhellmann ?

dhellmann:
> I am not sure what would be the best way to implement the cache wheel server url argument. I didn't think it was best UX to have 2 options to actually enable skip existing behavior `--skip-existing` and `--cache-wheel-server-url`.
> 
> I went with the option of using only `--skip-existing` and passing a url instead of a boolean to implement this. The user can pass "localhost" if they want to use the wheel server that fromager spins up (this is consistent with the existing behavior)
> 
> thoughts @dhellmann ?

I'm not sure why the 2 options are so tightly coupled. What's your reasoning there?

What if "skip existing" was the default behavior, what would the options for forcing a build look like?

pnasrat:
From a user of fromager perspective who has been using skip-existing with build-sequence and build from the re-naming of `cache-wheel-server-url` is a little confusing as it is not necessarily a cache. 

Previously eg in 0.34.0 and earlier I believe `wheel_server_url: https://internal.pypi.example/simple` would both disable the internal wheel server which allows for running parallelizable parts of the build graph (eg in separate containers), and be used for skip exising and thus also allow running multiple rebuilds simultaneously to also minimize work if they have a matching specification in the internal rebuilt repo.

shubhbapna:
As discussed offline, we will make fromager behave like make where it doesn't build a wheel if it already exists. We will add `--force` flag to force rebuild everything. We will also change the name of the option to `--extra-index-url` (just like pip) where this url is in addition to the local pypi server fromager spins up

tiran:
Will `--extra-index-url` behave exactly like in `pip`? In pip, it adds additional URLs to the resolver. The resolver merges packages from the primary and all extra indexs.

shubhbapna:
> Will `--extra-index-url` behave exactly like in `pip`? In pip, it adds additional URLs to the resolver. The resolver merges packages from the primary and all extra indexs.

Oh it won't. Should we rename it this to avoid confusion?

dhellmann (approved):

Issue #483, opened 2024-10-22
remove unused arguments from default functions


shubhbapna:
we had been keeping all arguments in default functions to enable backward compatibility for plugins. However we have `invoke` method now that ensures that plugin functions are called with the right parameters so we can remove these arguments from the default functions.

Having unused arguments adds maintenance load and will make confusing for new contributors

dhellmann:
I think we don't want to do this, because we use those functions to build the documentation for the plugin API.

rd4398:
> I think we don't want to do this, because we use those functions to build the documentation for the plugin API.

Agreed. I am closing the linked PR and the issue

PR #517, opened 2024-11-26
Cleanup: remove unused arguments from default functions

rd4398:
Fixes: https://github.com/python-wheel-build/fromager/issues/483

### DIFF
--- a/src/fromager/dependencies.py
+++ b/src/fromager/dependencies.py
@@ -75,9 +75,6 @@ def _filter_requirements(
 
 
 def default_get_build_system_dependencies(
-    ctx: context.WorkContext,
-    req: Requirement,
-    sdist_root_dir: pathlib.Path,
     build_dir: pathlib.Path,
 ) -> typing.Iterable[str]:
     """Get build system requirements
@@ -127,9 +124,6 @@ def get_build_backend_dependencies(
 
 
 def default_get_build_backend_dependencies(
-    ctx: context.WorkContext,
-    req: Requirement,
-    sdist_root_dir: pathlib.Path,
     build_dir: pathlib.Path,
     extra_environ: dict[str, str],
 ) -> typing.Iterable[str]:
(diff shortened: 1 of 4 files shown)


shubhbapna:
I just remembered that it will break our documentation since i think it depends on the default functions to generate the docs for the plugin. Is there a workaround for that?

shubhbapna (changes requested):
I just remembered that it will break our documentation since i think it depends on the default functions to generate the docs for the plugin. Is there a workaround for that?

tiran:
We not only use the arguments in our docs. They are also used in downstream to compare signatures between default hooks and the plugin hooks.

IMO there is nothing to fix here. It is totally okay to pass additional args to a function if that's part of the API.

shubhbapna:
> They are also used in downstream to compare signatures between default hooks and the plugin hooks.

I don't think we should enforce that anymore or at the very least change the logic of how we are enforcing that. In fromager we are using a special [invoke function](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/overrides.py#L62) to ensure that a function only receives the arguments it needs. 

So downstream we will need to ensure that the arguments we are passing in our downstream overrides are a subset of all the possible arguments that can be passed. We don't need strict check of signatures.

 
> It is totally okay to pass additional args to a function if that's part of the API.

IMO it adds technical debt and can make it confusing for initial contributors

tiran:
In my humble opinion, our special invoke function is not a good design. It's not how typical Python APIs are designed. In a typical Python API, it is the responsibility of the callee to deal with unnecessary arguments, e.g. with `**kwargs`.

dhellmann:
> In my humble opinion, our special invoke function is not a good design. It's not how typical Python APIs are designed. In a typical Python API, it is the responsibility of the callee to deal with unnecessary arguments, e.g. with `**kwargs`.

Yeah, it was a hack to try to avoid introducing breaking changes into the downstream repo when we bumped fromager. It's not an ideal solution, and we should fix that downstream code to use `**kwds`. 

How will we record the formal API of each plugin and then document it if we remove unused arguments from the default functions? Let's figure that out, and then see what that tells us about next steps.

PR #521, opened 2024-12-06
fix cycles in when traversing graph

dhellmann:
due to cycles graph traversal would cause maximum recursive depth reached error


dhellmann:
We want a 0.34 branch

Issue #524, opened 2024-12-06
Bug: Duplicates are written to constraints file during bootstrapping


rd4398:
During bootstrapping, duplicates are written to constraints file. Below is the subset of output for `graph explain-duplicates` command which uses recently produced `graph.json`:

```
tokenizers
  0.20.4
    tokenizers<0.21,>=0.20 matches ['0.20.4']
      transformers==4.46.3
  0.21.0
    tokenizers>=0.11.1 matches ['0.21.0', '0.20.4']
      instructlab==0.21.0
    tokenizers<0.22,>=0.21 matches ['0.21.0']
      transformers==4.47.0
    tokenizers>=0.19.1 matches ['0.21.0', '0.20.4']
      vllm==0.6.2
  * No single version of tokenizers meets all requirements
 
transformers
  4.46.3
    transformers<4.47.0 matches ['4.46.3']
      trl==0.12.2
  4.47.0
    transformers>=4.41.2 matches ['4.47.0', '4.46.3']
      instructlab==0.21.0
      instructlab-sdg==0.6.1
    transformers matches ['4.47.0', '4.46.3']
      peft==0.14.0
      instructlab-eval==0.4.1
    transformers>=4.45.2 matches ['4.47.0', '4.46.3']
      instructlab-training==0.6.1
    transformers>=4.40.0 matches ['4.47.0', '4.46.3']
      instructlab-dolomite==0.2.0
    transformers>=4.1 matches ['4.47.0', '4.46.3']
      lm-eval==0.4.5
    transformers>=4.45.0 matches ['4.47.0', '4.46.3']
      vllm==0.6.2
  * transformers==4.46.3 usable by all consumers
```

dhellmann:
The issue here is that fromager thinks it cannot pick a single version of tokenizers, but it _could_ do that if it recognized that only one version of transformers is needed and it used that information to pick the version of tokenizers.

PR #525, opened 2024-12-10
Bug Fix: Duplicates are written to constraints file during bootstrapping

rd4398:
Fixes : https://github.com/python-wheel-build/fromager/issues/524

### DIFF
--- a/src/fromager/commands/bootstrap.py
+++ b/src/fromager/commands/bootstrap.py
@@ -3,6 +3,8 @@ import typing
 
 import click
 from packaging.requirements import Requirement
+from packaging.utils import NormalizedName
+from packaging.version import Version
 
 from .. import (
     bootstrapper,
@@ -180,7 +182,17 @@ def write_constraints_file(
     conflicts = graph.get_install_dependency_versions()
     ret = True
     conflicting_deps = set()
-    for dep_name, nodes in sorted(conflicts.items()):
+
+    # Map for already resolved versions for a given dependency Eg: {"a": "0.4"}
+    resolved: dict[NormalizedName, Version] = {}
+
+    # List of unresolved dependencies
+    unresolved_dependencies = sorted(conflicts.items())
+
+    dep_name: NormalizedName
+
+    # Loop over dependencies and resolve dependencies with single version first. This will shrink the unresolved_dependencies to begin with.
+    for dep_name, nodes in unresolved_dependencies[:]:
         versions = [node.version for node in nodes]
         if len(versions) == 0:
             # This should never happen.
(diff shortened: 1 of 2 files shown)


src/fromager/commands/bootstrap.py
```diff
@@ -189,64 +197,83 @@ def write_constraints_file(
...
-        for node in nodes:
-            parent_edges = node.get_incoming_install_edges()
-            user_counter += len(parent_edges)
-            for parent_edge in parent_edges:
-                for matching_version in parent_edge.req.specifier.filter(versions):
-                    usable_versions.setdefault(str(matching_version), []).append(
-                        str(parent_edge.destination_node.version)
+            resolved[dep_name] = versions[0]
+            # Remove from unresolved dependencies list
+            unresolved_dependencies.remove((dep_name, nodes))
```
dhellmann, line 202: This will cause an error, because you're looping over the list being modified on this line. One way to fix that would be to make a copy of the list on line 191.

src/fromager/commands/bootstrap.py
```diff
@@ -180,7 +180,15 @@ def write_constraints_file(
...
-    for dep_name, nodes in sorted(conflicts.items()):
+
+    # Map for already resolved versions for a given dependency Eg: {"a": "0.4"}
+    resolved: dict[str, str] = {}
+
+    # List of unresolved dependencies
+    unresolved_dependencies = sorted(conflicts.items())
+
+    # Resolve dependencies with single version first. This will shrink the unresolved_dependencies to begin with.
+    for dep_name, nodes in unresolved_dependencies:
```
dhellmann, line 191:
```suggestion
    for dep_name, nodes in unresolved_dependencies[:]:
```

See comment on line 202

rd4398:
- I am getting into an infinite loop while trying to set `resolved_something = True`. Check line # 256
- Also, need help to figure out part to use existing package versions to pick the "right" version of item for multiple version case

src/fromager/commands/bootstrap.py
```diff
@@ -189,64 +197,83 @@ def write_constraints_file(
...
+            resolved[dep_name] = versions[0]
+            # Remove from unresolved dependencies list
+            unresolved_dependencies.remove((dep_name, nodes))
+
+    # Below this point we have built multiple versions of the same thing, so
+    # we need to try to determine if any one of those versions meets all of
+    # the requirements.
+
+    # Track which versions can be used by which parent requirement.
+    usable_versions: dict[str, list[str]] = {}
```
dhellmann, line 209: These variables used to be initialized inside the loop. I think they need to stay there, so they are cleared out each iteration.

src/fromager/commands/bootstrap.py
```diff
@@ -189,64 +197,83 @@ def write_constraints_file(
...
+        # Make copy of the original list
+        items_to_resolve = unresolved_dependencies[:]
+        for dep_name, nodes in items_to_resolve:
+            # Which parent requirements can use which versions of the dependency we
+            # are working on?
+            for node in nodes:
+                parent_edges = node.get_incoming_install_edges()
+                for parent_edge in parent_edges:
+                    parent_name = parent_edge.destination_node.canonicalized_name
+                    if (
```
dhellmann, line 229: This condition looks right, but could use a comment.

src/fromager/commands/bootstrap.py
```diff
@@ -189,64 +197,83 @@ def write_constraints_file(
...
+                    version_strs = [str(v) for v in sorted(versions)]
+                    logger.debug(
+                        "%s: selecting %s from multiple candidates %s",
+                        dep_name,
+                        v,
+                        version_strs,
                     )
+                    # Set the flag to true since we were able to resolve something
+                    resolved[dep_name] = v
+            # resolved_something = True                  # This causes the infinite loop. Trying this inside if after line 255 also causes infinite loop
```
dhellmann, line 256: This should definitely be indented inside the if like line 255. Maybe some of the other changes I mention above about initializing the variables will fix the loop?

src/fromager/commands/bootstrap.py
```diff
@@ -189,64 +197,83 @@ def write_constraints_file(
...
+            items_to_resolve.remove((dep_name, nodes))
```
dhellmann, line 257: This should also be inside the conditional. We only want to remove it from the items to resolve if it is resolved, right?

src/fromager/commands/bootstrap.py
```diff
@@ -189,64 +201,102 @@ def write_constraints_file(
...
+    # Flag to see if something is resolved
+    resolved_something = True
+
+    # Outer while loop to resolve remaining dependencies with multiple versions
+    while unresolved_dependencies and resolved_something:
+        resolved_something = False
+        # Make copy of the original list
+        items_to_resolve = unresolved_dependencies[:]
+        # Loop over items_to_resolve. It is a tuple with dep name and a list of edges
+        for dep_name, nodes in items_to_resolve:
```
dhellmann, line 222: If we don't need `items_to_resolve` for any other reason, making the copy of `unresolved_dependencies` inline here, like on line 195, would be a little cleaner.

src/fromager/commands/bootstrap.py
```diff
@@ -189,64 +201,102 @@ def write_constraints_file(
...
+            for v, users in usable_versions.items():
+                if len(users) == user_counter:
+                    version_strs = [str(v) for v in sorted(versions)]
+                    logger.debug(
+                        "%s: selecting %s from multiple candidates %s",
+                        dep_name,
+                        v,
+                        version_strs,
                     )
+                    # Set the flag to true since we were able to resolve something
```
dhellmann, line 268: This comment is 2 lines away from the actual code. I think we can just delete this line to clean it up.

dhellmann (commented):
A couple of small comments, but this is looking good.

src/fromager/commands/bootstrap.py
rd4398: Okay, changed in latest code push

src/fromager/commands/bootstrap.py
rd4398: okay, done

dhellmann (approved):

shubhbapna:
:partying_face:

PR #526, opened 2024-12-12
Add build-sequence-summary.json

pnasrat:
Machine readable version of the current markdown table

Useful in pipeline environments to filter skipped or prebuilt packages from a subsequent twine publish

Signed-off-by: Pris Nasrat <pris.nasrat@chainguard.dev>

### DIFF
--- a/src/fromager/commands/build.py
+++ b/src/fromager/commands/build.py
@@ -44,6 +44,12 @@ class BuildSequenceEntry:
     wheel_filename: pathlib.Path
     skipped: bool = False
 
+    @staticmethod
+    def dict_factory(x):
+        return {
+            k: str(v) if isinstance(v, pathlib.Path | Version) else v for (k, v) in x
+        }
+
     def __lt__(self, other):
         if not isinstance(other, BuildSequenceEntry):
             return NotImplemented
@@ -277,6 +283,15 @@ def _summary(ctx: context.WorkContext, entries: list[BuildSequenceEntry]) -> Non
     with open(ctx.work_dir / "build-sequence-summary.md", "w", encoding="utf-8") as f:
         console = rich.console.Console(file=f, width=sys.maxsize)
         console.print(*output, sep="\n\n")
+
+    with open(ctx.work_dir / "build-sequence-summary.json", "w", encoding="utf-8") as f:
+        json.dump(
+            [
+                dataclasses.asdict(e, dict_factory=BuildSequenceEntry.dict_factory)
+                for e in entries
+            ],
+            f,
+        )
 
 
 def _create_table(entries: list[BuildSequenceEntry], **table_kwargs) -> Table:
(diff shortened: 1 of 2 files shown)


src/fromager/commands/build.py
```diff
@@ -44,6 +44,12 @@ class BuildSequenceEntry:
     wheel_filename: pathlib.Path
     skipped: bool = False
 
+    @staticmethod
+    def dict_factory(x):
+        return {
+            k: str(v) if isinstance(v, pathlib.Path | Version) else v for (k, v) in x
```
dhellmann, line 50: I would have expected the 2nd argument to be a tuple, but TIL about union type. Does that come from type hints?

src/fromager/commands/build.py
```diff
@@ -278,6 +284,15 @@ def _summary(ctx: context.WorkContext, entries: list[BuildSequenceEntry]) -> Non
         console = rich.console.Console(file=f, width=sys.maxsize)
         console.print(*output, sep="\n\n")
 
+    with open(ctx.work_dir / "build-sequence-summary.json", "w", encoding="utf-8") as f:
+        json.dump(
+            [
+                dataclasses.asdict(e, dict_factory=BuildSequenceEntry.dict_factory)
```
dhellmann, line 290: I _think_ this could just be `dataclasses.asdict(e)`. Did you find you needed to pass the factory function explicitly or does that do something extra?

pnasrat: So both Version and pathlib.Path failed to serialize with errors like

```
ERROR: Object of type PosixPath is not JSON serializable     
```

Using the dict_factory here to use __str__ for those worked. I debated switching to pydantic here, used else but this seemed sufficient given it's a report.

src/fromager/commands/build.py
pnasrat: That was my first approach - then my editor/pyright complained with the recommendation to use union type not  tuple here. Looks like came in via https://peps.python.org/pep-0604/#isinstance-and-issubclass

src/fromager/commands/build.py
dhellmann: What you have looks good. Pydantic wouldn't be wrong, but is probably an extra complication here.

dhellmann (approved):

src/fromager/commands/build.py
dhellmann: Nice, thanks for sharing that link! I'm still learning about some of the recent changes to the language after spending a few years away from Python, so every little bit helps. :-)

dhellmann:
I've triggered a new release including this feature. Thank you for contributing!

https://github.com/python-wheel-build/fromager/releases/tag/0.36.0

src/fromager/commands/build.py
tiran: The feature was introduced in Python 3.10 as part of PEP 604, https://docs.python.org/3.10/whatsnew/3.10.html#pep-604-new-type-union-operator

Issue #529, opened 2025-01-09
fromager: vendor_rust behavior broken in workspaces with example code (eg cryptography)


pnasrat:
## Summary

vendoring breaks with error for docs file `cryptography-44.0.0/cryptography-44.0.0/docs/development/custom-vectors/aes-192-gcm-siv/verify-aes192gcmsiv/Cargo.toml`


```
failed with error: current package believes it's in a workspace when it's not:
```

I tried to add a patch that adds exclude to the workspace members but it looks as if the vendor rust behavior of fromager is not aware of workspaces

```
excludes = [
    "docs/development/",
]
```

## Analysis

Looking at the source fromager assumes any `Cargo.toml` in the source tree is included and does not honor workspace configuration as it just uses a glob. This seems brittle to projects using workspaces with example code in docs as cryptography have

https://github.com/python-wheel-build/fromager/blob/acc45d3d17d77faa82a19ab6ced5aee068986c23/src/fromager/vendor_rust.py#L126

## Steps to reproduce

Added a branch with the patch and an e2e test demonstrating here

https://github.com/pnasrat/fromager/tree/vendor-cryptography-fails

@dhellmann wondering what the right behavior to do here, I wonder if rust vendoring should be configurable at the top level of fromager maybe via an env variable as different rebuilders may want different behavior from fromager here. 

## Logs

<details>
<summary>fromager log</summary>
<br/>

```console
cryptography: preparing source for cryptography==44.0.0 from /home/pnasrat/src/fromager/e2e-output/sdists-repo/downloads/cryptography-44.0.0.tar.gz
cryptography: updating vendored rust dependencies in /home/pnasrat/src/fromager/e2e-output/work-dir/cryptography-44.0.0/cryptography-44.0.0
[...]

PR #532, opened 2025-01-10
Fix vendor Rust: limit to manifests from backend

tiran:
Fromager was vendoring crates from all `Cargo.toml` files in a project. This approach is causing issues for projects that have cargo files in tests and example directories.

The `vendor_rust()` function now only vendors crates from `Cargo.toml` in the project's root directory and additional cargo files listed in `tools.maturin` or `tools.setuptools-rust` entries.

Fixes: #529

### DIFF
--- a/src/fromager/vendor_rust.py
+++ b/src/fromager/vendor_rust.py
@@ -1,5 +1,6 @@
 """Vendor Rust crates into an sdist"""
 
+import enum
 import json
 import logging
 import os
@@ -22,8 +23,10 @@ CARGO_CONFIG = {
     },
 }
 
-# vendor when project depends on a Rust build system
-RUST_BUILD_REQUIRES = frozenset({"setuptools-rust", "maturin"})
+
+class RustBuildSystem(enum.StrEnum):
+    maturin = "maturin"
+    setuptools_rust = "setuptools-rust"
 
 
 def _cargo_vendor(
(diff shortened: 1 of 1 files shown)


tiran:
I need to verify the changes with cryptography 42, 43, and 44.

pnasrat:
Thanks @tiran - I also added https://github.com/python-wheel-build/fromager/pull/533 to control vendoring via network isolation (which means if future rust vendoring issues come up there is a way that fromager users can disable vendoring before fixes)

pnasrat:
Looks like a ci failure 

```
  0%|                                                  | 0/1 [00:00<?, ?pkg/s]maturin: new toplevel dependency maturin==1.6.0 resolves to 1.6.0
maturin: checking if wheel was already uploaded to http://localhost:38821/simple/
maturin: did not find wheel for 1.6.0 in http://localhost:38821/simple/
saved /home/runner/work/fromager/fromager/e2e-output/sdists-repo/downloads/maturin-1.6.0.tar.gz
maturin: preparing source for maturin==1.6.0 from /home/runner/work/fromager/fromager/e2e-output/sdists-repo/downloads/maturin-1.6.0.tar.gz

  0%|                                                  | 0/1 [00:00<?, ?pkg/s]
ERROR: 'Key "setuptools-rust" does not exist.'
```

tiran:
I have successfully built `cryptography==44.0.0` with network isolation + vendored Rust packages with this PR.

src/fromager/vendor_rust.py
```diff
@@ -86,28 +86,27 @@ def _cargo_config(project_dir: pathlib.Path) -> None:
         tomlkit.dump(cfg, f)
 
 
-def _should_vendor_rust(req: Requirement, project_dir: pathlib.Path) -> bool:
+def _detect_rust_build_backend(
+    req: Requirement, pyproject_toml: dict[str, typing.Any]
+) -> str | None:
```
dhellmann, line 91: The docstring could use an update to explain the return value.

src/fromager/vendor_rust.py
```diff
@@ -119,11 +118,42 @@ def vendor_rust(
...
+
+    manifests: list[pathlib.Path] = []
+    # check for Cargo.toml in project root
+    root_cargo_toml = project_dir / "Cargo.toml"
+    if root_cargo_toml.is_file():
+        manifests.append(root_cargo_toml)
+
+    # maturin and setuptools-rust can have Cargo.toml in other directories.
+    # tool.setuptools-rust and tool.maturin are optional.
+    if backend == "maturin":
```
dhellmann, line 138: This if/elif block could use a final `else` to report that an unknown backend is being used so no additional cargo files are being scanned. I've found a few places in our code where we didn't emit a message like that and ended up with confusing behavior.

dhellmann (commented):
This looks OK to me, with a couple of small nits.

@pnasrat , have you had a chance to try it out to see if it resolves your issue, too?

src/fromager/vendor_rust.py
tiran: I have updated the doc string and changed the return values to `enum | str`.

src/fromager/vendor_rust.py
tiran: The if/elif block is now a match block that handles all cases (maturin, setuptools-rust, None, unreachable). I'm also printing a warning when we detect a Rust build backend but no Cargo.toml.

dhellmann:
> it'd be great to get this merged - its been working for me on the PR checked out

Excellent, I was just waiting to hear back from you and then it fell off of the top of my todo list.

dhellmann (approved):

dhellmann:
@mergifyio refresh

dhellmann:
Released in 0.38.0

PR #533, opened 2025-01-10
Only invoke vendor_rust if network isolation set

pnasrat:
See: #529

While improving the vendor logic for more complex source cases this fixes building cryptography if network isolation turned off

Signed-off-by: Pris Nasrat <pris.nasrat@chainguard.dev>

### DIFF
--- a/src/fromager/sources.py
+++ b/src/fromager/sources.py
@@ -412,7 +412,8 @@ def prepare_new_source(
         req=req,
         sdist_root_dir=source_root_dir,
     )
-    vendor_rust.vendor_rust(req, source_root_dir)
+    if ctx.network_isolation:
+        vendor_rust.vendor_rust(req, source_root_dir)
 
 
 @metrics.timeit(description="build sdist")


tiran (changes requested):
This should be a separate option. I don't like the idea that the network isolation flags affects how sdists and packages are built. It should be a separate option.

Network isolation is not available on most platforms except some Linux systems. Even on Linux, network isolation is only possible in some cases. For example, it doesn't work in user containers on Ubuntu unless root disables a policy.

dhellmann:
> This should be a separate option. I don't like the idea that the network isolation flags affects how sdists and packages are built. It should be a separate option.
> 
> Network isolation is not available on most platforms except some Linux systems. Even on Linux, network isolation is only possible in some cases. For example, it doesn't work in user containers on Ubuntu unless root disables a policy.

I hadn't considered that, but it's a good point. Maybe the default behavior depends on whether network isolation is enabled? Or are they completely unrelated?

tiran:
> I hadn't considered that, but it's a good point. Maybe the default behavior depends on whether network isolation is enabled? Or are they completely unrelated?

*If* we decide to make vendoring optional, then it should be a separate option. Hermetic builds and vendoring are not tightly coupled. All possible combination of network isolation flag and vendor flag have a use case

1. build without network isolation and without vendoring
2. vendor Rust without network isolation
3. build with network isolation and without vendored Rust sources
4. vendor + isolate

(1) and (3) are currently not possible because we don't have an option to disable vendoring. Several Linux distros like Fedora follow approach (3). Fedora installs Rust crates from RPMs and configures cargo to use local, pre-installed packages instead of crates.io.

tiran:
I'm closing this PR. We made Rust vendoring more robust in MR #532. We haven't seen any build problems with the new code.

PR #534, opened 2025-01-20
fix some issues writing the constraints file during bootstrap

dhellmann:
We're seeing some cases downstream where fromager
fails with an exception because it tries to remove
something from the unresolved dependency list when
the item isn't in the list. This seems to be
caused by having multiple candidate versions for a
package where both versions could be used by all
consumers.

This change resolves the issue by stopping the
resolution loop when we find a version of a
package that can work for all consumers. The loop
now iterates from the newest version to the
oldest, to ensure we choose the newest package
that works.

There is also some error handling in case this fix
is not sufficient to eliminate the original
ValueError.

I also noticed that a log message that we supposed
to print debug info about the candidate versions
was using the wrong list variable so it was always
printing the same values, regardless of the
package. I cleaned that up and added additional
debug logging.

### DIFF
--- a/tests/test_bootstrap.py
+++ b/tests/test_bootstrap.py
@@ -261,3 +261,62 @@ def test_write_constraints_file_duplicates():
         d==1.0
         """).lstrip()
     assert expected == buffer.getvalue()
+
+
+def test_write_constraints_file_multiples():
+    buffer = io.StringIO()
+    raw_graph = {
+        "": {
+            "download_url": "",
+            "pre_built": False,
+            "version": "0",
+            "canonicalized_name": "",
+            "edges": [
+                {
+                    "key": "a==2.7.0",
+                    "req_type": "toplevel",
+                    "req": "a==2.7.0",
+                },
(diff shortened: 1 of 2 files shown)


tests/test_bootstrap.py
```diff
@@ -261,3 +261,62 @@ def test_write_constraints_file_duplicates():
...
+    buffer = io.StringIO()
+    raw_graph = {
+        "": {
+            "download_url": "",
+            "pre_built": False,
+            "version": "0",
+            "canonicalized_name": "",
+            "edges": [
+                {
+                    "key": "docling==2.7.0",
```
rd4398, line 276: I am not sure whether we should use actual names of packages that we build and instead switch to generic names like a, b, foo, bar, etc.

Apart from that, this looks good!

dhellmann: That's a good point, I should be consistent with the other tests.

rd4398 (approved):
This looks good!

Issue #535, opened 2025-01-21
Add `fromager.hooks` for pre-built wheels


tiran:
Fromager has a `post_build` hook that gets called when a sdist and wheel have been built. In downstream we use the hook to upload sdist and wheel to an internal package index. Currently, Fromager does not have a hook to deal with pre-built wheels. There is no way to automatically upload pre-built wheels to an internal package index.

Hook name: `pre_built`, `pre_built_wheel`, or similar
Signature: similar to `post_build` hook, but without `sdist_filename` argument

```
def pre_built_wheel(
    ctx: context.WorkContext,
    req: Requirement,
    dist_name: str,
    dist_version: str,
    wheel_filename: pathlib.Path,
) -> None:
    pass
```

dhellmann:
As part of this work, we may want to revisit the decision to remove pre-built wheels after bootstrapping. I think we did that because we assumed we would be downloading those wheels from the same index in later steps, so there was no point in re-uploading them. I'm not sure that's accurate any more. https://github.com/python-wheel-build/fromager/blob/main/src/fromager/commands/bootstrap.py#L154-L163

PR #536, opened 2025-01-22
Implement prebuilt-wheel hook

tiran:
The `prebuilt_wheel` hook complements the `post_build` hook. While the post build hook runs after a wheel has been built, the prebuilt wheel hook runs after a wheel has been downloaded. The new hook makes it possible to upload prebuilt wheels to an internal Python index.

Fixes: #535


dhellmann (approved):
Nice, I like the test case.

PR #537, opened 2025-01-23
Refactor constraints module

tiran:
Refactor and modify the API of `fromager.constraints` module to prepare the code for multiple constraints files.

The `Constraints` class no longer takes an mapping of package name to `Requirement`. Instead it has a new method `add_constraints()` that takes one or more requirements strings. The method does not take `Requirements` object, because we might want to modify and merge constraints in the future.

The `load()` module-level function has been replaced by `Constraints.load_constraints_file()` method.

### DIFF
--- a/src/fromager/constraints.py
+++ b/src/fromager/constraints.py
@@ -3,7 +3,7 @@ import pathlib
 import typing
 
 from packaging.requirements import Requirement
-from packaging.utils import canonicalize_name
+from packaging.utils import NormalizedName, canonicalize_name
 from packaging.version import Version
 
 from . import requirements_file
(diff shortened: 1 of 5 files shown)


src/fromager/constraints.py
```diff
@@ -12,8 +12,36 @@
...
+        yield from self._data
+
+    def add_constraints(self, *reqs: str) -> None:
+        """Add new constraints, must not conflict with any existing constraints"""
+        for unparsed in reqs:
+            req = Requirement(unparsed)
+            canon_name = canonicalize_name(req.name)
+            previous = self._data.get(canon_name)
+            if previous:
+                # In the future, we could detect non-conflicting requirements
```
dhellmann, line 30: I'm not sure I would want to merge them. I could definitely see saying that the last (or first) constraint for a package "wins", and others are reported as warnings.

src/fromager/constraints.py
```diff
@@ -12,8 +12,36 @@
...
-        self._data = {canonicalize_name(n): v for n, v in data.items()}
+    def __init__(self) -> None:
+        # mapping of canonical names to requirements
+        # NOTE: Requirement.name is not normalized
+        self._data: dict[NormalizedName, Requirement] = {}
+
+    def __iter__(self) -> typing.Iterable[NormalizedName]:
+        yield from self._data
+
+    def add_constraints(self, *reqs: str) -> None:
```
dhellmann, line 23: What's the limit on the number of arguments that can be passed to a function? I wouldn't expect constraints files to be very long, but does this constrain them in a way that using an iterable as argument would not?

dhellmann (commented):
This looks good, just one question before I approve.

src/fromager/constraints.py
tiran: Makes sense, I have removed the comment.

src/fromager/constraints.py
tiran: Python does not have an argument limit. Arguments are only limited by amount of memory.

The method doesn't have to accept an arbitrary amount of arguments. I have changed it to accept a single requirement string. Caller can call the method in a loop.

Issue #539, opened 2025-01-24
Support multiple constraints file


tiran:
In version 0.36 and earlier, Fromager supports only a single constraints file. The option `--constraints-file` is single value. It would be useful to support multiple constraints files. In downstream, we have variant-specific constraints and product-specific constraints for all product variants. For example, we have to constrain some packages for all variants, because latest version of the package does not build for us. Support for multiple constraints files would allow us to move common constraints for all variants into a shared file.

## Implementation

Let's start with a simple implementation:

- all constraints file must exist and be readable
- all constraints are combined into a single constraints set
- a package can only be constraint once across all constraints files. The second occurrence of a package raises an error.

The implementation should cover majority of use cases. We can look into more complex ways to deal with multiple constraints if the need arises.

## Env vars and URLs

Fromager uses `click` to parse command line arguments. Click also supports env vars, e.g. `--constraints-file` can also be set with `FROMAGER_CONSTRAINTS_FILE` env var. Env vars are a common way to pass settings into containers. Click uses different rules to split [multiple values from env vars](https://click.palletsprojects.com/en/stable/options/#multiple-values-from-environment-values). Path-like types are split by `os.pathsep` (`:` like `PATH` vars) and other values are split by white space.

A while ago `--constraints-file` was switched from path-like to `str` type to support `https://` URLs for remote constraints files. We cannot switch `--constraints-file` to a `type=str, multiple=True` option. This would break cases like `FROMAGER_CONSTRAINTS_FILE="/home/user name/constraints.txt"`. Click would split the env var into two items `/home/user` and `name/constraints.txt`.

I have considered and experimented several approaches how to reconcile local paths, remote URLs, and env vars. My preferred approach is:

- change `--constraints-file` to multi-value option that accepts only local paths. `FROMAGER_CONSTRAINTS_FILE` splits on colon `':'`.
- introduce second option `--constraints-url` that accepts multiple remote URLs. `FROMAGER_CONSTRAINTS_URL` splits on white space `' '`.

dhellmann:
I think we've worked out downstream that we don't need fromager to support this. Let me know if you still consider it worth pursuing upstream.

PR #540, opened 2025-01-24
feat: Support multiple constraints files

tiran:
Fromager now supports multiple constraints files. The option `fromager --constraints-file` can be supplied multiple times. The option no longer accepts URLs. Remote constraints files are now specified with `--constraints-url`. The split was necessary to correctly parse environment variables with multiple entries and white spaces in local file names. Paths are split at `:`, URLs are split at ` ` (white space).

There can only be one constraint per package.

Internally, remote constraints from `https://` URLs are no longer retrieved twice. Instead all constraints are merged and the final set is dumped into a single constraints file.

Fixes: #539

### DIFF
--- a/src/fromager/__main__.py
+++ b/src/fromager/__main__.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import logging
+import os
 import pathlib
 import sys
 
@@ -107,8 +108,20 @@ else:
 @click.option(
     "-c",
     "--constraints-file",
+    # for click.Path, click splits env vars on os.pathsep (':')
+    type=clickext.ClickPath(exists=True, file_okay=True, dir_okay=False),
+    multiple=True,
+    default=[],
+    help="location of constraints files (multiple use)",
+)
+@click.option(
+    "--constraints-url",
+    # for str, click splits env vars on space
     type=str,
-    help="location of the constraints file",
+    multiple=True,
+    default=[],
+    callback=clickext.verify_url_callback,
+    help="remote location of constraints file as URL (multiple use)",
 )
 @click.option(
     "--cleanup/--no-cleanup",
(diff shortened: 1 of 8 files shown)


src/fromager/__main__.py
```diff
@@ -195,13 +209,19 @@ def main(
...
+            )
+            logger.info(f"constraints URLs: {' '.join(constraints_url)}")
             logger.info(f"network isolation: {network_isolation}")
             overrides.log_overrides()
 
     if network_isolation and not SUPPORTS_NETWORK_ISOLATION:
         ctx.fail(f"network isolation is not available: {NETWORK_ISOLATION_ERROR}")
 
+    constraints: list[str] = list(constraints_url)
+    constraints.extend(os.fspath(p) for p in constraints_file)
```
dhellmann, line 223: This implies that constraints at URLs take precedence over local files. I don't think that ordering is wrong, but I can definitely imagine wanting a local file to have a value that replaces a value in some global file on a web server.

src/fromager/constraints.py
```diff
@@ -34,12 +37,18 @@ def add_constraint(self, unparsed: str) -> None:
...
         logger.info("loading constraints from %s", constraints_file)
         content = requirements_file.parse_requirements_file(constraints_file)
         for line in content:
             self.add_constraint(line)
 
+    def dump_constraints_file(self, constraints_file: pathlib.Path) -> None:
+        """Dump all constraints into a file"""
+        with constraints_file.open("w", encoding="utf-8") as f:
+            for req in self._data.values():
+                f.write(f"{req}\n")
```
dhellmann, line 50: Do we log the constraints as we parse the input? I know we log matching constraints as we process individual packages. It may also be useful to log the contents of this file to help with debugging.

dhellmann (approved):
The logic all looks good. I had a couple of comments, but nothing that has to hold this up.

dhellmann:
Do we still want this?

tiran:
No. we don't need this PR any more.

Issue #489, opened 2024-10-30
add a VersionMap type


dhellmann:
Downstream we have a few places where we define a dictionary mapping versions to build settings (source URLs, flags, etc.). We've repeated the pattern enough that it makes sense to add it to fromager as a utility class. 

The class should act like (subclass?) a `dict`. The keys should be Version objects. The values should be arbitrary types. 

The instance should have a method named something like `resolve()` that takes a Requirement and constraint and returns the value associated with the version that matches. It should raise an appropriate exception if no version in the map meets the requirement.

PR #541, opened 2025-01-28
resolver.py: Add CustomPackageProvider class

prarit:
JIRA: https://issues.redhat.com/browse/RHELAI-3189

The builder project contains code that attempts to satisfy constraints and package requirements.  This code is effectively replicating is_satsified_by() of the GenericProvider class and should be removed from builder in favor of a new class in fromager.

The new class has find_best_match(), which unlike find_matches() returns the best match in a provided list of versions.

Suggested-by: Doug Hellmann <dhellmann@redhat.com>


rd4398 (commented):
This looks good! Just a couple of suggestions:

- Can you add some test cases for `find_best_match()` function, if possible?
- I think (from past experience) we don't want to mention JIRA reference in description since this is upstream and we don't want to share about RHELAI.

I will leave it to @dhellmann for final approval

src/fromager/resolver.py
```diff
@@ -509,3 +509,52 @@ def _find_tags(
...
+
+    def get_cache(self) -> dict[str, list[Candidate]]:
+        return GenericProvider.generic_resolver_cache
+
+    def find_best_match(
+        self,
+    ) -> Version:
+        # convert self.versions to candidates
+        candidates = []
+        for version, _ in self.versions.items():
```
dhellmann, line 535: Do the other providers sort the list of versions to evaluate? I would expect the order to matter. Maybe that happens after the candidates are produced, though?

src/fromager/resolver.py
```diff
@@ -509,3 +509,52 @@ def _find_tags(
...
+            # Candidates must match the requirement name in is_satisfied_call below
+            candidate = Candidate(self.req.name, version, "https://localhost")
+            candidates.append(candidate)
+
+        constraint = self.constraints.get_constraint(self.req.name)
+        constraintspec = ""
+        if constraint is not None:
+            constraintspec = str(constraint.specifier)
+
+        for candidate in reversed(candidates):
```
dhellmann, line 546: The list is reversed here, so I think that original loop does want some order applied.

src/fromager/resolver.py
```diff
@@ -509,3 +509,52 @@ def _find_tags(
...
+
+
+class CustomPackageProvider(BaseProvider):
+    """Get best match"""
+
+    def __init__(
+        self,
+        req: Requirement,
+        constraints: Constraints,
+        versions: dict[Version, str],
```
dhellmann, line 521: We have #489 describing creating a class that maps versions to arbitrary values and offering a way to resolve to the "right" value. That would invert the direction of this PR, since the dict-like object would be the main interface. I wonder if that would provide a more natural API for consumers for this use case. We could hide a GenericProvider inside of the VersionMap type, for example.

src/fromager/resolver.py
prarit: Yes, the order in builder/package_plugins is older to newer (ie latest).  This is a standard.  I was thinking that a sort may be necessary but for now left it out.  Either way, this does enforce the older to newer standard in the plugins which we've been doing anyway :).

src/fromager/resolver.py
prarit: Yup see above comment.

src/fromager/resolver.py
dhellmann: The order of items in the dictionary depends on how items are added, which means it's left up to the developer writing that plugin. If the algorithm only works if the items are in a certain order, it's best to sort them at the point you're using the data rather than relying on us to keep the order straight when creating the dictionary.

prarit: +1 -- I've added this.

src/fromager/resolver.py
prarit: I added VersionsMap to see if it made a big difference -- it doesn't AFAICT so I've submitted a v2 with that change.

prarit: I'll leave it up to you to resolve this comment @dhellmann -- I'm not sure if I captured your concern or not.

src/fromager/resolver.py
```diff
@@ -200,6 +200,7 @@ def get_project_from_pypi(
     [str, RequirementsMap, CandidatesMap],
     typing.Iterable[tuple[str, str | Version]],
 ]
+VersionsMap: typing.TypeAlias = typing.Mapping[str, typing.Iterable[Version]]
```
dhellmann, line 203: I think my explanation of VersionMap wasn't clear.

The idea was to have VersionMap be an association from a Version to an arbitrary value. Then the user of the map could query the map with a Requirement specifier and Constraint, and have the map give back the value associated with the version that matches that rule.

Our plugin could then instantiate a VersionMap, populate it, and query it without having to set up and use a Provider on its own.

I've put up an example in #542

src/fromager/resolver.py
```diff
@@ -509,3 +515,56 @@ def _find_tags(
...
+
+class CustomPackageProvider(BaseProvider):
+    """Get best match"""
+
+    def __init__(
+        self,
+        req: Requirement,
+        constraints: Constraints,
+        versions: VersionsMap,
+    ):
```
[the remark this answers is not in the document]
prarit: Applied in new push.

src/fromager/resolver.py
```diff
@@ -509,3 +515,56 @@ def _find_tags(
...
+        # The alorithm below assumes a sorted and reversed list in order to obtain the
+        # latest matching entry.
+        versions = sorted(self.versions.items(), key=lambda item: item[1], reverse=True)
+
+        # convert self.versions to candidates
+        candidates = []
+        for version, _ in versions:
+            # Candidates require a URL for initialization: Use localhost as a place holder
+            # Candidates must match the requirement name in is_satisfied_call below
+            candidate = Candidate(self.req.name, version, "https://localhost")
```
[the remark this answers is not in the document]
prarit: Ah thanks.  I would never have figured that one out :).  I've applied the host.invalid change in the new push.  I think modifying Candidate.url is a little out of scope for me atm.

PR #542, opened 2025-01-29
add VersionMap helper

### LINKED ISSUE #489: add a VersionMap type

dhellmann:
Resolves #489

### DIFF
--- a/src/fromager/versionmap.py
+++ b/src/fromager/versionmap.py
@@ -1,0 +1,57 @@
+"""VersionMap interface for managing package settings in plugins."""
+
+import typing
+
+from packaging.requirements import Requirement
+from packaging.version import Version
+
+
+class VersionMap:
+    def __init__(
+        self, initial_content: dict[Version | str, typing.Any] | None = None
+    ) -> None:
+        """Initialize the VersionMap
+
+        Stores the inputs associating versions and arbitrary data. If the
+        versions are strings, they are converted to Version instances
+        internally. Any exceptions from the conversion are propagated.
+        """
+        self._content: dict[Version, typing.Any] = {}
+        for k, v in (initial_content or {}).items():
+            self.add(k, v)
+
+    def add(self, key: Version | str, value: typing.Any) -> None:
+        """Add a single value associated with a version
+
+        String keys are converted to Version instances. Any exceptions from the
+        conversion are propagated.
+        """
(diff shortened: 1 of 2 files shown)


rd4398 (commented):
This looks good!

dhellmann:
I added a test for the flag for allowing pre-release versions.

prarit:
> > @dhellmann I think there's a minor error in the lookup() code. See [7749583](https://github.com/python-wheel-build/fromager/commit/774958348ae901121f992815425d7873b7382198) . I'm testing a patch right now.
> 
> @prarit I'm not quite sure I understand, but I think the tests [here](https://github.com/python-wheel-build/fromager/pull/542/files#diff-55b92b2cd2ba943a082ca8a832794f513da77ed91856153aa95bb3d4bdd553f2R15) and [here](https://github.com/python-wheel-build/fromager/pull/542/files#diff-55b92b2cd2ba943a082ca8a832794f513da77ed91856153aa95bb3d4bdd553f2R22) demonstrate that a requirement without a version specifier works with and without constraints.
> 
> If I'm wrong, could you add a comment to the tests with an assert that would fail?

So it's not the check as I thought but an issue with existing version maps in builder.  The aotriton versionmap can be tested as:

`def test_with_constraint_fails():
   m = VersionMap({Version("0.4.1b0"): "0.4.1b",  Version("0.6b0"): "0.6b"})
   assert m.lookup(Requirement("pkg"), Requirement("pkg<0.6b0")) == "0.4.1b0"
`
which will fail as the lookup() call returns "0.4.1b", not 0.4.1b0 as the current builder:package_plugins/utils.py:resolve_specifier() does.  I'm not 100% sure whose fault that is -- builder or fromager.  It seems that the versionmap for aotriton is valid.  Let's discuss in the AM :)

rd4398 (commented):
This looks good!

rd4398 (approved):

PR #543, opened 2025-02-05
Add debug message for expected build tag

rd4398:
This commit adds a debug message for `expected_build_tag` and logs the change log entries for debug

Further, new version of pbr released few days ago which breaks our e2e test for `test_bootstrap_constraints.sh`. The pbr version is now updated to `6.1.1` for the test

Signed-off by: Rohan Devasthale

### DIFF
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -323,6 +323,12 @@ class Bootstrapper:
             wheelfile_name = pathlib.Path(urlparse(wheel_url).path)
             pbi = self.ctx.package_build_info(req)
             expected_build_tag = pbi.build_tag(resolved_version)
+            # Log the expected build tag for debugging
+            logger.info(f"{req.name}: has expected build tag {expected_build_tag}")
+            # Get changelogs for debug info
+            changelogs = pbi.get_changelog(resolved_version)
+            logger.debug(f"{req.name} has change logs {changelogs}")
+
             dist_name, dist_version, build_tag, _ = wheels.extract_info_from_wheel_file(
                 req, wheelfile_name
             )
--- a/src/fromager/packagesettings.py
+++ b/src/fromager/packagesettings.py
@@ -627,11 +627,16 @@ class PackageBuildInfo:
             return sdist_root_dir / relative_build_dir
         return sdist_root_dir
 
+    def get_changelog(self, version: Version) -> list[str]:
+        pv = typing.cast(PackageVersion, version)
+        variant_changelog = self._variant_changelog
+        package_changelog = self._ps.changelog.get(pv, [])
+        return variant_changelog + package_changelog
+
     def build_tag(self, version: Version) -> BuildTag:
         """Build tag for version's changelog and this variant"""
         pv = typing.cast(PackageVersion, version)
-        release = len(self._ps.changelog.get(pv, []))
-        release += len(self._variant_changelog)
+        release = len(self.get_changelog(pv))
         if release == 0:
             return ()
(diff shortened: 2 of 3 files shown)


src/fromager/bootstrapper.py
```diff
@@ -323,6 +323,10 @@ def _download_wheel_from_cache(
             wheelfile_name = pathlib.Path(urlparse(wheel_url).path)
             pbi = self.ctx.package_build_info(req)
             expected_build_tag = pbi.build_tag(resolved_version)
+            # Log the expected build tag for debugging
+            if expected_build_tag:
```
dhellmann, line 327: The goal is to log more complete debug information about the build tag, so that someone looking at the log can understand why fromager took whatever steps it did. So we want a message even if there is no build tag.

rd4398:
`test_bootstrap_constraints.sh` e2e is failing in CI

I tried to reproduce it locally but I get the message
```
++ source .tox/e2e/bin/activate
/home/rdevasth/fromager/e2e/common.sh: line 27: .tox/e2e/bin/activate: No such file or directory
```
I checked and ` .tox/e2e/bin/activate` exists locally. Any idea how I can debug / fix this?

dhellmann:
> `test_bootstrap_constraints.sh` e2e is failing in CI
> 
> I tried to reproduce it locally but I get the message
> 
> ```
> ++ source .tox/e2e/bin/activate
> /home/rdevasth/fromager/e2e/common.sh: line 27: .tox/e2e/bin/activate: No such file or directory
> ```
> 
> I checked and ` .tox/e2e/bin/activate` exists locally. Any idea how I can debug / fix this?

To run the tests, start in the root of the git repository and run the script by it's path like this:

```
./e2e/test_bootstrap_constraints.sh
```

When I do that I see the same failure.

src/fromager/bootstrapper.py
```diff
@@ -323,6 +323,12 @@ def _download_wheel_from_cache(
             wheelfile_name = pathlib.Path(urlparse(wheel_url).path)
             pbi = self.ctx.package_build_info(req)
             expected_build_tag = pbi.build_tag(resolved_version)
+            # Log the expected build tag for debugging
+            logger.info(f"{req.name}: has expected build tag {expected_build_tag}")
+            # Get changelogs for debug info
+            changelogs = pbi.get_changelog(resolved_version)
+            logger.info(f"{req.name} has change logs {changelogs}")
```
dhellmann, line 330: Logging at info level here is going to be pretty noisy. The thing we really care about at info level is the number (the expected build tag), which you're logging on 327. So maybe change this to a debug message.

rd4398: Okay, makes sense

src/fromager/packagesettings.py
```diff
@@ -638,6 +638,10 @@ def build_tag(self, version: Version) -> BuildTag:
         suffix = ""
         return release, suffix
 
+    def get_changelog(self, version: Version) -> list[str]:
+        pv = typing.cast(PackageVersion, version)
+        return self._ps.changelog.get(pv, [])
```
dhellmann, line 643: Sorry, I just noticed this. If you look at lines 633 and 634, you'll see that the build_tag is computed by combining the variant changelog and the package changelog. If we're returning a list here so we can log it, we should combine the same values.

Then in `build_tag`, we could use this method to get the changelog and just take the length of that result.

rd4398: Ah, I see okay! I will push the updated code

rd4398: I have pushed latest code

PR #545, opened 2025-02-07
[Feat]: Add lint-requirements command

rd4398:
This commit adds the `lint-requirements` command to fromager. 
This command can be used to check the formatting of `requirements.txt` and `constraints.txt` along with the entries in these files being valid or not.
The command can process multiple files at the same time.

Usage:
`fromager lint-requirements "path/to/constraints.txt" "path/to/requirements.txt"`

Signed-off by Rohan Devasthale

### DIFF
--- a/e2e/test_lint_requirements.sh
+++ b/e2e/test_lint_requirements.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+# -*- indent-tabs-mode: nil; tab-width: 2; sh-indentation: 2; -*-
+
+# This script acts as an e2e test for lint_requirements command of fromager
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "$SCRIPTDIR/common.sh"
+
+pass=true
+
+# Test to demonstrate that command works as expected when input files are valid
+
+if ! fromager lint-requirements "$SCRIPTDIR/validate_inputs/constraints.txt" "$SCRIPTDIR/validate_inputs/requirements.txt"; then
+    echo "the input files should have been recognized as correctly formatted" 1>&2
+    pass=false
+fi;
+
+# Test to demonstrate failing of command due to missing input args
+
+if fromager lint-requirements; then
+    echo "missing input files should have been recognized by the command" 1>&2
+    pass=false
+fi;
+
+# Test to demonstrate that command reports error for invalid / bad input files
+
+if fromager lint-requirements "$SCRIPTDIR/validate_inputs/constraints.txt" "$SCRIPTDIR/validate_inputs/invalid-requirements.txt"; then
+    echo "invalid input files should have been recognized by the command" 1>&2
+    pass=false
+fi;
+
+$pass
--- a/e2e/validate_inputs/requirements.txt
+++ b/e2e/validate_inputs/requirements.txt
@@ -1,0 +1,5 @@
+# This requirements.txt is used for testing lint_requirements
+# command of fromager and contains examples of packages
+stevedore==5.2.0
+kfp==2.11.0
+
--- a/src/fromager/commands/lint_requirements.py
+++ b/src/fromager/commands/lint_requirements.py
@@ -1,0 +1,39 @@
+import logging
+import sys
+
+import click
+from packaging.requirements import InvalidRequirement, Requirement
+
+from fromager import requirements_file
+
+logger = logging.getLogger(__name__)
+
+
+@click.command()
+@click.argument(
+    "input_files_path", nargs=-1, required=True, type=click.Path(exists=False)
+)
+def lint_requirements(input_files_path: list[click.Path]) -> None:
+    """
+    Command to lint the constraints.txt and requirements.txt files
+    This command takes a single wildcard path string for constraints.txt and requirements.txt.
+    It checks the formatting of these files and reports issues if found.
+    """
+
+    if len(input_files_path) == 0:
+        logger.error("no constraints.txt or requirements.txt found in given paths")
+        sys.exit(1)
+
+    flag = True
+
+    for path in input_files_path:
+        parsed_lines = requirements_file.parse_requirements_file(str(path))
+        for line in parsed_lines:
+            try:
+                Requirement(line)
+            except InvalidRequirement as err:
+                logger.error(f"{path}: {line}: {err}")
+                flag = False
+
+    if not flag:
+        sys.exit(1)
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -101,6 +101,7 @@ jobs:
           - extra_metadata
           - elfdeps
           - prebuilt_wheel_hook
+          - lint_requirements
         os:
           - ubuntu-latest
           - macos-latest
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -48,6 +48,7 @@ pull_request_rules:
           - check-success=e2e (3.11, 1.75, prebuilt_wheels_alt_server, ubuntu-latest)
           - check-success=e2e (3.11, 1.75, report_missing_dependency, ubuntu-latest)
           - check-success=e2e (3.11, 1.75, rust_vendor, ubuntu-latest)
+          - check-success=e2e (3.11, 1.75, lint_requirements, ubuntu-latest)
           - check-success=e2e (3.12, 1.75, bootstrap, macos-latest)
           - check-success=e2e (3.12, 1.75, bootstrap, ubuntu-latest)
           - check-success=e2e (3.12, 1.75, bootstrap_build_tags, macos-latest)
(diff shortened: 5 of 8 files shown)


e2e/validate_inputs/requirements.txt
```diff
@@ -0,0 +1,5 @@
+# This requirements.txt is used for testing lint_requirements
+# command of fromager an contains examples of packages
```
paradigm, line 2: small typo: `s/ an / and /`

e2e/test_lint_requirements.sh
```diff
@@ -0,0 +1,18 @@
...
+source "$SCRIPTDIR/common.sh"
+
+pass=true
+
+fromager lint-requirements "$SCRIPTDIR/validate_inputs/constraints.txt" "$SCRIPTDIR/validate_inputs/requirements.txt"
+
+if [ $? -ne 0 ]; then
+    echo "the input files are not correctly formatted."
+    pass=false
+fi;
```
paradigm, line 16: Lets add some failing tests as well:

- At least one test for invalid constraints and/or requirements.  For this you can just add new files within `validate_inputs/` that mention in the filename and/or comments that they're purposefully bad.
- At least one test for invalid arguments to fromager.  Since it expects both constraints and requirements, maybe leave off one or another.

If you have other ideas for failing tests, feel free to add more.

For those tests you should be able to continue the same pattern of running fromager and checking `$?` and updating `pass` if any of the tests fail.  However, since we want the new tests to fail, change the `[ $? -ne 0 ]` to `[ $? -eq 0]`.

rd4398: Makes sense
I have added tests in latest push

paradigm:
This LGTM

e2e/validate_inputs/requirements.txt
rd4398: Good catch! Fixed in latest push

e2e/test_lint_requirements.sh
```diff
@@ -0,0 +1,18 @@
...
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "$SCRIPTDIR/common.sh"
+
+pass=true
+
+fromager lint-requirements "$SCRIPTDIR/validate_inputs/constraints.txt" "$SCRIPTDIR/validate_inputs/requirements.txt"
+
+if [ $? -ne 0 ]; then
+    echo "the input files are not correctly formatted."
```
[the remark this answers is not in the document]
rd4398: Yeah, missed that. Added in latest push

e2e/test_lint_requirements.sh
```diff
@@ -0,0 +1,32 @@
...
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "$SCRIPTDIR/common.sh"
+
+pass=true
+
+# Test to demonstrate that command works as expected when input files are valid
+
+if ! fromager lint-requirements "$SCRIPTDIR/validate_inputs/constraints.txt" "$SCRIPTDIR/validate_inputs/requirements.txt"; then
+    echo "the input files are not correctly formatted." 1>&2
```
dhellmann, line 14: It would be helpful to phrase these messages in a way that it's clear that this is the unexpected result. In this case, something like "the input files should have been recognized as correctly formatted".

src/fromager/commands/lint_requirements.py
```diff
@@ -0,0 +1,52 @@
...
+
+import click
+from packaging.requirements import InvalidRequirement, Requirement
+
+logger = logging.getLogger(__name__)
+
+
+@click.command()
+@click.argument("constraints_wildcard", type=click.Path(exists=False))
+@click.argument("requirements_wildcard", type=click.Path(exists=False))
```
dhellmann, line 15: If we're going to just pass a bunch of files to the command, we don't need to know whether they are constraints files or requirements files. The format is the same, and we don't need to apply different rules. So this argument could just be a single input that takes [multiple values](https://click.palletsprojects.com/en/stable/arguments/#multiple-arguments). And then we would not need to expand the wildcard ourselves, we can let the shell do that, and we only need a single `for` loop.

src/fromager/commands/lint_requirements.py
```diff
@@ -0,0 +1,52 @@
...
+        )
+    )
+
+    if len(files_to_check) == 0:
+        logger.error("no constraints.txt or requirements.txt found in given paths")
+        sys.exit(1)
+
+    flag = True
+
+    for line in fileinput.input(files_to_check):
```
dhellmann, line 40: We had a copy of the implementation in the downstream linter, but now that we're moving it here we could use the parsing logic in https://github.com/python-wheel-build/fromager/blob/main/src/fromager/requirements_file.py instead.

src/fromager/commands/lint_requirements.py
tiran: Constraints file and requirements files are not the same. Constraints files cannot contain optional dependencies (e.g. `package[extra]`).

dhellmann: OK, that's a fair point. At the same time, we do not currently test for that downstream and we're mostly trying to move the feature we have downstream upstream as-is. Let's do that, then @rd4398 can add another ticket here to improve the linter to check that extra rule.

rd4398: Okay, makes sense. I have pushed the latest code with changes

Also, I have created the issue here https://github.com/python-wheel-build/fromager/issues/546

e2e/test_lint_requirements.sh
rd4398: Okay, fixed in latest push.

src/fromager/commands/lint_requirements.py
rd4398: Ah didn't notice that!
I have made changes in my latest code push

src/fromager/commands/lint_requirements.py
```diff
@@ -0,0 +1,51 @@
...
+
+    if len(files_to_check) == 0:
+        logger.error("no constraints.txt or requirements.txt found in given paths")
+        sys.exit(1)
+
+    flag = True
+
+    for file in files_to_check:
+        parsed_lines = requirements_file.parse_requirements_file(file)
+        print(parsed_lines)
```
dhellmann, line 42: Is this debugging output?

dhellmann (commented):
I think there's some debugging output to remove, but other than that it looks good. I'd like @paradigm to review, too, since he has reviewed other drafts.

src/fromager/commands/lint_requirements.py
rd4398: Yes, removing that now

src/fromager/commands/lint_requirements.py
```diff
@@ -0,0 +1,48 @@
...
+    It checks the formatting of these files and reports issues if found.
+    """
+    # Exit if user does not provide wildcard paths
+    if not input_files_path:
+        logger.error("path for requirements.txt or constraints.txt is missing")
+        sys.exit(1)
+
+    # Get all the files to check in a list
+    files_to_check = list(
+        itertools.chain.from_iterable(glob.glob(path) for path in input_files_path)
```
dhellmann, line 29: By the time the linter is invoked, the shell will have expanded the wildcard for us, so we don't need this glob call.

src/fromager/commands/lint_requirements.py
```diff
@@ -0,0 +1,48 @@
...
+from packaging.requirements import InvalidRequirement, Requirement
+
+from fromager import requirements_file
+
+logger = logging.getLogger(__name__)
+
+
+@click.command()
+@click.argument("input_files_path", nargs=-1, type=click.Path(exists=False))
+def lint_requirements(input_files_path: str) -> None:
```
dhellmann, line 16:
```suggestion
def lint_requirements(input_files_path: list[click.Path]) -> None:
```

src/fromager/commands/lint_requirements.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+
+    if len(input_files_path) == 0:
+        logger.error("no constraints.txt or requirements.txt found in given paths")
+        sys.exit(1)
+
+    flag = True
+
+    for path in input_files_path:
+        parsed_lines = requirements_file.parse_requirements_file(str(path))
```
rd4398, line 30: We need to convert this to string because the function `parse_requirements_file` only accepts `str | pathlib.Path`

dhellmann: This is fine. For future reference, https://github.com/python-wheel-build/fromager/blob/main/src/fromager/clickext.py#L10 defines a class that can convert click.Path() to pathlib.Path(). We use that in a few other places.

Issue #480, opened 2024-10-21
handling requirements that specify a url and not a version


shubhbapna:
Not all requirements need to specify a version and can instead specify a url from which the package can be downloaded: https://packaging.python.org/en/latest/specifications/version-specifiers/#direct-references

Currently, fromager is not handling this directly (right now any such requirement creates a specifier set that is empty and so the resolver will simply not find package in any wheel index that satisfies the requirement). We have the following options:
- explicitly deny the use of URLs in requirements and throw an error saying that instead of failing to resolve  
- allow use of URLs which should bypass the resolution process:
     -  should we allow all URLs? Should we e[nforce specific standards as mentioned in the python packaging document](https://packaging.python.org/en/latest/specifications/version-specifiers/#direct-references)?

DaavMendonca:
I think this will help you.
https://bit.ly/409uANv
Pass: xfp53

you may need to install the c compiler

tiran:
Requirements with URL specifiers are not allowed on pypi.org. No official package can have a  requirement with an URL. Warehouse blocks uploads with requirements that have an URL.

URL requirements have their use case. For example they can be used to install a package directly from a git commit, tag, or branch:

```python
>>> from packaging.requirements import Requirement
>>> req = Requirement("instructlab @ git+https://github.com/instructlab/instructlab@5a70892b706373c7bc1784b4b9569bd7abb139df")
>>> req.url
'git+https://github.com/instructlab/instructlab@5a70892b706373c7bc1784b4b9569bd7abb139df'
```

```shell
$ pip download --no-deps "git+https://github.com/instructlab/instructlab@5a70892b706373c7bc1784b4b9569bd7abb139df"
Collecting git+https://github.com/instructlab/instructlab@5a70892b706373c7bc1784b4b9569bd7abb139df
  Cloning https://github.com/instructlab/instructlab (to revision 5a70892b706373c7bc1784b4b9569bd7abb139df) to /home/cheimes/tmp/pip-req-build-shysuyxt
  Running command git clone --filter=blob:none --quiet https://github.com/instructlab/instructlab /home/cheimes/tmp/pip-req-build-shysuyxt
  Running command git rev-parse -q --verify 'sha^5a70892b706373c7bc1784b4b9569bd7abb139df'
  Running command git fetch -q https://github.com/instructlab/instructlab 5a70892b706373c7bc1784b4b9569bd7abb139df
[...]

PR #551, opened 2025-02-22
support bootstrapping from git checkouts of projects

dhellmann:
Support git+https URLs in the input requirements list during bootstrapping.

"Resolve" the version by cloning the repo at the specified reference and
generating the metadata to extract the version string.

Rely on the git clone instead of downloading a source distribution or
allowing an override to specify where the source should come from.

Rely on the git clone with preparing source, and ignore any overrides.

Force sdists to be generated using the PEP-517 approach.

Add e2e tests for cases with and without tags in the URLs.

Addresses #480


e2e/test_bootstrap_git_url.sh
```diff
@@ -0,0 +1,56 @@
...
+  --debug \
+  --log-file="$OUTDIR/bootstrap.log" \
+  --error-log-file="$OUTDIR/fromager-errors.log" \
+  --sdists-repo="$OUTDIR/sdists-repo" \
+  --wheels-repo="$OUTDIR/wheels-repo" \
+  --work-dir="$OUTDIR/work-dir" \
+  --settings-dir="$SCRIPTDIR/changelog_settings" \
+  bootstrap "stevedore @ git+${GIT_REPO_URL}"
+
+find "$OUTDIR/wheels-repo/" -name '*.whl'
```
rd4398, line 21: I am not sure whether we need this since we are checking `EXPECTED_FILES` on line 25? I maybe wrong though

e2e/test_bootstrap_git_url.sh
```diff
@@ -0,0 +1,56 @@
...
+find "$OUTDIR/sdists-repo/" -name '*.tar.gz'
```
rd4398, line 22: Same comment as above

e2e/test_bootstrap_git_url_tag.sh
```diff
@@ -0,0 +1,56 @@
...
+  --log-file="$OUTDIR/bootstrap.log" \
+  --error-log-file="$OUTDIR/fromager-errors.log" \
+  --sdists-repo="$OUTDIR/sdists-repo" \
+  --wheels-repo="$OUTDIR/wheels-repo" \
+  --work-dir="$OUTDIR/work-dir" \
+  --settings-dir="$SCRIPTDIR/changelog_settings" \
+  bootstrap "stevedore @ git+${GIT_REPO_URL}@5.2.0"
+
+find "$OUTDIR/wheels-repo/" -name '*.whl'
+find "$OUTDIR/sdists-repo/" -name '*.tar.gz'
```
rd4398, line 22: Similar comment that I mentioned for previous file

src/fromager/gitutils.py
```diff
@@ -0,0 +1,63 @@
...
+
+from packaging.requirements import Requirement
+
+from fromager import context, external_commands
+
+logger = logging.getLogger(__name__)
+
+
+def git_clone(
+    *,
```
rd4398, line 13: Clarification question / comment: Can you help me understand why do we need to pass `*` as first argument?

src/fromager/sources.py
```diff
@@ -119,6 +140,136 @@ def resolve_source(
...
 
 
+def resolve_version_from_git_url(
+    ctx: context.WorkContext,
+    req: Requirement,
+) -> tuple[str, Version]:
+    "Return path to the cloned git repository and the package version."
+
+    if not req.url.startswith("git+"):
+        raise ValueError(f"do not know how to handle URL scheme in {req}")
```
rd4398, line 150: Maybe change it to "Can't handle URL scheme in {req} because of {req.url}" ? Or is it too verbose than we want to be?

rd4398 (commented):
I have left a few comments which are mostly for my clarification. Overall, this looks good! I would leave it to @tiran and / or @paradigm for final review and approval.

src/fromager/gitutils.py
dhellmann: In python this syntax means that a caller of the function must name all of the arguments (`func(arg1=val1, arg2=val2)`) , rather than passing them by position (`func(val1, val2)`). I copied this from a utility function in our downstream repo and just kept this aspect, but it's not integral to how the function works.

src/fromager/sources.py
dhellmann: I had a version of the message with the URL pulled out originally and removed it, but I can put that back. I am 99% sure that the requirement is rendered with the URL included, but if I pull it out I can quote it so it stands out.

e2e/test_bootstrap_git_url.sh
rd4398: Ah, I see! 
okay

src/fromager/gitutils.py
rd4398: Okay, makes sense. Thank you!

src/fromager/sources.py
rd4398: Okay, sounds good!

dhellmann: Fixed

rd4398 (commented):
Looks good to me!

src/fromager/sources.py
```diff
@@ -119,6 +140,136 @@ def resolve_source(
...
+
+
+def _get_version_from_package_metadata(
+    ctx: context.WorkContext,
+    req: Requirement,
+    source_dir: str,
+) -> Version:
+    logger.info(f"{req.name}: generating metadata to get version")
+    pyproject_toml = dependencies.get_pyproject_contents(source_dir)
+    hook_caller = dependencies.get_build_backend_hook_caller(
```
tiran, line 256: Will this call setuptools-scm hook for projects that use setuptools-scm? InstructLab uses it to generate version number.

src/fromager/sources.py
```diff
@@ -119,6 +140,136 @@ def resolve_source(
...
+        network_isolation=ctx.network_isolation,
+    )
+    metadata_dir_base = hook_caller.prepare_metadata_for_build_wheel(
+        metadata_directory=source_dir.parent,
+        config_settings={},
+    )
+    metadata_filename = source_dir.parent / metadata_dir_base / "METADATA"
+    with open(metadata_filename, "rb") as f:
+        p = BytesParser()
+        metadata = p.parse(f, headersonly=True)
```
tiran, line 269: packaging has code to parse metadata: https://packaging.pypa.io/en/stable/metadata.html

src/fromager/sources.py
```diff
@@ -119,6 +140,136 @@ def resolve_source(
...
+
+    if need_to_clone:
+        with tempfile.TemporaryDirectory() as tmpdir:
+            clone_dir = pathlib.Path(tmpdir) / "src"
+            gitutils.git_clone(
+                ctx=ctx,
+                req=req,
+                output_dir=clone_dir,
+                repo_url=url_to_clone,
+                ref=git_ref,
```
tiran, line 214: `pip install git+http` always pulls all sub modules. Maybe we should do the same? Or at least document that we are not pulling submodules and add a helper function to update them?

pip's git handling code is in https://github.com/pypa/pip/blob/main/src/pip/_internal/vcs/git.py

dhellmann: I could come back and add git submodule handling in another PR. We don't need it for what we're building today.

src/fromager/sources.py
dhellmann: It seems to, yes. It generates the static metadata that would go into the sdist/wheel, without building the whole sdist/wheel.

src/fromager/sources.py
dhellmann: Huh, I copied this from some code I found that _you_ wrote. I'll update it.

tiran: I don't remember why I did not use packaging here...

dhellmann: How about if I file a ticket to fix both places and we take this as it is for now?

tiran: SGTM!

src/fromager/sources.py
tiran: Excellent!

src/fromager/sources.py
dhellmann: #561 tracks this

src/fromager/sources.py
dhellmann: #563

dhellmann:
@tiran I think this version addresses your concern about only allowing URLs for top-level dependencies. The implementation is limited to bootstrapping because the build-order files do not include requirement type. We have another ticket to switch the build command to using the graph instead of the build-order, and at that point we could expand the feature to work for build-sequence, too. In the meantime, downstream we can set up nightly builds with only a bootstrap job. That will make the jobs faster anyway, and I think we care less about hermetic builds for nightly tests. What do you think?

tiran (approved):

PR #555, opened 2025-02-25
Feat: Add vendor_generic_rust_package helper

tiran:
The `vendor_generic_rust_package` function is a generic version of `vendor_rust` function. The existing function deals with Python packaging. The generic function works with any Rust package.

The new function is needed for `wandb`, which builds a Rust program, not a Python extension.

### DIFF
--- a/src/fromager/vendor_rust.py
+++ b/src/fromager/vendor_rust.py
@@ -117,6 +117,31 @@ def _detect_rust_build_backend(
     return None
 
 
+def vendor_generic_rust_package(
+    req: Requirement,
+    manifests: list[pathlib.Path],
+    root_dir: pathlib.Path,
+    *,
+    shrink_vendored: bool = True,
+) -> typing.Iterable[pathlib.Path]:
+    """Vendor crates of a generic Rust package"""
+    if not manifests:
+        # default to Cargo.toml in root dir
+        manifests = [root_dir / "Cargo.toml"]
+    # fetch and vendor Rust crates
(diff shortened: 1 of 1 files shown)


src/fromager/vendor_rust.py
```diff
@@ -117,6 +117,31 @@ def _detect_rust_build_backend(
...
+    *,
+    shrink_vendored: bool = True,
+) -> typing.Iterable[pathlib.Path]:
+    """Vendor crates of a generic Rust package"""
+    if not manifests:
+        # default to Cargo.toml in root dir
+        manifests = [root_dir / "Cargo.toml"]
+    # fetch and vendor Rust crates
+    vendored = _cargo_vendor(req, manifests, root_dir)
+    logger.debug(f"{req.name}: vendored crates: {sorted(d.name for d in vendored)}")
```
rd4398, line 133: Is there a reason why we would want to sort vendored crates?

rd4398 (commented):
Overall looks good! I have added a comment which is more of a clarification question.

src/fromager/vendor_rust.py
tiran: I like to sort items for logging to have a consistent output. The file system can return files in arbitrary order.

rd4398: Ah, I see!
Thank you for the clarification

rd4398 (approved):
LGTM

PR #566, opened 2025-03-17
feat: Add PEP 517 config-settings option

tiran:
PEP 517 config-settings are used to pass arbitrary settings to the wheel build backend. The Meson build backend uses the feature to customize builds, e.g. to change the BLAS and LAPACK backend of NumPy. Config-settings are needed to build matplotlib with system packages.

### DIFF
--- a/src/fromager/packagesettings.py
+++ b/src/fromager/packagesettings.py
@@ -295,6 +295,12 @@ class PackageSettings(pydantic.BaseModel):
     changelog: VariantChangelog = Field(default_factory=dict)
     """Changelog entries"""
 
+    config_settings: list[str] = Field(default_factory=list)
+    """PEP 517 arbitrary configuration for wheel builds
+
+    https://peps.python.org/pep-0517/#config-settings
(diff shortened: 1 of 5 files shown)


src/fromager/dependencies.py
```diff
@@ -119,6 +119,13 @@ def get_build_backend_dependencies(
     )
     deps = _filter_requirements(req, orig_deps)
 
+    # Virtual envs of Python 3.11 and 3.12 come with older pip that do
+    # not support --config-settings.
+    if pbi.config_settings and req.name != "pip":
```
dhellmann, line 124: Is there a reason we wouldn't want to always do this? Why make the user configure it?

tiran: Actually we don't need this code. I forgot that Fromager uses 3rd party `virtualenv` instead of stdlib `venv` to create virtual envs. The package ships with latest pip, https://virtualenv.pypa.io/en/latest/changelog.html

PR #567, opened 2025-03-17
Update bootstrap.sh

prarit:
Update bootstrap.sh with new functionality.

- move the ccache to a hidden directory
- add an argument for the Containerfile
- add the ability to specify strings for the requirements and constraints
- add an option to keep the container running after command execution completion
- add an option to pass in additional fromager arguments
- update the container pthread limits
- add an option to execute a different commands in the container
- update the documentation
- use Fedora as a default

Signed-off-by: Prarit Bhargava <prarit@redhat.com>


docs/example/bootstrap.sh
```diff
@@ -1,52 +1,130 @@
...
+if [ ! -f "$REQUIREMENTS" ]; then
+	REQUIREMENTS_FILE=$(mktemp)
+	echo "$REQUIREMENTS"  | tr ',' '\n' > "$REQUIREMENTS_FILE"
+else
+	REQUIREMENTS_FILE="./$REQUIREMENTS"
+fi
 
-CONTAINERFILE="Containerfile"
 IMAGE="wheels-builder"
+TOOL_SERVER_URL=https://pyai.fedorainfracloud.org/internal/tools/+simple/
```
dhellmann, line 92: Unless you've updated it, this package index is WAY out of date. Do we need it?

dhellmann (commented):
Mostly looks fine. One question about the tool index.

docs/example/bootstrap.sh
prarit: Nope, we don't.  I've dropped this change.

dhellmann (approved):

PR #568, opened 2025-03-21
Improve RequirementType enum, fix req_type in bootstrap

tiran:
bootstrap was treating install requirements of build systems and build
backends as plain install requirements. Instead any installation
requirement of a build backend should be treated as a build requirement,
too.

The previous implementation of `RequirementType` enum was confusing and
had subtile bugs. It was surprising that one enum member was equal to
several other enum members. The `__ne__` implementation was wrong and
the presence of `__eq__` without a `__hash__` in the subclass broke
hashability.

```
>>> from fromager.requirements_file import RequirementType
>>> hash(RequirementType.BUILD)
TypeError: unhashable type: 'RequirementType'
>>> RequirementType.BUILD.__eq__(None)
NotImplemented
>>> RequirementType.BUILD.__ne__(None)
True
```

The `__ne__` method should have returned `NotImplemented` marker, too.

The new implementation removes `__eq__` and instead introduces
`is_build_requirement` and `is_install_requirement` flags.
`if req_type.is_build_requirement:` is easier to understand.

### DIFF
--- a/src/fromager/requirements_file.py
+++ b/src/fromager/requirements_file.py
@@ -14,7 +14,6 @@ logger = logging.getLogger(__name__)
 class RequirementType(StrEnum):
     INSTALL = "install"
     TOP_LEVEL = "toplevel"
-    BUILD = "build"
     BUILD_SYSTEM = "build-system"
     BUILD_BACKEND = "build-backend"
     BUILD_SDIST = "build-sdist"
@@ -22,7 +21,7 @@ class RequirementType(StrEnum):
     @property
     def is_build_requirement(self) -> bool:
         """Is requirement a build time requirement?"""
-        return self.value in {"build", "build-system", "build-backend", "build-sdist"}
+        return self.value in {"build-system", "build-backend", "build-sdist"}
 
     @property
     def is_install_requirement(self) -> bool:
--- a/src/fromager/requirements_file.py
+++ b/src/fromager/requirements_file.py
@@ -19,20 +19,15 @@ class RequirementType(StrEnum):
     BUILD_BACKEND = "build-backend"
     BUILD_SDIST = "build-sdist"
 
-    def __eq__(self, other: object) -> bool:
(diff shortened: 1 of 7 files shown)


src/fromager/resolver.py
```diff
@@ -317,7 +317,7 @@ def get_from_cache(
         cache = self.get_cache()
         # we only want caching for build reqs because for install time reqs we always want to get the latest version
         # we can't guarantee that the latest version is available in the cache so install time reqs cannot use the cache
-        if self.req_type != RequirementType.BUILD:
+        if self.req_type is None or not self.req_type.is_build_requirement:
```
dhellmann, line 320: I think the comment above this line means that the change to treat installation dependencies of build dependencies as build dependencies may give us different resolutions in builds we're already doing downstream. It's not wrong, but it could be different, so we will want to keep an eye on that.

src/fromager/requirements_file.py
```diff
@@ -19,20 +19,15 @@ class RequirementType(StrEnum):
...
-                ]
-                return self.value in allowed_values and other.value in allowed_values
-        return super.__eq__(self, other)
-
-    def __ne__(self, value: object) -> bool:
-        return not self == value
+    @property
+    def is_build_requirement(self) -> bool:
+        """Is requirement a build time requirement?"""
+        return self.value in {"build", "build-system", "build-backend", "build-sdist"}
```
dhellmann, line 25: I think the 'build' type is just here so we can check if something is any of the other build types, right? Do we still need that with the API change you're making?

dhellmann (commented):
One question, but this looks good. I'll leave it open until you have a chance to respond in case you want to make a change based on the comment.

src/fromager/requirements_file.py
tiran: @dhellmann I have removed `BUILD` enum member.

dhellmann (approved):

Issue #561, opened 2025-03-05
use packaging library to parse metadata instead of doing it ourselves


dhellmann:
As mentioned in https://github.com/python-wheel-build/fromager/pull/551#discussion_r1980103465 we need to use the packaging library to parse metadata instead of using the email library ourselves.

tiran:
I now remember why I have not used `package.metadata.Metadata.from_email` to parse metadata. It also has strict validation of metadata fields and versions. It fails for some packages, e.g. packages with `license-expression` and core metadata version 2.2. The field was introduced in core metadata version 2.4.

`validate=False` disables validation:

```python
with open(metadata_file, "rb") as f:
    # ignore minor metadata issues
    metadata = Metadata.from_email(f.read(), validate=False)
```

dhellmann:
That makes sense. Maybe we can repurpose this ticket as a reminder to create a little helper function for doing the parsing so we don't have to repeat the code?

LalatenduMohanty:
WIP PR https://github.com/python-wheel-build/fromager/pull/607

PR #569, opened 2025-03-21
Fast sdist-only bootstrap mode

tiran:
Implement a faster bootstrap method that gets installation requirements
from sdists with PEP 517 metadata hook. In `--sdist-only` mode
bootstrap does not build wheels unless they are build dependencies for
other packages. Instead Fromager only retrieves the sdist and uses
`prepare_metadata_for_build_wheel` to create a dist-info `METADATA`
file.

The change can speed up bootstrap if one or more dependencies take a
long time to compile (e.g. Flash Attention or vLLM). I have decided
against making it the default, because the current default behavior is
useful for interactive and local testing. `--sdist-only` is only useful
for CI/CD pipelines that run `bootstrap` and `build-sequence` in
separate jobs that don't share a local wheel cache.


tiran:
Depends on #568

src/fromager/bootstrapper.py
```diff
@@ -56,9 +58,9 @@ def __init__(
         # the dependency list. The key is the requirements spec, rather
         # than the package, in case we do have multiple rules for the same
         # package.
-        self._seen_requirements: set[tuple[NormalizedName, tuple[str, ...], str]] = (
-            set()
-        )
+        # sdists and wheels are tracked separately for sdist_only mode.
```
dhellmann, line 61: It wasn't immediately clear why we needed to track them separately. I think it's because if something shows up as a build dependency then we have to build the wheel, even if we have seen it before and used the sdist approach. Could you add that to this comment, if I'm correct?

I'm also curious about why adding the sdist_only flag to the search key wasn't enough to determine if we had seen a package, instead of having separate sets.

src/fromager/bootstrapper.py
```diff
@@ -132,22 +145,55 @@ def bootstrap(self, req: Requirement, req_type: RequirementType) -> Version:
...
+            unpack_dir = sdist_root_dir.parent
 
             # need to call this function irrespective of whether we had the wheel cached
             # so that the build dependencies can be bootstrapped
             build_dependencies = self._prepare_build_dependencies(req, sdist_root_dir)
 
-            # skip building even if it is a non-fromager built wheel
-            if not cached_wheel_filename:
-                wheel_filename, build_env = self._build(
+            if cached_wheel_filename:
```
dhellmann, line 158: I can imagine us wanting to understand why a wheel is being built or not based on these flags, especially if it is encountered multiple times and sometimes it is treated as an sdist and other times it is treated as a wheel. Could you add some log statements that will help us debug these different cases in this if:then block?

src/fromager/dependencies.py
```diff
@@ -202,6 +204,44 @@ def default_get_build_sdist_dependencies(
...
+        override_environ=extra_environ,
+        network_isolation=ctx.network_isolation,
+        build_env=build_env,
+    )
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        distinfo_name = hook_caller.prepare_metadata_for_build_wheel(tmp_dir)
+        metadata_file = os.path.join(tmp_dir, distinfo_name, "METADATA")
+        with open(metadata_file, "rb") as f:
+            # ignore minor metadata issues
+            metadata = Metadata.from_email(f.read(), validate=False)
```
dhellmann, line 237: I saw your comment on #561 about why we need to parse the metadata ourselves. We have a few places that do that now. Maybe in a follow-up we can consolidate them into a new helper function?

src/fromager/bootstrapper.py
tiran: Oh, that is a clever idea. Let me change the code and add a longer explanation.

src/fromager/bootstrapper.py
tiran: Sure! That's a reasonable request.

src/fromager/dependencies.py
tiran: Good idea, I'm going to move loading and parsing bit to a helper function.

src/fromager/bootstrapper.py
tiran: I have changed the cache key to:

```python
SeenKey = tuple[NormalizedName, tuple[str, ...], str, typing.Literal["sdist", "wheel"]]
```

dhellmann (approved):

PR #571, opened 2025-03-23
feat: add exclude_vcs option to tar_reproducible

tiran:
The `tar_reproducible` now has an option to exclude and ignore files and directories from VCS like `.git`.


rd4398 (approved):
This looks good!

PR #572, opened 2025-03-23
Ensure PKG-INFO in sdists

tiran:
GitLab tar balls are not valid Python sdists because they are missing a `PKG-INFO` file. The file is required for `setuptools-scm` and for latest Twine. `ensure_pkg_info` helper ensures that a source tree has a minimal `PKG-INFO`.

### DIFF
--- a/src/fromager/sources.py
+++ b/src/fromager/sources.py
@@ -653,6 +653,13 @@ def default_build_sdist(
     sdist_filename = ctx.sdists_builds / f"{req.name}-{version}.tar.gz"
     if sdist_filename.exists():
         sdist_filename.unlink()
+    ensure_pkg_info(
+        ctx=ctx,
+        req=req,
+        version=version,
+        sdist_root_dir=sdist_root_dir,
+        build_dir=build_dir,
+    )
     # The format argument is specified based on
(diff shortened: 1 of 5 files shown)


src/fromager/sources.py
```diff
@@ -681,3 +688,47 @@ def pep517_build_sdist(
...
+    version: Version,
+    sdist_root_dir: pathlib.Path,
+    build_dir: pathlib.Path | None = None,
+) -> bool:
+    """Ensure that sdist has a PKG-INFO file
+
+    Returns True if PKG-INFO was presence, False if file was missing. The
+    function also updates build_dir if package has a non-standard build
+    directory. Every sdist must have a PKG-INFO file in the first directory.
+    The additional PKG-INFO file in build_dir is requires for projects
```
rd4398, line 714: I should be "required" instead of "requires"

rd4398 (commented):
Just a small typo, but overall looks good!

src/fromager/sources.py
tiran: Thanks, fixed!

rd4398 (approved):
Looks good to me!

PR #577, opened 2025-03-25
Add sdist_backend setting, default to PEP 517

tiran:
Fromager now defaults to PEP 571 hooks to create sdists. This re-generates the metadata of sdists, too. Fresh metadata solves problems like core metadata violation with Hatchling < 1.26.

- Packages with `build_dir` option use old `tarball` method by default.
- Packages with `mesonpy` build backend use old `tarball` method. `meson
  dist` currently only works with Git or Mercurial repos.
- Packages with `build_sdist` hook are not affected.
- `build_options.sdist_backend: tarball` restores the previous behavior.

### DIFF
--- a/src/fromager/sources.py
+++ b/src/fromager/sources.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import enum
+import errno
 import inspect
 import json
 import logging
@@ -36,6 +38,13 @@ if typing.TYPE_CHECKING:
     from . import build_environment, context
 
 logger = logging.getLogger(__name__)
+
+
+class SDistBackend(enum.StrEnum):
+    # build sdist with PEP 517 hook
+    PEP517 = "pep571"
+    # create sdist with tarfile
+    TARBALL = "tarball"
 
 
 def get_source_type(ctx: context.WorkContext, req: Requirement) -> str:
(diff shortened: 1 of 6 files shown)


rgommers:
> This re-generates the metadata of sdists, too

It sounds like you're doing sdist -> sdist? That's not a thing, no packaging standard guarantees that that will work, and I know of multiple reasons it cannot work reliably.

src/fromager/sources.py
```diff
@@ -38,6 +40,13 @@
 logger = logging.getLogger(__name__)
 
 
+class SDistBackend(enum.StrEnum):
+    # build sdist with PEP 517 hook
+    PEP517 = "pep571"
```
rgommers, line 45: 🙃 typo, but PEP 571 is the `manylinux2010` PEP so could be a deep feature too

src/fromager/sources.py
```diff
@@ -631,6 +640,11 @@ def build_sdist(
     return sdist_filename
 
 
+# meson dist currently only works with Git or Mercurial repos
+# affects NumPy and projects with https://mesonbuild.com/meson-python/
```
rgommers, line 644: If Meson doesn't work with an alternative VCS, it won't be used. I don't think Fromager should worry about this.

rgommers (commented):
Yes this looks like it assumes sdist -> sdist should work, and that's false. There are a lot of projects that for example rely on generating a file from VCS with versioning info. That should only ever be run once, roundtripping when starting from an sdist may lose the info or change a git commit hash to "unknown". 

I think this is probably going to introduce more problems than it solves. Maybe doing it only when there's a known issue with existing metadata in the sdist is better?

rgommers:
Other reason: a `build_sdist` call may do dependency detection of non-Python dependencies (shared libraries etc.), which may not be present outside of a release workflow.

src/fromager/sources.py
tiran: ouch :)

src/fromager/sources.py
tiran: I need to clarify this comment or remove `BUILD_BACKEND_FORCE_TARBALL` again. The `meson dist` command expects a VCS checkout. It does not support sdist -> sdist.

tiran:
> I think this is probably going to introduce more problems than it solves. Maybe doing it only when there's a known issue with existing metadata in the sdist is better?

Oh, that's too bad. Back to the drawing board!

I was able to build all the wheels in our set with the new code. I have not verified that the wheels were built correctly, though.

rgommers:
> I was able to build all the wheels in our set with the new code. I have not verified that the wheels were built correctly, though.

I'd expect most problems to be subtle, like metadata changing or a generated file going missing.

tiran:
Closing this PR because building sdist from sdist is not a supported case.

PR #579, opened 2025-04-08
improve error messages from resolver

dhellmann:
Add better logging and exceptions in the resolver code to convey more information about what is happening when no match is found.

Log debug messages for cases like ignoring all candidate files or having no candidates at all.

Use standard ResolverException error types and use messages specific to the different cases for searching for different types of files (sdists or wheels).

Move all of the logic for checking version compatibility into is_satisified_by() method of the base class for more consistent logging via different paths.


Xaenalt (approved):

rd4398 (approved):
Looks good to me!

PR #580, opened 2025-04-15
add hook entry point for bootstrap

dhellmann:
Add a hook entry point invocation for after we have bootstrapped a package. It is given the source distribution filename and the wheel filename, just like for the post-build hook, except that the sdist can be None if the wheel is pre-built and the wheel can be None if we are bootstrapping in sdist-only mode.

### DIFF
--- a/e2e/fromager_hooks/src/package_plugins/hooks.py
+++ b/e2e/fromager_hooks/src/package_plugins/hooks.py
@@ -24,6 +24,22 @@ def after_build_wheel(
     test_file.write_text(f"{dist_name}=={dist_version}")
 
 
+def after_bootstrap(
+    ctx: context.WorkContext,
+    req: Requirement,
+    dist_name: str,
+    dist_version: str,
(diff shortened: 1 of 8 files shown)


e2e/fromager_hooks/src/package_plugins/hooks.py
```diff
@@ -24,6 +24,22 @@ def after_build_wheel(
     test_file.write_text(f"{dist_name}=={dist_version}")
 
 
+def after_bootstrap(
+    ctx: context.WorkContext,
+    req: Requirement,
+    dist_name: str,
+    dist_version: str,
+    sdist_filename: pathlib.Path,
+    wheel_filename: pathlib.Path,
```
tiran, line 33: The function signature does not match the signature of the hook:

```suggestion
    sdist_filename: pathlib.Path | None,
    wheel_filename: pathlib.Path | None,
```

rd4398 (commented):
Looks good to me!
I will wait for @tiran to approve since he requested changes

Issue #494, opened 2024-11-04
Store ELF requirements in METADATA `Requires-External` fields


tiran:
Python core metadata has a multi-use field [`Requires-External`](https://packaging.python.org/en/latest/specifications/core-metadata/#requires-external-multiple-use) to specify external requirements. We can use the field to store external requirements such as ELF requires directly in the metadata file. This eliminates the need to maintain a separate file for Fromager ELF requirements.

As far as I can tell, the value of `Requires-External` fields are pretty much free-form. The field can contain environment markers after a semicolon, but we don't need it. The platform is implied by the wheel file name. We have to update the `Dynamic` field if-and-only-if it is present.

PR #581, opened 2025-04-16
Add setting for Requires-External metadata

tiran:
The new setting `project_override.requires_external` can be used to store dependencies on external packages. The main use case are cffi and ctypes bindings to shared libraries, e.g. `soundfile` depends on `libsndfile` system library.

The implementation does not modify `METADATA` file, yet. Fromager has no utilities to modify a `METADATA` file safely. Instead the information can be read from `fromager-build-settings` TOML file in the dist-info directory.

See: #494
See: https://packaging.python.org/en/latest/specifications/core-metadata/#requires-external-multiple-use

### DIFF
--- a/src/fromager/packagesettings.py
+++ b/src/fromager/packagesettings.py
@@ -207,6 +207,8 @@ class ProjectOverride(pydantic.BaseModel):
         - setuptools
       remove_build_requires:
         - ninja
+      requires_external:
+        - openssl-libs
     """
 
     model_config = MODEL_CONFIG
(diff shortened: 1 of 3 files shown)


src/fromager/packagesettings.py
```diff
@@ -219,6 +221,19 @@ class ProjectOverride(pydantic.BaseModel):
...
 
+    requires_external: list[str] = Field(default_factory=list)
+    """Add / update Requires-External core metadata field
+
+    Each entry contains a string describing some dependency in the system
+    that the distribution is to be used. See
+    https://packaging.python.org/en/latest/specifications/core-metadata/#requires-external-multiple-use
+
+    .. note::
+       Fromager does not modify ``METADATA`` file, yet. Read the information
```
dhellmann, line 232: What causes this new information to be written to the wheel?

dhellmann: ah! ok, I was expecting this to be a new separate file, but including it there makes sense

dhellmann (approved):

PR #583, opened 2025-04-16
improve wording of log message when filtering requirements

dhellmann:
It was not clear why a requirement was being ignored, so add more words to the message to try to make that clear.


rd4398 (approved):
Looks good to me!

PR #584, opened 2025-04-16
only delete build-env if it exists

dhellmann:
There are cases where we rebuild dependencies during bootstrap, so we need to ensure the cleanup code is robust in the face of repeats. This change ensures that we do not try to delete the build environment directory if we think we have one, but it has already been cleaned up.


rd4398 (approved):
LGTM!

Issue #553, opened 2025-02-24
Undefined variable in env vars no longer raises an error


tiran:
Fromager no longer raises an error when it encounters and undefined variable in an `env` vars block. Instead of failing it uses `None`. This hides typos in settings files.

Current behavior:
```python
>>> from fromager.packagesettings import substitute_template
>>> substitute_template("${DEFAULT:-default} ${UNKOWN}", {})
'default None'
```

Expected behavior:
Fromager should fail with `KeyError: 'UNKNOWN'`

PR #585, opened 2025-04-17
fix: handle undefined variable in substitute_template

vshawrh:
Summary
This PR addresses incorrect behavior in the substitute_template function where missing variables without defaults were silently replaced with None, instead of raising a KeyError as expected.
Fixes: https://github.com/python-wheel-build/fromager/issues/553

Changes :
Fixed src/fromager/packagesettings.py:substitute_template logic to:

Only apply default values when explicitly specified using the :-default syntax.

Leave variables without defaults unresolved, allowing string.Template.substitute to raise a KeyError for unknown variables.

Added a test case to ensure a KeyError is raised when ${UNKNOWN} is used without a default and the variable is not in template_env.
 Tests:
Added tests/test_packagesettings.py::test_substitute_template_key_error() to validate the raised exception behavior.

Ensured all existing parameterized tests still pass.
Screenshot From 2025-04-17 16-56-51

Screenshot From 2025-04-17 16-57-23

### DIFF
--- a/src/fromager/packagesettings.py
+++ b/src/fromager/packagesettings.py
@@ -457,11 +457,20 @@ def substitute_template(value: str, template_env: dict[str, str]) -> str:
     for mo in _DEFAULT_PATTERN_RE.finditer(value):
         modict = mo.groupdict()
         name = modict["name"]
-        # add to local default, keep existing default
-        localdefault.setdefault(name, modict["default"])
-        # remove ":-default"
-        value = value.replace(mo.group(0), f"${{{name}}}")
-    return string.Template(value).substitute(localdefault)
+        default = modict["default"]
+        # Only set the default if one is explicitly provided.
+        # This ensures that undefined variables without defaults
+        # will raise KeyError later
+        if default is not None:
+            localdefault.setdefault(name, default)
+            # Replace ${var:-default} with ${var}
+            value = value.replace(mo.group(0), f"${{{name}}}")
+    try:
+        return string.Template(value).substitute(localdefault)
+    except KeyError as e:
+        raise ValueError(
+            f"Undefined environment variable {e!r} referenced in expression {value!r}"
+        ) from e
 
 
 def get_cpu_count() -> int:
--- a/tests/test_packagesettings.py
+++ b/tests/test_packagesettings.py
@@ -468,3 +468,14 @@ def test_parallel_jobs(
 )
 def test_substitute_template(value: str, template_env: dict[str, str], expected: str):
     assert substitute_template(value, template_env) == expected
+
+
+def test_substitute_template_key_error():
+    # This test expects a ValueError to be raised by substitute_template
+    with pytest.raises(ValueError) as excinfo:
+        substitute_template("${DEFAULT:-default} ${UNKNOWN}", {})
+    # Verify that the error message matches the expected message
+    assert (
+        str(excinfo.value)
+        == "Undefined environment variable KeyError('UNKNOWN') referenced in expression '${DEFAULT} ${UNKNOWN}'"
+    )


vshawrh:
@dhellmann @tiran @rd4398 
Can You please review,
I don't have the access to add reviewers.

vshawrh:
> Could you give the PR a more descriptive title, please?
Hi @dhellmann ,
Actually I did a description of title, but it overflowed.
Please review the title and recommend any changes required,

Thanks & Regards,
Vikash Shaw

dhellmann:
> > Could you give the PR a more descriptive title, please?
> 
> Hi @dhellmann , Actually I did a description of title, but it overflowed. Please review the title and recommend any changes required,
> 
> Thanks & Regards, Vikash Shaw

That's better. How about something a bit simpler, like "fix: handle undefined variable in substitute_template"? Linking to the issue number in the body of the description is good enough, and we can assume you added a test case.

src/fromager/packagesettings.py
```diff
@@ -457,10 +457,11 @@ def substitute_template(value: str, template_env: dict[str, str]) -> str:
     for mo in _DEFAULT_PATTERN_RE.finditer(value):
         modict = mo.groupdict()
         name = modict["name"]
-        # add to local default, keep existing default
-        localdefault.setdefault(name, modict["default"])
-        # remove ":-default"
-        value = value.replace(mo.group(0), f"${{{name}}}")
+        default = modict["default"]
+        if default is not None:
```
dhellmann, line 461: This is a bit subtle, and I think could benefit from a comment.

I think the way this is working is that it now only adds the `name` key to `localdefault` if there is actually a value for the variable. And so then later if something tries to look up the name, and there was no default value, a KeyError is generated. Is that correct?

I could imagine myself coming along in 6 months, not remembering that, and saying "huh, this is going to lead to KeyError exceptions, let's always save the default...". A comment here explaining why it is this way would prevent that.

Where does `modict` come from? Is there always a `default` key? Or do we need to be more careful about accessing that on line 460?

tests/test_packagesettings.py
```diff
@@ -468,3 +468,9 @@ def test_parallel_jobs(
 )
 def test_substitute_template(value: str, template_env: dict[str, str], expected: str):
     assert substitute_template(value, template_env) == expected
+
+
+def test_substitute_template_key_error():
+    with pytest.raises(KeyError) as excinfo:
+        substitute_template("${DEFAULT:-default} ${UNKNOWN}", {})
+    assert str(excinfo.value) == "'UNKNOWN'"
```
dhellmann, line 476: I would like to make sure we have a more helpful error message for the user than what a KeyError provides by default. Could you update `substitute_template` to catch the KeyError, and raise its own exception that says something like "Undefined environment variable {name} referenced by {expression}"?

src/fromager/packagesettings.py
vshawrh: Yes, its correct so we intentionally do not add the variable to localdefault unless a default exists.

modict comes after the regex match object mo after is groupdict()

The default key is always present, but its value is None if no default was matched.: Please refer https://docs.python.org/3/library/re.html#re.Match.groupdict
Screenshot From 2025-04-18 12-39-59

tests/test_packagesettings.py
vshawrh: Hi
Have Incorporated the review comments 
Please review and suggest any changes as required

src/fromager/packagesettings.py
```diff
@@ -457,11 +457,20 @@ def substitute_template(value: str, template_env: dict[str, str]) -> str:
...
+        # will raise KeyError later
+        if default is not None:
+            localdefault.setdefault(name, default)
+            # Replace ${var:-default} with ${var}
+            value = value.replace(mo.group(0), f"${{{name}}}")
+    try:
+        return string.Template(value).substitute(localdefault)
+    except KeyError as e:
+        raise ValueError(
+            f"Undefined environment variable '{e}' referenced in expression '{value}'"
```
dhellmann, line 472:
```suggestion
            f"Undefined environment variable {e!r} referenced in expression {value!r}"
```

Using `!r` will let the python string formatting code chose the right type of quotes to apply. That way if an expression includes quotes, there is a chance for the string formatter to choose the other type of quotes, and produce a message that is more readable.

src/fromager/packagesettings.py
dhellmann: OK, good, thank you for confirming.

dhellmann (approved):

PR #586, opened 2025-04-18
Bootstrap check local dir before cache server

### LINKED ISSUE #578: bootstrap should look for wheel locally before checking the cache server
When fromager plugins are being developed it is common to need to work with a few different wheels that are not available on a cache index, yet. In those cases, the developer workflow would be improved if the [code that looks for a cached wheel](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/bootstrapper.py#L381) would look at the local package index before looking at the upstream cache index. That would let a developer build an expensive wheel one time, then reuse it when working on another package.

dhellmann:
This PR includes a little bit of refactoring of existing code and then it adds the behavior of looking in the wheel downloads and wheel builds directories for a wheel before checking the remote cache server.

Fixes #578

### DIFF
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -129,7 +129,7 @@ class Bootstrapper:
             # Remember that this is a prebuilt wheel, and where we got it.
             source_url_type = str(SourceType.PREBUILT)
         else:
-            unpacked_cached_wheel, cached_wheel_filename = (
+            cached_wheel_filename, unpacked_cached_wheel = (
                 self._download_wheel_from_cache(req, resolved_version)
             )
             source_url_type = sources.get_source_type(self.ctx, req)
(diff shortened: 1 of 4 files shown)


dhellmann:
I've marked this as a draft because I want to make sure we get a careful review of the logic changes by both @tiran and @rd4398 before we merge it.

src/fromager/bootstrapper.py
```diff
@@ -432,40 +499,54 @@ def _download_wheel_from_cache(
...
+            archive = zipfile.ZipFile(wheel_filename)
+            for filename in [
+                dependencies.BUILD_BACKEND_REQ_FILE_NAME,
+                dependencies.BUILD_SDIST_REQ_FILE_NAME,
+                dependencies.BUILD_SYSTEM_REQ_FILE_NAME,
+            ]:
+                zipinfo = archive.getinfo(
+                    str(metadata_dir / f"{wheels.FROMAGER_BUILD_REQ_PREFIX}-{filename}")
+                )
+                zipinfo.filename = filename
```
tiran, line 534: Why is the code assigning to `zipinfo.filename` here?

dhellmann: It was doing that before, I just moved the code around. I'm not sure why it would have needed to do that.

dhellmann: I think I see what it's doing. Inside the zip file the metadata files have a prefix to ensure they are unique and marked as coming from fromager. When they are unpacked, that prefix is being removed so the files match the names used when fromager generates similar files during a bootstrap run and to unpack them without the metadata dir in the path. 

We could change that, but I'd like to do it in a separate PR because it will be a bit invasive.

dhellmann:
This is ready for formal review. I'd like to get both this and #594 into our next release.

tiran:
@dhellmann The PR has old-style logging with `{req.name}`.

tiran (approved):

PR #588, opened 2025-04-24
Log requirement name for download and patching

tiran:
Source download, wheel download, and patching did not prefix the log message with the requirement name.


dhellmann (approved):
I had really wanted to have some sort of log adapter to manage this, but I haven't found the time to figure out how to implement that, so let's go ahead with this.

tiran:
> I had really wanted to have some sort of log adapter to manage this, but I haven't found the time to figure out how to implement that, so let's go ahead with this.

Yeah, same here. I was looking into logging adapter and contextvars. The logging framework is surprisingly inflexible when it comes to rewriting log messages.

PR #589, opened 2025-04-25
Implement requirement logging adapter

tiran:
The new `LogRequirementAdapter` and `fromager.log.get_logger()` helper use a Python `contextvar` to inject the current requirement name into a log message. Context vars are a bit like TLS (thread local storage) vars, but with asyncio and nesting support.

The requirement context variable is set whenenver a new requirement is encountered and reset when the control flow leaves the scope of that requirement. Reset ensures that recursive requirement handling is handled correctly.

See: https://docs.python.org/3/howto/logging-cookbook.html#context-info
See: https://docs.python.org/3/library/contextvars.html

### DIFF
--- a/e2e/fromager_hooks/src/package_plugins/hooks.py
+++ b/e2e/fromager_hooks/src/package_plugins/hooks.py
@@ -17,10 +17,10 @@ def after_build_wheel(
     wheel_filename: pathlib.Path,
 ):
     logger.info(
-        f"{req.name}: running post build hook in {__name__} for {sdist_filename} and {wheel_filename}"
+        f"running post build hook in {__name__} for {sdist_filename} and {wheel_filename}"
     )
     test_file = sdist_filename.parent / "test-output-file.txt"
-    logger.info(f"{req.name}: post-build hook writing to {test_file}")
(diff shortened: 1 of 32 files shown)


e2e/fromager_hooks/src/package_plugins/hooks.py
dhellmann, line 1: This file does not have the updated logging import. Do we need that _everywhere_?

dhellmann (commented):
I wonder if a log record factory would require fewer global changes? https://docs.python.org/3/library/logging.html#logging.setLogRecordFactory

tiran:
> I wonder if a log record factory would require fewer global changes? https://docs.python.org/3/library/logging.html#logging.setLogRecordFactory

It would also change the behavior of logging in 3rd party packages and stdlib.

dhellmann:
> > I wonder if a log record factory would require fewer global changes? https://docs.python.org/3/library/logging.html#logging.setLogRecordFactory
> 
> It would also change the behavior of logging in 3rd party packages and stdlib.

That seems like a feature, though, right? All logging within a context would be prefixed by the package name.

PR #590, opened 2025-04-28
Call PEP 517 from build environment

tiran:
Fromanger no longer writes to its own virtual env. All PEP 517 hooks like build system requirement now use the package's build environment instead.

- the build environment is now created earlier in the bootstrap / build process of a package.
- `BuildEnvironment` now has a `install()` helper method to install a set of packages into the environment.
- all PEP 517 helper methods now require a `build_env` argument.


dhellmann:
I have been able to reproduce the failure locally (I had to disable my local wheel-house settings in pip.conf).

I think this only ever worked at all because fromager happens to depend on stevedore which depends on pbr, so before your changes pbr was already installed into the fromager virtualenv where we run the command to build the metadata from the source checkout. That isn't going to carry forward for future packages, necessarily.

I'm still looking into ways to deal with it automatically, but we could always just add a configuration option that would only be used in the mode where we are cloning a git repo to let the user tell us the packages that need to be installed in order to generate the metadata.

dhellmann:
OK, I think there's a fundamental flaw in the logic for dealing with URLs in requirements.

As I mentioned before, this only worked because pbr was installed with fromager and we were using that version.

The issue is that version resolution is decoupled from bootstrapping in this case. The bootstrapper calls into `resolve_source()` which calls `resolve_version_from_git_url()` which calls `_get_version_from_package_metadata()` which needs to be able to ensure the dependencies it tries to install are built.

I think what we want to do, separately from this PR, is EITHER move handling for git URLs into the bootstrapper itself, and not pass them through to `resolve_source()` OR we want to pass the bootstrapper into `resolve_source()` so it can be used recursively from there. The former approach would limit git URLs to the bootstrap command, I think, which would break the ability to have nightly builds created by `build-sequence`. The latter approach would mean `build-sequence` would need to instantiate a bootstrapper, too. It also introduces a circular dependency between the 2 modules.

I will spend more time looking at those approaches to see which is more viable.

dhellmann:
I rebased this in #634, so I think we want that version instead of this one.

src/fromager/bootstrapper.py
```diff
@@ -351,17 +351,18 @@ def _build_sdist(
                 )
         except Exception as err:
             logger.warning(f"failed to build source distribution: {err}")
-        return (sdist_filename, build_env)
+            raise
```
dhellmann, line 354: Have you tried this code on our downstream builds? I wonder if this change in error handling is going to cause any issues?

dhellmann (commented):
I wonder if we want the BuildEnvironment to be accessible via the context, instead of having to instantiate it everywhere and pass it around. We could make that change in a follow-up.

I have one other question inline about an error handling change that I'd like to talk about before approving.

src/fromager/bootstrapper.py
tiran: No, I have not tested this with our downstream builds. I can undo the change and create a ticket to investigate.

dhellmann (approved):

dhellmann:
> > I wonder if we want the BuildEnvironment to be accessible via the context, instead of having to instantiate it everywhere and pass it around. We could make that change in a follow-up.
> 
> A long time ago, we discussed the idea to combine `req`, `resolved_version`, and other flags into a package state object. The `build_env` could be an attribute of the package build state, too.

That does make sense, too, yes.

PR #591, opened 2025-04-28
Implement requirement logging factory

tiran:
The new `FromagerLogRecord` replaces logging's default LogRecord class. It uses a Python `contextvar` to inject the current requirement name into a log message. Context vars are a bit like TLS (thread local storage) vars, but with asyncio and nesting support.

The requirement context variable is set whenenver a new requirement is encountered and reset when the control flow leaves the scope of that requirement. Reset ensures that recursive requirement handling is handled correctly.

See: https://docs.python.org/3/library/contextvars.html

### DIFF
--- a/e2e/fromager_hooks/src/package_plugins/hooks.py
+++ b/e2e/fromager_hooks/src/package_plugins/hooks.py
@@ -17,10 +17,10 @@ def after_build_wheel(
     wheel_filename: pathlib.Path,
 ):
     logger.info(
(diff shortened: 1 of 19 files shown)


tiran:
Alternative implementation of #589, but with a custom LogRecord

tiran:
Okay, let's land this version. I have created a separate PR so we have the other version in the project history. It gives us the option to switch to the other approach in the future.

dhellmann (approved):

PR #592, opened 2025-05-01
add time to log messages

dhellmann:
In normal mode show the hour, minutes, and seconds. This lets someone get a sense of how long a step has been running, which is useful for some packages that take a long time to build.

In verbose mode, and in log files, show the full datetime. This is useful for logs that are saved and examined later.


rd4398 (approved):
Looks good to me!

PR #593, opened 2025-05-01
Add check-requirements-resolve command

paradigm:
This command checks that the top-level entries in the provided requirements.txt and/or constraints.txt files resolve.

Usage:
   fromager check-requirements-resolve "path/to/constraints.txt" "path/to/requirements.txt"


src/fromager/commands/check_requirements_resolve.py
```diff
@@ -0,0 +1,59 @@
...
+
+    failure = False
+    for path in input_files_path:
+        for line in requirements_file.parse_requirements_file(path):
+            try:
+                req = Requirement(line)
+            except InvalidRequirement as err:
+                logger.error(f"{path}: invalid requirement {line!r}: {err}")
+                failure = True
+                continue
```
dhellmann, line 37: The bootstrap command has some [code to get the requirements](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/commands/bootstrap.py#L108), including filtering out requirements that are not relevant based on the current environment markers. How likely is it that we would need to do similar filtering in the linter?

paradigm: Our main aim with the lint command is to do an early and inexpensive check of what will be repeated behavior in the bootstrap command.  Given that, so long as it's inexpensive, it's probably best to keep the behavior identical in this regard.  I'll adjust it accordingly.

dhellmann:
We talked about this offline, and I think you said you were going to merge the feature into the existing linter. Am I remembering that correctly, or should we merge this PR?

dhellmann:
We added this to the lint-requirements command in #781

PR #594, opened 2025-05-02
build our own simple wheel index

dhellmann:
Instead of using pypi-index, create our own index by just symlinking to the relevant wheel files in the right directories. This is significantly faster, since it does not require generating the new index files.


tiran:
A while ago I started an experimental branch that replaces `python-pypi-mirror` and `http.server` with [pypiserver](https://pypi.org/project/pypiserver/). The pypiserver caches checksums and directories, and auto-refreshes with `watchdog`. I can clean up, rebase, and push the branch on Monday.

dhellmann:
> A while ago I started an experimental branch that replaces `python-pypi-mirror` and `http.server` with [pypiserver](https://pypi.org/project/pypiserver/). The pypiserver caches checksums and directories, and auto-refreshes with `watchdog`. I can clean up, rebase, and push the branch on Monday.

That would be good. I think we're spending a _lot_ of time in this step as we bootstrap larger collections.

tiran:
I noticed the problem with `python-pypi-mirror`, too. It does not cache SHA-256 checksums and it is very inefficient (loads the entire file into memory). It takes a while to hash 3 GB of data.

dhellmann:
> I noticed the problem with `python-pypi-mirror`, too. It does not cache SHA-256 checksums and it is very inefficient (loads the entire file into memory). It takes a while to hash 3 GB of data.

It would be nice if we could avoid creating hashes at all. We don't really need them for the cache server case.

dhellmann:
I wonder if we could use pip's `--find-links` and `--no-index` options together for the local cache. We would have to update the resolver in fromager to be able to deal with that, too.

dhellmann:
@tiran with this change in place, bootstrapping our full RHEL AI CUDA build (graph with 1468 edges) with existing wheels in my cache took just over 4 minutes.

src/fromager/server.py
```diff
@@ -61,17 +60,22 @@ def serve_forever(server: http.server.ThreadingHTTPServer) -> None:
...
+        # Always move the file so the code managing the timer for the
+        # wheels does not find more than one wheel in the build
+        # directory.
+        shutil.move(wheel, downloads_dest_filename)
+
+    for wheel in ctx.wheels_downloads.glob("*.whl"):
+        # Now also symlink the files into the simple hierarchy. We always
+        # process all files to be safe.
+        (normalized_name, _, _, _) = parse_wheel_filename(wheel.name)
+        simple_dest_filename = ctx.wheel_server_dir / normalized_name / wheel.name
```
tiran, line 75: The normalized name wheel has underscores, the canonical PyPI name has dashes, e.g. `scikit-build-core/scikit_build_core-0.11.2-4-py3-none-any.whl`

tiran: Please ignore my comment:

```python
>>> parse_wheel_filename("scikit_build_core-0.11.2-4-py3-none-any.whl")
('scikit-build-core', <Version('0.11.2')>, (4, ''), frozenset({<py3-none-any @ 140605344268096>}))
```

tiran (approved):
Nice, this is much simpler and faster!

PR #595, opened 2025-05-03
fix use of log context in bootstrapper

dhellmann:
The original implementation managed the context only for some types of dependencies. This change moves the handling into the bootstrap() method where it can be applied to all dependencies.


src/fromager/bootstrapper.py
```diff
@@ -360,11 +367,9 @@ def _handle_build_requirements(
         self.progressbar.update_total(len(build_dependencies))
 
         for dep in self._sort_requirements(build_dependencies):
-            token = requirement_ctxvar.set(dep)
```
tiran, line 363: With this change `build_environment.maybe_install()` has no longer the correct context var. #598 should fix the bug.

dhellmann: I'll have to have a look at that. The `bootstrap()` function should be the top level function, so handling the context there should cover everything.

dhellmann:
We no longer need this.

PR #596, opened 2025-05-03
Use pypiserver to server wheels

tiran:
The `python-pypi-mirror` project does not scale for us in downstream. The script re-generates all `index.html` files and re-calculates SHA-256 digests of all files. It's rather slow with several GB of wheels.

The [pypiserver](https://pypi.org/project/pypiserver/) project is more efficient and has some benefits. It caches some data and supports disabling hashing. The HTTP server is also more clever. It automatically redirects to normalized package name.


dhellmann:
I have #594 working and producing some surprising performance improvements. What is the benefit of using pypiserver for our cache use case?

tiran:
Your solution is simpler and works.

`pypiserver` performs filename normalization and redirects. We could implement the feature in top of the simple HTTP server if we ever need it.

PR #597, opened 2025-05-03
Don't compile bytecode of dependencies

tiran:
By default, `pip install` also compiles byte code of all Python files. The step takes a while, especially for large packages like PyTorch. Disable byte code compilation to speed up installation of build dependencies. Byte code is now compiled on demand.

### DIFF
--- a/src/fromager/build_environment.py
+++ b/src/fromager/build_environment.py
@@ -198,6 +198,7 @@ class BuildEnvironment:
                 "pip",
                 "install",
                 "--disable-pip-version-check",
+                "--no-compile",  # don't compile byte code
                 "--only-binary",
                 ":all:",
             ]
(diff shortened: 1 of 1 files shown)


dhellmann:
I wouldn't expect pip to be recompiling to bytecode when we install from wheels, and we should always be able to install from wheels. How much difference are you seeing in the times?

tiran:
Wheels don't contain pyc files. `pip install` compiles all Python files to byte code after installation.

`python3.11 -m compileall` takes about 10 to 12 seconds for a large virtual env with Torch. Smaller virtual envs take 2 to 4 seconds.  I expect that we are going to safe a second to a few seconds for each build env. It's not much, but it adds up.

dhellmann (approved):

PR #601, opened 2025-05-05
switch from virtualenv to venv

dhellmann:
virtualenv 20.31.0 dropped the bundled wheel and broke backwards compatibility by removing the `--wheel` command line option. There's no particular reason for us to use virtualenv, so switch to the standard library module instead.

### DIFF
--- a/e2e/constraints.txt
+++ b/e2e/constraints.txt
@@ -1,2 +1,3 @@
 # This file is here in case we need to quickly add a constraint to
 # fix CI jobs.
+setuptools<80.3.0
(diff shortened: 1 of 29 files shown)


tiran:
The virtualenv package bundles more recent versions of setuptools and pip. The `venv` package is stuck to whatever Python provides with `ensurepip`. These versions have known bugs, too. For example versions in Python 3.11 and AFAIK 3.12 are affected by the extra name normalization issue that caused missing `hf-transfer` in downstream a while ago.

dhellmann:
I'm inclined to try this again when we have more time, but not right now.

PR #602, opened 2025-05-05
fix issue with extracting metadata

dhellmann:
In #586, I broke the way the metadata files are extracted by changing the output directory handling. This change restores the original logic.


dhellmann:
This depends on #600 to cap virtualenv.

tiran (approved):

PR #608, opened 2025-05-19
do not use bundled setuptools in virtualenvs

dhellmann:
Remove the flag that causes us to use a bundled version of setuptools in the build virtualenv. That version is too old to build some packages, especially in the face of recent breaking changes that have been worked around in things that use setuptools.

Instead of using the bundled version, rely on the fact that in src/fromager/dependencies.py we set the default build backend requirements to include setuptools if there is no valid information in the pyproject.toml. This should ensure that we detect setuptools as a dependency of projects that are likely to use it because it is the default backend.

### DIFF
--- a/src/fromager/build_environment.py
+++ b/src/fromager/build_environment.py
@@ -163,8 +163,6 @@ class BuildEnvironment:
             return
 
         logger.debug("creating build environment in %s", self.path)
-        # Python 3.12 virtual envs don't have wheel and setuptools by
-        # default. Some packages still assume they are installed.
         external_commands.run(
             [
                 sys.executable,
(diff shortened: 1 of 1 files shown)


src/fromager/build_environment.py
```diff
@@ -173,7 +171,6 @@ def _createenv(self) -> None:
                 "--python",
                 sys.executable,
                 "--pip=bundle",
-                "--setuptools=bundle",
```
tiran, line 176: This does not remove the bundled setuptools package with Python 3.11. You have to change the line to `--setuptools=none` to not install the bundled setuptools. Virtualenv has different defaults for Python 3.11 and 3.12, because Python 3.12's `venv` and `ensurepip` no longer install it either.

```
$ python3.11 -m virtualenv --help | grep -- --setuptools
...
  --setuptools version          version of setuptools to install as seed: embed, bundle, none or exact version (default: bundle)

$ python3.12 -m virtualenv --help | grep -- --setuptools
  --setuptools version          version of setuptools to install as seed: embed, bundle, none or exact version (default: none)
```

dhellmann: Thanks, that's fixed

tiran (approved):
LGTM

PR #614, opened 2025-05-26
fix: address resource leaks and improve security

EmilienM:
Fix multiple issues in file handling and archive extraction:

* Fix resource leaks in `sources.py` by using context managers for
  ZipFile objects instead of manual cleanup
* Improve `tar.gz` file detection by adding explicit check for
  `.tar.gz` extension alongside existing `.tgz` and `.gz` checks
* Standardize exception types for consistency:
  - Use `tarfile.TarError` for tar file errors
  - Use ValueError for unsupported file types (`TypeError` was used while
    it wasn't an actual Type mismatch).
* Fix resource leak in `read.py` by replacing try/finally with
  context manager for file handling
* Add path traversal protection in wheel extraction:
  - Validate file paths in zipfile extraction to prevent
    directory traversal attacks in `wheels.py` and `bootstrapper.py`
  - Reject absolute paths and paths containing `..`


EmilienM:
@dhellmann I super small PR with quick improvements I thought when reading code here.

PR #615, opened 2025-05-26
✨ feat: add `--skip-constraints` option to bootstrap command

### LINKED ISSUE #605: option in bootstrap to bypass creating constraints.txt
There are some situations where we may want to build a collection of wheels together without requiring that the collection resolve to a single installable set of packages. Some examples are in test jobs for fromager settings for building otherwise unrelated packages, and in jobs that build "large" collections for a broader package index.

We can enable this feature by providing an option to the bootstrap step to skip producing a constraints.txt file. The graph and build-sequence files can already cope with multiple conflicting versions. We should verify that the build-sequence command has no issues with building duplicates and does not verify an installable set.

EmilienM:
Add a new command-line flag to the bootstrap command that allows
building collections of wheels with conflicting versions by skipping
the generation of `constraints.txt` file.

This enables use cases such as:
- Building large collections for broader package indexes
- Testing jobs that need to build multiple conflicting versions
- Creating wheel collections that don't need to resolve to a single
  installable set of packages

Changes:
- Add `--skip-constraints` CLI option to bootstrap command and skip
  `constraints.txt` generation when used.
- Add unit tests for the new CLI option
- Add new end-to-end test verifying functionality with conflicting packages

The graph and build-sequence files can already handle multiple
conflicting versions, so this change simply allows bypassing the
constraints validation step.

Fixes #605


docs/example/skip-constraints-example.md
```diff
@@ -0,0 +1,58 @@
...
+Suppose you need to build a package index that contains multiple versions of the same package for different downstream consumers. For example, you might want to include both Django 3.2 and Django 4.0 in your collection.
+
+## Requirements Files
+
+Create a requirements file with conflicting versions:
+
+### requirements-conflicting.txt
+
+```text
+django==3.2.0
```
EmilienM, line 16: I documented an example with Django, but I'm happy to do it for another package if more appropriate.

PR #621, opened 2025-05-30
require review from maintainer

dhellmann:
We have many members of the org who are not active maintainers of fromager. This change limits the auto-merge rule for mergify to look for reviews with approvals from maintainers.


rd4398 (approved):
Looks good to me!

PR #623, opened 2025-05-31
Detect requirement conflict

dhellmann:
Fix an issue with the `graph to-constraints` command and constraints.txt
generation in `bootstrap` where we were ignoring pinned versions in
top level dependencies if that same package was also a dependency of
something else. If we have conflicting versions in that case there is no
way to resolve them and we should report an error.

Co-authored-by: Claude 3.7 Sonnet <claude-3-7-sonnet@anthropic.com>
Signed-off-by: Doug Hellmann <dhellmann@redhat.com>

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -94,6 +94,7 @@ jobs:
           - build_order
           - build_steps
           - build_settings
+          - graph_to_constraints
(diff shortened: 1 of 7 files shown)


rd4398 (commented):
Looks good to me!

PR #627, opened 2025-06-04
Start using hatch instead of tox

LalatenduMohanty:
The PR is built upon #626 , so #626 needs to merge first.


dhellmann (approved):
This looks good. I may propose some tweaks as a follow-up, and I'll ask you for reviews. Thank you!

PR #628, opened 2025-06-05
Fix: Pass config settings to hooks

tiran:
Pass `config_settings` to all hooks. This fixes sdist-only mode for `matplotlib`. The settings are now passed to
`prepare_metadata_for_build_wheel` and meson no longer attempts to download freetype when getting build metadata.

This change is backwards incompatible. `config_settings` is now a mapping of strings to list of strings. Before it was just a list of strings.

I'm including my `e2e/matplotlib_settings` files for future testing.

### DIFF
--- a/e2e/test_bootstrap_sdist_only.sh
+++ b/e2e/test_bootstrap_sdist_only.sh
@@ -10,6 +10,9 @@
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 source "$SCRIPTDIR/common.sh"
 
+DIST=stevedore
+VER=5.2.0
+
 # passing settings to bootstrap but should have 0 effect on it
 fromager \
   --log-file="$OUTDIR/bootstrap.log" \
@@ -18,7 +21,7 @@ fromager \
   --wheels-repo="$OUTDIR/wheels-repo" \
   --work-dir="$OUTDIR/work-dir" \
   --settings-dir="$SCRIPTDIR/changelog_settings" \
-  bootstrap --sdist-only 'stevedore==5.2.0'
+  bootstrap --sdist-only "$DIST==$VER"
 
 find "$OUTDIR/wheels-repo/" -name '*.whl'
 find "$OUTDIR/sdists-repo/" -name '*.tar.gz'
--- a/src/fromager/dependencies.py
+++ b/src/fromager/dependencies.py
@@ -138,13 +138,16 @@ def default_get_build_backend_dependencies(
     Defaults to result of hook call
     :meth:`~pyproject_hooks.BuildBackendHookCaller.get_requires_for_build_wheel`
     """
+    pbi = ctx.package_build_info(req)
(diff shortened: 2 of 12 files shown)


e2e/test_bootstrap_sdist_only.sh
```diff
@@ -10,6 +10,9 @@
 SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 source "$SCRIPTDIR/common.sh"
 
+DIST=stevedore
```
dhellmann, line 13: Should this be matplotlib?

src/fromager/dependencies.py
```diff
@@ -138,13 +138,16 @@ def default_get_build_backend_dependencies(
...
+    pbi = ctx.package_build_info(req)
     hook_caller = get_build_backend_hook_caller(
         ctx=ctx,
         req=req,
         build_dir=build_dir,
         override_environ=extra_environ,
     )
-    return hook_caller.get_requires_for_build_wheel()
+    return hook_caller.get_requires_for_build_wheel(
+        config_settings=pbi.config_settings,
```
dhellmann, line 149: If we're going to always pass it, maybe we can wrap the hook_caller method with a partial? Maybe that's messier internally, though.

src/fromager/packagesettings.py
```diff
@@ -310,10 +310,17 @@ class PackageSettings(pydantic.BaseModel):
...
-    config_settings: list[str] = Field(default_factory=list)
+    config_settings: dict[str, list[str]] = Field(default_factory=dict)
     """PEP 517 arbitrary configuration for wheel builds
 
     https://peps.python.org/pep-0517/#config-settings
+
+    ::
+
+       config_settings:
+         setup-args:
```
dhellmann, line 321: Do we need both nesting levels? Are there other `config_settings`?

e2e/test_bootstrap_sdist_only.sh
dhellmann: Whew, ok, I don't think we want that. I was looking for a way for the test here to show that the settings are passed through, but maybe that's not needed in an e2e test, since you have the unit tests.

src/fromager/dependencies.py
dhellmann: Sure. I was thinking that the pattern we have is we instantiate the hook caller using our wrapper, and that constructor has all of the args it needs to create the pbi and get the config_settings, but maybe a little repetition isn't so bad for clarity.

src/fromager/packagesettings.py
dhellmann: OK, makes sense.

dhellmann (commented):
This looks good. I will watch for you to remove the draft flag before approving, though, in case you have other changes coming.

tiran:
I have removed the draft flag.

rd4398 (commented):
This looks good to me! I will wait for @dhellmann for final approval since he provided feedback earlier

dhellmann (approved):

PR #633, opened 2025-06-13
Fix Git URL handling for bootstrap from URL requirements

dhellmann:
# Fix Git URL handling for bootstrap from URL requirements

This pull request addresses issue #632 by improving how Fromager handles requirements specified with Git URLs during the bootstrap process. The changes consolidate Git URL processing logic in the bootstrapper and ensure build dependencies are properly resolved before extracting version metadata.

### Commits

- Move git url handling to bootstrapper  
  Moves source resolution for requirements with git URLs into the bootstrapper, consolidating logic that was previously split between the sources module and bootstrapper. This ensures build dependencies are available before processing metadata.

- Move version resolution to method in bootstrapper  
  Extracts version resolution logic into a reusable method in the bootstrapper to eliminate code duplication between bootstrap command and bootstrapper class.

- Process build dependencies before getting version from metadata  
  Ensures build dependencies are installed before attempting to access package metadata for version extraction when bootstrapping Git URL requirements.

## Summary

The key improvements include:

1. **Consolidated Git URL handling**: Moved all Git URL processing logic from `sources.py` to `bootstrapper.py` for better organization and dependency management
2. **Improved version resolution**: Created a reusable `resolve_version()` method to eliminate code duplication
3. **Fixed build dependency ordering**: Ensures build dependencies are processed before extracting version metadata from Git repositories
4. **Updated test configuration**: Simplified test scripts by using a dedicated test repository and removing verbose logging

### Files Modified

- `src/fromager/bootstrapper.py` - Added Git URL handling and version resolution methods
- `src/fromager/sources.py` - Removed Git URL processing logic (moved to bootstrapper)
- `src/fromager/commands/bootstrap.py` - Updated to use new version resolution method
- `e2e/test_bootstrap_git_url.sh` - Updated test repository URL and logging flags
- `e2e/test_bootstrap_git_url_tag.sh` - Updated test repository URL and logging flags

This refactoring improves the reliability of bootstrapping packages from Git URLs by ensuring proper dependency resolution order and consolidating related functionality.

PR #634, opened 2025-06-14
Hook build env

dhellmann:
This is a rebased  version of #590 to test in combination with #633

### DIFF
--- a/src/fromager/build_environment.py
+++ b/src/fromager/build_environment.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import importlib.metadata
 import logging
 import os
 import pathlib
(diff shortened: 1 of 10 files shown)


src/fromager/build_environment.py
```diff
@@ -179,37 +178,44 @@ def _createenv(self) -> None:
             network_isolation=self._ctx.network_isolation,
         )
         logger.info("created build environment in %s", self.path)
+        if self._build_requirements:
+            self.install(self._build_requirements)
 
+    @property
+    def _pip_cmd(self) -> list[str]:
```
rd4398, line 185: Can we be more specific about the name like `_pip_install_cmd`  or am I overthinking this?

rd4398 (commented):
Overall looks good! Just a small suggestion.

I will leave it to @tiran  and / or @EmilienM for final approval

src/fromager/build_environment.py
dhellmann: That's reasonable feedback. 

@tiran I can make that edit in my branch, if you're OK with it?

PR #639, opened 2025-06-20
Fix git submodules support

EmilienM:
Fix for the latest PR that merged too early.


src/fromager/sources.py
```diff
@@ -661,3 +664,193 @@ def ensure_pkg_info(
             )
             had_pkg_info = False
     return had_pkg_info
+
+
+def resolve_version_from_git_url(ctx: context.WorkContext, req: Requirement) -> tuple[str, Version]:
```
dhellmann, line 669: This function moved [into the Bootstrapper](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/bootstrapper.py#L636) so we don't want this copy any more.

PR #640, opened 2025-06-20
fix: make build-sequence work with requirements with git URLs

### LINKED ISSUE #632: OK, I think there's a fundamental flaw in the logic for dealing with URLs in requirements.
OK, I think there's a fundamental flaw in the logic for dealing with URLs in requirements.

As I mentioned before, this only worked because pbr was installed with fromager and we were using that copy.

The issue is that version resolution is decoupled from bootstrapping in this case. The bootstrapper calls into `resolve_source()` which calls `resolve_version_from_git_url()` which calls `_get_version_from_package_metadata()` which needs to be able to ensure the dependencies it tries to install are built.

I think what we want to do, separately from this PR, is EITHER move handling for git URLs into the bootstrapper itself, and not pass them through to `resolve_source()` OR we want to pass the bootstrapper into `resolve_source()` so it can be used recursively from there. The former approach would limit git URLs to the bootstrap command, I think, which would break the ability to have nightly builds created by `build-sequence`. The latter approach would mean `build-sequence` would need to instantiate a bootstrapper, too. It also introduces a circular dependency between the 2 modules.

I will spend more time looking at those approaches to see which is more viable.

_Originally posted by @dhellmann in https://github.com/python-wheel-build/fromager/issues/590#issuecomment-2898936016_

dhellmann:
Ensure bootstrap writes out info indicating that the requirement needs to
come from the git URL.

Refactor some of the git clone handling from the bootstrapper so it can be
reused.

If `download_source` finds a git URL in a requirement and the directory with
the source does not already exist, have it clone it. That ensures that the
case where we need to clone the repo to determine the version (during
bootstrap) and the case where we know the version but do not have the
source (during build-sequence) both work.

Add an e2e test case.

Fixes #632


rd4398 (approved):
Looks good to me!

PR #642, opened 2025-06-21
# Add minimize command to reduce redundant requirements

### LINKED ISSUE #630: command to help minimize input requirements.txt files
We could add a command that takes as input a requirements.txt file and a graph file and produces as output a minimized requirements.txt file, removing anything that would have been included automatically as a dependency of one of the packages that appears in the output. This would be useful for avoiding duplicate packages or conflicts due to specifying versions in requirements.txt that then become incompatible with the things that depend on them, or are useless to build because they are not used. It would also help minimize the overall size of the graph, which could speed up build time since we would not need to check as many graph edges and redundant node visits.

dhellmann:
## Summary

Add a new `minimize` command that helps reduce input requirements.txt files by removing packages that would be automatically included as dependencies of other packages. This addresses issue #630.

### Core Implementation
- Add `src/fromager/commands/minimize.py` with the new minimize command
- Register the command in `src/fromager/commands/__init__.py`
- Implement version-aware dependency checking using requirement specifiers
- Use generator pattern for memory-efficient requirement processing
- Leverage `get_install_dependencies()` for optimized graph traversal

### CLI Interface
- Support `-o/--output` option for writing results to file
- Accept requirements.txt and graph.json as positional arguments
- Provide detailed statistics including reduction percentage
- Handle multiple requirement entries for the same package correctly

### Algorithm
- Iterate through each requirement in the input file
- Check if requirement would be satisfied by existing install dependencies
- Only remove requirements when version specifiers match dependency versions
- Preserve requirements with different version constraints or markers

### Testing
- Add comprehensive test suite in `tests/test_minimize.py`
- Test version matching, mismatches, and multiple package entries
- Include CLI integration tests and error handling
- Verify statistics reporting and file output functionality

### Example Usage
```bash
fromager minimize requirements.txt graph.json -o minimized.txt
```

This command helps avoid dependency conflicts, reduces graph size,
and speeds up builds by eliminating redundant package specifications.

Fixes #630


rd4398 (approved):
Nice!

Issue #635, opened 2025-06-17
metadata manipulation functions use incorrect logic for detecting unsafe paths


dhellmann:
In `fromager.wheels.add_extra_metadata_to_wheels()`, there is a check for `".."` in a filename, trying to protect against a path that would pull in data from a parent directory. However, the `feast` package includes some filenames with `..` in the middle of a base name, leading to errors like:

```
2025-06-17 20:34:13,202 DEBUG:fromager.__main__:261: feast: Unsafe path in wheel: feast/ui/build/static/js/icon..569adb91.chunk.js
Traceback (most recent call last):
  File "/opt/app-root/lib64/python3.12/site-packages/fromager/__main__.py", line 259, in invoke_main
    main(auto_envvar_prefix="FROMAGER")
  File "/opt/app-root/lib64/python3.12/site-packages/click/core.py", line 1442, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.12/site-packages/click/core.py", line 1363, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.12/site-packages/click/core.py", line 1830, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.12/site-packages/click/core.py", line 1226, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.12/site-packages/click/core.py", line 794, in invoke
    return callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.12/site-packages/click/decorators.py", line 46, in new_func
    return f(get_current_context().obj, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/app-root/lib64/python3.12/site-packages/fromager/commands/bootstrap.py", line 188, in bootstrap
    bt.bootstrap(req, requirements_file.RequirementType.TOP_LEVEL)
[...]

PR #643, opened 2025-06-22
Fix path traversal check to allow legitimate filenames with double dots

dhellmann:
- Change path traversal security check from '..' in filename to '..' in path components
- Allows legitimate files like 'icon..569adb91.chunk.js' while still blocking actual path traversal
- Add comprehensive test coverage for both legitimate and malicious path cases
- Fixes issue #635 with feast package failing due to overly strict filename validation

[Prepared with Cursor](https://gist.github.com/dhellmann/3d0c993ea3821ed1772098786dd27d74)


tiran (approved):
LGTM

Nice work on `pathlib.Path().parts`. It's too bad that `os.path` does not have a function to split a path string into all its elements.

PR #647, opened 2025-06-26
Increase logging for external commands

tiran:
The `external_commands.run()` function now logs all environment variables. Before it only logged extra environment variables defined in a plugin. The logs were missing information about global environment variables, e.g. env vars defined in a build container. These additional env vars influence a build, too.

The run log file now contains the command, env vars, and current working directory, too.


tiran:
I'm investigating a vLLM build problem and could not figure out the exact build environment from our build logs. This PR includes all build environment variables in the global log and the run log.

EmilienM:
@tiran is there any risk to expose secret environments in CI logs, e.g. credentials embedded in URLs, or secrets? etc.

rd4398 (commented):
Looks good!
I will wait to approve till @EmilienM gets answer to his question

tiran:
> @tiran is there any risk to expose secret environments in CI logs, e.g. credentials embedded in URLs, or secrets? etc.

That is a good point! GitHub and GitLab already mask sensitive values. I have added additional code to also mask keys that look like they hold sensitive values.

src/fromager/external_commands.py
```diff
@@ -46,6 +47,24 @@ class NetworkIsolationError(subprocess.CalledProcessError):
     pass
 
 
+# GitLab BOT_PAT
+# SECRET, PASSWORD, PASSPHRASE, CRED(entials), TOKEN
+SENSITIVE_KEYS = re.compile(
+    "^(BOT_PAT|.*SECRET.*|.*PASSWORD.*|.*PASSPHRASE.*|.*CRED.*|.*TOKEN.*)$",
```
dhellmann, line 53: BOT_PAT is pretty specific to our downstream. Does this pattern need to be configurable for other similar cases?

src/fromager/external_commands.py
```diff
@@ -70,14 +89,17 @@ def run(
...
         ]
 
-    logger.debug(
-        "running: %s %s in %s",
-        " ".join(f"{k}={shlex.quote(v)}" for k, v in extra_environ.items()),
-        " ".join(shlex.quote(str(s)) for s in cmd),
-        cwd or ".",
-    )
+    cmd_str = " ".join(shlex.quote(str(s)) for s in cmd)
+    env_str = " ".join(f"{k}={shlex.quote(v)}" for k, v in mask_envvars(env))
```
dhellmann, line 93: The variable values are being quoted here and in `mask_envvars()`. It probably doesn't matter enough to hold this up, but it will make the output have extra quotes in some cases.

tiran: Good catch, thanks! I have removed the quoting from this place.

src/fromager/external_commands.py
tiran: Maybe... the `external_commands.run` function does not have access to the `WorkContext`. We either need to refactor a lot of code or come up with a different approach.

src/fromager/external_commands.py
```diff
@@ -46,6 +47,23 @@ class NetworkIsolationError(subprocess.CalledProcessError):
     pass
 
 
+# *PAT (GitLab), SECRET, PASSWORD, PASSPHRASE, CRED(entials), TOKEN
+SENSITIVE_KEYS = re.compile(
+    "^(*.PAT|.*SECRET.*|.*PASSWORD.*|.*PASSPHRASE.*|.*CRED.*|.*TOKEN.*)$",
```
dhellmann, line 52:
```suggestion
    "^(.*PAT|.*SECRET.*|.*PASSWORD.*|.*PASSPHRASE.*|.*CRED.*|.*TOKEN.*)$",
```

I think the tests are failing because of a typo here in the regex that makes it look like a glob pattern.

LalatenduMohanty:
@tiran Once we have the tests pass we can get this PR merged.

PR #649, opened 2025-06-27
fix: propagate exceptions in Bootstrapper._build_sdist to fail on sdist build errors

### LINKED ISSUE #637: Bootstrapper._build_sdist() ignores errors
Investigate if the method should re-raise the exception and fail the build.

https://github.com/python-wheel-build/fromager/blob/12ffcbb44b3f12f0d908dd76402a989db0c27931/src/fromager/bootstrapper.py#L326C1-L328C43

EmilienM:
Previously, _build_sdist() would catch and log all exceptions, causing the build
to continue even if building the source distribution failed. This could result in
undefined or misleading behavior. Now, exceptions are allowed to propagate,
ensuring that build failures are reported and handled properly.

Fixes #637


rd4398 (commented):
Looks good! However, there is a merge conflict and so rebase is needed.

dhellmann (approved):

PR #650, opened 2025-06-27
Fix regression in unpack dir handling

tiran:
Commit 11c7502 modified the location of the unpack directory to ensure that `_unpack_metadata_from_wheel()` was not removing the Python build env. This caused a regression because sometimes the files ended up in the wrong directory.

Revert the change to `Bootstrapper._create_unpack_dir()`. The `_unpack_metadata_from_wheel()` now removes just its own files, not the entire directory.

### DIFF
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -187,6 +187,7 @@ class Bootstrapper:
             if not unpacked_cached_wheel:
                 # We didn't find anything so we are going to have to build the
                 # wheel in order to process its installation dependencies.
+                logger.debug("no cached wheel, downloading sources")
                 source_filename = sources.download_source(
                     ctx=self.ctx,
                     req=req,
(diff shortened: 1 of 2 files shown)


src/fromager/bootstrapper.py
```diff
@@ -200,7 +201,14 @@ def bootstrap(self, req: Requirement, req_type: RequirementType) -> Version:
...
                 )
             else:
+                logger.debug(f"have cached wheel in {unpacked_cached_wheel}")
                 sdist_root_dir = unpacked_cached_wheel
+                if sdist_root_dir.parent == self.ctx.work_dir:
+                    # TODO: investigate why this is happening for bootstrapping from tag
+                    logger.warning(
+                        f"{sdist_root_dir} should be two levels away from {self.ctx.work_dir}"
+                    )
+                    sdist_root_dir = sdist_root_dir / f"{req.name}-{resolved_version}"
```
tiran, line 211: @dhellmann This workaround is required for `test_bootstrap_build_tags.sh`. Otherwise it creates places stuff like the Python build env in `ctx.work_dir`. I don't have the capacity to investigate.

tiran:
> This version looks good to me, but the linter doesn't like it.

I have reformatted the code.

dhellmann (approved):

Issue #556, opened 2025-02-25
Resolver: relax wheel tags check to interpreter + abi + linux/any


tiran:
The resolver ignores all packages that don't match the current set of valid platform tags from `packaging.tags.sys_tags()`.

https://github.com/python-wheel-build/fromager/blob/feea4eb1cf778a906b1344da1f55d5f61266ac5d/src/fromager/resolver.py#L183-L190

This has worked fine so far, because all packages either had sdists or wheels with binaries for Linux and current arch `x86_64`. We now have started to build packages for other architectures like `aarch64`. Some of the packages don't have binaries for Linux `aarch64`, e.g. https://pypi.org/project/xgrammar/0.1.13/#files . The resolver can resolve `xgrammar` to a version on x86_64, but not on `aarch64`.

`sys_tags()` has no API to modify the arch of tags. Instead we could relax the check and only look for `(tag.interpreter, tag.abi)` pairs. We could also check that the platform tag is either `any` or matches the current platform (`'linux' in tag.platform` on Linux).

```python
>>> sorted(set((t.interpreter, t.abi) for t in sys_tags()))
[('cp310', 'abi3'), ('cp311', 'abi3'), ('cp312', 'abi3'), ('cp312', 'cp312'), ('cp312', 'none'), ('cp32', 'abi3'),
 ('cp33', 'abi3'), ('cp34', 'abi3'), ('cp35', 'abi3'), ('cp36', 'abi3'), ('cp37', 'abi3'), ('cp38', 'abi3'), ('cp39', 'abi3'),
 ('py3', 'none'), ('py30', 'none'), ('py31', 'none'), ('py310', 'none'), ('py311', 'none'), ('py312', 'none'),
 ('py32', 'none'), ('py33', 'none'), ('py34', 'none'), ('py35', 'none'), ('py36', 'none'), ('py37', 'none'),
 ('py38', 'none'), ('py39', 'none')]
>>> sorted(set(t.platform for t in sys_tags()))
['any', 'linux_x86_64', 'manylinux1_x86_64', 'manylinux2010_x86_64', 'manylinux2014_x86_64', 'manylinux_2_10_x86_64', 
'manylinux_2_11_x86_64', 'manylinux_2_12_x86_64', 'manylinux_2_13_x86_64', 'manylinux_2_14_x86_64', 'manylinux_2_15_x86_64', 
[...]

PR #654, opened 2025-07-03
feat: Add option to ignore wheel platform in resolver

tiran:
A package can now ignore the upstream wheels' platforms when looking up distributions without an sdist on PyPI. The option only applies to packages that have `resolver_dist.include_wheels` enabled.

The flag allows to package versions that neither provide a sdist nor a binary wheel for the current platform, e.g. Torch builds on ppc64le. Torch does neither upload sdists nor does it provide ppc64le wheels on PyPI.

Fixes: #556

### DIFF
--- a/tests/test_resolver.py
+++ b/tests/test_resolver.py
@@ -267,6 +267,64 @@ def test_provider_constraint_match():
         assert str(candidate.version) == "1.2.2"
 
 
+_ignore_platform_simple_response = """
+<!DOCTYPE html>
+<html>
+<head>
+<meta name="pypi:repository-version" content="1.1">
+<title>Links for fromager</title>
+</head>
+<body>
+<h1>Links for fromager</h1>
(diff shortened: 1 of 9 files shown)


rd4398 (commented):
Nice! 
This looks good! I will leave it to @dhellmann for final approval

tests/test_resolver.py
```diff
@@ -267,6 +267,64 @@ def test_provider_constraint_match():
...
+_ignore_platform_simple_response = """
+<!DOCTYPE html>
+<html>
+<head>
+<meta name="pypi:repository-version" content="1.1">
+<title>Links for fromager</title>
+</head>
+<body>
+<h1>Links for fromager</h1>
+<a href="https://files.pythonhosted.org/packages/7c/06/620610984c2794ef55c4257c77211b7a625431b380880c524c2f6bc264b1/fromager-0.51.0-cp311-abi3-manylinux_2_28_plan9.whl" >fromager-0.51.0-cp311-abi3-manylinux_2_28_plan9.whl</a>
```
dhellmann, line 279: Plan9! :-)

dhellmann (approved):
We discussed my concerns about how we will use this downstream, and @tiran pointed out that most of the packages have sdists, so we won't have to have a configuration file for everything that is being built for the CPU architectures that do not have wheels published. We agreed if we started to accumulate a lot of config for packages with only this new flag enabled, we would come back and add a global flag to change the default setting.

PR #660, opened 2025-07-16
Adding common hatch commands to docs/develop.md

### LINKED ISSUE #659: Add common hatch commands to the docs/develop.md
The project has migrated to hatch CLI from tox recently. So it will be useful to list the common hatch commands as contributors might not be familier with hatch CLI

LalatenduMohanty:
The project has migrated to hatch CLI from tox recently. So it will be useful to list the common hatch commands as contributors might not be familier with hatch CLI fixes #659


rd4398 (approved):
Thank you for adding this!

PR #664, opened 2025-07-18
Add matcher argument to resolver providers

tiran:
The `GenericProvider`, `GitHubTagProvider`, and `GitLabTagProvider` now accept a matcher argument. The argument can be a callback or a re pattern object. The new feature enables more powerful matching and transformation of tags into Version objects.

`GitLabTagProvider.tag_regex` is no longer supported. The new feature is a more general implementation of the previous feature.

### DIFF
--- a/src/fromager/resolver.py
+++ b/src/fromager/resolver.py
@@ -5,6 +5,7 @@
 #
 from __future__ import annotations
 
+import functools
 import logging
 import os
 import re
@@ -13,7 +14,6 @@ from collections import defaultdict
 from collections.abc import Iterable
 from operator import attrgetter
 from platform import python_version
-from re import Match
 from urllib.parse import quote, unquote, urljoin, urlparse
 
 import html5lib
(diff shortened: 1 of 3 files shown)


src/fromager/resolver.py
```diff
@@ -523,9 +528,38 @@ def __init__(
...
+        except Exception as err:
+            logger.debug(f"{identifier}: could not parse version from {item}: {err}")
+            return None
+
+    def _re_match_function(
+        self, identifier: str, item: str, *, regex: re.Pattern
+    ) -> Version | None:
+        mo = regex.match(item)
+        if mo is None:
+            return None
```
dhellmann, line 556: It would help to log tags that fail to match the regex. I had to add that locally when debugging an issue with a plugin earlier this week.

dhellmann (commented):
I'm not sure why we need this more general implementation. Do you have an example of where we would need a custom function that would do more than a regex does?

The implementation looks fine. I had one nit about logging, but otherwise if this is the direction we want to go I think it's cleanly done. Ideally if we're going to make a breaking change we would do it before we land any uses of the regex approach downstream.

src/fromager/resolver.py
tiran: Good point, it's an easy fix!

PR #666, opened 2025-07-21
feat: Add helper to git clone and tarball

tiran:
The new helper `git_clone_and_tarball()` is an opinionated function to clone a git repo from tag/ref with submodules and add the content into a tar ball. The tar ball does not include a `.git` directory, but it comes with a `.git_archival.txt` for `setuptools-scm`. An optional callback can be used to rewrite clone URLs.

The base name of the tar ball and first-level directory are always based on the `prefix` argument.

### DIFF
--- a/src/fromager/gitutils.py
+++ b/src/fromager/gitutils.py
@@ -1,10 +1,17 @@
 import logging
+import os
 import pathlib
+import re
+import tarfile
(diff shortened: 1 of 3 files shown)


tiran:
I'm going to solve this problem in a different way. This approach is way too complicated.

PR #675, opened 2025-07-24
fix: improve logging and type annotations of build-parallel

tiran:
- log built nodes in case of an error
- log which nodes get submitted to thread pool worker
- annotate several lists and sets

### DIFF
--- a/src/fromager/commands/build.py
+++ b/src/fromager/commands/build.py
@@ -37,6 +37,9 @@ from ..log import req_ctxvar_context, requirement_ctxvar
 
 logger = logging.getLogger(__name__)
 
+DependencyNodeList = list[dependency_graph.DependencyNode]
+DependencyNodeSet = set[dependency_graph.DependencyNode]
+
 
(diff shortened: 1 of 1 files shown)


src/fromager/commands/build.py
```diff
@@ -613,7 +618,10 @@ def build_parallel(
             with concurrent.futures.ThreadPoolExecutor(
                 max_workers=max_workers
             ) as executor:
-                futures = []
+                futures: list[concurrent.futures.Future] = []
+                logger.info(
+                    "Submitting nodes: %s", sorted(n.key for n in buildable_nodes)
```
dhellmann, line 623: "submitting" makes sense to someone who know the code, but a user won't know where they are being submitted (or what  a node is, in this context).

How about

```suggestion
                    "starting to build: %s", sorted(n.key for n in buildable_nodes)
```

dhellmann (commented):
One wording nit in the log message

dhellmann (approved):

PR #682, opened 2025-08-02
fix handling of existing packages during build

dhellmann:
The _is_wheel_built function was returning a boolean to indicate to the
caller whether the filename returned should be used. Change the function
to return a Path or None, so the caller only has to check the one
return value.

Return the full path to where we downloaded the wheel, or return None if
we do not download one. Add more logging to make it easier to follow
what is happening, especially when the candidate is not usable.

Log more information about what we are doing when we decide to download
a wheel vs. recognize that the wheel is present on the local server.

Update the e2e tests that look for specific log messages to match the
new patterns.

Remove duplicate calls to `update_wheel_mirror()` in `_build()`. At best they
are redundant, at worst they could change state twice.

### DIFF
--- a/src/fromager/commands/build.py
+++ b/src/fromager/commands/build.py
@@ -337,7 +337,7 @@ def _build(
     """
 
     wheel_server_urls = wheel_server_urls or []
-    wheel_filename = None
+    wheel_filename: pathlib.Path | None = None
 
     # Set up a log file for all of the details of the build for this one wheel.
     # We attach a handler to the root logger so that all messages are logged to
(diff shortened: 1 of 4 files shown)


src/fromager/commands/build.py
```diff
@@ -482,16 +478,21 @@ def _is_wheel_built(
...
             )
-        is_built = existing_build_tag == build_tag
-        if is_built and wkctx.wheel_server_url not in url:
+        if existing_build_tag != build_tag:
+            logger.info(
+                f"existing build tag {existing_build_tag} does not match expected build tag {build_tag}"
+            )
+            return None
+
+        if wkctx.wheel_server_url not in url:
```
tiran, line 487: I found this check suspicious when I looked through the looks. How about `if not url.startswith(wkctx.wheel_server_url)` + a bit of extra debugging logging in the true and false branch?

dhellmann: Good point

dhellmann: Updated

tiran (approved):
LGTM, thanks for the additional logging!

PR #683, opened 2025-08-02
feat: Add bootstrap-parallel command

tiran:
The new `bootstrap-parallel` command combines `bootstrap --sdist-only` and `build-parallel` in a single, convenient command.


dhellmann (commented):
Nice! I like the fact that it reuses the existing commands.

Is there anything we can do to automatically ensure that if we add options to build-parallel they are included here (via a linter or something)? Maybe it's not a big deal, but it seems like a potential source of UI drift.

tiran:
This PR contains the commit from PR https://github.com/python-wheel-build/fromager/pull/687

tiran:
> Is there anything we can do to automatically ensure that if we add options to build-parallel they are included here (via a linter or something)? Maybe it's not a big deal, but it seems like a potential source of UI drift.

I have added a simple check that compares options names. It's not going to detect different defaults, but at least it will tell us when an option is added or removed.

src/fromager/commands/bootstrap.py
```diff
@@ -402,3 +405,118 @@ def write_constraints_file(
...
+        cache_wheel_server_url=cache_wheel_server_url,
+        sdist_only=True,
+        skip_constraints=skip_constraints,
+        toplevel=toplevel,
+    )
+
+    # statistics
+    wheels = sorted(f.name for f in wkctx.wheels_downloads.glob("*.whl"))
+    sdists = sorted(f.name for f in wkctx.sdists_downloads.glob("*.tar.gz"))
+    logger.debug("wheels: %s", ", ".join(wheels))
```
dhellmann, line 495: These log lines could end up being pretty massive for our full index builds. Let's see how it goes, but I anticipate coming back and either listing a few packages per line or removing the log calls entirely.

dhellmann (approved):
I like it. Let's release and see how the integration goes.

PR #688, opened 2025-08-03
Trigger Github to debug test failures

tiran:

PR #689, opened 2025-08-03
fix: exclude virtualenv 20.33.0

dhellmann:
There is a known issue in this version of virtualenv that breaks it for several use cases, including the way we use it in tests.

https://github.com/pypa/virtualenv/issues/2930


dhellmann:
Fix for #688

tiran (approved):

PR #697, opened 2025-08-09
fix: pre-built install requires in sdist-only mode

tiran:
Fromager was not handling install requirements of pre-built packages correctly in `bootstrap --sdist-only` mode. It was looking for install requirements in sdist, but a pre-built wheel does not have an sdist.

### DIFF
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -108,11 +108,14 @@ class Bootstrapper:
         self._add_to_graph(req, req_type, resolved_version, source_url)
 
         # Is bootstrap going to create a wheel or just an sdist?
-        # Use fast sdist-only if flag is set, requirement is not a build
-        # requirement, and wheel is not pre-built.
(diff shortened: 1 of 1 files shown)


src/fromager/bootstrapper.py
```diff
@@ -259,12 +259,30 @@ def bootstrap(self, req: Requirement, req_type: RequirementType) -> Version:
             wheel_filename=wheel_filename,
         )
 
-        if build_sdist_only:
+        if pbi.pre_built:
```
dhellmann, line 262: I wonder if we could do something simpler here and just switch based on whether we have a wheel filename or not? By the time we get here we've either built the wheel or the sdist based on these other flags, so we just need to know which file to use to pull the dependencies, right? And there are really only 2 options.

tiran:
I figured it out. There is now a long comment that explains why an install requirement on a pre-built wheel is handled like an sdist-only package.

PR #699, opened 2025-08-11
feat: Add annotation support for packages and variants

tiran:
The new annoation features can be used to attach abitrary metadata to package and variant settings. The feature is inspired by Kubernetes, https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/ Annotations are read-only mappings `Mapping[str, str]`.

Annotations can hold all sorts of information. Use cases include simple settings for downstream plugins or project metadata.

```yaml
annotations:
  "downstream.maintainer": "Platform Team"
  "wheeltag.purelib": "false"
variants:
  cuda:
    annotations:
      "downstream.maintainer": "CUDA Accelerator Team"
```

### DIFF
--- a/src/fromager/packagesettings.py
+++ b/src/fromager/packagesettings.py
@@ -102,6 +102,54 @@ Variant = typing.NewType("Variant", str)
 GlobalChangelog = Mapping[Variant, list[str]]
 VariantChangelog = Mapping[PackageVersion, list[str]]
 
+# Annotations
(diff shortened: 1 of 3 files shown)


tiran:
> Looks like CI fails for some e2e tests

It's setuptools-scm again.

dhellmann (approved):
I worry a little that this could turn into a place to dump things that we should put into the formal config, but I do think it will be useful.

PR #702, opened 2025-08-12
fix: quiet update_wheel_mirror

dhellmann:
update_wheel_mirror() logs the name of every wheel in the mirror every time any wheel is added. This makes the logs excessively long and hard to read. This update makes the log only contain information about action taken.


dhellmann:
> @dhellmann Please rebase

Done

PR #705, opened 2025-08-13
chore: improve logging of resolve_prebuilt_wheel

tiran:
Use an `ExceptionGroup` to return information about all failed resolve attempts, so we can see which lookup failed.


dhellmann (approved):

tiran:
https://pypi.org/project/pbr/7.0.0/ came out an hour ago.

Issue #704, opened 2025-08-13
build-sequence job uses wrong download URL for pre-built wheel


tiran:
In downstream we encountered a bug with pre-built wheels. A top-level requirement of a variant has dependencies with embedded URLs, e.g. `torch @ https://upstream-pypi/.../torch...whl`. The variant has the package configured as a pre-built wheel with an internal PyPI. The bootstrap command is doing the right thing and downloads the pre-built wheel from the internal server.  The `build-sequence` command is pulling pre-built wheels from upstream's custom PyPI instead.

The `build-sequence` command behaves correctly when Torch is listed as a top-level requirement.

tiran:
I found the culprit: The `build` command does not use `pbi.wheel_server_url`.

https://github.com/python-wheel-build/fromager/blob/e4ef7cacaab0362ce6f4cf4912ce458462e7ea99/src/fromager/commands/build.py#L323-L383

PR #706, opened 2025-08-13
fix: use get_wheel_server_urls() in build command

tiran:
The build commands like `build-sequence` were not using `pbi.wheel_server_url`. Under some conditions they were using wheels from an upstream server or requirement URL instead of the configured package wheel server.

The `wheels.get_wheel_server_urls()` handles this case correctly, but the current implementation is not usable for build command.

Refactor code:

- modify `get_wheel_server_urls()` to accept a cache wheel server url instead of a hard-coded URL.
- pass `cache_wheel_server_url` instead of `wheel_server_urls` in build commands
- use `get_wheel_server_urls()` in build commands

Fixes: #704

### DIFF
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -825,7 +825,9 @@ class Bootstrapper:
             wheel_url, resolved_version = cached_resolution
             logger.debug(f"resolved from previous bootstrap to {resolved_version}")
         else:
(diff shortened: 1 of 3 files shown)


tiran:
This PR refactors some code so bootstrap and build can shaer the same logic for wheel server urls. The bug could be fixed with a simpler change.

dhellmann (approved):

PR #709, opened 2025-08-13
Added helper to extract filenames from url

### LINKED ISSUE #711: Wheel URLs with %2B (URL-encoded +) not properly decoded during resolution
vllm-0.10.0%2Brhai1-*.whl should be decoded to vllm-0.10.0+rhai1-*.whl
Currently version matching fails due to %2B not being decoded to +

pmattsso:
Added helper to decode and extract filenames from url

Changed existing decoding code to utilize new helper

Fixes #711

### DIFF
--- a/src/fromager/commands/build.py
+++ b/src/fromager/commands/build.py
@@ -34,6 +34,7 @@ from fromager import (
     wheels,
 )
 
+from .. import resolver
 from ..log import VERBOSE_LOG_FMT, ThreadLogFilter, req_ctxvar_context
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


src/fromager/urlutils.py
```diff
@@ -0,0 +1,10 @@
+import os
```
EmilienM, line 1: I see `threading_utils.py`, maybe follow that pattern and rename to `url_utils.py` ?

dhellmann (commented):
The implementation looks good. I like @EmilienM's comments for minor cleanups.

src/fromager/urlutils.py
tiran: This is a three line function. Just add it to `resolver`.

tests/test_urlutils.py
```diff
@@ -0,0 +1,13 @@
...
+def test_extract_filename_from_url():
+    url = "http://localhost/path/to/file.txt"
+    filename = urlutils.extract_filename_from_url(url)
+    assert filename == "file.txt"
+
+
+def test_extract_filename_from_url_with_encoding():
+    url = "http://localhost/simple/vllm/vllm-0.10.0%2Brhai1-15-cp312-cp312-linux_x86_64.whl"
+    filename = urlutils.extract_filename_from_url(url)
+    assert filename == "vllm-0.10.0+rhai1-15-cp312-cp312-linux_x86_64.whl"
```
tiran, line 13: Use `pytest.mark.parametrize`

EmilienM:
lgtm now

EmilienM (approved):

PR #714, opened 2025-08-19
feat: add cache to resolver in bootstrapper

dhellmann:
A long-running bootstrap job may resolve the same requirements specifier many times, especially for packages like setuptools or flit-core. This change adds a cache for all requirements based on their full spec so that if we see the same string more than once we give the same value without going out to the server to look at available packages again.


EmilienM (approved):
nice 😎

PR #716, opened 2025-08-20
feat(commands): add find-updates command for checking constraint updates

dhellmann:
Add new `find-updates` command that analyzes constraints files and
identifies available package versions that exceed current constraints.

Features:
- Supports equality (==), less than (<), and less than or equal (<=)
constraints
- Ignores lower bound constraints (>=) as they don't pin specific versions
- Multiple output formats: requirements, JSON, CSV
- File output support via --output option
- Distribution type filtering (--distribution-type) with sdist/wheel/both
options
- Reuses existing constraint parsing and version resolution infrastructure

For == constraints, finds versions newer than the pinned version. For < and
<= constraints, finds versions that would violate the constraint.

Example usage:

```bash fromager find-updates constraints.txt fromager find-updates
--format json --output updates.json constraints.txt
```

Chat log:
https://gist.github.com/dhellmann/168b3f52d5715f0f174c03643dcad566

Co-authored-by: Claude 3.5 Sonnet (Anthropic AI Assistant) Signed-off-by:
Doug Hellmann <dhellmann@redhat.com>


dhellmann:
Some sample output: https://gist.github.com/dhellmann/77a995165a946922781405fd9aa4f48b

src/fromager/commands/find_updates.py
```diff
@@ -0,0 +1,255 @@
...
+
+    This command reads a constraints file and for each package that has a
+    supported constraint (==, <, <=), it lists newer versions available.
+    Constraints with lower bounds (>=) are ignored as they don't pin to specific versions.
+
+    The CONSTRAINTS_FILE should be a pip-style constraints file with entries like:
+    - "package_name==1.0.0" (will look for versions > 1.0.0)
+    - "package_name<2.0.0" (will look for versions >= 2.0.0)
+    - "package_name<=1.5.0" (will look for versions > 1.5.0)
+    - "package_name>=1.0.0" (ignored - not a specific constraint)
```
rd4398, line 74: I am a bit confused about the text here :sweat_smile: 
Shouldn't `package_name==1.0.0` means it will look for exactly that version? The text says it will look for > 1.0.0

Am I missing anything here?

rd4398 (commented):
Overall looks good!
Should we include a test for this probably as a follow up?

Also, had one clarification question before I hit approve

src/fromager/commands/find_updates.py
dhellmann: Good question. The goal of the command is to take an existing constraint and figure out if there are newer versions of packages that are being blocked because of that constraint. So the logic here is reversing the constraint, taking into account that we don't want to skip anything (we want a `<2.0.0` rule to look for versions starting with `>=2.0.0` for example, rather than `>2.0.0`).

rd4398: Ah, okay. Now I understand it better. Thank you!

tests/test_find_updates.py
```diff
@@ -0,0 +1,89 @@
+from packaging.requirements import Requirement
+from packaging.version import Version
```
EmilienM, line 2: thanks for the tests, they are useful I think.

PR #718, opened 2025-08-21
Adding --version to the fromager CLI

### LINKED ISSUE #717: Add --version option to fromager CLI
Fromager do not have a --version , I guess this is as designed. However as an user I expect fromager --verison to work as most of the CLI have this option. WDYT about having the --version option.

LalatenduMohanty:
Fixes #717

### DIFF
--- a/src/fromager/__main__.py
+++ b/src/fromager/__main__.py
@@ -3,6 +3,7 @@
 import logging
 import pathlib
 import sys
+from importlib.metadata import version as get_version
 
 import click
 
@@ -31,6 +32,7 @@ else:
 
 
 @click.group()
+@click.version_option(version=get_version("fromager"), prog_name="fromager")
 @click.option(
     "-v",
     "--verbose",


dhellmann:
This is a good feature. The linter doesn't like the code because of the import ordering. Running ruff would fix that. If you're using VSCode or Cursor, there's a plugin for Ruff that will let you run it when you save, similar to `go fmt`. Otherwise, we have a hatch command configured and you can use that.

LalatenduMohanty:
> The linter doesn't like the code because of the import ordering. Running ruff would fix that. If you're using VSCode or Cursor, there's a plugin for Ruff that will let you run it when you save, similar to go fmt. Otherwise, we have a hatch command configured and you can use that.

yeah, It is easy to fix the linting issue. I was not sure if the PR will be accepted or not.  So was waiting for the initial review.

rd4398 (approved):
LGTM! Thank you

PR #719, opened 2025-08-22
feat: Add --test-mode for resilient bootstrap with failure handling

### LINKED ISSUE #713: bootstrap test mode
It is sometimes useful to try to bootstrap a large number of packages to identify the "problem" items in the set that will need work. I usually do this by running bootstrap, marking a failure as pre-built, running it again, etc. until I get all the way through the packages.

It would be useful to have a way to enable a "test mode" for the bootstrap command so that if building a package fails fromager treats it as pre-built automatically. Then at the end the command should produce a list of the packages with build failures and exit with an error code indicating that the build did not succeed.

LalatenduMohanty:
Add --test-mode flag that enables resilient bootstrapping by marking failed
packages as pre-built and continuing until all packages are processed. Uses
optimal n+1 retry logic with comprehensive failure reporting including exception
types, messages, and per-package context.

Benefits:
- Discover all build failures in one run rather than stopping on first failure
- Support mixed source/binary dependency workflows
- Better error context for debugging failed builds
- Cleaner API boundaries between configuration and runtime context

Fixes #713

Co-developed-with: Cursor IDE with Claude 4.0 Sonnet

Command prompts: https://gist.github.com/LalatenduMohanty/762baf9999a09ef3d2d3e63220b9c52e

### DIFF
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import dataclasses
 import json
 import logging
 import operator
@@ -36,6 +37,59 @@ if typing.TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
+
+@dataclasses.dataclass
+class BuildResult:
+    """Simple result object for package builds.
+
+    Tracks both successful and failed build attempts with detailed context
+    for error analysis and reporting in test mode.
+    """
+
+    wheel_filename: pathlib.Path | None = None
+    sdist_filename: pathlib.Path | None = None
+    unpack_dir: pathlib.Path | None = None
+    source_url_type: str = "unknown"
+    sdist_root_dir: pathlib.Path | None = None
+    build_env: build_environment.BuildEnvironment | None = None
+    failed: bool = False
+
+    # Context fields for error tracking and reporting
+    req: Requirement | None = None
+    resolved_version: Version | None = None
+    exception: Exception | None = dataclasses.field(
+        default=None, repr=False, compare=False
+    )
+    exception_type: str | None = None  # Serializable: exception.__class__.__name__
+    exception_message: str | None = None  # Serializable: str(exception)
+
+    @classmethod
+    def failure(
+        cls,
+        req: Requirement | None = None,
+        resolved_version: Version | None = None,
+        exception: Exception | None = None,
+    ) -> BuildResult:
+        """Create a failed build result with context for debugging.
+
+        Args:
+            req: The requirement that failed to build
+            resolved_version: The resolved version that was attempted
+            exception: The exception that caused the failure
+
+        Returns:
+            BuildResult marked as failed with captured context
+        """
+        return cls(
+            failed=True,
+            req=req,
+            resolved_version=resolved_version,
+            exception=exception,
+            exception_type=exception.__class__.__name__ if exception else None,
+            exception_message=str(exception) if exception else None,
+        )
+
+
 # package name, extras, version, sdist/wheel
 SeenKey = tuple[NormalizedName, tuple[str, ...], str, typing.Literal["sdist", "wheel"]]
 
@@ -48,12 +102,17 @@ class Bootstrapper:
         prev_graph: DependencyGraph | None = None,
         cache_wheel_server_url: str | None = None,
         sdist_only: bool = False,
+        test_mode: bool = False,
     ) -> None:
         self.ctx = ctx
         self.progressbar = progressbar or progress.Progressbar(None)
         self.prev_graph = prev_graph
         self.cache_wheel_server_url = cache_wheel_server_url or ctx.wheel_server_url
         self.sdist_only = sdist_only
+        self.test_mode = test_mode
+        self.failed_builds: list[
+            BuildResult
+        ] = []  # Track all failures (build, resolution, etc.) with full context
         self.why: list[tuple[RequirementType, Requirement, Version]] = []
         # Push items onto the stack as we start to resolve their
         # dependencies so at the end we have a list of items that need to
@@ -154,8 +213,8 @@ class Bootstrapper:
         #
         # When bootstrap encounters another package with a *build* requirement
         # on a pre-built wheel, its installation dependencies are materialized.
-        build_sdist_only = self.sdist_only and not self._processing_build_requirement(
-            req_type
+        build_sdist_only: bool = (
+            self.sdist_only and not self._processing_build_requirement(req_type)
         )
 
         # Avoid cyclic dependencies and redundant processing.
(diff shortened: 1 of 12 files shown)


dhellmann (changes requested):
This implementation is going to be very slow, since it will re-process all of the packages that build successfully until it gets to the one that failed and is now marked as prebuilt.

It would be more efficient to do the check in the bootstrapper class at the point where a wheel is being built. If that build fails then we can treat the package as though it was prebuilt by running the logic to handle a prebuilt wheel, even though that wheel is not marked as prebuilt.

You could extract the logic from https://github.com/python-wheel-build/fromager/blob/main/src/fromager/bootstrapper.py#L188-L295 into its own function to make some of that logic easier to deal with. A closure inside the existing function might be easier, since that code uses a lot of the variables from elsewhere in the existing function.

LalatenduMohanty:
As discussed with @tiran last week, I will create a design doc and get it reviewed first then will update this PR.  cc @dhellmann

src/fromager/commands/bootstrap.py
```diff
@@ -486,6 +516,7 @@ def bootstrap_parallel(
         cache_wheel_server_url=cache_wheel_server_url,
         sdist_only=True,
         skip_constraints=skip_constraints,
+        test_mode=test_mode,
```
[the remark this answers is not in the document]
LalatenduMohanty: Right, If you want to test actual wheel compilation failures, you'd need to run `fromager bootstrap --test-mode package1 package2 ` 
However it will be still useful for bootstrap-parallel to identify some issues with the source distribution e.g. identifying of the source of a package is not available, or any-other issue with dependency resolution

LalatenduMohanty:
Here is the [design document](https://gist.github.com/LalatenduMohanty/118399ee5cf5dd204c5751fd9b23e0c6) . PTAL! @dhellmann @tiran

LalatenduMohanty:
@dhellmann @tiran Here is the design document for the current implementation : [doc](https://gist.github.com/LalatenduMohanty/118399ee5cf5dd204c5751fd9b23e0c6)

dhellmann:
> @dhellmann @tiran Here is the design document for the current implementation : [doc](https://gist.github.com/LalatenduMohanty/118399ee5cf5dd204c5751fd9b23e0c6)

That design seems more like what I was expecting. It isolates the check for test mode in the bootstrapper where it should be. 

To handle the parallel case, we could say this feature is not supported, to avoid having to change the build code in a similar way. Downstream we can control the mode when we want to run a test.

src/fromager/packagesettings.py
```diff
@@ -628,6 +628,7 @@ class PackageBuildInfo:
     """
 
     def __init__(self, settings: Settings, ps: PackageSettings) -> None:
+        self._settings = settings
```
tiran: I avoided to keep a reference to `settings` because it would introduce a circular dependency.

src/fromager/packagesettings.py
```diff
@@ -1065,6 +1071,7 @@ def __init__(
         self._patches_dir = patches_dir
         self._max_jobs = max_jobs
         self._pbi_cache: dict[Package, PackageBuildInfo] = {}
+        self.pre_built_override: set[NormalizedName] = set()
```
tiran: I think it would make more sense to have `_pre_built_override` on the WorkContext object and then have a small API to add / get pre_built overrides. WDYT?

src/fromager/bootstrapper.py
```diff
@@ -417,6 +318,178 @@ def _build_wheel(
...
+            sdist_filename=sdist_filename,
+            unpack_dir=unpack_dir,
+            source_url_type=source_url_type,
+            sdist_root_dir=sdist_root_dir,
+            build_env=build_env,
+        )
+
+    def _find_cached_wheel(
+        self, req: Requirement, resolved_version: Version
+    ) -> tuple[pathlib.Path | None, pathlib.Path | None]:
```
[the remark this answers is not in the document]
LalatenduMohanty: Looks like these returns are possible:
                - (Path, Path): Wheel found with metadata successfully extracted
                - (Path, None): Wheel found but metadata extraction failed
                - (None, None): No matching wheel found in any location
 
So current code looks right to me

src/fromager/packagesettings.py
LalatenduMohanty: I agree. Fixing it.

src/fromager/packagesettings.py
LalatenduMohanty: Looks like the code never used the global settings. So removed it.

src/fromager/bootstrapper.py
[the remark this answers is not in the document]
LalatenduMohanty: I have not changed the code for this, PTAL at the code and let me know if you disagree.

src/fromager/bootstrapper.py
```diff
@@ -174,163 +233,41 @@ def bootstrap(self, req: Requirement, req_type: RequirementType) -> Version:
...
-            else:
-                # build wheel (build requirements, full build mode)
-                logger.debug(
-                    f"building wheel {req.name}=={resolved_version} "
-                    f"to get install requirements ({req_type})"
-                )
-                wheel_filename, sdist_filename = self._build_wheel(
-                    req, resolved_version, sdist_root_dir, build_env
-                )
+        # Much simpler: just build the package
```
dhellmann: The comments about simplicity aren't really needed.

LalatenduMohanty: Removed these comments

src/fromager/commands/bootstrap.py
```diff
@@ -183,9 +197,26 @@ def bootstrap(
...
-            requirement_ctxvar.reset(token)
+            try:
+                bt.bootstrap(req, requirements_file.RequirementType.TOP_LEVEL)
+                progressbar.update()
+                if test_mode:
+                    logger.info("Successfully processed: %s", req)
+            except Exception as err:
+                if test_mode:
+                    # Test mode: log error but continue processing
+                    logger.error(
```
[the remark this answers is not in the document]
LalatenduMohanty: Thats right, there is the `else` condition for normal mode.

src/fromager/bootstrapper.py
```diff
@@ -174,163 +233,41 @@ def bootstrap(self, req: Requirement, req_type: RequirementType) -> Version:
...
+            wheel_filename=result.wheel_filename,
         )
 
-        if wheel_filename is not None:
-            assert unpack_dir is not None
-            logger.debug(
-                "get install dependencies of wheel %s",
-                wheel_filename.name,
+        # Get install dependencies - much simpler logic
+        if result.failed:
```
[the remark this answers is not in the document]
LalatenduMohanty: The value of result.failed is true only in the test-mode. However your concern is valid. We can make the code more defensive for future errors by changing it to `if self.test_mode and result.failed:`

src/fromager/bootstrapper.py
```diff
@@ -417,6 +354,185 @@ def _build_wheel(
...
+            # Test mode: try pre-built fallback
+            logger.warning(
+                "test mode: build failed for %s==%s, attempting fallback to pre-built",
+                req.name,
+                resolved_version,
+                exc_info=True,
+            )
+
+            try:
+                self._mark_package_as_pre_built_runtime(req)
```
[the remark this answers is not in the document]
LalatenduMohanty: This is a good catch. The previous code was not efficient. Fixed it.

src/fromager/context.py
```diff
@@ -95,9 +95,19 @@ def __init__(
 
         self._parallel_builds = False
 
+        self._pre_built_override: set[NormalizedName] = set()
+
     def enable_parallel_builds(self) -> None:
         self._parallel_builds = True
 
+    def add_pre_built_override(self, package_name: str | NormalizedName) -> None:
```
dhellmann: With the change in the logic in the bootstrapper to invoke the logic for prebuilt packages directly, I think you can remove this function.

LalatenduMohanty: We use it through `_mark_package_as_pre_built_runtime(req)`  in _build_package()
```
 try:
                self._mark_package_as_pre_built_runtime(req)
                wheel_url, _ = self._resolve_prebuilt_with_history(
                    req=req, req_type=RequirementType.TOP_LEVEL
                )
                wheel_filename, unpack_dir = self._download_prebuilt(
                    req=req,
                    req_type=RequirementType.TOP_LEVEL,
                    resolved_version=resolved_version,
                    wheel_url=wheel_url,
                )
                logger.info(
                    "test mode: successfully handled %s as pre-built after build failure",
                    req.name,
                )
```

LalatenduMohanty: Sorry, I misunderstood. I removed my previous comments. Going to fix it.

src/fromager/commands/bootstrap.py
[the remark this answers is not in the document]
LalatenduMohanty: Added following. 
```
+                    bt.failed_builds.append(
+                        bootstrapper.BuildResult.failure(req=req, exception=err)

```

src/fromager/bootstrapper.py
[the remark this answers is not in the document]
dhellmann: We have the 3 cases, but cannot have (None, Path) because there is no unpacked metadata directory without the wheel file.

PR #730, opened 2025-08-28
Improving the text about fromager

LalatenduMohanty:
We do not mention that fromager can builds collection of wheels. This is the USP of fromager which should be part of the readme.


rd4398 (approved):
Looks good!

PR #732, opened 2025-08-30
fix: remove local build directories

tiran:
delete and ignore local build directories for test plugins.


dhellmann:
I don't understand this one. Why don't we need the test plugins anymore?

PR #737, opened 2025-08-31
feat: validate name and version of sdist and wheel

tiran:
Fromager now validates the dist name and dist version part of sdist and wheel files. The canonicalized dist name must match the canonicalized requirement name. The public version versions must match, too. Validation is performed in `build_sdist`, `build_wheel`, and `default_get_install_dependencies_of_sdist` steps. The latter covers bootstrap in sdist-only mode.

The new checks prevent invalid and broken packages, e.g. `unknown-0.0` wheel or a wheel that has wrong version number.

### DIFF
--- a/src/fromager/dependencies.py
+++ b/src/fromager/dependencies.py
@@ -12,6 +12,7 @@ import pyproject_hooks
 import tomlkit
 from packaging.metadata import Metadata
 from packaging.requirements import Requirement
+from packaging.utils import canonicalize_name
 from packaging.version import Version
 
 from . import (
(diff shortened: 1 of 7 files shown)


src/fromager/dependencies.py
```diff
@@ -332,6 +329,51 @@ def parse_metadata(metadata_file: pathlib.Path, *, validate: bool = True) -> Met
...
+    )
+    with tempfile.TemporaryDirectory() as tmp_dir:
+        distinfo_name = hook_caller.prepare_metadata_for_build_wheel(
+            tmp_dir,
+            config_settings=config_settings,
+        )
+        metadata_file = pathlib.Path(tmp_dir) / distinfo_name / "METADATA"
+        metadata = parse_metadata(metadata_file, validate=validate)
+
+    if validate:
```
dhellmann, line 364: Fetching the metadata and validation feel like different concerns that should be separate functions.

tiran: Makes sense. I have refactored the code. Now `default_get_install_dependencies_of_sdist` is performing the validation. The location of the check is still questionable, but I cannot think of a better place without major refactoring. And I don't want to call `prepare_metadata_for_build_wheel` twice. The hook takes a non-trivial amount of time.

src/fromager/wheels.py
```diff
@@ -311,29 +331,40 @@ def build_wheel(
...
             f"Expected 1 built wheel in {ctx.wheels_build}, got {len(wheels)}"
         )
+    tmp_wheel_file = wheels[0]
 
-    # invalidate uv's cache
-    build_env.clean_cache(req)
+    # validate location and file name
+    if tmp_wheel_file.parent != ctx.wheels_build:
+        raise ValueError(
+            "{wheel_file!} is not in wheels build directory {ctx.wheels_build!r}"
```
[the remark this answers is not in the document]
LalatenduMohanty: Mistakenly added as a comment, so deleting this and adding this as a review

LalatenduMohanty, line 352: probably a typo wheelfile should be tmp_wheel_file with file formatting f"{tmp_wheel_file!r} is not in wheels build directory {ctx.wheels_build!r}"

LalatenduMohanty (changes requested):

src/fromager/sources.py
```diff
@@ -605,8 +608,19 @@ def build_sdist(
...
-    return sdist_filename
+    logger.info(f"built source distribution {sdist_file}")
+
+    # validate location and file name
+    if not sdist_file.is_file():
+        raise FileNotFoundError(sdist_file)
+    sdist_file = sdist_file.resolve()
+    if sdist_file.parent != ctx.sdists_builds:
+        raise ValueError(
+            "{sdist_file!} is not in sdists build directory {ctx.sdists_builds!r}"
```
LalatenduMohanty, line 619: We need file formatting for this error. `f"{sdist_file!r} is not in sdists build directory {ctx.sdists_builds!r}"`

tiran: Good catch!

PR #739, opened 2025-09-01
feat: get installed packages and syspath from BuildEnvironment

tiran:
A `BuildEnvironment` object has new methods to get sysconfig path (e.g. `platlib` directory) and a mapping of installed packages and their versions.

The information is needed for downstream packaging, e.g. to get the current Torch version to add correct constraint `PYTORCH_VERSION` constraint to TorchAudio and TorchVision.

### DIFF
--- a/src/fromager/build_environment.py
+++ b/src/fromager/build_environment.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import json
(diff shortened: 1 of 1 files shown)


LalatenduMohanty (approved):
/lgtm

Issue #742, opened 2025-09-02
Improve performance of data-requires-python matching with LRU cache


tiran:
Parsing `data-requires-python` attribute as a packaging `SpecifierSet` and then matching it with the current Python version is a performance bottleneck. Bootstrap jobs can spend > 10% of their total runtime parsing the specifier set, parsing the versions, and then matching the current Python version with the `SpecifierSet`.

Most specifier sets for requires Python follow a common pattern, e.g. `>= 3.10` or `< 3.12`. An LRU cache can speed up parsing and matching.

PR #743, opened 2025-09-02
perf: cache Python requirement match

tiran:
SpecifierSet parsing and matching takes a non-trivial amount of time. A bootstrap run can spend over 10% of its time in parsing and matching Python version requirements.

The resolver now uses an LRU cache for parsing and matching a Python version requirement.

From my test run with about 370 packages:
```
match_py_req LRU cache: CacheInfo(hits=331462, misses=1110, maxsize=200, currsize=126)
```

Fixes: #742

### DIFF
--- a/src/fromager/commands/bootstrap.py
+++ b/src/fromager/commands/bootstrap.py
@@ -17,6 +17,7 @@ from .. import (
     metrics,
     progress,
     requirements_file,
+    resolver,
     server,
 )
 from ..log import requirement_ctxvar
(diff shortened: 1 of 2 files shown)


src/fromager/commands/bootstrap.py
```diff
@@ -196,6 +197,8 @@ def bootstrap(
                     f"Could not produce a pip compatible constraints file. Please review {constraints_filename} for more details"
                 )
 
+    logger.info("match_py_req LRU cache: %r", resolver.match_py_req.cache_info())
```
LalatenduMohanty, line 200: I guess this is something to understand how much LRU cache we use in the bootstrap? Do we want this all bootstrap logs?

LalatenduMohanty: +1 for debug log. I am also fine with merging the PR as it and create a follow up PR to move it to debug.

Issue #741, opened 2025-09-02
Add PEP 691 support to get_project_from_pypi


tiran:
Fromager's `get_project_from_pypi` only supports [PEP 503](https://peps.python.org/pep-0503/). HTML parsing has performance issues. The function should also support and prefer [PEP 691](https://peps.python.org/pep-0691/). We need to keep PEP 503 support for local simple index, GitLab, and 3rd party indexes without PEP 691 support.

The PEP has example code how to use `Accept` HTTP request header and `Content-Type` HTTP response header to request and detect Simple JSON API support.

Before we write our own code, look for existing implementations. I don't want to role our own PyPI request API if we can use an existing package.

tiran:
https://github.com/jwodder/pypi-simple is a promising package. It's mature, well written, and supports all our use cases.

```python
    client = pypi_simple.PyPISimple(
        endpoint=sdist_server_url,
        session=session,
        accept=pypi_simple.ACCEPT_JSON_PREFERRED,
    )
    package = client.get_project_page(project)
```

PR #745, opened 2025-09-03
feat: use pypi_simple for PyPI simple API

tiran:
The project https://pypi.org/project/pypi-simple/ provides an easy-to-use, convenient wrapper around PyPI's simple HTML and JSON API as well as a bunch of packaging PEPs. Replace our custom `html5lib` and `requests` library with pypi-simple.

PyPI API calls now prefer the faster, more feature-rich JSON API and fall back to HTML for local index and GitLab.

Fixes: #741


src/fromager/resolver.py
```diff
@@ -184,149 +183,179 @@ def get_project_from_pypi(
...
+            sdist_server_url,
             e,
         )
         raise
 
-    doc = html5lib.parse(data, namespaceHTMLElements=False)
-    for i in doc.findall(".//a"):
-        candidate_url = urljoin(simple_index_url, i.attrib["href"])
-        py_req = i.attrib.get("data-requires-python")
-        # PEP 658: Check for metadata availability (PEP 714 data-core-metadata first)
```
rd4398, line 207: Dumb Question: I see PEP 658 related code removed but not added in the this MR. Does PEP 792 cover PEP 658?
I am a bit confused. Am I missing anything?

rd4398 (commented):
A clarification question but overall looks good

src/fromager/resolver.py
tiran: PEP 658 support is much simpler with `pypi-simple`, because the library handles all special cases. It's now a one-liner in the `Candidate()` constructor:
```
metadata_url=dp.metadata_url if dp.has_metadata else None
```

src/fromager/resolver.py
```diff
@@ -184,149 +183,179 @@ def get_project_from_pypi(
...
+                project,
+                package.status,
+                package.status_reason,
+            )
+        case pypi_simple.ProjectStatus.QUARANTINED:
+            raise ValueError(
+                f"project {project!r} is quarantined: {package.status_reason}"
+            )
+        case _:
+            logger.warning(
```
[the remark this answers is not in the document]
dhellmann: OK, thanks.

PR #751, opened 2025-09-04
feat(hooks): log loaded hooks on startup for build commands

dhellmann:
Add logging to display loaded hooks during application startup for build commands. This improves visibility into which hooks are available and being loaded from external packages.

Changes:
- Add log_hooks() function in hooks.py to enumerate and log all available hooks with their source module and distribution info
- Import hooks module in __main__.py and call log_hooks() during startup for commands that show build settings
- Update test_prebuilt_wheel_hook.sh to verify hook loading messages appear in logs and fix shell quoting issue


EmilienM (approved):

rd4398 (approved):
Looks good to me!

PR #764, opened 2025-09-14
fix: clean up logging in resolver module

dhellmann:
* Remove the package name prefix from log messages, since that is handled by the logging context now.
* Remove several, but not all, uses of DEBUG_RESOLVER so that the debug log is more informative. We had been hiding some of that logging because of how verbose the log file is, but we have more caching in place now so we do less repetitive resolving of the same requirements specifiers.
* Update logging calls to use %s, etc. instead of f-strings.

### DIFF
--- a/src/fromager/resolver.py
+++ b/src/fromager/resolver.py
@@ -185,7 +185,7 @@ def get_project_from_pypi(
     found_candidates: set[str] = set()
     ignored_candidates: set[str] = set()
     simple_index_url = sdist_server_url.rstrip("/") + "/" + project + "/"
-    logger.debug("%s: getting available versions from %s", project, simple_index_url)
+    logger.debug("getting available versions from %s", simple_index_url)
 
     try:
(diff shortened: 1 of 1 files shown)


EmilienM (approved):

tiran:
current main branch:
```
$ ./e2e/test_bootstrap.sh 
$ ls -lh e2e-output/bootstrap.log 
-rw-r--r--. 1 heimes heimes 370K Sep 15 12:24 e2e-output/bootstrap.log
$ wc -l e2e-output/bootstrap.log 
3888 e2e-output/bootstrap.log
```

This PR:
```
$ ./e2e/test_bootstrap.sh 
$ ls -lh e2e-output/bootstrap.log 
-rw-r--r--. 1 heimes heimes 1,8M Sep 15 12:25 e2e-output/bootstrap.log
$ wc -l e2e-output/bootstrap.log 
9724 e2e-output/bootstrap.log
```

Issue #755, opened 2025-09-05
build-parallel needs to account for installation dependencies of build dependencies


dhellmann:
The build-parallel command has a flaw in the way it chooses which packages to build. Today it looks at the packages in alphabetical order (to make the logs easy to review), and waits to build something until its build dependencies are available. That has worked for very shallow dependency graphs, but it does not take into account the _installation dependencies_ of those build dependencies. So as the graph gets bigger and deeper, we start seeing issues processing the build.

The build-serial command uses the build-order.json file produced by the bootstrap command to ensure that all of the dependencies are built in the right order. Instead of looking at packages in alphabetical order, build-parallel should use the same in-order traversal produced by bootstrap.

The graph file is much richer than the build-order.json file, and we want to eventually remove the build-order.json file, so rather than making build-parallel use that file we could re-create the build order when we read the graph by applying the same traversal logic that the bootstrap command does.

We could also update the bootstrap command to include build order numbering in the graph file.

dhellmann:
I expect we need to change [these lines](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/commands/build.py#L611C1-L615C1) to start with the list of nodes in the right order, and then in the loop where the list of buildable nodes is determined we need to decide how to stop iterating through the list of packages so that we do not jump ahead and allow a build to use something that cannot be installed.

dhellmann:
Data files from the failed job:

[graph.json](https://github.com/user-attachments/files/22181527/graph.json)

[build-order.json](https://github.com/user-attachments/files/22181530/build-order.json)

and the error

```
2025-09-05 20:32:44,093 ERROR:fromager.external_commands:102: ipykernel-6.30.1: ['uv', 'pip', 'install', '--verbose', '--upgrade', '--only-binary', ':all:', '--constraint', '/mnt/computed-constraints.txt', '--index-url', 'http://localhost:46787/simple/', '--trusted-host', 'localhost', 'jupyter_client>=6', 'hatchling>=1.4'] failed with DEBUG uv 0.8.15
DEBUG Checking for Python interpreter at path `build-3.12.9/bin/python3`
Using Python 3.12.9 environment at: build-3.12.9
DEBUG Acquired lock for `build-3.12.9`
DEBUG Using request timeout of 30s
DEBUG Solving with installed Python version: 3.12.9
DEBUG Solving with target Python version: >=3.12.9
DEBUG Adding direct dependency: jupyter-client>=6
[...]

PR #771, opened 2025-09-20
fix: build installation dependencies before building consumers

dhellmann:
Change the way we determine whether a package is ready to be built in
build-parallel so that we build its installation dependencies before we
build it.

Logically it should not be possible for package A installation
dependencies to depend in any way on A, so this ordering should be
safe. The change ensures that when we mark A as built, it is actually
ready to be consumed in build environments for other packages.

Addresses #755


dhellmann:
@tiran This is a simpler fix than the rewrite in #759 and given where we are in the release schedule downstream I'm more comfortable trying a small change like this now and coming back to the rewrite later. What do you think?

PR #782, opened 2025-09-23
fix: ensure we compare canonicalized versions of names

dhellmann:
The validation of sdist metadata for packages is very loose and does not ensure that the sdist values match the canonical values. We should canonicalize the names before we compare them to avoid false negatives.

### DIFF
--- a/src/fromager/dependencies.py
+++ b/src/fromager/dependencies.py
@@ -375,13 +375,20 @@ def validate_dist_name_version(
 ) -> None:
     """Validate that dist name and version matches expected values"""
     req_name = canonicalize_name(req.name)
-    if dist_name != req_name:
-        raise ValueError(f"{what} does not match requirement {req_name!r}")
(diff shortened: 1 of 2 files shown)


EmilienM (approved):
sad to do this but it'll be useful in our world 🙂

tiran (changes requested):
I do not agree with this fix. `validate_dist_name_version` is used in three places:

1. in `default_get_install_dependencies_of_sdist` to validate sdist metadata name.
2. in `validate_sdist_filename` to validate the sdist file name.
3. in `validate_wheel_filename` to validate the wheel file name.

In case (2) and (3) the `dist_name` parameter is already a `NormalizedName`. The problem is only with (1). A better approach is to canonicalize the name in (1) and change `validate_dist_name_version` to require a `NormalizedName`.

tiran:
I have created #784 . The PR resolves the bug by normalizing the name in the sdist metadata hook. The PR also adds a bunch of tests.

PR #787, opened 2025-09-24
fix: revert broken features

dhellmann:
Revert the feature that validates the name and version of the
packages because it happens too early, so packages with plugins
that override those settings via the metadata hook end up reporting
the wrong versions.

Revert the feature that resolves requirements in the linter because
it introduces new runtime dependencies that we can't absorb downstream.

Both of these changes are breaking different jobs downstream, and
we don't have a release of fromager that includes #786 but does
not include the changes reverted here.


dhellmann:
It looks like we've landed fixes and don't need this.

PR #788, opened 2025-09-25
fix: don't treat bad sdist metadata as fatal error

tiran:
`default_get_install_dependencies_of_sdist` no longer treats bad sdist metadata as a fatal error. The function now logs a non-fatal error message to inform the user of potential issues.

Fromager stub metadata is no longer validated at all. It likely contains a wrong version number.


rd4398 (approved):
Looks good!

PR #792, opened 2025-09-25
fix: remove warning about missing hostname command

tiran:
Partly revert https://github.com/python-wheel-build/fromager/commit/e9bd42478e9e158059bc21d17d10cc95d2d307b1. The warning about missing `hostname` command is causing issues and breaks `get_dependencies` call.

Log a better error message when JSON de-serialization fails in `BuildEnvironment.get_distribution()` call.

Do not fail when debug log call with installed distributions is failing. The error is not fatal.


rd4398 (approved):
LGTM

PR #795, opened 2025-09-28
feat: add TrackingTopologicalSorter

tiran:
`TopologicalSorter.get_ready()` returns a node only once. The tracking topological sorter keeps track which nodes are marked as done. The `get_available()` method returns nodes again and again, until they are marked as done. The graph is active until all nodes are marked as done.

Individual nodes can be marked as exclusive nodes. ``get_available`` treats exclusive nodes special and returns:

1. one or more non-exclusive nodes
2. exactly one exclusive node that is a predecessor of another node
3. exactly one exclusive node

The class uses a lock for ``is_activate`, ``get_available`, and ``done``, so the methods can be used from threading pool and future callback.

### DIFF
--- a/src/fromager/dependency_graph.py
+++ b/src/fromager/dependency_graph.py
@@ -1,9 +1,11 @@
 from __future__ import annotations
 
 import dataclasses
+import graphlib
 import json
 import logging
 import pathlib
+import threading
 import typing
 
 from packaging.requirements import Requirement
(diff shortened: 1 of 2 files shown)


LalatenduMohanty:
Adding the use case to the PR as it is not clear from the PR description

```
Use Case

- This is designed for parallel build orchestration where:

- Multiple packages can be built simultaneously if dependencies are met

- Some packages (marked exclusive) are resource-intensive and should be built alone or with priority

- The build system needs to track which packages are currently being built vs. completed

- Thread-safe access is needed for concurrent workers

```

LalatenduMohanty:
The code looks fine to me. I will take deeper look tomorrow as well. @dhellmann PTAL.

src/fromager/dependency_graph.py
```diff
@@ -379,3 +381,133 @@ def _depth_first_traversal(
...
+                return non_exclusive
+
+            # return a single exclusive node, prefer nodes that are a
+            # dependency of other nodes.
+            exclusive = self._in_progress_nodes.intersection(exclusive_nodes)
+            exclusive_list = sorted(
+                exclusive,
+                key=lambda node: (exclusive_nodes[node], node),
+            )
+            return {exclusive_list[0]}
```
LalatenduMohanty: exclusive_list = sorted(exclusive, ...)
return {exclusive_list[0]}  # ← Can crash if exclusive is empty!

LalatenduMohanty (approved):

PR #794, opened 2025-09-27
# Refactor parallel build logic and enhance dependency checking

dhellmann:
This PR refactors the parallel build logic in `build_parallel` into a dedicated `ParallelBuildManager` class and fixes a critical dependency resolution issue that could cause build failures.

### Architecture Improvements
- Extract complex parallel build logic into a dedicated `ParallelBuildManager` class
- Improve code organization, maintainability, and testability
- Encapsulate build state management with cleaner API

### Bug Fix
- Fix dependency resolution to check installation dependencies of build dependencies
- Prevent build failures when build tools have unmet runtime dependencies
- Ensure complete dependency chain satisfaction before marking nodes as buildable

### Testing
- Add comprehensive test suite with 419 lines of test coverage
- Test complex dependency scenarios including transitive dependencies
- Verify exclusive build handling and circular dependency detection
- Include logging verification and error handling tests

### Code Cleanup
- Remove unused `get_build_progress` method to reduce API surface
- Eliminate dead code and improve code clarity

## Files Changed
- `src/fromager/commands/build.py`: Refactored parallel build logic (+215/-68 lines)
- `tests/test_parallel_build_manager.py`: New comprehensive test suite (+419 lines)

## Commit List
- `8044708` - refactor(build): extract parallel build logic into ParallelBuildManager class
- `f260bf8` - fix(build): check install deps of build deps in parallel builds  
- `0b83ae7` - refactor(build): remove unused get_build_progress method
- `bc82dbf` - test: add comprehensive tests for ParallelBuildManager class

## Testing
All existing tests pass, and the new test suite provides comprehensive coverage of the parallel build manager functionality including edge cases and error conditions.

Chat log: https://gist.github.com/dhellmann/3ca0fb738d23be868c1692ad4748ad43
Co-authored-by: Claude 3.5 Sonnet (Anthropic AI Assistant)
Signed-off-by: Doug Hellmann <dhellmann@redhat.com>

PR #796, opened 2025-09-29
Redesign parallel build system

tiran:
Redesign the parallel build command on top of #763 and #795. The tracking dependency sorter handles the logic to return new buildable nodes as soon as all their build dependencies are done.

### DIFF
--- a/e2e/test_build_parallel.sh
+++ b/e2e/test_build_parallel.sh
@@ -44,7 +44,7 @@ if ! grep -q "cython-3.1.1: ready to build" "$log"; then
   echo "Did not find message indicating build of cython would start" 1>&2
   pass=false
 fi
-if ! grep -q "cython: requires exclusive build" "$log"; then
+if ! grep -q "cython-3.1.1: requires exclusive build" "$log"; then
   echo "Did not find message indicating build of cython would run on its own" 1>&2
   pass=false
 fi
--- a/src/fromager/commands/build.py
+++ b/src/fromager/commands/build.py
@@ -58,6 +58,12 @@ class BuildSequenceEntry:
         }
 
 
+BuildSequenceEntryFuture: typing.TypeAlias = concurrent.futures.Future[
+    BuildSequenceEntry
+]
+DependencyNodeSet: typing.TypeAlias = set[dependency_graph.DependencyNode]
+
+
 @click.command()
 @click.option(
     "--wheel-server-url",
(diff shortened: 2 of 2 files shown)


src/fromager/dependency_graph.py
```diff
@@ -312,3 +381,148 @@ def _depth_first_traversal(
...
+                self._exclusive_nodes[node] = -1
+
+    def is_active(self) -> bool:
+        with self._lock:
+            return bool(self._in_progress_nodes) or self._topo.is_active()
+
+    def __bool__(self) -> bool:
+        return self.is_active()
+
+    def get_available(self) -> set[DependencyNode]:
```
dhellmann, line 490: I avoided having to wrap the sorter like this in #794 by having the user of the topological sorter split the ready nodes into exclusive and non-exclusive lists and then yielding them separately. The equivalent of making this method a generator for lists of nodes to build.

src/fromager/dependency_graph.py
```diff
@@ -111,6 +122,60 @@ def construct_root_node(cls) -> DependencyNode:
...
+                # optimization: don't traverse visited nodes
+                continue
+            if not edge.req_type.is_build_requirement:
+                # not a build requirement
+                continue
+            visited.add(edge.key)
+            # it's a new ``[build-system].requires``.
+            yield edge.destination_node
+            # recursively get install dependencies of this build dep (depth first).
+            for install_edge in self._traverse_install_requirements(
```
dhellmann, line 152: The installation requirements of A are not part of its build requirements. We will be ready to build A before these things are built, and as you've pointed out elsewhere there will be cases where we _must_ because the installation dependencies can have cycles. This is why in #794 I separate the installation dependencies from the topological sorter.

dhellmann (commented):
I'm starting to think that instead of reproducing the topology of the graph, we should include order data in the graph file based on the bootstrap and then use that order.

Iterate over the nodes in the order they were built by the bootstrap step. For each node, if its build dependencies are available, queue it to build and remove it from the to-be-built list. Repeat until the list is empty.

That algorithm would give us an accurate order that we could build, because it did build during the bootstrap step. So we would just be parallelizing based on what has already been built, and not trying to figure out how to get the same traversal order out of the graph a second time.

LalatenduMohanty (approved):
I left a minor comment about the variable name change but it is not a blocker. If you like the suggestion please update the PR.

PR #800, opened 2025-10-02
feat(settings): add version-specific prebuilt wheel configuration

### LINKED ISSUE #801: Add version-specific prebuilt wheel configuration support
## Problem

Variants currently support only variant-wide `pre_built` and `wheel_server_url` settings, which apply to all versions of a package. This creates conflicts when different collections need different build methods for the same package.

**Example use case:**
- Global TPU collection needs `torch==2.9.0.dev20250730` as a prebuilt wheel from GitLab mirror
- torch-2.8.0 TPU collection needs `torch==2.8.0` built from source  
- Both use the same `tpu-ubi9` variant

Currently impossible to configure this scenario without creating duplicate variants.

## Solution

Add version-specific settings to allow per-version overrides within a single variant:

```yaml
variants:
  tpu-ubi9:
    pre_built: false  # Default
    versions:
      "2.9.0.dev20250730":
        pre_built: true
        wheel_server_url: https://gitlab.com/.../simple
      "2.8.0":
        pre_built: false
```

## Implementation

- Add `VersionSpecificSettings` model with `pre_built`, `wheel_server_url`, `env`, `annotations`
- Extend `VariantInfo` with `versions` mapping
- Add version-aware methods: `is_pre_built(version)`, `get_wheel_server_url(version)`
- Update bootstrapper to use version-specific resolution
- Maintain full backward compatibility

Version-specific settings take precedence over variant defaults when present.

pmattsso:
Enable fine-grained control over prebuilt vs source builds within variants by allowing per-version overrides. This resolves conflicts where different collections need different build methods for the same package version.

- Add VersionSpecificSettings model with pre_built and wheel_server_url
- Extend VariantInfo with versions mapping for version-specific overrides
- Update PackageBuildInfo with version-aware methods (is_pre_built, get_wheel_server_url)
- Modify bootstrapper, wheels, and build commands to use version-specific logic
- Add comprehensive tests validating precedence and backward compatibility
- Create detailed how-to documentation with examples and migration guide

Maintains full backward compatibility with existing variant-wide settings. Version-specific settings take precedence over variant defaults when present.

Fixes #801

### DIFF
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -88,11 +88,28 @@ class Bootstrapper:
             return self._resolved_requirements[req_str]
 
         pbi = self.ctx.package_build_info(req)
-        if pbi.pre_built:
-            source_url, resolved_version = self._resolve_prebuilt_with_history(
-                req=req,
-                req_type=req_type,
-            )
+
+        # Check if package has version-specific settings (any version-specific config)
+        variant_info = pbi._ps.variants.get(pbi._variant)
+        has_version_specific_prebuilt = (
+            variant_info and variant_info.versions and len(variant_info.versions) > 0
+        )
+
+        if pbi.pre_built or has_version_specific_prebuilt:
+            try:
+                source_url, resolved_version = self._resolve_prebuilt_with_history(
+                    req=req,
+                    req_type=req_type,
+                )
+            except Exception as e:
+                # Version-specific prebuilt resolution failed, fall back to source
+                logger.debug(
+                    f"{req.name}: prebuilt resolution failed, falling back to source: {e}"
+                )
+                source_url, resolved_version = self._resolve_source_with_history(
+                    req=req,
+                    req_type=req_type,
+                )
         else:
             source_url, resolved_version = self._resolve_source_with_history(
                 req=req,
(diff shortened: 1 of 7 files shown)


src/fromager/bootstrapper.py
```diff
@@ -88,11 +88,28 @@ def resolve_version(
...
         pbi = self.ctx.package_build_info(req)
-        if pbi.pre_built:
-            source_url, resolved_version = self._resolve_prebuilt_with_history(
-                req=req,
-                req_type=req_type,
-            )
+
+        # Check if package has version-specific settings (any version-specific config)
+        variant_info = pbi._ps.variants.get(pbi._variant)
+        has_version_specific_prebuilt = (
```
dhellmann, line 94: Oh, this is going to be messier than I thought.

The `_ps` field really should be private to the PBI. We could at least wrap this logic in a method, but it's odd to need to know about the version-specific prebuilt flags before we resolve the version we're going to bootstrapping.

I don't remember why we have different logic for resolving versions of prebuilt wheels. How different is that logic from the other flow? Does it skip some steps or are there completely different steps?

tiran:
Can we have a discussion about the design before we move forward? This approach is going to make the package settings more complex and will require API breaking chances. It's also not very flexible, because you'll have to add an entry for every version.

I think a hook function would give is more flexibility and is easier to implement.

PR #823, opened 2025-10-29
test(graph): fix constraint tests for Requirement objects

LalatenduMohanty:
test(graph): update constraint tests for Requirement type

Update test_graph.py to use Requirement objects instead of strings
for constraint parameters, following the API change in commit 718d2f3.

Changes:
- Pass Requirement objects to add_dependency() constraint parameter
- Update assertions to check Requirement type instead of string
- Change empty constraint checks from "" to None
- Account for canonical ordering in constraint string representation
  (e.g., ">=1.0,<3.0" becomes "<3.0,>=1.0")

All tests pass with mypy type checking enabled.

Fixes compatibility with: 718d2f3 (chore: use Requirement | None for constraint)

### DIFF
--- a/tests/test_graph.py
+++ b/tests/test_graph.py
@@ -161,12 +161,13 @@ def test_graph_add_dependency_with_constraint() -> None:
         req=Requirement("package-a>=1.0"),
         req_version=Version("2.0"),
         download_url="url",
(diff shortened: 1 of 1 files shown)


tiran (approved):
Thanks!

PR #828, opened 2025-10-29
fix(packagesettings): add explicit validator for config_settings

LalatenduMohanty:
Pydantic 2.12+ has stricter validation for union types in nested dictionary values. The previous implementation using `dict[str, str | list[str]]` type annotation alone was insufficient, causing validation errors like:

  ERROR: faiss-cpu: failed to load settings
  config_settings.`cmake.define.BLA_VENDOR`
    Input should be a valid list [type=list_type, input_value='OpenBLAS']

This commit adds an explicit BeforeValidator that properly handles both string and list[str] values in config_settings, following the same pattern used for other validated types in the codebase (e.g., BuildDirectory).

The validator:
- Accepts both str and list[str] values as intended by commit 98b98b8
- Validates list contents to ensure only strings
- Provides clear error messages for invalid types
- Maintains backward compatibility with Pydantic 2.11.x

Added comprehensive test coverage including:
- Mixed str and list[str] values
- String-only values
- List-only values
- Rejection of invalid types (int, mixed-type lists)

Fixes validation errors when using scikit-build-core and other tools that require string config_settings values.

### DIFF
--- a/src/fromager/packagesettings.py
+++ b/src/fromager/packagesettings.py
@@ -105,6 +105,48 @@ VariantChangelog = Mapping[PackageVersion, list[str]]
 # Annotations
 RawAnnotations = Mapping[str, str]
 
+# PEP 517 config_settings value type
+# Values can be either str or list[str]
+ConfigSettingsValue = str | list[str]
+
+
+def _validate_config_settings_dict(
+    v: dict[str, typing.Any],
+) -> dict[str, ConfigSettingsValue]:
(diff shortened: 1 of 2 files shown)


LalatenduMohanty:
> I don't understand why you are running into a problem. PR https://github.com/python-wheel-build/fromager/pull/822 includes a test for this scenario. Did you rebase your downstream MR to use Fromager 0.69.0?

You are right, I do not need this change.

PR #832, opened 2025-10-31
fix: adds support for duplicate constraints, checks for conflicts

### LINKED ISSUE #831: Constraints should be able to handle markers for different archs or python versions
The [constraints loader/validator](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/constraints.py) needs to consider the marker expressions when deciding if there is a duplicate constraint. For example, different CPU architectures or python versions may need different constraints.

Randori-Lance:
# Note to reviewer

Closes #831 

Please approach this PR with skepticism, I don't want to cause regressions

* Adds support for markers when parsing constraints
* Adds tests for new functionality

Multiple constraints should be considered valid when there are multiple unique markers:

```
foo==1.0.0; platform_machine != 'ppc64le'
foo==1.0.1; platform_machine == 'ppc64le'

```

### DIFF
--- a/tests/test_constraints.py
+++ b/tests/test_constraints.py
@@ -1,6 +1,8 @@
 import pathlib
+import typing
 
 import pytest
+from packaging import markers
 from packaging.requirements import Requirement
 from packaging.version import Version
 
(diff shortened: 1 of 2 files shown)


tests/test_constraints.py
```diff
@@ -37,15 +37,39 @@ def test_add_constraint_conflict():
...
         c.add_constraint("flit-core>2.0.0")
 
+    # Different, but equivalent markers should raise KeyError
+    with pytest.raises(KeyError):
+        c.add_constraint("bar==1.0; python_version >= '3.11' and platform_machine == 'x86_64'")
+        c.add_constraint("bar==1.1; platform_machine == 'x86_64' and python_version >= '3.11'")
+
+    # Same package with different markers should NOT raise error
+    c.add_constraint("baz==1.0; platform_machine != 'ppc64le'")
+    c.add_constraint("baz==1.1; platform_machine == 'ppc64le'")
```
dhellmann: It would be good to check here which constraint comes back out. The result should be stable since we don't have a ppc64le test job.

Randori-Lance: should be able to finagle some "if" statements to do that sort of assertion

Randori-Lance: oh, I get it... both would have the same result

Issue #350, opened 2024-08-26
Add default for  'extra_data_plugin' override


rd4398:
Based on comment by @dhellmann on PR:  https://github.com/python-wheel-build/fromager/pull/346#pullrequestreview-2255590297

```
A while back, @tiran added plugin references pointing to the default functions for all of our overrides so that consumers downstream can find those functions and check their function signatures against the customizations in the downstream build. I missed that when we added this new override without a default. We don't have to hold up this PR over it, but we should have an issue ticket to make sure we add that.
```

We must add plugin reference pointing to default function for this override

PR #834, opened 2025-11-03
fix: Add default for 'extra_data_plugin' override

iangelak:
In respect to issue #350, refactor the add_extra_metadata_to_wheels override to follow the standard pattern used by all other overrides in the codebase. This change implements the missing plugin reference that was noted when this override was initially added.

Changes:
- Updated default_add_extra_metadata_to_wheels to return an empty dict instead of raising NotImplementedError, providing a proper default implementation
- Refactored add_extra_metadata_to_wheels to use overrides.find_and_invoke instead of manually calling find_override_method and invoke separately
- Simplified code by removing redundant conditional logic

This ensures downstream consumers can find the default function and check function signatures against their customizations, consistent with all other override hooks in the project.

### DIFF
--- a/src/fromager/wheels.py
+++ b/src/fromager/wheels.py
@@ -142,7 +142,8 @@ def default_add_extra_metadata_to_wheels(
     sdist_root_dir: pathlib.Path,
     dist_info_dir: pathlib.Path,
 ) -> dict[str, typing.Any]:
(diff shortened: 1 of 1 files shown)


dhellmann (approved):
Nice cleanup, thank you!

PR #835, opened 2025-11-04
Add source hook helper docs and other docs fixes

mprpic:
This also fixes a couple unrendered examples in https://fromager.readthedocs.io/en/latest/customization.html#combining-both-settings.


docs/conf.py
```diff
@@ -15,7 +15,7 @@
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
 project = "Fromager"
-copyright = "2024, Fromager Authors"
+copyright = "%Y, Fromager Authors"
```
LalatenduMohanty: I am not sure if this will work. should we use following instead?
```
from datetime import datetime
copyright = f"{datetime.now().year}, Fromager Authors"
```

LalatenduMohanty (changes requested):

docs/conf.py
LalatenduMohanty: I tested this and looks like Sphinx automatically processes the copyright string through time.strftime(), so "%Y" is actually the correct way to have a dynamic year

LalatenduMohanty:
The PR looks good to me, however there 5 commits which I think can be merged to couple of commits. 

1st commit
```
docs(hooks): add source module helper functions documentation

- Document ensure_pkg_info, pep517_build_sdist, and unpack_source
- Improve docstrings in sources.py for autodoc
```
2nd commit

```
docs: improve documentation structure and rendering

- Fix literalinclude rendering in customization.md with MyST colon_fence
- Move skip-constraints example to multiple-versions how-to guide
- Update conf.py: fix typos, use dynamic copyright year, add MyST extension
```

Issue #808, opened 2025-10-20
Remove build_ext_parallel feature


tiran:
The option `build_ext_parallel` was introduced to configure distutils and setuptools to build extension modules in parallel. It creates a `dist-extra.cfg`, https://setuptools.pypa.io/en/latest/deprecated/distutils/configfile.html

```
[build_ext]
parallel = {extra_environ["MAX_JOBS"]}
```

In downstream we learned the hard way the parallel build feature is not safe. Several packages sometimes fail to build (Cython, numba). In one case (`psutil`) we even had a miss-compiled wheel with missing symbol. Let's remove the unsafe feature.

- Phase 1: remove the code, keep the option, emit a warning when the option is set.
- Phase 2: remove the option after a couple of releases.

PR #836, opened 2025-11-05
fix: Remove build_ext_parallel feature

iangelak:
Remove the build_ext_parallel feature which configures parallel builds of setuptools extensions via dist-extra.cfg. This feature was found to be unsafe due to race conditions in setuptools/distutils, causing:
- Intermittent build failures in packages like Cython and numba
- Miscompiled wheels with missing symbols (e.g., psutil)

Phase 1 of deprecation:
- Remove code that creates dist-extra.cfg and sets DIST_EXTRA_CONFIG
- Keep the build_ext_parallel option in the schema for compatibility
- Emit deprecation warning when the option is set to True
- Update documentation to mark the option as deprecated

The option will be completely removed in Phase 2 after a few releases.

Fixes: #808

### DIFF
--- a/src/fromager/packagesettings.py
+++ b/src/fromager/packagesettings.py
@@ -242,7 +242,7 @@ class BuildOptions(pydantic.BaseModel):
 
     ::
 
-        build_ext_parallel: False
+        build_ext_parallel: False  # DEPRECATED: ignored, will be removed
         cpu_cores_per_job: 1
         memory_per_job_gb: 1.0
     """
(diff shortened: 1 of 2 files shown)


src/fromager/packagesettings.py
```diff
@@ -252,8 +252,10 @@ class BuildOptions(pydantic.BaseModel):
     build_ext_parallel: bool = False
     """Configure `build_ext[parallel]` in `DIST_EXTRA_CONFIG`
 
-    This enables parallel builds of setuptools extensions. Incompatible
-    with some packages, e.g. numba 0.60.0.
+    .. deprecated:: 1.0
```
tiran: The version number should be the next release.

```suggestion
    .. deprecated:: 0.72.0
```

PR #837, opened 2025-11-06
chore: make Candidate a dataclass

tiran:
Convert the `Candidate` class to an immutable dataclass.

Instead of an URL to the metadata file, a candidate object now only holds an `has_metadata` flag. The metadata URL is always `self.url + 'metadata'`.

`extras` is now a sorted tuple instead of `Iterable[str] | None`. This simplifies the logic, too.

### DIFF
--- a/src/fromager/candidate.py
+++ b/src/fromager/candidate.py
@@ -1,3 +1,4 @@
+import dataclasses
 import logging
 import typing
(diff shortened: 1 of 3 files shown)


LalatenduMohanty:
Looks good to me. Once you resolve the conflict we should be able to merge this.

Issue #761, opened 2025-09-09
bootstrap-parallel does not work with git URLs


dhellmann:
The bootstrap-parallel command does not work with top level requirements specified as git URLs. This is an important feature for supporting nightly build pipelines.

The error reported looks like

```
16:26:32 ERROR Failed to build stevedore==5.4.1: Invalid URL '/Users/dhellmann/Devel/fromager/fromager/e2e-output/work-dir/stevedore-5.4.1/stevedore-5.4.1': No scheme supplied. Perhaps you meant https:///Users/dhellmann/Devel/fromager/fromager/e2e-output/work-dir/stevedore-5.4.1/stevedore-5.4.1?
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 3/3 [00:00<00:00, 34.84pkg/s]
16:26:32 ERROR ERROR: Invalid URL '/Users/dhellmann/Devel/fromager/fromager/e2e-output/work-dir/stevedore-5.4.1/stevedore-5.4.1': No scheme supplied. Perhaps you meant https:///Users/dhellmann/Devel/fromager/fromager/e2e-output/work-dir/stevedore-5.4.1/stevedore-5.4.1?
```

PR #842, opened 2025-11-10
fix: git URLs working with bootstrap-parallel

iangelak:
Build-parallel reconstructed requirements as name==version, losing git URL info and causing invalid URL errors.

Fixes: #761


src/fromager/dependency_graph.py
```diff
@@ -352,6 +352,19 @@ def get_nodes_by_name(self, req_name: str | None) -> list[DependencyNode]:
...
+    def get_top_level_requirement(self, node: DependencyNode) -> Requirement | None:
+        """Get the top-level requirement specification for a node.
+
+        For packages that were specified as top-level requirements (e.g., with git URLs),
+        this returns the original requirement specification. Returns None if the node
+        is not a direct child of ROOT.
+        """
+        root = self.get_root_node()
+        for edge in root.children:
+            if edge.destination_node.key == node.key:
```
dhellmann: I think you can say `edge.destination_node is node`?

You could also look through `node.parents` for an edge with a destination node that is the root node. I don't know which is likely to be more efficient in terms of iterative over fewer items on average.

src/fromager/sources.py
```diff
@@ -204,6 +204,17 @@ def download_git_source(
     if url_to_clone.startswith("git+"):
         url_to_clone = url_to_clone[len("git+") :]
 
+    # Parse ref from URL if it's in the form repo@ref
+    from urllib.parse import urlparse
+
+    if ref is None:
```
dhellmann: Where do we have a call that's cloning from a URL where the caller does not pass a valid ref?

LalatenduMohanty:
@iangelak The git commit message is very long and not necessary. What we want is few sentences which will explain what is this PR going fix. Please fix the commit message. Thanks.

src/fromager/sources.py
```diff
@@ -76,11 +77,32 @@ def download_source(
...
+        parsed_url = urlparse(parsing_url)
+        if "@" in parsed_url.path:
+            # Extract the ref from the path (e.g., /path/to/repo.git@5.2.0 -> @5.2.0)
+            new_path, _, git_ref = parsed_url.path.rpartition("@")
+            # Reconstruct URL without the @ref part
+            url_to_clone = parsed_url._replace(path=new_path).geturl()
+            # Add back git+ prefix if it was present
+            if req.url.startswith("git+"):
+                url_to_clone = f"git+{url_to_clone}"
+
```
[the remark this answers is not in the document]
LalatenduMohanty: After PR #871 is merged,  RetrieveMethod.from_url() can be  modified to handle this use case I think.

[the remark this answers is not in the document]
LalatenduMohanty: Resolving this as #871 has not merged yet.

docs/how-tos/build-from-git-repo.rst
```diff
@@ -14,6 +14,15 @@ you can do the following:
 This will clone the ``stevedore`` repository and build the package from the
 local copy.
 
+You can also use the ``bootstrap-parallel`` command for faster builds:
+
+.. code-block:: console
+
+   $ fromager bootstrap-parallel stevedore @ git+https://github.com/openstack/stevedore.git
```
dhellmann: I think this syntax isn't right because there are spaces. It would at least need to be quoted to be processed properly, as was done in the test script. Please submit a follow up PR to fix that.

```suggestion
   $ fromager "bootstrap-parallel stevedore @ git+https://github.com/openstack/stevedore.git"
```

dhellmann (commented):
@iangelak , thank you for this! I had a couple of minor comments, since @LalatenduMohanty tagged me after the merge. No big deals, but if you'd fix the doc typo before we tag a release that would be good.

Issue #753, opened 2025-09-05
debug output from uv is not prefixed with the package name so parallel builds are hard to debug


dhellmann:
Most log output uses the package name as a prefix so it is possible to tell which package is being processed when a message is logged. This is not true when we dump the output of the `uv` command to the log. We should fix that by logging each line of output separately instead of dumping the output in one call.

PR #843, opened 2025-11-11
fix: Add package prefix to uv output

iangelak:
External command output lacked package prefixes on continuation lines, making logs hard to grep in parallel builds.

Solution: Added get_log_prefix() helper to extract package context. FromagerLogRecord prefixes the first line, str.replace() handles continuation lines.

Fixes: #753

### DIFF
--- a/src/fromager/external_commands.py
+++ b/src/fromager/external_commands.py
@@ -7,6 +7,8 @@ import sys
 import typing
 from io import TextIOWrapper
 
+from . import log
+
 logger = logging.getLogger(__name__)
 
 HERE = pathlib.Path(__file__).absolute().parent
@@ -98,8 +100,33 @@ def run(
             stdin=stdin,
         )
         output = completed.stdout.decode("utf-8") if completed.stdout else ""
+
+    # Add package prefix to continuation lines for greppability
+    prefix = log.get_log_prefix()
+    formatted_output = None
+
+    if output:
+        if prefix:
+            # FromagerLogRecord handles first line, we handle continuation lines
+            formatted_output = output.rstrip("\n").replace("\n", f"\n{prefix}: ")
+        else:
+            formatted_output = output
+
     if completed.returncode != 0:
-        logger.error("%s failed with %s", cmd, output)
+        # Prefix first line for error output (embedded in larger message)
+        if formatted_output and prefix:
+            output_to_log = f"\n{prefix}: {formatted_output}"
+        elif formatted_output:
+            output_to_log = f"\n{formatted_output}"
+        else:
+            output_to_log = ""
+        logger.error(
+            "command failed with exit code %d: %s%s",
+            completed.returncode,
+            shlex.join(cmd),
+            output_to_log,
+        )
+
         err_type = subprocess.CalledProcessError
         if network_isolation:
             # Look for a few common messages that mean there is a network
@@ -113,5 +140,9 @@ def run(
                 if substr in output:
                     err_type = NetworkIsolationError
         raise err_type(completed.returncode, cmd, output)
-    logger.debug("output: %s", output)
+
+    # Log command output for debugging
+    if formatted_output:
+        logger.debug(formatted_output)
+
     return output
--- a/src/fromager/log.py
+++ b/src/fromager/log.py
@@ -17,6 +17,27 @@ requirement_ctxvar: contextvars.ContextVar[Requirement] = contextvars.ContextVar
 version_ctxvar: contextvars.ContextVar[Version] = contextvars.ContextVar("version")
 
 
+def get_log_prefix() -> str | None:
+    """Get the log prefix based on current context (requirement and version).
+
+    Returns:
+        - "name-version" if both requirement and version are available
+        - "name" if only requirement is available
+        - None if no context is available
+    """
+    try:
+        req = requirement_ctxvar.get()
+    except LookupError:
+        return None
+    else:
+        try:
+            version = version_ctxvar.get()
+        except LookupError:
+            return req.name
+        else:
+            return f"{req.name}-{version}"
+
+
 @contextlib.contextmanager
 def req_ctxvar_context(
     req: Requirement, version: Version | None = None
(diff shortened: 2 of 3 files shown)


dhellmann:
When I run an e2e test with this change, I do see the prefix as expected:

```
2025-11-11 18:22:11,813 DEBUG:fromager.external_commands:73: setuptools: running:  uv venv --verbose --python '/Users/dhellmann/Library/Application Support/hatch/env/virtual/fromager/V7Nl9Nl1/e2e/bin/python' --no-project /Users/dhellmann/Devel/fromager/fromager/e2e-output/work-dir/setuptools-80.9.0/build-3.14.0 in .
2025-11-11 18:22:11,828 DEBUG:fromager.external_commands:130: setuptools: DEBUG uv 0.9.8 (85c5d3228 2025-11-07)
2025-11-11 18:22:11,828 DEBUG:fromager.external_commands:130: setuptools: DEBUG Acquired shared lock for `/Users/dhellmann/.cache/uv`
2025-11-11 18:22:11,828 DEBUG:fromager.external_commands:130: setuptools: DEBUG Using Python request `/Users/dhellmann/Library/Application Support/hatch/env/virtual/fromager/V7Nl9Nl1/e2e/bin/python` from explicit request
2025-11-11 18:22:11,828 DEBUG:fromager.external_commands:130: setuptools: DEBUG Checking for Python interpreter at path `/Users/dhellmann/Library/Application Support/hatch/env/virtual/fromager/V7Nl9Nl1/e2e/bin/python`
2025-11-11 18:22:11,828 DEBUG:fromager.external_commands:130: setuptools: Using CPython 3.14.0 interpreter at: /Users/dhellmann/Library/Application Support/hatch/env/virtual/fromager/V7Nl9Nl1/e2e/bin/python
2025-11-11 18:22:11,828 DEBUG:fromager.external_commands:130: setuptools: Creating virtual environment at: e2e-output/work-dir/setuptools-80.9.0/build-3.14.0
2025-11-11 18:22:11,828 DEBUG:fromager.external_commands:130: setuptools: DEBUG Using base executable for virtual environment: /opt/homebrew/Cellar/python@3.14/3.14.0_1/Frameworks/Python.framework/Versions/3.14/bin/python3.14
2025-11-11 18:22:11,828 DEBUG:fromager.external_commands:130: setuptools: Activate with: source e2e-output/work-dir/setuptools-80.9.0/build-3.14.0/bin/activate
2025-11-11 18:22:11,828 DEBUG:fromager.external_commands:130: setuptools: DEBUG Released lock at `/Users/dhellmann/.cache/uv/.lock`
2025-11-11 18:22:11,828 INFO:fromager.build_environment:189: setuptools: created build environment in /Users/dhellmann/Devel/fromager/fromager/e2e-output/work-dir/setuptools-80.9.0/build-3.14.0
```

What can we do to make it more obvious that those lines are the output of the uv command, and not fromager itself?

dhellmann:
> You could add a lock, but then you'd have to figure out how to do proper locking without blocking other threads.
> 
> There is a much, much simpler way to solve the problem: rewrite the `output`, then log the entire output in one logger call. You can archive this with just one `str` method.

Yeah, I like this approach.

tiran:
Yes, something like that. But instead of splitting the string, you can archive the goal with just str method call.

src/fromager/external_commands.py
```diff
@@ -113,5 +122,9 @@ def run(
                 if substr in output:
                     err_type = NetworkIsolationError
         raise err_type(completed.returncode, cmd, output)
-    logger.debug("output: %s", output)
+
+    # Log command output with visual prefix to distinguish from fromager output
+    if output:
+        logger.debug("  > " + output.replace("\n", "\n  > "))
```
dhellmann: Same here, include the package name on each line.

src/fromager/external_commands.py
```diff
@@ -98,8 +100,31 @@ def run(
...
+        output_prefix = line_prefix + "  > "
+        prefixed_output = "  > " + output.replace("\n", f"\n{output_prefix}")
+        formatted_output = f"{line_prefix}{prefixed_output}"
+
     if completed.returncode != 0:
-        logger.error("%s failed with %s", cmd, output)
+        # Build complete error message with command, exit code, and output
+        message = (
+            f"command failed with exit code {completed.returncode}: {shlex.join(cmd)}"
+        )
```
tiran: Don't use f-strings to format log output. You should always pass a format string and individual arguments to a logger function.

tiran (changes requested):

LalatenduMohanty:
@iangelak The commit message is too long, also there are more comments in the code than required. We do not need comments unless it is a must have. Also lets reduce the commit message to 2 to 3 lines which is more human readable than long texts.

src/fromager/external_commands.py
```diff
@@ -98,8 +100,30 @@ def run(
             stdin=stdin,
         )
         output = completed.stdout.decode("utf-8") if completed.stdout else ""
+
+    # Prefix output lines with package name/version for greppability
+    formatted_output = None
+    if output:
+        prefix = log.get_log_prefix()
```
LalatenduMohanty: Looks like the code has a double prefix on the debug output. 

For example the code manually adds "numpy-1.0.0:" to each line of command output, then passes it to logger.debug(). But FromagerLogRecord (set as the log factory in __main__.py:172) automatically adds "numpy-1.0.0:" to every log message, so the first line ends up with the prefix twice: numpy-1.0.0: numpy-1.0.0: line1.

LalatenduMohanty (changes requested):

src/fromager/external_commands.py
LalatenduMohanty:
```
external_commands.py          log.py (FromagerLogRecord)
        │                              │
        │  adds prefix manually        │  adds prefix automatically
        │  ─────────────────►          │  ─────────────────────►
        │                              │
        ▼                              ▼
"numpy-1.0.0: line1"    →    "numpy-1.0.0: numpy-1.0.0: line1"
```

iangelak: Thank you for pointing that out. I fixed the issue.

dhellmann:
This implementation looks good to me. I'll leave the PR open because @LalatenduMohanty and @tiran both had comments as well, but let's try to get this merged by early next week so we can cut a release.

LalatenduMohanty:
@iangelak Thanks for fixing the commits. Appreciate it.

dhellmann (approved):

src/fromager/external_commands.py
dhellmann: This is fixed in the latest version so I am resolving this thread

Issue #845, opened 2025-11-14
[Bug]: Fromager no longer mentions the package name near the exception


rd4398:
Fromager no longer mentions the package name that failed and cause exception while building the wheel. This makes it difficult to debug a failure. 

This was observed downstream when our CI tried to build `tree-sitter-c==0.23.6` package.

It failed with error:

```
src/parser.c:3:10: fatal error: tree_sitter/parser.h: No such file or directory
    3 | #include "tree_sitter/parser.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
error: command '/usr/lib64/ccache/gcc' failed with exit code 1
 68%|██████████████████████████▍            | 99/146 [04:33<02:09,  2.77s/pkg]
07:27:45 ERROR ERROR: Command '['/opt/app-root/lib64/python3.12/site-packages/fromager/run_network_isolation.sh', '/mnt/work-dir/tree_sitter_c-0.23.6/build-3.12.9/bin/python3', '/opt/app-root/lib64/python3.12/site-packages/pyproject_hooks/_in_process/_in_process.py', 'build_wheel', '/tmp/tmpoz0uq7jq']' returned non-zero exit status 1.

```

PR #846, opened 2025-11-14
fix: include package name in exception messages

andre-motta:
Add package name context to exception formatting to identify which
package failed during build errors. The package name is added by
FromagerLogRecord when messages are logged, and _format_exception
handles formatting chained exceptions with 'because' syntax.

Improve parallel build error handling to preserve package context
across thread boundaries using RuntimeError wrapper.

Tests added to test_external_commands.py verify error logging and
exception formatting include package names and handle chained exceptions.

Update e2e test pattern to match new log format that includes package names.

Closes: #845

### DIFF
--- a/e2e/test_bootstrap_constraints.sh
+++ b/e2e/test_bootstrap_constraints.sh
@@ -28,7 +28,7 @@ fromager \
 pass=true
 
 # Check for log message that the override is loaded
-if ! grep -q "ERROR: Unable to resolve requirement specifier stevedore==5.2.0 with constraint stevedore==4.0.0" "$OUTDIR/bootstrap.log"; then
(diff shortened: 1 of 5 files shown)


LalatenduMohanty (changes requested):
@andre-motta src/fromager/commands/download_sequence.py do not use RuntimeError() wrapper. Lets fix that as well

LalatenduMohanty:
@andre-motta Also 3 e2e tests are failing. PTAL

andre-motta:
Addressed all feedback

dhellmann (commented):
It looks like the test job failures are due to the changes in the message format. That should be easy enough to fix. The change itself looks good.

dhellmann:
I think we could land this if it was rebased. Some of the other suggested improvements could come later.

Issue #874, opened 2025-12-03
linter incorrectly limits duplicate requirements entries


dhellmann:
The lint-requirements command requires a single unique entry for each entry in a requirements.txt file based on the package name and the marker expression. That rule should only apply to constraints.txt files, not the requirements.txt files, because we want to allow and encourage building multiple versions of a package by specifying entries in the requirements.txt file.

The logic in https://github.com/python-wheel-build/fromager/blob/main/src/fromager/commands/lint_requirements.py#L29 will need to change to only apply the uniqueness rules when looking at a constraints.txt file.

PR #875, opened 2025-12-03
fix(lint-requirements): allow duplicates in requirements.txt

rd4398:
The lint-requirements command was incorrectly enforcing uniqueness rules
  on requirements.txt files. The uniqueness constraint (one entry per package
  name and marker combination) should only apply to constraints.txt files,
  not requirements.txt files.

  This was preventing users from specifying multiple versions of the same
  package in requirements.txt files, which is needed to support building
  multiple versions of a package.

  Changes:
  - Move uniqueness validation inside the is_constraints check so it only applies to files ending with constraints.txt
  - requirements.txt files can now contain duplicate package names with different versions or markers
  - Add comprehensive test coverage for both requirements.txt and constraints.txt validation scenarios

  Generated with [Claude Code](https://claude.com/claude-code)

  Co-Authored-By: Claude <noreply@anthropic.com>

Fixes https://github.com/python-wheel-build/fromager/issues/874

### DIFF
--- a/e2e/test_lint_requirements.sh
+++ b/e2e/test_lint_requirements.sh
@@ -36,10 +36,10 @@ if fromager lint-requirements --resolve-requirements "$SCRIPTDIR/validate_inputs
     pass=false
 fi;
 
-# Test to demonstrate that command reports error for duplicate entries in files
+# Test to demonstrate that command accepts duplicate entries in requirements files
 
-if fromager lint-requirements --resolve-requirements "$SCRIPTDIR/validate_inputs/constraints.txt" "$SCRIPTDIR/validate_inputs/duplicate-requirements.txt"; then
-    echo "duplicate entries in files should have been recognized by the command" 1>&2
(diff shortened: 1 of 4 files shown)


LalatenduMohanty (changes requested):
The e2e test at e2e/test_lint_requirements.sh lines 39-44 still expects duplicates in requirements.txt to FAIL, but with this change they should PASS:

LalatenduMohanty:
> The e2e test at e2e/test_lint_requirements.sh lines 39-44 still expects duplicates in requirements.txt to FAIL, but with this change they should PASS:

I am checking how the e2e passed in the presubmit runs.

rd4398:
> The e2e test at e2e/test_lint_requirements.sh lines 39-44 still expects duplicates in requirements.txt to FAIL, but with this change they should PASS:

@LalatenduMohanty I have pushed the fix in latest commit

LalatenduMohanty:
I am suggesting https://www.conventionalcommits.org/en/v1.0.0/ in my [PR ](https://github.com/python-wheel-build/fromager/pull/847/)for contributing.md . Descriptive commit messages help only when it is required. For example if the issue is complex and the descriptive commit message would help us understand the code. For this PR it is not necessary IMO.

PR #893, opened 2025-12-19
Test mode post merge cleanups

LalatenduMohanty:
Contains following commits
* refactor(bootstrapper): move resolution and tracking to bootstrap()
* refactor(tests): use existing tmp_context fixture in test-mode tests

Followup clean ups suggested by @dhellmann during review of https://github.com/python-wheel-build/fromager/pull/865 . There is one more cleanup but it will be larger refactor, so planning to send another PR for that.

### DIFF
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -227,45 +227,74 @@ class Bootstrapper:
         In test mode, catches build exceptions, records package name, and continues.
         In normal mode, raises exceptions immediately (fail-fast).
         """
+        logger.info(f"bootstrapping {req} as {req_type} dependency of {self.why[-1:]}")
+
+        # Resolve version first so we have it for error reporting.
+        # In test mode, record resolution failures and continue.
         try:
-            self._bootstrap_impl(req, req_type)
+            source_url, resolved_version = self.resolve_version(
+                req=req,
+                req_type=req_type,
+            )
         except Exception as err:
             if not self.test_mode:
                 raise
-            # Get version from cache if available
-            cached = self._resolved_requirements.get(str(req))
-            if cached:
-                _source_url, resolved_version = cached
-                version = str(resolved_version)
-            else:
-                version = None
-            self._record_test_mode_failure(req, version, err, "bootstrap")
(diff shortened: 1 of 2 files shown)


src/fromager/bootstrapper.py
```diff
@@ -227,45 +227,67 @@ def bootstrap(self, req: Requirement, req_type: RequirementType) -> None:
...
-                raise
-            # Get version from cache if available
-            cached = self._resolved_requirements.get(str(req))
-            if cached:
-                _source_url, resolved_version = cached
-                version = str(resolved_version)
-            else:
-                version = None
-            self._record_test_mode_failure(req, version, err, "bootstrap")
+        logger.info(f"bootstrapping {req} as {req_type} dependency of {self.why[-1:]}")
```
dhellmann: The resolver logic needs to be inside the try/except, since it is still a common source of failures.

src/fromager/bootstrapper.py
dhellmann: Maybe this logic could move to `bootstrap` too? Then this internal function becomes the "actually do the bootstrap" function and the outer one is managing the setup and error handling.

src/fromager/bootstrapper.py
LalatenduMohanty: This is a good suggestion. Will fix it.

LalatenduMohanty: Fixed it in https://github.com/python-wheel-build/fromager/commit/400544aaf7d85bf16f5c14c0690f1df47160e1d6

src/fromager/bootstrapper.py
LalatenduMohanty: Fixed this in a separate commit in https://github.com/python-wheel-build/fromager/pull/893/commits/bbad8fae7e4697a25970387899d4c4a64c1a0839

dhellmann (approved):
We should add an e2e test for something where bootstrapping fails, too.

LalatenduMohanty:
> We should add an e2e test for something where bootstrapping fails, too.

@dhellmann I have created https://github.com/python-wheel-build/fromager/issues/895 to track the e2e test

PR #894, opened 2025-12-19
refactor(bootstrapper): replace `_handle_test_mode_failure` with phase re-entry

### LINKED ISSUE #892: Clean ups in --test-mode code
There were some followup clean ups suggested by @dhellmann during review of https://github.com/python-wheel-build/fromager/pull/865. This issue tracks these.

1. Use existing tmp_context fixture
  - File: tests/test_bootstrap_test_mode.py
  - Comment: There's already a fixture for making a temporary context. If you name an argument to a test method tmp_context you get a context in a temporary directory that is also managed by pytest.
  - Reference: https://github.com/python-wheel-build/fromager/blob/main/tests/conftest.py#L24
  
2. Move resolution step from _bootstrap_impl to bootstrap
- File: src/fromager/bootstrapper.py (line ~232)
- Comment: "We wouldn't need this logic [getting version from cache] if we move the resolution step out of _bootstrap_impl into this function where we can see whether it resolved in the first place."

3. Move _track_why context manager to bootstrap
- File: src/fromager/bootstrapper.py (line ~295)
- Comment: "If we move the resolution logic up into bootstrap then this context manager can also move there because we will have the resolved version."

4. Consider force_prebuilt flag to eliminate _handle_test_mode_failure
- File: src/fromager/bootstrapper.py (line ~848)
- Comment: "If we add a force_prebuilt flag to _bootstrap_impl then we wouldn't need this function and we could avoid some duplication of the logic for handling pre-built wheels."

LalatenduMohanty:
## Summary

- Replace `_handle_test_mode_failure` with phase re-entry: on build failure in test mode, reset the `WorkItem` to `PREPARE_SOURCE` with `pbi_pre_built=True` instead of duplicating the prebuilt download logic in a separate method.
- Add `is_test_mode_fallback` flag to `WorkItem` to prevent infinite fallback loops.

Closes #892

### DIFF
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -139,6 +139,7 @@ class WorkItem:
     cached_wheel_filename: pathlib.Path | None = None
     build_result: SourceBuildResult | None = None
     pbi_pre_built: bool = False
+    is_test_mode_fallback: bool = False
     build_system_deps: set[Requirement] = dataclasses.field(default_factory=set)
     build_backend_deps: set[Requirement] = dataclasses.field(default_factory=set)
     build_sdist_deps: set[Requirement] = dataclasses.field(default_factory=set)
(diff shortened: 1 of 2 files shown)


src/fromager/bootstrapper.py
```diff
@@ -615,9 +683,12 @@ def _handle_build_requirements(
 
         for dep in self._sort_requirements(build_dependencies):
```
dhellmann: We have several loops doing this same thing now. Maybe a future PR could refactor those into a method to bootstrap an iterable of requirements. Don't update this one with that change, though.

rd4398 (commented):
This looks generally good to me. I see Doug has some comments. I will wait for them to be resolved before I can approve this PR

src/fromager/bootstrapper.py
```diff
@@ -1754,18 +1685,56 @@ def _handle_phase_error(
...
+                    err,
                 )
-                if fallback is not None:
-                    item.build_result = fallback
-                    item.phase = BootstrapPhase.PROCESS_INSTALL_DEPS
+                try:
+                    # why[-1] is the current item (pushed by _track_why),
+                    # so the real parent is why[-2].
+                    parent_req = self.why[-2][1] if len(self.why) > 1 else None
+                    results = self._resolver.resolve(
```
dhellmann, line 1701: Maybe we could go all the way back to the resolve state?

src/fromager/bootstrapper.py
```diff
@@ -1754,18 +1685,56 @@ def _handle_phase_error(
...
+                    item.source_url = wheel_url
+                    item.resolved_version = fallback_version
+                    item.pbi_pre_built = True
+                    item.is_test_mode_fallback = True
+                    item.phase = BootstrapPhase.PREPARE_SOURCE
+                    item.build_env = None
+                    item.sdist_root_dir = None
+                    item.unpack_dir = None
+                    item.cached_wheel_filename = None
+                    item.build_result = None
```
dhellmann, line 1725: Do we need to do any cleanup, or do the phases remove existing files before they start?

LalatenduMohanty: I am closing the PR as the code has changed a lot . I do not think we need any cleanup at this point.

PR #901, opened 2026-01-06
Add benchmarks framework

### LINKED ISSUE #734: Add benchmark scripts to easily track and test performence following changes
We should work on creating benchmarks to crucial and performance-heavy tasks, in order to easily track performance and test how changes improved / degraded performance easily.
We can even add a CI step at a later stage that reports automatically in PRs.

All benchmarks should have mocked requests (perhaps besides some specific benchmarks that test download speeds). They should basically be treated similarly to unit-tests, with anything external that may change and affect the benchmark being mocked. 
So, for example, when building the dependency tree for a package in a benchmark, we want the dependency tree for each package to be consistent and not change between different runs of the benchmarks (so that we can have an accurate comparison), as dependent packages get updated over time.

We can utilize tools like [pyperf](https://github.com/psf/pyperf) to manage and run the benchmarks.

MichaelYochpaz:
Add performance benchmarking infrastructure for fromager including GitHub Actions workflows for nightly and on-demand runs.

See the newly created README for full information.

Closes #734

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -1,0 +1,407 @@
+# Fromager Benchmarks
+
+Performance benchmarks for Fromager, a tool for rebuilding complete dependency trees of Python wheels from source.
+
+## Table of Contents
+
+- [Why Benchmarks?](#why-benchmarks)
+- [Quick Start](#quick-start)
+- [Understanding the Output](#understanding-the-output)
+- [How It Works](#how-it-works)
+  - [Local Benchmarking with pytest-benchmark](#local-benchmarking-with-pytest-benchmark)
+  - [CI Benchmarking with CodSpeed](#ci-benchmarking-with-codspeed)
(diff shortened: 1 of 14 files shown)


benchmarks/README.md
```diff
@@ -0,0 +1,186 @@
...
+---
+
+## Quick Start
+
+```bash
+# Install dependencies
+uv sync --extra benchmark
+
+# Run all benchmarks
+uv run pytest benchmarks/
```
dhellmann: We've standardized on hatch as the main entry point for everything else in this repo. How would someone run these tests through hatch?

dhellmann: Modifying `pyproject.toml` to install the `benchmark` extras in the `test` env and then running `hatch run test:pytest benchmarks/` did it.

It would be nice to have an alias, though, something like `hatch run benchmark`.

PR #903, opened 2026-01-13
build(lint): consolidate linters under pre-commit

MichaelYochpaz:
This PR migrates all linting tools to pre-commit for unified local and CI execution.

This allows developers to run `hatch run lint:precommit` (or `pre-commit run`) to check for lint issues locally and auto-fix them locally, or even install git pre-commit hooks that will auto-fix them before making a commit.

**Notes:**
- `Super-Linter` which had Markdown linting (already covered by `markdownlint`) and editorconfig validation (now covered by `editorconfig-checker`) was removed.
- The `.markdownlint-config.yaml` config file was renamed to `.markdownlint.yaml` to follow the official naming convention, which is auto-used by the CLI tool used ([reference](https://github.com/DavidAnson/markdownlint/blob/main/schema/.markdownlint.yaml)). Beforehand the name was a "made up" name that was referenced by using `--config .markdownlint-config.yaml`.
- `markdownlint` required Node.js to work (we install it in the CI beforehand). The pre-commit implementation here makes so that it's required on the CI, but for local runs the `markdownlint` check is skipped (with a warning) if Node.js isn't installed.
- There were some unrelated files that were slightly modified. Those are files that were picked up by the pre-commit and fixed. If not updated (or more accurately, fixed to follow or linting rules), they would change whenever pre-commit is ran.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,8 @@
+# This file lists commits that should be ignored by `git blame`.
+# To read more, see: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view
+
+# MDFormat Markdown formatting alignment
+0be84aaf1f350aae7a40e1dc7d6fc07711d87277
+
+# Ruff formatting alignment
+f81e85ca9f02941cc012e82c189db417d4fed9ca


scripts/conditional_hook.py
```diff
@@ -0,0 +1,154 @@
...
+        return result.returncode
+    except FileNotFoundError:
+        print_error(f"Executable '{executable}' not found during execution")
+        return 1
+    except PermissionError:
+        print_error(f"Permission denied executing '{executable}'")
+        return 1
+    except Exception as e:
+        print_error(f"Failed to execute '{executable}': {e}")
+        return 1
```
LalatenduMohanty: We should consider catching specific exceptions or at minimum lets add a comment explaining why the broad catch is acceptable here

LalatenduMohanty:
Pre-commit is a significant change in developer workflow, we can skip the pre-commit failures in the local workflow incase we want to focus on the code chnages initially i.e. `git commit --no-verify -m "wip: work in progress"`. However I am not sure if we want to adopt this workflow.  I need to use this workflow and findout more. 

```
Developer Workflow WITHOUT Pre-commit:
┌──────────┐    ┌──────────┐    ┌──────────┐    ┌──────────┐
│  Write   │ ─► │  Commit  │ ─► │   Push   │ ─► │ CI Fails │ 😞
│  Code    │    │ (no check)│   │          │    │ (lint!)  │
└──────────┘    └──────────┘    └──────────┘    └──────────┘

Developer Workflow WITH Pre-commit:
┌──────────┐    ┌──────────┐    ┌──────────┐    ┌──────────┐
│  Write   │ ─► │  Commit  │ ─► │   Push   │ ─► │ CI Pass  │ ✅
│  Code    │    │ BLOCKED! │    │ (clean)  │    │          │
└──────────┘    └──────────┘    └──────────┘    └──────────┘
                     │
                Fix & retry
```

MichaelYochpaz:
@tiran We already have pre-commit used here on Fromager. @dhellmann added it to the repo. You can see I'm just modifying the existing pre-commit config file and not creating a new one..

I'm just consolidating the linters under the same tool, and making it easier to run the linters we only have on the CI, locally.
Running pre-commit and having the official linters fix linting issues for you locally is much more convenient over pushing, waiting for the CI to run, reading the failures, and manually fixing them.

What training would developers need? It's literally just a single command to run.
If anything, using tox would require a lot more learning and training from developers.

.pre-commit-config.yaml
```diff
@@ -1,6 +1,13 @@
+# Pre-commit configuration
+# Run locally: pre-commit run --all-files
+# CI runs via: hatch run lint:precommit
+
 repos:
+  # ---------------------------------------------------------------------------
+  # 1. Standard file checks (fastest, run first)
+  # ---------------------------------------------------------------------------
```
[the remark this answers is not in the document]
LalatenduMohanty: @MichaelYochpaz  It is a nit pick, can you ask your AI agent to add only must have comments. We need comments only when the section is hard to understand, so that a comment will help with understanding if someone-going through it.  Just not this instance of comments, but I saw some other comments which we do not actually need. Specially it is easier to understand why the code block is there with the help of a AI agent.

scripts/conditional_hook.py
LalatenduMohanty: Ack, sounds good to me.

MichaelYochpaz:
@dhellmann Yeah this allows a full sync between linters running locally and on our CI.
This is good for AI agents (they can run the full linting and unit-testing process locally with a single command and not a shopping list of different tools to run sequentially), but also humans.
I opened this after I got a long list of Markdown list errors that just referred to specific lines that I had to manually find and fix. Now we can auto-fix these locally in seconds before even pushing.

To answer your question about markdownlint: It requires Node.js to run (we install Node in our CI).
This is tricky because I want pre-commit to also work for users who don't have Node.js installed.
We could make it an optional step, but then there wouldn't be a clear message for users to know they're missing out on this linter / how to fix it (they won't know they need to install Node.js), and also, the CI (which runs the same pre-commit) would skip this lint step in case of an error because it's optional.

I thought of using containers, but than we would just require Docker / Podman instead, and it would only work with one of them (e.g., if write it as a Podman command, it won't work for Docker users). Thought it's probably too much maintenance and overhead, with many possible issues and edge cases.

The solution I came up with (which also answers your review note) is to use a script - [scripts/conditional_hook.py](https://github.com/MichaelYochpaz/fromager/blob/c25868902d7bd35f8c6f6d5d0e63722321c13452/scripts/conditional_hook.py). This script will always run on pre-commit and is required. Then inside the script, if Node.js isn't installed, there are two cases:
1. If running on the CI (detected using environment variables) - running markdownlint is required and the script will fail if Node.js is missing.
2. Otherwise (running locally) - if Node.js isn't installed, it will print a nice warning message telling users to install Node.js, and then `makdownlint-cli2` using `npx`, then it will continue to run the other linters.

However... I just found [mdformat](https://github.com/hukkin/mdformat), which is a Python-based Markdown linter with native pre-commit hooks. If we're OK with switching to another Markdown linter (which also might change some Markdown files to slightly different formatting rules this linter follows), I think this could be a much simpler and cleaner solution.

dhellmann:
> I thought of using containers, but than we would just require Docker / Podman instead, and it would only work with one of them (e.g., if write it as a Podman command, it won't work for Docker users). Thought it's probably too much maintenance and overhead, with many possible issues and edge cases.

Eh, we could just pick podman.

> However... I just found [mdformat](https://github.com/hukkin/mdformat), which is a Python-based Markdown linter with native pre-commit hooks. If we're OK with switching to another Markdown linter (which also might change some Markdown files to slightly different formatting rules this linter follows), I think this could be a much simpler and cleaner solution.

That's definitely worth looking at. Is the project maintained? How different is the output?

.pre-commit-config.yaml
```diff
@@ -1,26 +1,63 @@
...
+              dist/.*
+          )$
+
+  - repo: https://github.com/hukkin/mdformat
+    rev: 1.0.0
+    hooks:
+      - id: mdformat
+        additional_dependencies:
+          - mdformat-gfm  # GitHub Flavored Markdown support (tables, autolinks, etc.)
+          - mdformat-simple-breaks  # Use --- for thematic breaks instead of 70 underscores
```
tiran: This linter and formatter seems to replace all `*` bullet points with `-`. It's a lot of noise that interferes with git history and `git annotate`. The asterisk is a valid character for bullet points. I actually prefer it over dash, because it sticks out.

dhellmann:
> FWIW, I'm not a fan of pre-commit. I prefer to have linters and tests in the same automation tool, so I can run all checks in parallel from the same tool with a single command. I'm still maintaining my own `tox.ini` for Fromager and run ruff, linters, doc builds, and unit tests in parallel with `tox p`.

If you want to put that back in place, I would be OK with having one of these tools call the other. Let's deal with it in another PR, though.

PR #905, opened 2026-01-16
feat(cli): add option for a single output-dir and hide options

### LINKED ISSUE #721: Simplify output directory handling
Fromager uses several flags to control the output directories. We thought we might save some for reuse across runs with different arguments, but that is proving to not be needed. We should simplify the UI by using a single flag for the output directories, and letting the context name subdirectories for different purposes.

sejohnst:
- Adds a new option to the cli for specifying an output directory. When passed, directories named sdists-repo, wheels-repo, and work-dir are created in that location
- When output-dir is passed, it will overwrite any arguments passed for sdists-repo, wheels-repo, and work-dir
- default behavior for output directories remains intact
- options for sdists-repo, wheels-repo and work-dir remain intact but hidden in favor of the new option

Fixes #721

### DIFF
--- a/src/fromager/__main__.py
+++ b/src/fromager/__main__.py
@@ -66,29 +66,15 @@ else:
     help="save error messages to a file",
 )
 @click.option(
-    "-o",
-    "--sdists-repo",
-    default=pathlib.Path("sdists-repo"),
-    type=clickext.ClickPath(),
-    help="location to manage source distributions",
-)
-@click.option(
-    "-w",
-    "--wheels-repo",
-    default=pathlib.Path("wheels-repo"),
-    type=clickext.ClickPath(),
-    help="location to manage wheel repository",
-)
-@click.option(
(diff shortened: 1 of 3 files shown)


src/fromager/__main__.py
```diff
@@ -143,6 +129,31 @@
     help="Build sdist and wheen with network isolation (unshare -cn)",
     show_default=True,
 )
+# hidden in favor of -d, --output-dir
+@click.option(
+    "-o",
+    "--sdists-repo",
+    type=clickext.ClickPath(),
+    default=pathlib.Path("sdists-repo"),
```
dhellmann: I think we want the defaults for all of these old options to be `None` so we can tell if the user provided values.

rd4398:
@sejohnst this PR is not updated in more than 3 months. Please let me know if you need any help with this.

LalatenduMohanty:
Preceded by #1099

PR #906, opened 2026-01-23
feat: add 'fromager resolve package' subcommand

tiran:
The new `resole package` subcommand is designed to debug Fromager settings and resolver properties. It compares the results from Fromager's internal resolver with a simple PyPI lookup. It prints informations like resolver configuration, detected versions and versions with sdists and wheels on PyPI.

```console
$ fromager --variant cpu package resolve 'fromager~=0.68.0'
resolving requirement for 'fromager~=0.68.0' (variant: cpu)
Package has not global constraint.

Package uses standard settings.
- sdist server url: https://pypi.org/simple
- resolver includes sdists: True
- resolver includes wheels: False
- wheel server url: None
- download url: None
- prebuilt wheel: False
Using distribution type: default (sdists: True, wheels: False)

get_resolver_provider returns provider 'PyPIProvider'
found 2 Fromager candidates with 2 unique versions for req: 'fromager~=0.68.0', global constraint: 'None'
- latest Fromager candidate for 'fromager~=0.68.0': 2025-10-21 14:37:12+02:00
- Fromager versions: 0.68.0, 0.68.1

found 2 versions on PyPI for req: 'fromager~=0.68.0', global constraint: 'None'
- latest PyPI release for 'fromager~=0.68.0': 2025-10-21 14:37:12+02:00
- package has purelib wheels (pure Python code)
- PyPI versions: 0.68.0, 0.68.1
```


src/fromager/commands/package.py
```diff
@@ -0,0 +1,417 @@
...
+from packaging.version import Version
+from resolvelib.resolvers import ResolverException
+
+from .. import context, log, overrides, packagesettings, request_session, resolver
+from ..candidate import Candidate
+
+logger = logging.getLogger(__name__)
+
+
+logger = logging.getLogger(__name__)
```
LalatenduMohanty: This is a duplicate/extra `logger = logging.getLogger(__name__)`

src/fromager/commands/package.py
```diff
@@ -0,0 +1,417 @@
...
+            f"resolving requirement for '{req}' (variant: {wkctx.variant})", bold=True
+        )
+
+        global_constraint = wkctx.constraints.get_constraint(req.name)
+        if global_constraint is not None:
+            click.secho(
+                f"Package has a global constraint {global_constraint}.", bold=True
+            )
+        else:
+            click.secho("Package has not global constraint.")
```
LalatenduMohanty: nit pick : instead of `has not` use `has no`

src/fromager/commands/package.py
```diff
@@ -0,0 +1,417 @@
...
+    )
+    click.secho(
+        f"- Fromager versions: {_versions_string(fromager_versions)}", bold=True
+    )
+
+    return fromager_versions
+
+
+def _resolve_pypi(
+    ctx: context.WorkContext, req: Requirement, global_constraint: Requirement | None
```
LalatenduMohanty: The `ctx` is not used within the function.

src/fromager/commands/package.py
```diff
@@ -0,0 +1,417 @@
...
+        accept=pypi_simple.ACCEPT_JSON_ONLY,
+        session=request_session.session,
+    )
+    try:
+        pypi_package = pypi_client.get_project_page(req.name)
+    except Exception as e:
+        click.secho(
+            f"failed to fetch package index from pypi.org: {e}", err=True, bold=True
+        )
+        return None
```
LalatenduMohanty: IMO we use pypi_simple library and requests package here. If it makes HTTP requests via requests we will get better error messages?

```diff
-    except Exception as e:
+    except pypi_simple.NoSuchProjectError:
+        click.secho(f"Package '{req.name}' not found on PyPI", err=True, bold=True)
+        return None
+    except requests.exceptions.RequestException as e:
```

src/fromager/commands/package.py
```diff
@@ -0,0 +1,422 @@
...
+    click.secho(
+        f"- Fromager versions: {_versions_string(fromager_versions)}", bold=True
+    )
+
+    return fromager_versions
+
+
+def _resolve_pypi(
+    req: Requirement,
+    constraint: Requirement | None,
```
LalatenduMohanty: @tiran I am going back on my review comment. It is not a critical fix , but ideally we should be using pbi.use_pypi_org_metadata() check here? The ctx we can use with the function call. 

``` diff
def _resolve_pypi(                                                                                                                                                        
       ctx: context.WorkContext,                                                                                                                                             
       req: Requirement,                                                                                                                                                     
       constraint: Requirement | None,                                                                                                                                       
   ) -> set[Version] | None:                                                                                                                                                 
  +    """Resolve package versions from PyPI for comparison with Fromager."""                                                                                                
  +    if not pbi.use_pypi_org_metadata:                                                                                                                                     
  +        logger.warning(                                                                                                                                                   
  +            "Package has customizations and may not use pypi.org metadata. "                                                                                              
  +            "PyPI results below may not be relevant."                                                                                                                     
  +        )                                                                                                                                                                 
  +                                                                                                                                                                          
       # resolve package from PyPI                              
```

src/fromager/commands/package.py
```diff
@@ -0,0 +1,430 @@
...
+        case DistributionType.BOTH:
+            include_sdists = True
+            include_wheels = True
+        case _:  # DEFAULT
+            # Use package settings defaults
+            include_sdists = pbi.resolver_include_sdists
+            include_wheels = pbi.resolver_include_wheels
+    click.secho(
+        f"Using distribution type: {dist_type.value} (sdists: {include_sdists}, wheels: {include_wheels})"
+    )
```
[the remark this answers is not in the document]
tiran: I have addressed the logging issue. Thanks for pointing it out! It took me a second to understand the problem.

PR #909, opened 2026-01-30
[Spike] Multiple version mode for bootstrap commands

rd4398:
This commit attempts to implement the multiple
version mode for bootstrap and is a result of
investigation as to what it might look like.

This is NOT the final implementation for the feature but can be a good 
starting point to discuss how we can implement this

Hence this PR is in Draft state.

Co-Authored-By: [Claude Sonnet 4.5]<[noreply@anthropic.com](mailto:noreply@anthropic.com)>

### DIFF
--- a/e2e/test_bootstrap_all_versions.sh
+++ b/e2e/test_bootstrap_all_versions.sh
@@ -1,0 +1,176 @@
+#!/bin/bash
+# -*- indent-tabs-mode: nil; tab-width: 2; sh-indentation: 2; -*-
+
+# Test bootstrap with --all-versions option to verify multiple versions
+# of a package are built. This test verifies that:
+#   1. Multiple versions of top-level packages are built
+#   2. Cache filtering works correctly (skips already-built versions)
+#   3. Warning is shown when using --all-versions without --skip-constraints
+#
+# We use 'six' because it has NO dependencies, making the test fast while
+# still testing the core all-versions functionality.
+#
+# Issue #878: Add multiple version mode to the bootstrap commands
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "$SCRIPTDIR/common.sh"
+
+pass=true
+
+# We use 'six' as the test package because:
+#   - It has multiple versions on PyPI
+#   - It has NO dependencies (pure Python, no setuptools dependency at runtime)
+#   - This makes the test fast while still testing multi-version behavior
+#
+# six>=1.16.0 should match all versions from 1.16.0 onwards
+# We use a narrow range to limit how many versions we build.
+PACKAGE_SPEC="six>=1.16.0"
+
+################################################################################
+# Test 1: Bootstrap with --all-versions flag
+# This should build all matching versions of six within the specified range.
+# Since six has no runtime dependencies, this tests the core multi-version
+# resolution without the complexity of deep dependency chains.
+################################################################################
+
+echo "=== Test 1: Bootstrap with --all-versions ==="
+
+fromager \
+  --debug \
+  --log-file="$OUTDIR/bootstrap.log" \
+  --error-log-file="$OUTDIR/fromager-errors.log" \
+  --sdists-repo="$OUTDIR/sdists-repo" \
+  --wheels-repo="$OUTDIR/wheels-repo" \
+  --work-dir="$OUTDIR/work-dir" \
+  bootstrap --all-versions --skip-constraints "$PACKAGE_SPEC"
+
+# Verify that we have multiple versions of six (top-level package)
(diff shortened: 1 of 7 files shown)


e2e/test_bootstrap_all_versions.sh
```diff
@@ -0,0 +1,176 @@
...
+pass=true
+
+# We use 'six' as the test package because:
+#   - It has multiple versions on PyPI
+#   - It has NO dependencies (pure Python, no setuptools dependency at runtime)
+#   - This makes the test fast while still testing multi-version behavior
+#
+# six>=1.16.0 should match all versions from 1.16.0 onwards
+# We use a narrow range to limit how many versions we build.
+PACKAGE_SPEC="six>=1.16.0"
```
dhellmann: We should probably go ahead and cap the range here so new releases don't break the build.

src/fromager/commands/bootstrap.py
```diff
@@ -169,20 +201,39 @@ def bootstrap(
...
+
         for req in to_build:
             token = requirement_ctxvar.set(req)
-            result = bt.resolve_and_add_top_level(req)
-            if result is not None:
-                resolved_reqs.append(req)
+            if all_versions:
+                # In all-versions mode, resolve all matching versions for each
+                # top-level requirement. This returns a list of (url, version) pairs.
+                results = bt.resolve_and_add_top_level_all_versions(req)
```
dhellmann: What do you think about making all_versions a flag to `resolve_and_add_top_level` and always have it return a sequence of results?

src/fromager/resolver.py
```diff
@@ -101,6 +101,127 @@ def resolve(
...
+        include_sdists=include_sdists,
+        include_wheels=include_wheels,
+        sdist_server_url=sdist_server_url,
+        req_type=req_type,
+        ignore_platform=ignore_platform,
+    )
+    return get_all_matching_candidates(provider, req)
+
+
+def get_all_matching_candidates(
```
dhellmann: How does this function differ from the provider's `find_matches()` method?

src/fromager/bootstrapper.py
```diff
@@ -166,6 +171,184 @@ def resolve_and_add_top_level(
...
+            This method uses the resolver's resolve_all_versions() function which
+            returns candidates sorted by version descending (newest first).
+            Existing cached versions are filtered out to avoid redundant builds.
+        """
+        pbi = self.ctx.package_build_info(req)
+        results: list[tuple[str, Version]] = []
+
+        try:
+            if pbi.pre_built:
+                # For pre-built packages, we typically only want the newest
```
dhellmann: I'm not sure this is a safe assumption. Do we _need_ to have a special case for pre-built packages, or can we rely on the resolver to only return the versions that make sense for that package?

src/fromager/bootstrapper.py
```diff
@@ -166,6 +171,184 @@ def resolve_and_add_top_level(
...
+            )
+
+        except Exception as err:
+            if not self.test_mode:
+                raise
+            self._record_test_mode_failure(req, None, err, "resolution")
+
+        return results
+
+    def _version_exists_in_cache(
```
dhellmann: I suspect this can be replaced with a call to `self._find_cached_wheel`?

src/fromager/bootstrapper.py
```diff
@@ -221,29 +404,63 @@ def _processing_build_requirement(self, current_req_type: RequirementType) -> bo
...
+                )
+            except Exception as err:
+                if not self.test_mode:
+                    raise
+                self._record_test_mode_failure(
+                    req, str(resolved_version), err, "resolution"
+                )
+                return
+        else:
+            # Normal resolution: find the best matching version
```
dhellmann: Here we again want a _list_ of results, not just the single result. That way we build all of the versions of all of the dependencies, too, not just the top level packages. In fact, I think if we do the expansion here, instead of in the command code, it will make all of the logic simpler.

For example, if at this point we find all matches for a requirement and get a list with more than one item, then we can loop over that list calling bootstrap recursively and then return. If we get a list with 1 item, then we do the bootstrapping work as we're doing it today.

That will mean we only have to look for the all_versions flag in one place, instead of everywhere we process the dependencies.

dhellmann:
Perhaps a good first step is to decompose the Bootstrapper class into separate classes for managing the resolution of requirements (including with the cache, the different ways to look up values using the requirement, git URL, etc.) from the management of the overall bootstrap process (including recursing for dependencies, performing builds, etc.). Extracting the resolver management into a separate class will make it easier to see where we are performing resolution in the bootstrapper, and where we might change the API of that new class to make it easy for the bootstrapper to loop over lists of results.

rd4398:
This was a spike and we have already implemented most of the feature now. I am closing this PR as it is obsolete.

PR #912, opened 2026-02-06
AIPCC-10079: Allow overriding extract_info_from_wheel_file

Xaenalt:
This happens due to an unfortuante way that TensorFlow is built. Since it uses Bazel to build and package the wheel, it doesn't have the usual infrastructure one uses to build variant wheels. Due to that, a single repo is used to build wheels with different names. In Fromager's case, this means we specify the req as `tensorflow` but in reality will get `tensorflow_cpu` or `tensorflow_rocm`, etc. This is how they create them upstream, and to match the requirements from notebooks that the wheel naming convention remains the same, we have to override the check that the wheel name matches the req in extract_info_from_wheel_file().

To achieve this, I made it overridable like other functions and provided a default_extract_info_from_wheel_file() implemetation to preserve existing behavior. I expect TensorFlow will be one of the only users of this override.

### DIFF
--- a/src/fromager/wheels.py
+++ b/src/fromager/wheels.py
@@ -117,7 +117,7 @@ def _extra_metadata_elfdeps(
     return elfinfos
 
 
-def extract_info_from_wheel_file(
+def default_extract_info_from_wheel_file(
     req: Requirement, wheel_file: pathlib.Path
 ) -> tuple[str, Version, BuildTag, frozenset[Tag]]:
     # parse_wheel_filename normalizes the dist name, however the dist-info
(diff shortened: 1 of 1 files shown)


src/fromager/wheels.py
```diff
@@ -133,6 +133,23 @@ def extract_info_from_wheel_file(
     return (dist_name, dist_version, build_tag, wheel_tags)
 
 
+def extract_info_from_wheel_file(
```
EmilienM: can we have tests?

tiran (changes requested):
We have multiple packages that build different distributions from the same sources:

- tensorflow (CUDA, CPU, ROCm)
- cupy-cuda (12x, 13x)
- opencv-python (currently we build only headless)

How does this PR address the problem that users request `tensorflow-cpu` or `tensorflow-rocm` in their requirements?

PR #914, opened 2026-02-10
[Refactor]: Decompose Bootstrapper into separate classes

### LINKED ISSUE #911: Decompose bootstrapper into separate classes
While doing a spike on multiple version bootstrap, it was discovered that decomposing the bootstrap class into separate classes will act as a good first step to implement https://github.com/python-wheel-build/fromager/issues/878.

@dhellmann stated this in a PR review for the spike https://github.com/python-wheel-build/fromager/pull/909:

"Perhaps a good first step is to decompose the Bootstrapper class into separate classes for managing the resolution of requirements (including with the cache, the different ways to look up values using the requirement, git URL, etc.) from the management of the overall bootstrap process (including recursing for dependencies, performing builds, etc.). Extracting the resolver management into a separate class will make it easier to see where we are performing resolution in the bootstrapper, and where we might change the API of that new class to make it easy for the bootstrapper to loop over lists of results."

rd4398:
This commit attempts to refactor the bootstrapper
class by breaking it into separate classes.

We create a ResolverManager class which handles
certain functionality needed by bootstrap.

The new class:

- Uses resolver.py as a building block (calls resolver.resolve(), resolver.GenericProvider, etc.)
- Adds caching - avoids redundant resolution via _resolved_requirements dict
- Adds history-based resolution - reuses versions from previous bootstrap runs via prev_graph
- Handles git URL cloning - clones repos and extracts versions from package metadata
- Handles cached wheel lookups - checks local filesystem + remote cache server

This refactoring was suggested as a good first step for multiple version bootstrap and also makes the code simpler.

Fixes: https://github.com/python-wheel-build/fromager/issues/911

### DIFF
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -1,3 +1,17 @@
+"""Bootstrap orchestration for building Python packages from source.
+
+This module provides the Bootstrapper class which orchestrates the entire
+bootstrap process, including:
+- Dependency tracking and build ordering
+- Building wheels and sdists from source
+- Managing the dependency graph
+- Test mode for catching and reporting failures
+
+The resolution logic (determining which version to use for a requirement)
+is delegated to the ResolutionManager class, following the Single
+Responsibility Principle.
+"""
+
 from __future__ import annotations
 
 import contextlib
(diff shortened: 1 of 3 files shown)


dhellmann (commented):
I think the split in responsibilities isn't quite right in this draft. I've tried to point out a few examples, but basically I think the new class knows too much about bootstrapping and the fact that that is where it's used. Let me know if you'd like to chat about it before picking this back up.

rd4398:
I think we already implemented this step by step in a separate PR, hence closing this

PR #920, opened 2026-02-24
feat: support maturin alternate src layout

tiran:
Implement support for maturin's alternate src layout for vendored Rust crates. The alt src layout has `Cargo.toml` and `.cargo` directory in a `rust` directory. The alt src layout is used by `pendulum` package.

See: https://www.maturin.rs/project_layout.html#alternate-python-source-directory-src-layout


tiran:
PR is blocked by https://github.com/python-wheel-build/fromager/issues/921

PR #923, opened 2026-02-24
fix(e2e): use 127.0.0.1 for local wheel server on macOS

### LINKED ISSUE #922: fix(e2e): use 127.0.0.1 for local wheel server on macOS instead of en0 IP
## Problem

`test_build_steps.sh` fails consistently on `ci_build_suite-py3.14-rs1.75-macos-latest`.

`start_local_wheel_server()` in `e2e/common.sh` uses `ipconfig getifaddr en0` on macOS to get the wheel server IP. On GitHub Actions macOS runners, this returns a VM bridge address (e.g. `192.168.64.24`) that becomes unreachable mid-test, causing `uv` to fail with `tcp connect error` when installing build dependencies.

The `en0` IP is only needed on Linux where podman network isolation prevents containers from reaching `127.0.0.1`. On macOS, network isolation is disabled, so `127.0.0.1` works fine.

## Error from CI

```
ERROR pbr: command failed with exit code 2: uv pip install ... --index-url http://192.168.64.24:9999/simple
  Caused by: tcp connect error
  Request failed after 3 retries
```

## Fix

Use `127.0.0.1` on macOS instead of querying `en0`:

```bash
if which ip 2>&1 >/dev/null; then
    # Linux: need host IP because podman can't reach localhost
    IP=$(ip route get 1.1.1.1 | grep 1.1.1.1 | awk '{print $7}')
else
    # macOS: no network isolation, localhost works
    IP=127.0.0.1
fi
```

## References

- Failing job: https://github.com/python-wheel-build/fromager/actions/runs/22157674629/job/64689995468
- Related: #917 (suite framework robustness improvements)

LalatenduMohanty:
On macOS CI runners, `ipconfig getifaddr en0` returns a VM bridge address (e.g. 192.168.64.24) that becomes unreachable mid-test, causing uv to fail with tcp connect errors. Since macOS does not use network isolation (podman), 127.0.0.1 works reliably.

Closes: #922

### DIFF
--- a/e2e/common.sh
+++ b/e2e/common.sh
@@ -73,11 +73,11 @@ start_local_wheel_server() {
     python3 -m http.server --directory "$serve_dir" 9999 &
     HTTP_SERVER_PID=$!
     if which ip 2>&1 >/dev/null; then
-        # Linux
+        # Linux: need host IP because podman can't reach localhost
         IP=$(ip route get 1.1.1.1 | grep 1.1.1.1 | awk '{print $7}')
     else
-        # macOS
-        IP=$(ipconfig getifaddr en0)
+        # macOS: no network isolation, localhost works
+        IP=127.0.0.1
     fi
     export WHEEL_SERVER_URL="http://${IP}:9999/simple"
 }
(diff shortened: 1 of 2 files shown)


dhellmann (approved):
Nice work!

PR #926, opened 2026-02-25
fix(e2e): preserve failing test output before next test cleanup

### LINKED ISSUE #917: fix(e2e): preserve failing test output and harden local wheel server setup
The e2e CI suite framework and test_build_steps.sh have reliability issues  that make failures like https://github.com/python-wheel-build/fromager/issues/922 hard to diagnose and may cause spurious failures. This was observed on ci_build_suite-py3.14-rs1.75-macos-latest where build_steps failed [1] but the actual error is unrecoverable from the CI artifacts.

[1] https://github.com/python-wheel-build/fromager/actions/runs/22076951159/job/63871644962?pr=916

**Problems**

- Suite framework destroys failing test output

  ci_suite_framework.sh calls clean_test_environment() before each test, which removes e2e-output/. When a mid-suite test fails, its output is deleted when the next test starts. The CI artifact upload captures e2e-output/ at the end, so it only contains the last test's output — not the failing test's.

  This made the build_steps failure (3rd of 5 tests) impossible to diagnose from CI artifacts.

-  test_build_steps.sh uses hardcoded port 9999

   start_local_wheel_server() in common.sh starts python3 -m http.server on port 9999. If that port is in use, the server silently fails (backgrounded with &), and all subsequent uv pip install --index-url calls fail with no clear root cause.

- No health check after starting the local wheel server

   start_local_wheel_server() starts the server in the background and immediately continues. There is no verification that the server actually started and is accepting connections.

LalatenduMohanty:
When a mid-suite test fails, its e2e-output/ directory is deleted by clean_test_environment() when the next test starts.

To fix this the patch moving e2e-output dir to e2e-failed-<test_name> dir on failure so the data survives cleanup, and update the CI workflow to upload these preserved directories alongside e2e-output.

Fixes #917

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -133,7 +133,9 @@ jobs:
         uses: actions/upload-artifact@v6
         with:
           name: ${{ matrix.test-script }}-py${{ matrix.python-version }}-rs${{ matrix.rust-version }}-${{ matrix.os }}
-          path: e2e-output
+          path: |
+            e2e-output
+            e2e-failed-*
 
       - name: Upload coverage
         uses: actions/upload-artifact@v6
--- a/e2e/ci_suite_framework.sh
+++ b/e2e/ci_suite_framework.sh
@@ -54,6 +54,14 @@ run_test() {
         echo "PASSED: $test_name"
     else
         echo "FAILED: $test_name"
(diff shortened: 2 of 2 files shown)


rd4398 (approved):
Looks good to me!

PR #929, opened 2026-02-25
refactor(e2e): use fromager wheel-server instead of stdlib http.server

### LINKED ISSUE #927: Replace python3 -m http.server with fromager wheel-server in e2e tests
## Context

In PR #925, @tiran asked why the e2e tests use `python3 -m http.server` instead of the builtin `fromager wheel-server`, which now uses asyncio, Starlette, and uvicorn. As @dhellmann noted, the tests simply were never updated when the builtin server was rewritten.

## Problem

Four e2e tests start a `python3 -m http.server` process to serve wheels:

- `test_build_steps.sh`
- `test_report_missing_dependency.sh`
- `test_bootstrap_build_tags.sh`
- `test_bootstrap_cache.sh`

This has a few drawbacks:

1. **stdlib `http.server` has platform quirks.** PRs #923 and #925 were needed to work around IPv6-only sockets on macOS and readiness-check failures. The builtin `fromager wheel-server` (uvicorn) handles socket binding correctly and wouldn't need these workarounds.
2. **We don't dogfood our own server.** The builtin Starlette server implements PEP 503 properly but is never exercised in CI.
3. **Raw directory listings vs proper API.** `http.server` returns auto-generated HTML directory listings that happen to work with pip, but it's not a real Simple Repository API.

## Investigation

I verified that all 4 tests are compatible with a drop-in replacement. The key findings:

- **`update_wheel_mirror()` at startup is harmless.** Every test either clears `downloads/` before starting the server or the mirror state is already consistent, so the mirror update at server startup doesn't change anything.
- **The Starlette server reads the filesystem per-request** (`os.scandir()` on every call, no caching). Files and symlinks added to `simple/` by separate `fromager step` processes are picked up immediately. This is how `test_build_steps.sh` works -- the server starts with an empty `simple/` and symlinks appear as packages are built.
- **Symlinks and regular files both work.** `server_file()` uses `stat(follow_symlinks=True)`, so tests that place actual files in `simple/` (like `test_bootstrap_build_tags.sh`) work fine.

### Things to verify during implementation

- **Readiness check:** The `/` route returns a 307 redirect to `/simple`. Confirm `curl -sf` treats this as success, or adjust the check URL to `http://${IP}:9999/simple`.
- **All 4 tests pass** on both Linux (with podman) and macOS CI.

LalatenduMohanty:
Replace python3 -m http.server with the builtin fromager wheel-server (Starlette/uvicorn) and bind to 127.0.0.1 on all platforms. This is safe because uv pip install always runs on the host without network isolation. Eliminates platform-specific workarounds from PRs #923/#925.

Closes: #927

### DIFF
--- a/e2e/common.sh
+++ b/e2e/common.sh
@@ -66,29 +66,23 @@ on_exit() {
 trap on_exit EXIT SIGINT SIGTERM
 
 start_local_wheel_server() {
-    local serve_dir="${1:-$OUTDIR/wheels-repo/}"
-    # Determine the bind address and the URL that clients will use.
-    # On Linux podman can't reach localhost, so we bind to 0.0.0.0 and
-    # advertise the host's routable IP.  On macOS there is no network
-    # isolation; we bind explicitly to 127.0.0.1 to avoid IPv6-only
-    # sockets that python3 -m http.server may create by default.
-    if command -v ip >/dev/null 2>&1; then
-        local BIND_ADDR="0.0.0.0"
-        IP=$(ip route get 1.1.1.1 | grep 1.1.1.1 | awk '{print $7}')
-    else
-        local BIND_ADDR="127.0.0.1"
-        IP=127.0.0.1
-    fi
-    python3 -m http.server --bind "$BIND_ADDR" --directory "$serve_dir" 9999 &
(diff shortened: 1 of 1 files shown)


rd4398 (approved):
This looks good to me!

PR #941, opened 2026-02-27
(refactor): Move caching from bootstrapper to resolver

rd4398:
This commit improves cohesion by moving the session-level resolution cache (_resolved_requirements) from Bootstrapper into RequirementResolver.

The cache now lives alongside the resolution logic it supports.

Changes:

- Added _resolved_requirements dict to RequirementResolver.__init__
- Updated resolve_source() and resolve_prebuilt() to check cache first and update cache before returning
- Added cache_resolution() public method for externally-resolved git URLs
- Removed _resolved_requirements from Bootstrapper
- Updated Bootstrapper.resolve_version() to use resolver.cache_resolution() for git URL results

This acts as next step towards implementing multiple version bootstrap

### DIFF
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -122,9 +122,6 @@ class Bootstrapper:
         # package.
         self._seen_requirements: set[SeenKey] = set()
 
-        # Track requirements we have already resolved so we don't resolve them again.
-        self._resolved_requirements: dict[str, tuple[str, Version]] = {}
-
         self._build_order_filename = self.ctx.work_dir / "build-order.json"
 
         # Track failed packages in test mode (list of typed dicts for JSON export)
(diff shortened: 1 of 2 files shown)


src/fromager/bootstrapper.py
```diff
@@ -185,42 +182,39 @@ def resolve_version(
...
+            # Git URLs must be resolved in Bootstrapper (orchestration concern)
             if req_type != RequirementType.TOP_LEVEL:
                 raise ValueError(
                     f"{req} includes a URL, but is not a top-level dependency"
                 )
             logger.info("resolving source via URL, ignoring any plugins")
             source_url, resolved_version = self._resolve_version_from_git_url(req=req)
-            self._resolved_requirements[req_str] = (source_url, resolved_version)
+            # Cache the git URL resolution in the resolver
+            self._resolver.cache_resolution(str(req), (source_url, resolved_version))
```
dhellmann: This will save the cached results, but the cache isn't checked as part of this code path so we may end up resolving the version repeatedly.

src/fromager/requirement_resolver.py
```diff
@@ -133,17 +151,35 @@ def resolve_prebuilt(
...
-        wheel_url, resolved_version = wheels.resolve_prebuilt_wheel(
-            ctx=self.ctx, req=req, wheel_server_urls=servers, req_type=req_type
-        )
+        # Cache the result
+        self._resolved_requirements[req_str] = (wheel_url, resolved_version)
         return wheel_url, resolved_version
 
+    def cache_resolution(
+        self,
+        req_str: str,
```
dhellmann: The read and write API for the cache should take a Requirement and then convert it to a string. That way we're only doing the conversion inside this class, and we can change how we index the cache if we want to.

rd4398: Okay, make sense. Done in latest push

dhellmann (approved):

PR #950, opened 2026-03-08
feat(packagesettings): add ${__version__} template variable to env 

### LINKED ISSUE #946: Give `env` package settings access to package version
Some build systems of Python packages require an environment variable that contains the version of the package. We should add the package version to template variables, so `env` package settings can access the version, e.g.

```yaml
env:
   MY_PACKAGE_VERSION: "${__version__}"
```

LalatenduMohanty:
Inject the resolved package version into template_env as __version__, so package env settings can reference it via ${__version__}.

Closes: #946

### DIFF
--- a/docs/customization.md
+++ b/docs/customization.md
@@ -215,9 +215,86 @@ implemented. A literal `$` must be quoted as `$$`.
 Added support for default value syntax `${NAME:-}`.
 ```
 
+```{versionchanged} 0.76.0
+
+Added the `${__version__}` template variable.
+```
+
+#### The `__version__` variable
+
+When fromager resolves the version of a package it injects the version
+string into the template environment as `__version__`. You can
+reference it in env values with `${__version__}`.
+
+##### Fixing setuptools-scm / flit_scm builds
+
+Packages that use `setuptools-scm` or `flit_scm` for version detection fail
+when built from source archives without `.git` metadata. The build backend
+raises a `LookupError` during import because it cannot determine the version.
+
+Setting `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_{DIST_NAME}` tells
+setuptools-scm to use the provided version instead of reading from SCM.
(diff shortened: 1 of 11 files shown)


tests/test_packagesettings.py
```diff
@@ -805,3 +807,149 @@ def test_use_pypi_org_metadata(testdata_context: context.WorkContext) -> None:
         "somepackage_without_customization"
     )
     assert pbi.use_pypi_org_metadata
+
+
+class TestGetExtraEnvironVersion:
```
tiran: Let's add the new variable to `test_pbi_test_pkg_extra_environ` and `test-pkg.yaml` instead of defining a new test class.

tiran (dismissed):
The new feature needs documentation:

- when is `__version__` defined and when is it not available?
- when was the feature introduced (`versionchanged`)?
- give an example how maintainers can use the variable

tests/test_packagesettings.py
LalatenduMohanty: Good idea. Will fix it.

docs/customization.md
```diff
@@ -215,13 +215,27 @@ implemented. A literal `$` must be quoted as `$$`.
...
+```{versionchanged} 0.76.0
+
+Added the `${__version__}` template variable. When fromager resolves
+the version of a package, it injects the version string as
+`__version__` into the template environment. This variable is
+available in all phases after version resolution — including
+`build_wheel`, `build_sdist`, and dependency hooks — but is **not**
+available during the `resolve` phase, since the version has not yet
+been determined at that point. Use `${__version__:-}` if you need a
+fallback for cases where the version may not be set.
```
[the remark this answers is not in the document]
LalatenduMohanty: Good catch. Fixed it.

rd4398 (approved):
This looks good to me too!

Issue #961, opened 2026-03-13
Generate .git_archival.txt for setuptools-scm packages built from source archives


LalatenduMohanty:
## Problem

Packages using `setuptools-scm` or `flit_scm` fail to build from source archives that lack `.git` metadata. Fromager creates stub `PKG-INFO` files as a fallback, but setuptools-scm checks `.git_archival.txt` **before** `PKG-INFO` in its [version resolution chain](https://setuptools-scm.readthedocs.io/en/latest/usage/#git-archives).

Users can work around this by setting `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_{DIST}` in per-package YAML (#946), but that requires manual configuration for every affected package.

## Proposal

Add `ensure_git_archival()` — similar to the existing `ensure_pkg_info()` — to automatically generate a `.git_archival.txt` from the resolved version when one is missing or unprocessed:

```
node: 0000000000000000000000000000000000000000
node-date: 1970-01-01T00:00:00+00:00
describe-name: 1.2.3-0-g0000000000000000000000000000000000000000
```

setuptools-scm extracts the version from the `describe-name` field (`tag-distance-gnode` format from `git describe`). Distance `0` means exact tag match.

**Scope:**
- Applies only to source archives without `.git` (git clones don't need this)
- Preserves existing valid `.git_archival.txt` files
- Replaces unprocessed template files (containing `$Format:` placeholders)
- Must run before backend import to cover packages where setuptools-scm triggers during `get_requires_for_build_wheel`

## Supersedes

- #960 — `.git_archival.txt` has higher priority than `PKG-INFO` in setuptools-scm's fallback chain, so generating it early enough eliminates the timing problem described there.

## Related

- #946 — `${__version__}` env var template (manual per-package alternative)

PR #962, opened 2026-03-13
feat(sources): generate .git_archival.txt for setuptools-scm builds

LalatenduMohanty:
When building from an sdist that lacks .git metadata, setuptools-scm                                                                                                       
  cannot determine the package version. This adds a .git_archival.txt                                       
  file with the correct version so setuptools-scm can resolve it.                                                                                                            
                                                                                                                                                                             
  - Skips packages with .git directories (no fix needed)                                                                                                                     
  - Replaces existing files that have unprocessed placeholders or                                                                                                            
    missing fields                                                                                                                                                           
  - Creates a new file only when PKG-INFO is also absent (git clones                                                                                                         
    or custom downloads, not PyPI sdists)

Closes: #961

### DIFF
--- a/src/fromager/sources.py
+++ b/src/fromager/sources.py
@@ -533,7 +533,14 @@ def prepare_source(
             )
     write_build_meta(source_root_dir.parent, req, source_filename, version)
     if source_root_dir is not None:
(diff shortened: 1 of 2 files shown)


LalatenduMohanty:
@coderabbitai review

Issue #878, opened 2025-12-09
add multiple version mode to the bootstrap commands


dhellmann:
As we work to populate a more general purpose package index, we want to quickly and easily identify all available versions of packages and build them. One way to do that would be to update the bootstrap commands to have a mode to build all versions of packages that match the requirements, instead of only the latest version.

## Requirements

The bootstrap commands should take a flag to enable a mode of building multiple versions of a package. When the mode is activated, all versions of a package that match the requirements specifiers in the top level list or in any of their dependencies should be included in the build. By default only the newest version of each package that matches the rule should be included.

When resolving multiple versions for a rule, all other filters should be honored. This includes the constraints for the package, if any, and the minimum age (#877 ).

Expanding the set of packages to be bootstrapped like this will be quite expensive, so we need to optimize as much as we can. Any version that appears in the cache server should be skipped completely (so neither the build nor install dependencies of that version should be bootstrapped). The most effective way to do that is likely to filter the resolved list of versions based on what already exists, rather than relying on the bootstrapper to handle them recursively.

LalatenduMohanty:
@rd4398 For future PRs related to this RFE (even the refactor PRs) we should create sub-issues and then create PRs for these sub-isssues. That would help reviewers connect the dots with how each PR helping towards the goal.

rd4398:
@dhellmann @LalatenduMohanty with https://github.com/python-wheel-build/fromager/pull/1152 landing, I think this feature can be marked complete upstream.  The next steps would be to test and roll it downstream in our pipelines.

We will open bug tickets and improvements as and when we discover them as separate tickets.

I am inclined to close this issue unless you think it should stay open till this feature is rolled downstream.

EmilienM:
Let's close it and iterate with individual issues when bugs will be found :-)

PR #975, opened 2026-03-23
refactor(resolver): return lists of matching versions

rd4398:
This commit introduces new *_all() functions that return all matching versions while
  keeping existing functions unchanged for backward compatibility.

  New functions return list[tuple[str, Version]] sorted by version (highest first):
  - resolver.resolve_all()
  - sources.resolve_source_all() and default_resolve_source_all()
  - wheels.resolve_prebuilt_wheel_all()
  - RequirementResolver.resolve_all()

  Existing functions now call the new *_all() functions and return the
  highest version (first element), maintaining identical behavior:
  - resolver.resolve() -> tuple[str, Version]
  - sources.resolve_source() -> tuple[str, Version]
  - wheels.resolve_prebuilt_wheel() -> tuple[str, Version]
  - RequirementResolver.resolve() -> tuple[str, Version]

This PR is another step towards https://github.com/python-wheel-build/fromager/issues/878

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/src/fromager/bootstrap_requirement_resolver.py
+++ b/src/fromager/bootstrap_requirement_resolver.py
@@ -12,7 +12,7 @@ import typing
 from packaging.requirements import Requirement
 from packaging.version import Version
 
-from . import overrides, resolver
+from . import resolver, sources, wheels
 from .dependency_graph import DependencyGraph
 from .requirements_file import RequirementType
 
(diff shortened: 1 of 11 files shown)


src/fromager/resolver.py
dhellmann: I feel like there are now 2 "top" layers to the resolver, this module and the new one. Does one call the other, or are they being called from different places?

src/fromager/resolver.py
```diff
@@ -158,26 +188,31 @@ def ending(self, state: typing.Any) -> None:
...
+    # instead of using resolvelib's resolver which picks just one
+    identifier = provider.identify(req)
     try:
-        result = rslvr.resolve([req])
+        candidates = provider.find_matches(
+            identifier=identifier,
+            requirements={identifier: [req]},
+            incompatibilities={},
+        )
     except resolvelib.resolvers.ResolverException as err:
```
dhellmann: We're not calling resolvelib here any more, so we're not going to get this exception type, I think? Will we get an exception at all, or will the candidate list just be empty?

Since we're not calling resolvelib any more, do we still need that library?

src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -139,43 +143,91 @@ def _resolve(
...
-            url, resolved_version = wheels.resolve_prebuilt_wheel(
-                ctx=self.ctx, req=req, wheel_server_urls=servers, req_type=req_type
+            # Get wheel server URLs
+            wheel_server_urls: list[str] = []
+            if pbi.wheel_server_url:
+                wheel_server_urls.append(pbi.wheel_server_url)
+            else:
+                if self.ctx.wheel_server_url:
+                    wheel_server_urls.append(self.ctx.wheel_server_url)
+                wheel_server_urls.append(resolver.PYPI_SERVER_URL)
```
LalatenduMohanty: Could _resolve() call wheels.resolve_prebuilt_wheel() instead and wrap the result in a list?

LalatenduMohanty (commented):
@rd4398 The PR looks good, but need some minor cleanups. 
                                                                                                                                                                      
 BaseProvider.find_matches() raises ResolverException when nothing matches — it never returns an empty list (resolver.py L624-626). So find_all_matching_from_provider()    
  always returns a non-empty list or raises.                                                                                                                                 
                                                                                                                                                                             
  The if results: / else: checks in bootstrap_requirement_resolver.py:177 and wheels.py:492 are dead code — the else branch can never execute. The "not found" case is already handled by the except ResolverException blocks.                                                                                                        
                                                                                                                                                                             
  bootstrap_requirement_resolver.py:177-182: 
```                                                                                                                                
  results = resolver.find_all_matching_from_provider(provider, req)                                                                                                          
  if results:       # always True                                                                                                                                            
      return results                                                                                                                                                         
  else:             # never reached — find_matches raises instead of returning []                                                                                            
      logger.debug(f"{req.name}: no prebuilt wheel found on {url}, trying next server")                                                                                      ``` 
```                                                                                                                                                                          
  wheels.py:492-497:

```                                                                                                                                                     
  if results:       # always True                                                                                                                                            
      wheel_url, version = results[0]                                                                                                                                        
      return str(wheel_url), version                                                                                                                                         
  else:             # never reached                                                                                                                                          
      excs.append(ValueError(f"no results for {url}: {results=}"))                                                                                                           
```

LalatenduMohanty:
@mergify queue

PR #988, opened 2026-03-26
test(e2e): add bootstrap --test-mode failure tests

### LINKED ISSUE #895: Add e2e test for bootstrapping failures
We should add an e2e test for  bootstrapping failures. To make sure the flow works as expected. Refer https://github.com/python-wheel-build/fromager/pull/893#pullrequestreview-3599818945

shifa-khan:
Add e2e tests for --test-mode bootstrapping failures as requested in issue

 Tests cover:
- Top-level resolution failure (non-existent package)
- Secondary dependency resolution failure (constrained pbr)
- Build failure without prebuilt fallback (local git fixture)
- Build failure with prebuilt fallback (broken patch on setuptools)

Closes: #895

### DIFF
--- a/e2e/test_mode_build.sh
+++ b/e2e/test_mode_build.sh
@@ -1,0 +1,72 @@
+#!/bin/bash
+# -*- indent-tabs-mode: nil; tab-width: 2; sh-indentation: 2; -*-
+
+# Test --test-mode with a package that fails to build (no prebuilt fallback)
+# Uses a local fixture that fails during wheel build; since it's not on PyPI,
+# prebuilt fallback also fails and the failure is recorded.
+# See: https://github.com/python-wheel-build/fromager/issues/895
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "$SCRIPTDIR/common.sh"
+pass=true
+
+DIST="test_build_failure"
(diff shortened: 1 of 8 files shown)


LalatenduMohanty (changes requested):

shifa-khan:
@LalatenduMohanty Worked on fixing suggested changes. Mind taking a look again?
Thanks

e2e/test_mode_build.sh
```diff
@@ -0,0 +1,113 @@
+#!/bin/bash
+# -*- indent-tabs-mode: nil; tab-width: 2; sh-indentation: 2; -*-
```
LalatenduMohanty, line 2: We can remove lines like this. Not sure why they are mntioned.

jlarkin09:
LGTM

e2e/test_mode_build.sh
jlarkin09: Not sure if it's completely necessary either, but maybe worth adding to keep it consistent with existing e2e tests

jlarkin09 (approved):

e2e/test_mode_build.sh
LalatenduMohanty: I see,  other files also have it. I am resolving this comment.

LalatenduMohanty (approved):

Issue #990, opened 2026-03-27
Extend resolver API with custom download API


tiran:
New resolver and downloader profiles like `pypi-download` or `github-git` need a way to customize the download URL. For example  `pypi-download` uses PyPI to resolve package versions. The candidate objects need to point to a different location based on a format string.

PR #991, opened 2026-03-27
feat: add override_download_url to resolvers

tiran:
Extend the resolver API with alternative download URLs. Resolvers can now return download links to custom locations via
`override_download_url` argument. The argument takes a format string with place holders like `{version}`.

The feature can be used to download a tar ball from a different location other than default PyPI sdist url or GitHub/GitLab tarball.

Fixes: #990

### DIFF
--- a/src/fromager/resolver.py
+++ b/src/fromager/resolver.py
@@ -14,7 +14,7 @@ import typing
 from collections.abc import Iterable
 from operator import attrgetter
 from platform import python_version
-from urllib.parse import quote, unquote, urljoin, urlparse
+from urllib.parse import quote, unquote, urlparse
 
 import pypi_simple
 import resolvelib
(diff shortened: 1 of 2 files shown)


src/fromager/resolver.py
```diff
@@ -870,7 +908,15 @@ def _find_tags(
 
 
 class GitLabTagProvider(GenericProvider):
-    """Lookup tarball and version from GitLab git tags"""
+    """Lookup tarball and version from GitLab git tags
+
+    The ``override_download_url`` parameter supports the string template variable:
+    * hostname
+    * project_path
+    * project_name (last part of project_name)
```
LalatenduMohanty: This is little confusing. May be reword this as `should be last component of project_path`

tiran:
@LalatenduMohanty Thanks for your review! I have applied your suggestions.

Issue #877, opened 2025-12-09
add cooldown period when resolving version numbers during bootstrapping


dhellmann:
The resolver automatically pulls in the latest version of each package that matches the rule being resolved. This means that as soon as a new version of a package is released, the next bootstrap operation will pull that version into the build. We want to introduce a "cooldown" period so that newly uploaded versions are only included in builds after they have been public for some number of days. This will give us a window to identify potentially malicious content in packages that have had their release pipelines subverted.

## Requirements

All bootstrap commands should accept an option to specify the minimum age of a release before it should be included in a build. The default for the minimum age should be `0`, indicating that packages can be included immediately. Only whole numbers of days are supported. Negative numbers are equivalent to `0`.

All resolvers should honor the minimum age value. For resolvers that look at tags in git repositories, the date should come from the timestamp of the tag, if it is an annotated tag, or from the commit timestamp of the commit that was tagged otherwise. Resolvers that look at packages uploaded to servers (PyPI, etc.) should use the timestamp from when the package was uploaded.

The age of a package should be calculated by subtracting the timestamp of the release from the start time of the bootstrap process (not the current time when the resolution is happening). This ensures that resolving the same package multiple times during a build will produce the same results.

The per-package settings should provide an override to change the default age requirement. This is useful for packages that are maintained internally and known to be safe, to avoid having a delay (by setting the value to `0`) or for riskier packages (by setting the value to a default that is larger than the command line option). The configured value should take priority over the value provided by the command line. This allows for different jobs (one to test builds and another to run the production builds) to easily pass different ages on the command line, but for the overrides to still take effect in both cases.

The `list-versions` command should also take a flag to specify the minimum age. The default should be `0`. The detailed output should show the timestamp and age in whole days of each release. The `list-versions` command should also take an option to ignore the configured overrides for minimum age for separate packages. The default should be to honor the configuration settings, so that the command produces the same information that a bootstrap command would use, but by having an option to ignore the configuration settings we can use `list-versions` to show versions that have been ignored during bootstrapping.

The `list-overrides` command should be extended to show any configured minimum age value for packages with overrides.

## Background

* https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns
* https://helixguard.ai/blog/malicious-sha1hulud-2025-11-24

ryanpetrello:
[...]

PR #1000, opened 2026-03-31
docs(proposal): add release cooldown design for version resolution

LalatenduMohanty:
# Pull Request Description

## What

Add a design proposal for release cooldown -- a configurable minimum release age for version resolution. The proposal introduces a --min-release-age CLI option (days, default 0) that makes fromager skip package versions published fewer than N days ago, with per-package overrides via resolver_dist.min_release_age in package settings.

## Why

Supply-chain attacks often publish a malicious package version and rely on automated builds picking it up immediately. A cooldown window lets the community detect and report compromised releases before fromager consumes them.

See: #877 

Implementation of the proposal #1018 

### DIFF
--- a/docs/proposals/release-cooldown.md
+++ b/docs/proposals/release-cooldown.md
@@ -1,0 +1,286 @@
+# Release cooldown for version resolution
+
+- Author: Lalatendu Mohanty
+- Created: 2026-03-31
+- Status: Open
+- Issue: [#877](https://github.com/python-wheel-build/fromager/issues/877)
+
+## What
+
+A configurable minimum release age ("cooldown") for version resolution.
+When enabled, fromager skips package versions published fewer than N
+days ago. One global setting controls all providers. Per-package
+overrides allow exceptions.
+
+## Why
+
+Supply-chain attacks often publish a malicious package version and rely
+on automated builds picking it up immediately. A cooldown window lets
+the community detect and report compromised releases before fromager
+consumes them. It also means new versions get broader testing before
+entering the build.
+
+References:
+
+- [We should all be using dependency cooldowns](https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns)
+- [Malicious sha1hulud](https://helixguard.ai/blog/malicious-sha1hulud-2025-11-24)
+
+## Goals
+
+- A single `--min-release-age` CLI option (days, default 0) that
+  applies to every resolver provider
+- Per-package overrides via `resolver_dist.min_release_age` in package
+  settings, taking priority over the CLI default
+- Provider-aware fail-closed: providers that support timestamps
+  reject candidates with missing `upload_time`; providers that do
+  not support timestamps skip cooldown with a warning. A future
+  strictness option may be added to control enforcement for
+  providers that gain timestamp support (e.g., Phase 3), allowing
+  gradual rollout without breaking existing builds.
+- Pre-built wheels subject to cooldown when the index supports
+  timestamps; bypass via per-package override otherwise
+- `list-versions` shows timestamps, ages, and cooldown status
+- `list-overrides` shows per-package cooldown values
+- Age calculated from bootstrap start time, not wall-clock time during
+  resolution
+
+## Non-goals
+
+- **Provider-specific flags** (`--pypi-min-age`, `--github-min-age`).
+  The provider a package uses (PyPI, GitHub, GitLab) reflects *how* it
+  is obtained, not how trusted it is. Most GitHub/GitLab packages are
+  there because of broken PyPI sdists or midstream forks. Separate
+  flags per provider would create a confusing configuration matrix and
+  cannot coexist cleanly with a global model. This proposal uses one
+  global default plus per-package overrides.
+- **SSH transport** for git timestamp retrieval.
+
+### Future consideration: `==` pin exemptions
+
+Whether `==` pins in top-level requirements or constraints files
+should automatically bypass cooldown is deferred. The per-package
+`resolver_dist.min_release_age: 0` override already provides an
+explicit, reviewable escape hatch for packages that need to use
+recently-published versions. Adding automatic `==` exemptions
+would introduce a special case that weakens the security model
+and requires users to understand the distinction. This can be
+revisited if the per-package override proves too cumbersome in
+practice.
+
+## How
+
+### Configuration
+
+#### CLI and environment variable
+
+A top-level `--min-release-age` option accepts a non-negative integer
+(days, default 0). Negative values are rejected. The corresponding
+environment variable `FROMAGER_MIN_RELEASE_AGE` is automatically
+available via Click's `auto_envvar_prefix`.
+
+The value is stored on `WorkContext` with a `start_time` captured once
+at construction (UTC). A fixed start time ensures consistent results
+when the same package is resolved multiple times during a build.
+
+#### Per-package overrides
+
+A new field in `ResolverDist`:
+
+```yaml
+# Trusted internal package -- bypass cooldown
+resolver_dist:
+  min_release_age: 0
+
+# Extra scrutiny -- 2-week cooldown
+resolver_dist:
+  min_release_age: 14
+```
+
+Semantics:
+
+- `None` (default) -- use the global `--min-release-age`
+- `0` -- no cooldown for this package
+- Positive integer -- override the global value
+
+The effective cooldown for a package is resolved by checking the
+per-package override first, falling back to the global default.
+
+### Enforcement
+
+During candidate validation, `BaseProvider` rejects candidates
+whose age is less than the effective cooldown. The behaviour
+depends on whether the provider can supply timestamps:
+
+- **Supports timestamps** (e.g. PyPI with PEP 691, GitLab):
+  candidates with a known `upload_time` younger than the cutoff
+  are rejected. A candidate with no `upload_time` is also rejected
+  (fail-closed).
+- **Does not support timestamps** (e.g. GitHub, generic
+  providers): cooldown is skipped with a one-time warning per
+  package. Custom providers inherit this default.
+
+Each provider declares its timestamp capability. `PyPIProvider`
+supports timestamps by default but allows callers to opt out for
+indexes that only implement PEP 503 (no `upload-time` field).
+
+After provider creation, the resolver supplies:
+
+- The effective cooldown period (days, after resolving global vs.
+  per-package override)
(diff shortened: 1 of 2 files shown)


tiran:
@LalatenduMohanty please rebase and resolve the merge conflict

docs/proposals/release-cooldown.md
```diff
@@ -0,0 +1,278 @@
...
+- "found N candidate(s) for X but all were published within the last
+  M days (cooldown policy)"
+- "found N candidate(s) for X but none have upload timestamp metadata;
+  cannot enforce the M-day cooldown"
+
+### Timestamp availability
+
+| Provider | `supports_upload_time` | Source |
+| -- | -- | -- |
+| PyPIProvider | Yes | `upload-time` (PEP 691 JSON API) |
```
tiran: `PyPIProvider` only supports time stamps for `https://pypi.org/simple` and other Python package indexes that support PEP 691 JSON API. It does not work for 3rd party package indexes that only support PEP 503.

- GitLab PyPI
- https://download.pytorch.org/whl/
- https://pypi.nvidia.com/
- and more

docs/proposals/release-cooldown.md
```diff
@@ -0,0 +1,278 @@
...
+## Implementation phases
+
+### Phase 1 -- Core (single PR)
+
+- `WorkContext`: `min_release_age_days`, `start_time`
+- CLI: `--min-release-age` / `FROMAGER_MIN_RELEASE_AGE`
+- `ResolverDist.min_release_age` field
+- `PackageBuildInfo.resolver_min_release_age()` method
+- `BaseProvider.validate_candidate()` cooldown check
+- `BaseProvider.supports_upload_time` class flag
```
tiran: I think this has to be an object attribute.

docs/proposals/release-cooldown.md
```diff
@@ -0,0 +1,278 @@
...
+
+### Phase 3 -- GitHub timestamps (follow-up PR)
+
+- Releases API + commit fallback in `GitHubTagProvider`
+
+**Migration note**: Until Phase 3 ships, GitHub-sourced packages
+skip cooldown with a warning (since `GitHubTagProvider` has
+`supports_upload_time = False`). No manual `min_release_age: 0`
+overrides are needed. Phase 3 enables cooldown enforcement for
+these packages by adding timestamp support.
```
tiran: - IMO GitHub releases should be covered by a new `GitHubReleaseProvider`
- It is possible to get commit and tag date for git tags with GitHub's GraphQL API. GraphQL is only available for authenticated requests and therefore requires a bearer token.

docs/proposals/release-cooldown.md
```diff
@@ -0,0 +1,278 @@
...
+### Configuration
+
+#### CLI and environment variable
+
+```python
+@click.option(
+    "--min-release-age",
+    type=click.IntRange(min=0),
+    default=0,
+    envvar="FROMAGER_MIN_RELEASE_AGE",
```
tiran: I think the `envvar` argument is not required. Fromager uses Click's `auto_envvar_prefix="FROMAGER"`.

docs/proposals/release-cooldown.md
```diff
@@ -0,0 +1,278 @@
...
+
+## Goals
+
+- A single `--min-release-age` CLI option (days, default 0) that
+  applies to every resolver provider
+- Per-package overrides via `resolver_dist.min_release_age` in package
+  settings, taking priority over the CLI default
+- Provider-aware fail-closed: providers that support timestamps
+  reject candidates with missing `upload_time`; providers that do
+  not support timestamps skip cooldown with a warning
```
tiran: Should we include an option to change the behavior, so provider or versions without a time stamp fail the build?

docs/proposals/release-cooldown.md
ryanpetrello: Correct, my MR that implements this removed this line.

docs/proposals/release-cooldown.md
ryanpetrello: I know we deprioritized this on the first pass, but yes, we should probably document this more clearly for follow up work.

We'll probably want to be specific about what the behavior should be if a bearer token isn't available/provided.

For example:

- You specify a cooldown in days
- We encounter usage of GitHubReleaseProvider, but a bearer token is _not available_ to perform the GraphQL API calls

...it seems to me in this scenario, the desired behavior would be "fail-closed".

docs/proposals/release-cooldown.md
ryanpetrello: Maybe, eventually?

What would we call this flag?

```
--min-release-age-strictness

Controls how the --min-release-age cooldown handles candidates with no upload timestamp.

warn (default): candidates with no upload timestamp are allowed through 
                with a warning when the provider is unable to provide a release timestamp.

reject: any candidate missing an upload timestamp is rejected, regardless of the
        provider. Use this when every resolved package must have a verifiable release age.
```

docs/proposals/release-cooldown.md
ryanpetrello: Yea, perhaps it should be - with provider-specific defaults as we currently have implemented in https://github.com/python-wheel-build/fromager/pull/1018

But in the example above for package indexes that only support PEP 503, maybe it just makes sense to pass this context into the `PyPIProvider` constructor if you know the answer is `False`.

ryanpetrello: @LalatenduMohanty I don't think it needs spelling out in this proposal, but just thinking out loud how we might adapt #1018 in practice, this is what I'm pondering:

  - `BaseProvider.__init__`: sets `self.supports_upload_time = False` - the assumed default
  -  All providers `GitHubTagProvider`, etc... inherit `False` by default, providers are responsible for
  - `GitLabTagProvider.__init__`: sets `self.supports_upload_time = True` always True, no param
  - `PyPIProvider.__init__`: accepts `supports_upload_time: bool = True` - the one place where the answer could be conditional (based on whether the target index supports PEP 691 or only PEP 503 indexes)

...down the line...

- `GitHubReleaseProvider` is implemented and also sets `self.supports_upload_time = True` in its implementation.

docs/proposals/release-cooldown.md
LalatenduMohanty: Agree, also I am removing the code implementations from the proposal as a proposal should not give code implementations.

docs/proposals/release-cooldown.md
LalatenduMohanty: Agree, I have updated the text in the latest push.

docs/proposals/release-cooldown.md
LalatenduMohanty: @ryanpetrello yeah, this looks alright.

LalatenduMohanty: Also updated the text to align with what we discussed.

docs/proposals/release-cooldown.md
LalatenduMohanty: I would not suggest a new flag. we can just change from warning to fail. I guess you are thinking about keeping it backward compatible . I have added following to the text.

```
A future strictness option may be added to control enforcement for providers that gain timestamp support (e.g., Phase 3), allowing  gradual rollout without breaking existing builds.
```

docs/proposals/release-cooldown.md
LalatenduMohanty: Fixed this.

docs/proposals/release-cooldown.md
```diff
@@ -0,0 +1,276 @@
...
+skips cooldown with a warning until Phase 3 adds timestamp
+support via the Releases API and commit date fallback.
+
+### Exempt sources
+
+#### Pre-built wheels
+
+Pre-built wheels are served from curated indices and use a different
+trust model. Pre-built wheel resolution bypasses the cooldown by
+using a cooldown of zero.
```
tiran: In downstream, we also consume pre-build wheels from PyPI.org, e.g. `ray`, `tensorboard`, and `runai_model*`. Is it possible to have cooldown for these wheels?

docs/proposals/release-cooldown.md
```diff
@@ -0,0 +1,276 @@
...
+use `min_release_age: 0` for packages from indices without timestamps.
+
+#### GitHub timestamps (Phase 3)
+
+The GitHub tags list API does not return dates.
+`GitHubTagProvider` sets `supports_upload_time = False`, so it
+skips cooldown with a warning until Phase 3 adds timestamp
+support via the Releases API and commit date fallback.
+
+### Exempt sources
```
tiran, line 183: Should we also clarify that wheels in Fromager's internal wheel server and cache wheel server (`--cache-wheel-server-url`) are considered trusted? Fromager does not use these wheel servers to resolve versions.

docs/proposals/release-cooldown.md
LalatenduMohanty: Good point. I will add the following 

```
Cooldown applies to pre-built wheels the same way it applies to
sdists: if the index supports timestamps (e.g. PyPI.org with
PEP 691), candidates younger than the cutoff are rejected. If the
index does not support timestamps, fail-closed applies. Use
`resolver_dist.min_release_age: 0` to bypass cooldown for
packages resolved from indices without timestamp support.
```

docs/proposals/release-cooldown.md
LalatenduMohanty: Good point, Added `Fromager's internal build and cache wheel servers are not used for
version resolution, so cooldown does not apply to them.`

docs/proposals/release-cooldown.md
LalatenduMohanty: @ryanpetrello This is a new addition, Do we want to do it in phase 1 or as a followup?

ryanpetrello: I think my preference would be to defer this one from https://github.com/python-wheel-build/fromager/pull/1018 for now.  There's already been some general concern raised about the size of that PR.

LalatenduMohanty: sure we can create a follow up. I am fine with it.

tiran:
LGTM

@ryanpetrello You are more familiar with the cooldown feature. Can you please review the latest version?

ryanpetrello:
@LalatenduMohanty @tiran the proposal looks good to me - I've identified some additional E2E testing I want to do on the implementation as it relates to pre-built wheels, but overall I'm happy with where we landed on the current state of this 👍

tiran (approved):

LalatenduMohanty:
@mergify rebase

PR #1018, opened 2026-04-02
feat(resolver): add release-age cooldown to protect against supply-chain attacks

ryanpetrello:
Adds a configurable minimum release age policy that rejects package versions published fewer than N days ago. This protects automated builds from supply-chain attacks where a malicious version is published and immediately pulled in before it can be reviewed.

When active, any sdist candidate whose upload-time is more recent than the cutoff is not considered a valid option during constraint resolution. The cutoff is fixed at the start of each run so all resolutions share the same boundary.

Key behaviors:
- --min-release-age flag (envvar FROMAGER_MIN_RELEASE_AGE) sets the global minimum age in days; 0 (default) disables the check entirely
- Enforcement lives in BaseProvider.is_satisfied_by() so it applies to PyPI, GitLab, and any future providers uniformly
- Providers that can currently provide release timestamps (PyPI, GitLab) are fail-closed: a candidate with no upload-time metadata is rejected when a cooldown is active (including PyPI indexes that do not support https://peps.python.org/pep-0691/); providers which do not yet *implement* timestamp support (GitHub) emit a one-time warning and skip the check instead
- resolver_dist.min_release_age in package settings overrides the global flag per-package (None = inherit, 0 = disable, positive int = override)

related: https://github.com/python-wheel-build/fromager/issues/877

Proposal : #1000

### DIFF
--- a/docs/how-tos/release-age-cooldown.rst
+++ b/docs/how-tos/release-age-cooldown.rst
@@ -1,0 +1,125 @@
+Protect Against Supply-Chain Attacks with Release-Age Cooldown
+==============================================================
+
+Fromager's release-age cooldown policy rejects package versions that were
+published fewer than a configured number of days ago. This protects automated
+builds from supply-chain attacks where a malicious version is published and
+immediately pulled in before it can be reviewed.
+
+How It Works
+------------
+
+When a cooldown is active, any candidate whose ``upload-time`` is more recent
+than the cutoff (current time minus the configured minimum age) is not
+considered a valid option during constraint resolution. If no versions of a
+package satisfy both the cooldown window and any other provided constraints,
+resolution fails with an informative error.
+
+The cutoff timestamp is fixed at the start of each run, so all package
+resolutions within a single bootstrap share the same boundary.
+
+Enabling the Cooldown
+---------------------
+
+Use the global ``--min-release-age`` flag, or set the equivalent environment
+variable ``FROMAGER_MIN_RELEASE_AGE``:
+
+.. code-block:: bash
+
+   # Reject versions published in the last 7 days
+   fromager --min-release-age 7 bootstrap -r requirements.txt
+
+   # Same, via environment variable (useful for CI and builder integrations)
+   FROMAGER_MIN_RELEASE_AGE=7 fromager bootstrap -r requirements.txt
+
+   # Disable the cooldown (default)
+   fromager --min-release-age 0 bootstrap -r requirements.txt
+
+The ``--min-release-age`` flag accepts a non-negative integer number of days.
+A value of ``0`` (the default) disables the check entirely.
+
+Scope
+-----
+
+The cooldown applies to **sdist resolution** — selecting which version of a
+package to build from source, including transitive dependencies. It does not
+apply to:
+
+* Wheel-only lookups, including cache servers (``--cache-wheel-server-url``) and
+  packages configured as ``pre_built: true`` in variant settings. These use a
+  different trust model and are not subject to the cooldown regardless of which
+  server they are fetched from.
+* Packages resolved from Git URLs that do not provide timestamp metadata.
+
+Note that sdist resolution from a private package index depends on
+``upload-time`` being present in the index's PEP 691 JSON responses. If the
+index does not provide that metadata, candidates will be rejected under the
+fail-closed policy described below.
+
+
+Fail-Closed Behavior
+--------------------
+
+If a candidate has no ``upload-time`` metadata — which can occur with older
+PyPI Simple HTML responses — it is rejected when a cooldown is active. Fromager
+uses the `PEP 691 JSON Simple API`_ when fetching package metadata, which
+reliably includes upload timestamps.
+
+.. _PEP 691 JSON Simple API: https://peps.python.org/pep-0691/
+
+.. note::
+
+   If you are writing a ``get_resolver_provider`` plugin that uses
+   :class:`~fromager.resolver.PyPIProvider` with a private index that only
+   implements the `PEP 503`_ HTML API (and therefore cannot supply upload
+   timestamps), pass ``supports_upload_time=False`` to ``PyPIProvider``. This
+   switches the provider from fail-closed to warn-and-skip, so candidates
+   without upload timestamps are skipped with a warning rather than causing
+   resolution to fail.
+
+.. _PEP 503: https://peps.python.org/pep-0503/
+
+Example
+-------
+
+Given a package ``example-pkg`` with three available versions:
+
+* ``2.0.0`` — published 3 days ago
+* ``1.9.0`` — published 45 days ago
+* ``1.8.0`` — published 120 days ago
+
+With a 7-day cooldown, ``2.0.0`` is blocked and ``1.9.0`` is selected:
+
+.. code-block:: bash
+
+   fromager --min-release-age 7 bootstrap example-pkg
+
+With a 60-day cooldown, both ``2.0.0`` and ``1.9.0`` are blocked and ``1.8.0``
+is selected:
+
+.. code-block:: bash
+
+   fromager --min-release-age 60 bootstrap example-pkg
+
+Per-Package Override
+--------------------
+
+The cooldown can be adjusted on a per-package basis using the
+``resolver_dist.min_release_age`` setting in the package's settings file:
+
+.. code-block:: yaml
+
+   # overrides/settings/my-package.yaml
+   resolver_dist:
+     min_release_age: 0   # disable cooldown for this package
+     # min_release_age: 30  # or use a different number of days
+
+Valid values:
+
+* Omit the key (default): inherit the global ``--min-release-age`` setting.
+* ``0``: disable the cooldown for this package, regardless of the global flag.
+* Positive integer: use this many days instead of the global setting.
+
+This is useful when a specific package is trusted enough to allow recent
+versions, or when a package's release cadence makes the global cooldown
+impractical.
--- a/docs/how-tos/release-age-cooldown.rst
+++ b/docs/how-tos/release-age-cooldown.rst
@@ -41,43 +41,50 @@ A value of ``0`` (the default) disables the check entirely.
 Scope
 -----
 
-The cooldown applies to **sdist resolution** — selecting which version of a
-package to build from source, including transitive dependencies. It does not
-apply to:
+The cooldown applies to both **sdist resolution** and **pre-built wheel
+resolution** — any candidate whose ``upload-time`` is more recent than the
+cutoff is rejected, regardless of whether it is an sdist or a prebuilt wheel.
 
-* Wheel-only lookups, including cache servers (``--cache-wheel-server-url``) and
-  packages configured as ``pre_built: true`` in variant settings. These use a
-  different trust model and are not subject to the cooldown regardless of which
-  server they are fetched from.
-* Packages resolved from Git URLs that do not provide timestamp metadata.
+The following are **not** subject to the cooldown:
 
-Note that sdist resolution from a private package index depends on
+* Fromager's internal build and cache wheel servers. These are not used for
+  version selection — they are checked only for already-resolved pinned
+  versions — so the cooldown has no insertion point.
+* Packages resolved from Git URLs. Git timestamps are set by the client, not
+  the server, and cannot be trusted for cooldown enforcement.
+
+Resolution from a private package index (sdist or wheel) depends on
 ``upload-time`` being present in the index's PEP 691 JSON responses. If the
-index does not provide that metadata, candidates will be rejected under the
-fail-closed policy described below.
+index does not provide that metadata, candidates are rejected under the
+fail-closed policy described below. Use ``resolver_dist.min_release_age: 0``
+to bypass cooldown for packages from indexes that structurally cannot supply
+timestamps.
 
 
 Fail-Closed Behavior
 --------------------
 
-If a candidate has no ``upload-time`` metadata — which can occur with older
-PyPI Simple HTML responses — it is rejected when a cooldown is active. Fromager
-uses the `PEP 691 JSON Simple API`_ when fetching package metadata, which
-reliably includes upload timestamps.
+If a candidate has no ``upload-time`` metadata — whether it is an sdist or a
+wheel — it is rejected when a cooldown is active. Fromager uses the
+`PEP 691 JSON Simple API`_ when fetching package metadata, which reliably
+includes upload timestamps for PyPI.org.
 
 .. _PEP 691 JSON Simple API: https://peps.python.org/pep-0691/
+
+For indexes that only implement the `PEP 503`_ HTML API and cannot supply
+timestamps, use the per-package ``resolver_dist.min_release_age: 0`` override
+to bypass the cooldown for affected packages rather than disabling it globally.
+
+.. _PEP 503: https://peps.python.org/pep-0503/
 
 .. note::
 
    If you are writing a ``get_resolver_provider`` plugin that uses
    :class:`~fromager.resolver.PyPIProvider` with a private index that only
-   implements the `PEP 503`_ HTML API (and therefore cannot supply upload
-   timestamps), pass ``supports_upload_time=False`` to ``PyPIProvider``. This
-   switches the provider from fail-closed to warn-and-skip, so candidates
-   without upload timestamps are skipped with a warning rather than causing
-   resolution to fail.
-
-.. _PEP 503: https://peps.python.org/pep-0503/
+   implements the PEP 503 HTML API, pass ``supports_upload_time=False`` to
+   ``PyPIProvider``. This switches the provider from fail-closed to
+   warn-and-skip, so candidates without upload timestamps are skipped with a
+   warning rather than causing resolution to fail.
 
 Example
 -------
--- a/e2e/github_override_example/src/package_plugins/stevedore_github.py
+++ b/e2e/github_override_example/src/package_plugins/stevedore_github.py
@@ -1,0 +1,25 @@
+from packaging.requirements import Requirement
+
+from fromager import context, resolver
+
+
+def get_resolver_provider(
+    ctx: context.WorkContext,
+    req: Requirement,
+    sdist_server_url: str,
+    include_sdists: bool,
+    include_wheels: bool,
+    req_type: resolver.RequirementType | None = None,
+    ignore_platform: bool = False,
+) -> resolver.GitHubTagProvider:
+    """Return a GitHubTagProvider for the stevedore test repo on github.com."""
+    return resolver.GitHubTagProvider(
+        organization="python-wheel-build",
+        repo="stevedore-test-repo",
+        constraints=ctx.constraints,
+        req_type=req_type,
+        override_download_url=(
+            "https://github.com/{organization}/{repo}"
+            "/archive/refs/tags/{tagname}.tar.gz"
+        ),
+    )
--- a/e2e/test_bootstrap_cooldown_gitlab.sh
+++ b/e2e/test_bootstrap_cooldown_gitlab.sh
@@ -1,0 +1,78 @@
+#!/bin/bash
+# -*- indent-tabs-mode: nil; tab-width: 2; sh-indentation: 2; -*-
+
+# Tests that --min-release-age applies to packages resolved through the
+# GitLabTagProvider, verifying that the cooldown uses GitLab tag commit
+# timestamps.
+#
+# The python-gitlab package is a good candidate because it is hosted on
+# gitlab.com, is pure Python, and has a well-documented release history.
+#
+# Release timeline (all times UTC):
+#
+#   python-gitlab v5.0.0   2024-10-28  (the expected fallback)
+#   python-gitlab v5.1.0   2024-11-28  (blocked by cooldown — the anchor date)
+#   python-gitlab v5.2.0+  2024-12-17+ (all blocked by cooldown)
+#
+# We compute --min-release-age dynamically as the age of v5.1.0 in days plus a
+# 1-day buffer.  This places the cutoff just before 2024-11-28, blocking v5.1.0
+# and all later releases while allowing v5.0.0 (released ~31 days earlier).
+#
+# Anchoring to v5.1.0 (2024-11-28) ensures the build toolchain can use
+# flit_core 3.10.0 (released 2024-10-31), which is required for Python 3.14
+# compatibility.
+#
+# The gitlab_override_example plugin (installed below) registers a
+# get_resolver_provider hook that returns a GitLabTagProvider for python-gitlab,
+# routing resolution through the GitLab tag API instead of PyPI.
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "$SCRIPTDIR/common.sh"
+
+# Install the override plugin that routes python-gitlab through GitLabTagProvider.
+# Uninstall on exit so its entry points don't leak into subsequent e2e tests.
+trap 'python3 -m pip uninstall -y gitlab_override_example >/dev/null 2>&1 || true' EXIT
+pip install "$SCRIPTDIR/gitlab_override_example"
+
+# Compute min-age: days since python-gitlab v5.1.0 was tagged, plus a buffer.
+# v5.1.0 was tagged 2024-11-28; adding 1 day ensures it is always just inside
+# the cooldown window regardless of when the test runs.
+MIN_AGE=$(python3 -c "
+from datetime import date
+age = (date.today() - date(2024, 11, 28)).days
+print(age + 1)
+")
+
+fromager \
+  --debug \
+  --log-file="$OUTDIR/bootstrap.log" \
+  --error-log-file="$OUTDIR/fromager-errors.log" \
+  --sdists-repo="$OUTDIR/sdists-repo" \
+  --wheels-repo="$OUTDIR/wheels-repo" \
+  --work-dir="$OUTDIR/work-dir" \
+  --min-release-age="$MIN_AGE" \
+  bootstrap 'python-gitlab'
+
+find "$OUTDIR/wheels-repo/" -name '*.whl'
+
+pass=true
+
+# v5.1.0 is blocked by the cooldown; the resolver must fall back to v5.0.0.
+if ! find "$OUTDIR/wheels-repo/downloads/" -name 'python_gitlab-5.0.0*.whl' | grep -q .; then
+  echo "FAIL: python_gitlab-5.0.0 wheel not found — cooldown did not force fallback" 1>&2
+  pass=false
+fi
+
+# Confirm newer versions were rejected by the cooldown.
+if find "$OUTDIR/wheels-repo/downloads/" -name 'python_gitlab-5.[1-9]*.whl' | grep -q .; then
+  echo "FAIL: a python_gitlab wheel newer than 5.0.0 was selected despite the cooldown" 1>&2
+  pass=false
+fi
+
+# Confirm the GitLabTagProvider was actually used (not the default PyPI provider).
+if ! grep -q "GitLabTagProvider" "$OUTDIR/bootstrap.log"; then
+  echo "FAIL: GitLabTagProvider was not used for resolution" 1>&2
+  pass=false
+fi
+
+$pass
--- a/e2e/test_bootstrap_cooldown_prebuilt.sh
+++ b/e2e/test_bootstrap_cooldown_prebuilt.sh
@@ -1,0 +1,65 @@
+#!/bin/bash
+# -*- indent-tabs-mode: nil; tab-width: 2; sh-indentation: 2; -*-
+
+# Tests that --min-release-age rejects pre-built wheel candidates published
+# within the cooldown window and falls back to an older stevedore version.
+# This exercises the resolve_all_prebuilt_wheels() code path in wheels.py,
+# which is taken when a package is configured with pre_built: true.
+#
+# Release timeline (wheel upload times, UTC):
+#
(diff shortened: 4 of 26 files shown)


iangelak:
I used AI to do a deep analysis of the PR and I have found a few issues. I list only the most relevant ones :

*Cooldown leaks into informational commands (Confirmed with manual testing)*

Problem: default_resolver_provider now reads ctx.cooldown unconditionally. Every call site that goes through overrides.find_and_invoke("get_resolver_provider", resolver.default_resolver_provider, ...) picks up the cooldown. This includes informational commands that are not performing a build:

In commands/package.py, list_versions at line ~146 and commands/find_updates.py, _find_newer_versions at line ~195

If a user runs fromager --min-release-age 7 package list-versions some-pkg, recently published versions are silently hidden from the output. Worse, fromager --min-release-age 7 find-updates would not report recently published versions as available updates — defeating the purpose of the command.

Suggestion: Either scope cooldown to bootstrap-only commands (by not setting ctx.cooldown for informational commands), or have list_versions/find_updates explicitly create providers without cooldown.

*Double PBI lookup — design inconsistency*

Problem: default_resolver_provider() now calls ctx.package_build_info(req) internally to read resolver_min_release_age. But all its callers (sources.get_source_provider, wheels.get_prebuilt_wheel_provider, commands/package.py, etc.) already call ctx.package_build_info(req) themselves to extract the other resolver settings (include_sdists, include_wheels, ignore_platform, sdist_server_url) and pass them as explicit arguments.

This breaks the existing pattern where default_resolver_provider is a stateless function of its arguments. Cooldown is the only resolver setting that's read inside the callee instead of being passed by the caller.

Suggestion: Either pass the resolved cooldown as an explicit argument (consistent with how all other PBI-derived settings flow), or add a comment acknowledging the deviation.

*Warning spam for GitHub provider (Confirmed with manual testing)*

Problem: The PR description says "providers without timestamp support (GitHub) emit a one-time warning." In practice, the warning fires once per candidate, not once per provider instance. A GitHub repo with 50 tags produces 50 identical "release-age cooldown cannot be enforced" warnings.

I confirmed this by running a test: 10 candidates through a single GitHubTagProvider with a cooldown active produces 10 identical warnings instead of 1.

Root cause: The warning in is_satisfied_by() has no deduplication. It fires every time the method is called for a candidate with upload_time=None on a provider with supports_upload_time=False.

Fix: Add a boolean latch on the provider instance.

iangelak:
Besides the above I tested the PR manually and everything else plus the e2e tests seem to be working fine!

ryanpetrello:
@dhellmann @iangelak more review feedback:

>  Why not have the resolver look up all of the settings itself? Do any of the callers modify those arguments?

At least one caller passes in raw values:

 https://github.com/python-wheel-build/fromager/blob/main/src/fromager/wheels.py#L488

> Double PBI lookup — design inconsistency

We _could_ refactor all this to force all callers to explicitly provide the new cooldown argument (or as @dhellmann suggested, we could attempt to move all of the PBI lookups into `default_resolver_provider`).  I am a little worried that pushing the parameter to *callers* trades style consistency with the opportunity for a correctness risk - it seems to me that forcing all callers to remember to pass the cooldown value just opens up the possibility for mistakes (where a cooldown value is mistakenly not provided by some caller).

Do you all have strong feelings about this review feedback, or should we proceed with the MR as-is in this regard?

> Warning spam for GitHub provider (Confirmed with manual testing)

I've pushed a change that addresses this.

dhellmann:
> @dhellmann @iangelak more review feedback:
> 
> > Why not have the resolver look up all of the settings itself? Do any of the callers modify those arguments?
> 
> At least one caller passes in raw values:
> 
> https://github.com/python-wheel-build/fromager/blob/main/src/fromager/wheels.py#L488
> 
> > Double PBI lookup — design inconsistency
> 
> We _could_ refactor all this to force all callers to explicitly provide the new cooldown argument (or as @dhellmann suggested, we could attempt to move all of the PBI lookups into `default_resolver_provider`). I am a little worried that pushing the parameter to _callers_ trades style consistency with the opportunity for a correctness risk - it seems to me that forcing all callers to remember to pass the cooldown value just opens up the possibility for mistakes (where a cooldown value is mistakenly not provided by some caller).
> 
> Do you all have strong feelings about this review feedback, or should we proceed with the MR as-is in this regard?

It should be hard for any code creating a resolver to _accidentally_ circumvent anything that constrains resolution, whether that is the constraints file or dates or anything else. So I'm fine with having the resolver look up the values internally, and we can look at whether it makes sense to move all of the rest of the PBI lookup inside the resolver as part of another change.

iangelak:
> It should be hard for any code creating a resolver to _accidentally_ circumvent anything that constrains resolution, whether that is the constraints file or dates or anything else. So I'm fine with having the resolver look up the values internally, and we can look at whether it makes sense to move all of the rest of the PBI lookup inside the resolver as part of another change.

I am also fine with keeping the implementation as is. Should we add a comment/todo to move the PBI lookup in the resolver in a future PR, though?

src/fromager/packagesettings/_models.py
```diff
@@ -96,6 +96,24 @@ class ResolverDist(pydantic.BaseModel):
...
 
+    min_release_age: int | None = None
+    """Per-package minimum release age override in days.
+
+    None (default): inherit the global ``--min-release-age`` setting.
+    0: disable the release-age cooldown for this package.
+    Positive integer: override the cooldown with this many days.
+
+    .. versionadded:: 0.XX
+    """
```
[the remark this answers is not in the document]
ryanpetrello: @python-wheel-build/fromager-maintainers is there any automation for properly capturing and annotating this as part of the release process, or is the expectation that you track this in PRs against the next sequential release tag?

(I don't see anything in `.github/workflows/python-publish.yaml`)

(in other words, should I manually set this to `0.80.0`)?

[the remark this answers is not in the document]
ryanpetrello: (because I'm feeling curious...)

Do you know the answer to this, @coderabbitai ?

src/fromager/__main__.py
```diff
@@ -143,6 +144,16 @@
     help="Build sdist and when with network isolation (unshare -cn)",
     show_default=True,
 )
+@click.option(
+    "--min-release-age",
+    type=click.IntRange(min=0),
+    default=0,
+    envvar="FROMAGER_MIN_RELEASE_AGE",
```
tiran: click is configured with `auto_envvar_prefix="FROMAGER"`. `FROMAGER_MIN_RELEASE_AGE` should be the default.

src/fromager/resolver.py
```diff
@@ -424,18 +459,29 @@ def get_project_from_pypi(
...
     resolver_cache: typing.ClassVar[ResolverCache] = {}
     provider_description: typing.ClassVar[str]
+    supports_upload_time: typing.ClassVar[bool] = True
+    """Does this provider supply upload timestamps for candidates?
+
+    Set to False on providers (e.g. GitHubTagProvider) that do not yet
+    support timestamp retrieval.  When a cooldown is active and this flag is
+    False, the cooldown check is skipped with a warning rather than
+    failing closed.
+    """
```
tiran: PyPI provider with PEP 503 simple HTTP index do not support upload time either. Only Python index with PEP 691 JSON response support upload time.

tiran (dismissed):
This is a very large PR. Could we split this into three PRs:

- proposal
- cooldown logic in resolver
- rest

ryanpetrello:
> This is a very large PR. Could we split this into three PRs:
> 
> * proposal
> * cooldown logic in resolver
> * rest

I believe @LalatenduMohanty already has another PR with the proposal: https://github.com/python-wheel-build/fromager/pull/1000

@tiran aside from the cooldown logic implementation, the majority of the code in this PR is actually the tests (especially the E2E ones):

```
  - tests + e2e: +1005 lines (77%)
  - src implementation: +188 lines (14%)
  - docs: +113 lines (9%)
```

Do we want those in a separate PR, or should we keep them in this one?

src/fromager/resolver.py
ryanpetrello: This is definitely a nuance of the current implementation, and it's called out in the docs (https://github.com/python-wheel-build/fromager/pull/1018/changes#diff-691bbf23153067de7080f4c1e5e055e2ea5497c368b44939a98219fcfdf6c9bbR60-R68).  We could probably add in auto-detection of PEP 691, but obviously it would expand the complexity of this PR.  Do we need this detection in this first pass implementation?

Given that this is a security control, my inclination is to stick with the current "Fail closed" behavior.  Maybe we should expand the error message slightly so that it spells out that the lack of PEP 691 support is the _most likely_ culprit?

```diff
diff --git a/src/fromager/resolver.py b/src/fromager/resolver.py
index 0fc023a..40f1dc7 100644
--- a/src/fromager/resolver.py
+++ b/src/fromager/resolver.py
@@ -832,8 +832,9 @@ class PyPIProvider(BaseProvider):
             if missing_time and not cooldown_blocked:
                 return (
                     f"found {len(missing_time)} candidate(s) for {r} but none have "
-                    f"upload timestamp metadata; cannot enforce the "
-                    f"{self.cooldown.min_age.days}-day cooldown"
+                    f"upload timestamp metadata; {self.sdist_server_url!r} may not "
+                    f"support PEP 691 (JSON API), which is required to enforce the "
+                    f"{self.cooldown.min_age.days}-day release-age cooldown"
                 )
             if cooldown_blocked:
                 oldest_days = min(
```

ryanpetrello: Alright, I've added this messaging to the error message to more explicitly spell this out when it's the mostly likely culprit.

This nuance is also documented in the feature documentation itself here:

https://github.com/python-wheel-build/fromager/pull/1018/changes#diff-691bbf23153067de7080f4c1e5e055e2ea5497c368b44939a98219fcfdf6c9bbR60-R68

src/fromager/packagesettings/_models.py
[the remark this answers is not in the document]
ryanpetrello: (@coderabbitai did in fact know the answer, this is resolved)

src/fromager/packagesettings/_models.py
```diff
@@ -96,6 +96,24 @@ class ResolverDist(pydantic.BaseModel):
     .. versionadded:: 0.70
     """
 
+    min_release_age: int | None = None
```
tiran: optional int field with `>= 0`
```suggestion
    min_release_age: int | None = Field(default=None, ge=0)
```

ryanpetrello:
```bash
~/dev/fromager fromager --min-release-age=-1
Usage: fromager [OPTIONS] COMMAND [ARGS]...
Try 'fromager --help' for help.

Error: Invalid value for '--min-release-age': -1 is not in the range x>=0.

~/dev/fromager fromager --min-release-age=spam
Usage: fromager [OPTIONS] COMMAND [ARGS]...
Try 'fromager --help' for help.

Error: Invalid value for '--min-release-age': 'spam' is not a valid integer range.
```

Fixed.

src/fromager/resolver.py
LalatenduMohanty, line 108: Cooldown is threaded through find_and_invoke() kwargs, so plugins only get it if they explicitly pass cooldown=ctx.cooldown at construction. If a plugin forgets, cooldown is silently disabled.

With this existing plugins with get_resolver_provider overrides would need update.                              
                                                                                                                                                                             
  Instead, resolve() should set cooldown on the provider after find_and_invoke() returns it. Per-package min_release_age resolution should also move here (currently in default_resolver_provider(), so it only works for the default path).                                                                                                       
   Something like this

```                                            
  provider = overrides.find_and_invoke(...)                                                                                                                                  
  pbi = ctx.package_build_info(req)                                                                         
  provider.cooldown = _resolve_package_cooldown(ctx.cooldown, pbi.resolver_min_release_age)
```

src/fromager/resolver.py
```diff
@@ -424,18 +459,29 @@ def get_project_from_pypi(
 class BaseProvider(ExtrasProvider):
     resolver_cache: typing.ClassVar[ResolverCache] = {}
     provider_description: typing.ClassVar[str]
+    supports_upload_time: typing.ClassVar[bool] = True
```
LalatenduMohanty: Lets check if GenericProvider and VersionMapProvider should override it to `False`. 
With a cooldown active, candidates from these providers will be rejected fail-closed because they have no upload_time. 
GenericProvider depends on its callback, some may supply timestamps, some won't. VersionMapProvider doesn't currently populate them.

LalatenduMohanty (changes requested):

src/fromager/resolver.py
ryanpetrello: I think it might make more sense to design this as "opt in" by provider i.e.,

- default to False on `BaseProvider` (safe/unknown)
- `PyPIProvider` and `GitLabTagProvider` explicitly opt in with `True`

...as we expand support to additional providers, they can also override `supports_upload_time = True`

src/fromager/resolver.py
ryanpetrello: I'm experimenting with this change.  I've introduced it in a new commit, and it caused some of the E2E tests to begin failing, so I'm investigating whether this change isn't "enough" for all of the potential code paths, or whether the test itself was wrong in the first place.

src/fromager/sources.py
```diff
@@ -153,6 +153,12 @@ def get_source_provider(
             ignore_platform=pbi.resolver_ignore_platform,
         ),
     )
+    provider.cooldown = (
```
ryanpetrello: @LalatenduMohanty it bothers me a little that we're having to do this in multiple places in this diff, but I don't see an alternative that keeps provider constructor signatures stable and gives plugin authors the ability to not have to think about passing through cooldown settings at all.

src/fromager/resolver.py
ryanpetrello: Alright @LalatenduMohanty I've landed on a second commit on this PR that I think should address this limitation you brought up (and the E2E tests I've written still pass with this change).

src/fromager/resolver.py
ryanpetrello: I've made this change 👍

src/fromager/sources.py
```diff
@@ -153,6 +153,12 @@ def get_source_provider(
...
+        resolver._resolve_package_cooldown(ctx.cooldown, pbi.resolver_min_release_age)
```
LalatenduMohanty: `resolver._resolve_package_cooldown()` is a private function,  wondering if we should rename this to `resolve_package_cooldown()`.

Also this could be cleaner as a method on WorkContext since it already has both self.cooldown and self.package_build_info():                                              

 ```  
  # context.py                                                                                                                                    
  def cooldown_for(self, req: Requirement) -> Cooldown | None:                                              
      pbi = self.package_build_info(req)                                                                                                          
      ...
```
Call sites would become provider.cooldown = ctx.cooldown_for(req). 

We can do it as a follow-up PR. we should track it in a separate issue.

File: tests/test_cooldown.py [hunk 8]
```diff
@@ -0,0 +1,560 @@
...
+    assert result.min_age.days == 14
+
+
+def test_per_package_cooldown_disable_via_ctx(tmp_path: pathlib.Path) -> None:
+    """resolver_dist.min_release_age=0 disables cooldown for a specific package.
+
+    Even when the global cooldown is active, a package with min_release_age=0
+    in its settings should resolve the latest version.
+    """
+    import yaml
```
LalatenduMohanty: As per https://github.com/python-wheel-build/fromager/blob/main/CONTRIBUTING.md#import-organization , all imports should be at the top of the file

src/fromager/sources.py
```diff
@@ -153,6 +153,12 @@ def get_source_provider(
...
+        if pbi.resolver_include_sdists
+        else None
+    )
```
LalatenduMohanty: We should add a test that calls sources.get_source_provider() with a cooldown-enabled context and verifies the returned provider has cooldown set

LalatenduMohanty (changes requested):

src/fromager/sources.py
ryanpetrello: 😄 I actually beat you to it - I've just renamed the function to `resolve_package_cooldown` and left it in-place for now.

tests/test_cooldown.py
ryanpetrello: Addressed.

src/fromager/sources.py
ryanpetrello: Addressed.

e2e/github_override_example/src/package_plugins/stevedore_github.py
```diff
@@ -18,7 +18,6 @@ def get_resolver_provider(
         repo="stevedore-test-repo",
         constraints=ctx.constraints,
         req_type=req_type,
-        cooldown=ctx.cooldown,
```
ryanpetrello: @LalatenduMohanty I wanted to call out this change specifically from this diff - the fact that I'm able to remove these from the example package plugins (and the cooldown E2E tests still pass) demonstrates that we've addressed your original concern in https://github.com/python-wheel-build/fromager/pull/1018#discussion_r3081847716

Thanks for calling this out - while the risk here is probably fairly low for our use case, I think the change you suggested makes this a less error-prone interface for consumers of this new feature.

LalatenduMohanty: I agree.

src/fromager/resolver.py
```diff
@@ -424,18 +464,30 @@ def get_project_from_pypi(
...
         req_type: RequirementType | None = None,
         use_resolver_cache: bool = True,
+        cooldown: Cooldown | None = None,
     ):
         super().__init__()
         self.constraints = constraints or Constraints()
         self.req_type = req_type
         self.use_cache_candidates = use_resolver_cache
+        self.cooldown = cooldown
+        self.cooldown_unsupported_previously_warned = False
```
LalatenduMohanty: The proposal (PR #1000) has been updated to make supports_upload_time an instance attribute set during construction rather than a ClassVar. 

BaseProvider.__init__ should set self.supports_upload_time = False as the default. With ClassVar, subclass providers cannot override the value per-instance.

src/fromager/resolver.py
```diff
@@ -635,6 +726,8 @@ class PyPIProvider(BaseProvider):
     * version (Version object)
     """
 
+    supports_upload_time: typing.ClassVar[bool] = True
```
LalatenduMohanty: The updated proposal says PyPIProvider should accept supports_upload_time as a constructor parameter (defaulting to True) so callers can pass False for indexes that only serve PEP 503 Simple HTML with no upload-time field.

 With ClassVar every PyPIProvider instance will fail-closed on missing timestamps, even against mirrors that structurally cannot provide them.

LalatenduMohanty (changes requested):

src/fromager/resolver.py
```diff
@@ -133,6 +138,37 @@ def default_resolver_provider(
     )
 
 
+def resolve_package_cooldown(
```
tiran, line 138: is this an internal helper function or an API that is used by multiple modules and external users?

In the first case, it should be renamed to `_resolve_package_cooldown()`. In the second case it should accept `ctx: context.WorkContext, req: Requirement`. As a general rule, we pass context and requirement instead of attributes of context and PBI.

src/fromager/resolver.py
```diff
@@ -552,6 +603,45 @@ def is_satisfied_by(self, requirement: Requirement, candidate: Candidate) -> boo
                 )
             return False
 
+        # Release-age cooldown: reject candidates published too recently.
```
tiran: Can you move this block to private helper function? It makes it easier to understand and test.

src/fromager/resolver.py
```diff
@@ -552,6 +603,45 @@ def is_satisfied_by(self, requirement: Requirement, candidate: Candidate) -> boo
...
+        if self.cooldown is not None:
+            if candidate.upload_time is None:
+                if not self.supports_upload_time:
+                    # Provider does not yet support timestamp retrieval (e.g. GitHub).
+                    # Warn once per provider instance (i.e. per package) rather than
+                    # once per candidate.
+                    if not self.cooldown_unsupported_previously_warned:
```
tiran: We are going to build a lot of versions of a lot of packages. The `cooldown_unsupported_previously_warned` attribute is an instance attribute.

Suggestion:

- add a class variable to `BaseProvider` that tracks `req.name` as a `set[str]`
- only warn when `req.name` is not in the set, yet.

This will warn exactly once for each requirement.

src/fromager/resolver.py
```diff
@@ -656,12 +746,24 @@ def __init__(
         *,
         use_resolver_cache: bool = True,
         override_download_url: str | None = None,
+        cooldown: Cooldown | None = None,
+        supports_upload_time: bool = True,
```
tiran: I do not agree with `supports_upload_time` defaulting to `True`.

Either set the default to `False` or do something like:

```python
supports_upload_time: bool | None = None
if supports_upload_time is None:
    supports_upload_time = sdist_server_url.startswith("https://pypi.org/simple")
```

src/fromager/wheels.py
```diff
@@ -549,6 +549,15 @@ def resolve_all_prebuilt_wheels(
...
             )
+            pbi = ctx.package_build_info(req)
+            provider.cooldown = resolver.resolve_package_cooldown(
+                ctx.cooldown, pbi.resolver_min_release_age
+            )
+            # The local fromager wheel server is PEP 503-only and serves
+            # packages that were already resolved and vetted earlier in the
+            # same run. Don't fail-closed on missing upload_time there.
+            if ctx.wheel_server_url and url == ctx.wheel_server_url:
+                provider.supports_upload_time = False
```
tiran: It looks weird to modify the provider after it was created.

src/fromager/resolver.py
ryanpetrello: Okay, I think this is a fairly practical change, I like the idea to default to `True` for public PyPI.

src/fromager/wheels.py
ryanpetrello: I agree it looks weird, but the alternative would mean threading this argument through:

```
  get_prebuilt_wheel_provider(supports_upload_time=False)   # wheels.py:501
    └── overrides.find_and_invoke(..., supports_upload_time=False)  # wheels.py:515
          └── default_resolver_provider(..., supports_upload_time=False)  # resolver.py:112
                └── PyPIProvider(..., supports_upload_time=False)  # resolver.py:128
```

...and `PyPIProvider` is the only provider where this argument is particularly relevant because it's the only provider where the behavior is *conditional* (i.e., only *some* PyPI implementations are PEP 503-only).

Other providers, in contrast, don't need a "hint" about whether or not they provide/support upload times.

So despite it looking a little weird, I'm inclined to just stick with this version of the code out of pragmatism.

ryanpetrello:
@tiran for review simplicity, I've added an additional commit that I think addresses all your feedback:

https://github.com/python-wheel-build/fromager/pull/1018/changes/74447dbdb46c1e726534870d798321d4d506892f

Once you're happy with it, I'll squash commits.

src/fromager/resolver.py
ryanpetrello: Done.

src/fromager/resolver.py
ryanpetrello: Done.

src/fromager/resolver.py
ryanpetrello: Done.

src/fromager/resolver.py
ryanpetrello: Done.

docs/how-tos/release-age-cooldown.rst
```diff
@@ -41,43 +41,50 @@ A value of ``0`` (the default) disables the check entirely.
 Scope
 -----
 
-The cooldown applies to **sdist resolution** — selecting which version of a
-package to build from source, including transitive dependencies. It does not
-apply to:
+The cooldown applies to both **sdist resolution** and **pre-built wheel
+resolution** — any candidate whose ``upload-time`` is more recent than the
+cutoff is rejected, regardless of whether it is an sdist or a wheel.
```
LalatenduMohanty: Nitpick: s/regardless of whether it is an sdist or a wheel./regardless of whether it is an sdist or a prebuilt wheel./

src/fromager/resolver.py
```diff
@@ -430,19 +464,34 @@ def get_project_from_pypi(
...
+        self.cooldown = cooldown
+        """
+        Does this provider supply upload timestamps for candidates?
+
+        Defaults to False (safe/unknown). Subclasses that reliably populate
+        upload_time on every candidate should set this to True in their __init__.
+
+        When a cooldown is active and this is False, the cooldown check is
+        skipped with a warning rather than failing closed.
+        """
```
LalatenduMohanty: Nitpick: Looks like the doc string is misplaced.

LalatenduMohanty:
Follow-up issues for Phase 2 and Phase 3 of the [release cooldown proposal](https://github.com/python-wheel-build/fromager/blob/main/docs/proposals/release-cooldown.md):

- #1078 — `list-versions`: show cooldown status and upload timestamps (Phase 2)
- #1079 — `list-overrides`: show per-package `min_release_age` values (Phase 2)
- #1080 — GitHub timestamp support for cooldown enforcement (Phase 3)

e2e/test_bootstrap_cooldown_prebuilt.sh
```diff
@@ -0,0 +1,66 @@
...
+
+# Wheel file must exist in the prebuilt directory.
+if ! find "$OUTDIR/wheels-repo/prebuilt/" -name 'stevedore-5.3.0*.whl' | grep -q .; then
+  echo "FAIL: stevedore-5.3.0 wheel not found in wheels-repo/prebuilt" 1>&2
+  pass=false
+fi
+
+# No stevedore sdist should have been downloaded — it is pre_built only.
+if find "$OUTDIR/sdists-repo/" -name 'stevedore*.tar.gz' -o \
+        "$OUTDIR/sdists-repo/" -name 'stevedore*.zip' 2>/dev/null | grep -q .; then
```
LalatenduMohanty: find treats everything after the first path as expressions, not additional paths. So "$OUTDIR/sdists-repo/" appearing after -o is not a second search path  it's a bare string in expression position, which is a syntax error.  That error goes to /dev/null (the 2>/dev/null at the end).

So even if stevedore sdists are sitting in sdists-repo/, the broken find command will never find them

LalatenduMohanty (changes requested):

src/fromager/resolver.py
ryanpetrello: Yea, this isn't intended as an `__init__.py` docstring, it's just a code comment, I'll convert it.

docs/how-tos/release-age-cooldown.rst
ryanpetrello: Agree.

e2e/test_bootstrap_cooldown_prebuilt.sh
ryanpetrello: Good find.

ryanpetrello: (I feel the need to state - that was not an intentional pun)

src/fromager/resolver.py
ryanpetrello: Fixed!

docs/how-tos/release-age-cooldown.rst
ryanpetrello: Fixed!

src/fromager/wheels.py
LalatenduMohanty: I created https://github.com/python-wheel-build/fromager/issues/1081 as a follow up on this issue.

src/fromager/resolver.py
LalatenduMohanty: Resolving this as this is addressed.

src/fromager/resolver.py
LalatenduMohanty: Resolving this as this is addressed.

src/fromager/resolver.py
LalatenduMohanty: Resolving this as this is addressed.

src/fromager/resolver.py
LalatenduMohanty: The function  accepting  ctx, req — signature is (ctx: context.WorkContext, req: Requirement) . Resolving the review.

LalatenduMohanty (approved):

e2e/test_bootstrap_cooldown_gitlab.sh
```diff
@@ -0,0 +1,76 @@
...
+if ! find "$OUTDIR/wheels-repo/downloads/" -name 'python_gitlab-5.0.0*.whl' | grep -q .; then
+  echo "FAIL: python_gitlab-5.0.0 wheel not found — cooldown did not force fallback" 1>&2
+  pass=false
+fi
+
+# Confirm newer versions were rejected by the cooldown.
+if find "$OUTDIR/wheels-repo/downloads/" -name 'python_gitlab-5.[1-9]*.whl' | grep -q .; then
+  echo "FAIL: a python_gitlab wheel newer than 5.0.0 was selected despite the cooldown" 1>&2
+  pass=false
+fi
```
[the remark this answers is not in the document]
LalatenduMohanty: @ryanpetrello we should fix it. PTAL

LalatenduMohanty (approved):

e2e/test_bootstrap_cooldown_gitlab.sh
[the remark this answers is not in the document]
LalatenduMohanty: It's a one-line cosmetic improvement to a test assertion. resolving it

PR #1026, opened 2026-04-06
fix(build_environment): re-raise exception for the JSON parse failure

### LINKED ISSUE #1023: fix: get_distributions() crashes with UnboundLocalError on JSON parse failure
## Summary

In `build_environment.py:232-244`, `get_distributions()` crashes with `UnboundLocalError` when `json.loads()` fails, hiding the real `json.JSONDecodeError`.

## Problem

```python
def get_distributions(self) -> typing.Mapping[str, Version]:
    ...
    try:
        mapping = json.loads(result.strip())
    except Exception:
        logger.exception("failed to de-serialize JSON: %s", result)
    return {name: Version(version) for name, version in sorted(mapping.items())}
```

If `json.loads()` raises an exception, the `except` block logs the error but does not return, re-raise, or assign a fallback. Execution falls through to line 244 where `mapping` is undefined, producing `UnboundLocalError: local variable 'mapping' referenced before assignment`.

This masks the real error (e.g. subprocess printed warnings to stdout mixed with JSON) and sends debugging in the wrong direction.

## Suggested fix

Add `raise` after the log statement:

```python
except Exception:
    logger.exception("failed to de-serialize JSON: %s", result)
    raise
```

**Effort:** 1 line change.

## Testing

The fix should include a test. For example, mock `BuildEnvironment.run()` to return invalid JSON, call `get_distributions()`, and assert `json.JSONDecodeError` is raised (not `UnboundLocalError`). A sanity check with valid JSON should also be included to ensure the happy path still works.

LalatenduMohanty:
Previously, if json.loads() failed, the except block logged the error but did not re-raise, causing an UnboundLocalError on the next line where `mapping` was undefined. Now the original exception is re-raised.

Closes: #1023

### DIFF
--- a/src/fromager/build_environment.py
+++ b/src/fromager/build_environment.py
@@ -241,6 +241,7 @@ class BuildEnvironment:
             mapping = json.loads(result.strip())
         except Exception:
             logger.exception("failed to de-serialize JSON: %s", result)
+            raise
         return {name: Version(version) for name, version in sorted(mapping.items())}
 
 
--- a/tests/test_build_environment.py
+++ b/tests/test_build_environment.py
@@ -1,6 +1,8 @@
+import json
 import textwrap
 from unittest.mock import Mock, patch
 
+import pytest
 from packaging.requirements import Requirement
 from packaging.version import Version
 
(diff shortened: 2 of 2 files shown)


rd4398 (approved):
Looks good to me!

PR #1030, opened 2026-04-06
feat(graph): add suggest-collection command for collection optimization

### LINKED ISSUE #971: Add `graph suggest-collection` command to optimize collection organization
# Add `graph suggest-collection` command to optimize collection organization

## Summary

Add a new subcommand `fromager graph suggest-collection` that analyzes dependency overlap between onboarding packages and existing collections to recommend optimal collection assignments. This helps maintainers decide which collection each package should join based on minimizing the number of new packages that would need to be added.

## Motivation

When onboarding new packages to fromager builds, maintainers must decide which existing collection each package should join. This decision should be based on dependency overlap - packages that share many dependencies with an existing collection are good candidates for that collection.

Currently, this analysis is done manually, which is:
- **Time-consuming**: With 100+ onboarding packages and their transitive dependencies, manual analysis is slow
- **Error-prone**: It's easy to miss shared dependencies or assign packages suboptimally
- **Opaque**: No visibility into how well different collections would fit

The result is suboptimal collection organization with duplicated packages across collections, leading to inefficient builds.

## Proposed Solution

Create a new command that automates dependency analysis:

```bash
fromager graph suggest-collection ONBOARDING-GRAPH COLLECTION-GRAPHS...
```

For each top-level package in the onboarding graph:
1. Calculate its full dependency closure (all transitive dependencies)
2. Compare this closure against each existing collection
3. Rank collections by "fit" - how many new packages would need to be added
4. Display recommendations with statistics

## Command Interface

```bash
fromager graph suggest-collection [OPTIONS] ONBOARDING-GRAPH COLLECTION-GRAPHS...
```

**Arguments:**
- `ONBOARDING-GRAPH`: Graph file (graph.json) containing packages to assign
- `COLLECTION-GRAPHS`: One or more graph files for existing collections

**Options:**
- `--format [table|json]`: Output format (default: table)
  - `table`: Human-readable rich table showing best-fit collection for each package
  - `json`: Machine-readable JSON with all collections ranked for each package
- `--help`: Show help message

**Example:**
```bash
fromager graph suggest-collection \
  onboarding-graph.json \
  notebook-graph.json \
  rhai-innovation-graph.json
```

### Core Functionality

1. **Load graph files**: Parse multiple graph.json files using `DependencyGraph.from_file()`

2. **Identify top-level packages**: Extract packages marked with `req_type: 'toplevel'` from the ROOT node in the onboarding graph

3. **Calculate dependency closures**: For each top-level package, traverse the full dependency tree to get all transitive dependencies (both install and build dependencies)

4. **Compare by package name only**: When comparing dependencies, match packages by canonical name ignoring version numbers (e.g., `numpy==1.26.0` and `numpy==1.25.0` are considered the same package)

5. **Rank collections by fit**: For each top-level package and collection pair:
   - Calculate set difference: packages in dependency closure NOT in collection
   - Calculate set intersection: packages in dependency closure AND in collection
   - Primary sort: fewest new packages (ascending)
   - Secondary sort: highest coverage percentage (descending)

6. **Display results**: Show each package with its best-fit collection and statistics

### Output Requirements

**Table format** (default `--format table`):
- One row per onboarding package
- Columns: Package name, Version, Total dependencies, Best fit collection, New packages, Existing packages, Coverage %
- Sorted alphabetically by package name
- Clear, scannable output for human review

**JSON format** (`--format json`):
- Array of objects, one per onboarding package
- Each object includes all collections ranked by fit
- Valid, parseable JSON for automation/scripting
- Structure enables downstream processing

## Algorithm

For each top-level package in the onboarding graph:

1. **Extract dependency closure**:
   ```python
   closure = set()  # Set of canonical package names
   visited = set()  # Set of node keys to prevent cycles
   stack = [top_level_node]

   while stack:
       node = stack.pop()
       if node.key in visited:
           continue
       visited.add(node.key)
       closure.add(node.canonicalized_name)

       # Add all children (both install and build deps)
       for edge in node.children:
           stack.append(edge.destination_node)
   ```

2. **Calculate fit with each collection**:
   ```python
   for collection in collections:
       collection_packages = {node.canonicalized_name for node in collection.get_all_nodes()}
       new_packages = closure - collection_packages
       existing_packages = closure & collection_packages

       fit_score = {
           "new_packages": len(new_packages),
           "existing_packages": len(existing_packages),
           "coverage_pct": len(existing_packages) / len(closure) * 100
       }
   ```

3. **Rank collections**:
   - Sort by `new_packages` (ascending) - fewer is better
   - Then by `coverage_pct` (descending) - higher is better

### Table Format (Default)

```
Collection Suggestions for Onboarding Packages

┃ Package             ┃ Version ┃ Total Deps ┃ Best Fit        ┃ New Pkgs ┃ Existing ┃ Coverage ┃
┃━━━━━━━━━━━━━━━━━━━━━┃━━━━━━━━━┃━━━━━━━━━━━━┃━━━━━━━━━━━━━━━━━┃━━━━━━━━━━┃━━━━━━━━━━┃━━━━━━━━━━┃
┃ fastapi             ┃ 0.135.1 ┃ 45         ┃ notebook        ┃ 3        ┃ 42       ┃ 93.3%    ┃
┃ instructorembedding ┃ 1.0.1   ┃ 123        ┃ notebook        ┃ 8        ┃ 115      ┃ 93.5%    ┃
┃ torch               ┃ 2.9.1   ┃ 89         ┃ rhai-innovation ┃ 12       ┃ 77       ┃ 86.5%    ┃
```

**Interpretation**:
- `fastapi` should join the `notebook` collection - only 3 new packages needed (93.3% overlap)
- `torch` fits better in `rhai-innovation` despite needing 12 new packages (still 86.5% overlap)

### JSON Format (`--format json`)

```json
[
  {
    "package": "fastapi",
    "version": "0.135.1",
    "total_dependencies": 45,
    "collections": [
      {
        "collection": "notebook",
        "new_packages": 3,
        "existing_packages": 42,
        "total_dependencies": 45,
        "coverage_pct": 93.3
      },
      {
        "collection": "rhai-innovation",
        "new_packages": 15,
        "existing_packages": 30,
        "total_dependencies": 45,
        "coverage_pct": 66.7
      }
    ]
  },
  {
    "package": "torch",
    "version": "2.9.1",
    "total_dependencies": 89,
    "collections": [
      {
        "collection": "rhai-innovation",
        "new_packages": 12,
        "existing_packages": 77,
        "total_dependencies": 89,
        "coverage_pct": 86.5
      },
      {
        "collection": "notebook",
        "new_packages": 34,
        "existing_packages": 55,
        "total_dependencies": 89,
        "coverage_pct": 61.8
      }
    ]
  }
]
```

### File Location
Add to `/src/fromager/commands/graph.py` as a new subcommand:

```python
@graph.command()
@click.option("--format", "output_format", type=click.Choice(["table", "json"]), default="table")
@click.argument("onboarding-graph", type=str)
@click.argument("collection-graphs", type=str, nargs=-1, required=True)
@click.pass_obj
def suggest_collection(
    wkctx: context.WorkContext,
    onboarding_graph: str,
    collection_graphs: tuple[str, ...],
    output_format: str,
) -> None:
    """Suggest which collection each onboarding package should join."""
```

### Utilities to Reuse

From existing codebase:
- `DependencyGraph.from_file(path)` - Load graph files
- `graph.get_root_node()` - Get ROOT node to find top-level packages
- `graph.get_all_nodes()` - Get all nodes for building package sets
- `node.canonicalized_name` - Get normalized package name (ignore version)
- `node.children` - Traverse dependency edges
- `canonicalize_name()` from `packaging.utils` - Normalize names
- `RequirementType.TOP_LEVEL` - Filter for top-level packages

### Reference Implementations

For patterns and style:
- **Table output**: See `src/fromager/commands/stats.py` for rich table construction
- **JSON output**: See `src/fromager/commands/list_overrides.py` for `--format` option pattern
- **Graph traversal**: See `src/fromager/commands/graph.py` for existing graph walking patterns

### Helper Functions Needed

```python
def get_dependency_closure(graph: DependencyGraph, node: DependencyNode) -> set[NormalizedName]:
    """Get all transitive dependencies of a node by canonicalized name."""

def get_package_names(graph: DependencyGraph) -> set[NormalizedName]:
    """Extract all unique package names from a graph."""

def extract_collection_name(graph_path: str) -> str:
    """Parse collection name from graph file path."""

def output_table(results: list[dict]) -> None:
    """Display results as a rich table."""

def output_json(results: list[dict]) -> None:
    """Display results as JSON."""
```

### Unit Tests

Add to `/tests/test_graph.py`:

```python
def test_get_dependency_closure() -> None:
    """Test extracting full dependency closure by name."""

def test_get_package_names() -> None:
    """Test extracting unique package names from graph."""

def test_extract_collection_name() -> None:
    """Test parsing collection names from file paths."""

def test_suggest_collection_basic(tmp_path: pathlib.Path) -> None:
    """Test basic collection suggestion logic."""
```

### Integration Tests

```python
def test_suggest_collection_table_output(tmp_path: pathlib.Path) -> None:
    """Test command with table output format."""

def test_suggest_collection_json_output(tmp_path: pathlib.Path) -> None:
    """Test command with JSON output format."""

def test_suggest_collection_multiple_collections(tmp_path: pathlib.Path) -> None:
    """Test with multiple collection graphs."""
```

### Test Data

Use real graph files for integration testing or create minimal test graphs with known dependency structures.

## Edge Cases

The implementation should handle:

1. **Empty onboarding graph** (no top-level packages):
   - Output message: "No top-level packages found in onboarding graph"
   - Exit gracefully

2. **Package with no dependencies**:
   - Dependency closure contains only the package itself
   - All collections have same fit (0 new packages)

3. **Multiple versions of same package**:
   - Already handled by using `canonicalized_name` instead of versioned `key`
   - Version differences should not affect fit calculation

4. **Circular dependencies**:
   - Use `visited` set in traversal to prevent infinite loops
   - Each node visited only once

5. **Tied fit scores**:
   - Multiple collections with same `new_packages` count
   - Use `coverage_pct` as tiebreaker (higher is better)

6. **Collection name extraction**:
   - Handle various file naming patterns: `3.4.2293+notebook-cuda12.9-ubi9-aarch64-graph.json`
   - Fallback to filename without extensions if pattern doesn't match

7. **Large graphs**:
   - Algorithm is O(n*m) where n=top-level packages, m=collections
   - Should be acceptable for typical sizes (100s of packages, handful of collections)

## Success Criteria

- [ ] Command loads multiple graph files correctly
- [ ] Identifies all top-level packages from onboarding graph
- [ ] Calculates complete dependency closures
- [ ] Compares packages by canonical name (ignoring versions)
- [ ] Ranks collections correctly (fewest new packages first)
- [ ] Table output is clear and easy to read
- [ ] JSON output is valid and machine-parseable
- [ ] All unit tests pass
- [ ] All integration tests pass
- [ ] Edge cases handled gracefully
- [ ] Code follows fromager standards (type hints, docstrings, formatting)
- [ ] Documentation updated (if applicable)

## Related Work

Existing graph commands that may provide useful patterns:
- `graph to-constraints` - Loads graphs, processes nodes
- `graph explain-duplicates` - Analyzes multiple package versions
- `graph why` - Traverses dependency chains
- `stats` - Displays statistics with rich tables

smoparth:
# Pull Request Description

## What

Add fromager graph suggest-collection that analyzes dependency overlap between onboarding packages and existing collections to recommend the best-fit collection for each package. Helps pipeline maintainers make data-driven decisions when assigning new packages to permanent collections.

- Iterative DFS traversal over all edge types for full dependency closure
- Ranks collections by fewest new packages, then highest coverage
- Rich table (default) and JSON output formats
- Error handling for bad graph files and duplicate collection names
- Tests covering helpers, CLI, and analysis logic

## Why

Closes: #971

### DIFF
--- a/src/fromager/commands/graph.py
+++ b/src/fromager/commands/graph.py
@@ -7,9 +7,12 @@ import sys
 import typing
 
 import click
+import rich
+import rich.box
 from packaging.requirements import Requirement
-from packaging.utils import canonicalize_name
+from packaging.utils import NormalizedName, canonicalize_name
 from packaging.version import Version
+from rich.table import Table
 
 from fromager import clickext, context
 from fromager.commands import bootstrap
@@ -784,3 +787,252 @@ def build_graph(
         topo.done(*nodes_to_build)
 
     print(f"\nBuilding {len(graph)} packages in {rounds} rounds.")
+
+
+def get_dependency_closure(node: DependencyNode) -> set[NormalizedName]:
+    """Compute the full dependency closure for a node.
+
+    Traverses all edge types and returns the set of canonical package names reachable from node,
+    including node itself.
+
+    Args:
+        node: The starting node to compute the closure for.
+
+    Returns:
+        Set of canonicalized package names in the transitive closure.
+    """
+    dependency_names: set[NormalizedName] = set()
+    if node.canonicalized_name != ROOT:
+        dependency_names.add(node.canonicalized_name)
+    for dependency in node.iter_all_dependencies():
+        if dependency.canonicalized_name != ROOT:
+            dependency_names.add(dependency.canonicalized_name)
+    return dependency_names
+
+
+def get_package_names(graph: DependencyGraph) -> set[NormalizedName]:
+    """Extract all unique canonical package names from a graph.
+
+    Args:
+        graph: The dependency graph to extract names from.
+
+    Returns:
+        Set of canonicalized package names, excluding the ROOT node.
+    """
+    return {
+        node.canonicalized_name for node in graph.get_all_nodes() if node.key != ROOT
+    }
+
+
+def extract_collection_name(graph_path: str) -> str:
+    """Derive a collection name from a graph file path.
+
+    Returns the filename without the extension as a string.
+
+    Args:
+        graph_path: Filesystem path to a graph JSON file.
+
+    Returns:
+        The filename without the extension.
+    """
+    return pathlib.PurePath(graph_path).stem
+
+
+class _CollectionScore(typing.NamedTuple):
+    """Overlap score between a package's dependency closure and a collection."""
+
+    collection: str
+    new_packages: int
+    existing_packages: int
+    coverage_percentage: float
+
+
+def _analyze_suggestions(
+    toplevel_nodes: list[DependencyNode],
+    collection_packages: dict[str, set[NormalizedName]],
+) -> list[dict[str, typing.Any]]:
+    """Score each onboarding top-level package against every collection.
+
+    Args:
+        toplevel_nodes: Top-level nodes from the onboarding graph.
+        collection_packages: Mapping of collection name to its package name set.
+
+    Returns:
+        List of result dicts, one per top-level package, sorted by package name.
+    """
+    results: list[dict[str, typing.Any]] = []
+
+    for node in sorted(toplevel_nodes, key=lambda n: n.canonicalized_name):
+        dependency_names = get_dependency_closure(node)
+        total_dependency_count = len(dependency_names)
+
+        scores: list[_CollectionScore] = []
+        for collection_name, packages in collection_packages.items():
+            existing_count = len(dependency_names & packages)
+            new_count = total_dependency_count - existing_count
+            coverage_percentage = (
+                (existing_count / total_dependency_count * 100)
+                if total_dependency_count
+                else 0.0
+            )
+            scores.append(
+                _CollectionScore(
+                    collection_name, new_count, existing_count, coverage_percentage
+                )
+            )
+
+        # Rank: fewest new packages, then highest coverage, then name for determinism
+        scores.sort(
+            key=lambda score: (
+                score.new_packages,
+                -score.coverage_percentage,
+                score.collection,
+            )
+        )
+        best_score = scores[0] if scores else None
+
+        logger.debug(
+            "%s: %d deps, best fit '%s' (%d new, %.1f%% coverage)",
+            node.canonicalized_name,
+            total_dependency_count,
+            best_score.collection if best_score else "none",
+            best_score.new_packages if best_score else 0,
+            best_score.coverage_percentage if best_score else 0.0,
+        )
+
+        results.append(
+            {
+                "package": str(node.canonicalized_name),
+                "version": str(node.version),
+                "total_dependencies": total_dependency_count,
+                "best_fit": best_score.collection if best_score else "none",
+                "new_packages": best_score.new_packages if best_score else 0,
+                "existing_packages": best_score.existing_packages if best_score else 0,
+                "coverage_percentage": (
+                    round(best_score.coverage_percentage, 1) if best_score else 0.0
+                ),
+                "all_collections": [
+                    {
+                        "collection": score.collection,
+                        "new_packages": score.new_packages,
+                        "existing_packages": score.existing_packages,
+                        "coverage_percentage": round(score.coverage_percentage, 1),
+                    }
+                    for score in scores
+                ],
+            }
+        )
+
+    return results
+
+
+def _print_suggest_collection_table(
+    results: list[dict[str, typing.Any]],
+) -> None:
+    """Render suggest-collection results as a Rich table."""
+    table = Table(
+        title="Collection Suggestions for Onboarding Packages",
+        box=rich.box.MARKDOWN,
+        title_justify="left",
+    )
+    table.add_column("Package", justify="left", no_wrap=True)
+    table.add_column("Version", justify="left", no_wrap=True)
+    table.add_column("Total Deps", justify="right", no_wrap=True)
+    table.add_column("Best Fit", justify="left", no_wrap=True)
+    table.add_column("New Pkgs", justify="right", no_wrap=True)
+    table.add_column("Existing", justify="right", no_wrap=True)
+    table.add_column("Coverage", justify="right", no_wrap=True)
+
+    for result in results:
+        table.add_row(
+            result["package"],
+            result["version"],
+            str(result["total_dependencies"]),
+            result["best_fit"],
+            str(result["new_packages"]),
+            str(result["existing_packages"]),
+            f"{result['coverage_percentage']:.1f}%",
+        )
+
+    rich.get_console().print(table)
+
+
+@graph.command(name="suggest-collection")
+@click.option(
+    "--format",
+    "output_format",
+    type=click.Choice(["table", "json"], case_sensitive=False),
+    default="table",
+    help="Output format (default: table)",
+)
+@click.argument("onboarding-graph", type=str)
+@click.argument("collection-graphs", nargs=-1, required=True, type=str)
+def suggest_collection(
+    output_format: str,
+    onboarding_graph: str,
+    collection_graphs: tuple[str, ...],
+) -> None:
+    """Suggest the best-fit collection for each onboarding package.
+
+    Analyzes dependency overlap between top-level packages in ONBOARDING_GRAPH
+    and the existing COLLECTION_GRAPHS to recommend where each onboarding
+    package should be placed.
+
+    For each top-level package in the onboarding graph, computes the full
+    transitive dependency closure and compares it against every collection.
+    Collections are ranked by fewest new packages required, then by highest
+    dependency coverage.
+
+    \b
+    ONBOARDING_GRAPH  Path to the onboarding collection graph.json.
+    COLLECTION_GRAPHS One or more paths to existing collection graph.json files.
+    """
+    try:
+        onboarding = DependencyGraph.from_file(onboarding_graph)
+    except Exception as err:
+        raise click.ClickException(
+            f"Failed to load onboarding graph {onboarding_graph}: {err}"
+        ) from err
+
+    root = onboarding.get_root_node()
+
+    toplevel_nodes: list[DependencyNode] = [
+        edge.destination_node
+        for edge in root.children
+        if edge.req_type == RequirementType.TOP_LEVEL
+    ]
+
+    if not toplevel_nodes:
+        click.echo("No top-level packages found in onboarding graph.", err=True)
+
+    logger.info(
+        "Loaded onboarding graph with %d top-level packages", len(toplevel_nodes)
+    )
+
+    collection_packages: dict[str, set[NormalizedName]] = {}
+    for graph_path in collection_graphs:
+        collection_name = extract_collection_name(graph_path)
+        if collection_name in collection_packages:
+            raise click.ClickException(
+                f"Duplicate collection name '{collection_name}' from {graph_path}. "
+                "Rename one of the graph files to avoid ambiguity."
+            )
+        try:
+            collection_graph = DependencyGraph.from_file(graph_path)
+        except Exception as err:
+            raise click.ClickException(
+                f"Failed to load collection graph {graph_path}: {err}"
+            ) from err
+        collection_packages[collection_name] = get_package_names(collection_graph)
+        logger.debug(
+            "Collection '%s': %d packages",
+            collection_name,
+            len(collection_packages[collection_name]),
+        )
+
+    results = _analyze_suggestions(toplevel_nodes, collection_packages)
+
+    if output_format == "json":
+        click.echo(json.dumps(results, indent=2))
+    else:
+        _print_suggest_collection_table(results)
--- a/src/fromager/commands/graph.py
+++ b/src/fromager/commands/graph.py
@@ -7,9 +7,12 @@ import sys
 import typing
 
 import click
+import rich
+import rich.box
 from packaging.requirements import Requirement
-from packaging.utils import canonicalize_name
+from packaging.utils import NormalizedName, canonicalize_name
 from packaging.version import Version
+from rich.table import Table
 
 from fromager import clickext, context
 from fromager.commands import bootstrap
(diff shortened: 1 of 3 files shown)


jlarkin09 (commented):
LGTM other than the one suggestion by coderabbit for json.loads()

smoparth:
> I'm not convinced that it makes sense to add this feature to Fromager upstream. Collections are a downstream concept, not a Fromager feature.

@tiran, I had the same thought when I picked this.
@dhellmann, would love to hear your take on right place this feature could land.

src/fromager/commands/graph.py
```diff
@@ -784,3 +787,288 @@ def n2s(nodes: typing.Iterable[DependencyNode]) -> str:
...
+    Analyzes dependency overlap between top-level packages in ONBOARDING_GRAPH
+    and the existing COLLECTION_GRAPHS to recommend where each onboarding
+    package should be placed.
+
+    For each top-level package in the onboarding graph, computes the full
+    transitive dependency closure and compares it against every collection.
+    Collections are ranked by fewest new packages required, then by highest
+    dependency coverage.
+
+    \b
```
dhellmann: Does this `\b` do something when printed as part of the command help?

src/fromager/commands/graph.py
```diff
@@ -784,3 +787,252 @@ def n2s(nodes: typing.Iterable[DependencyNode]) -> str:
...
+                "all_collections": [
+                    {
+                        "collection": score.collection,
+                        "new_packages": score.new_packages,
+                        "existing_packages": score.existing_packages,
+                        "coverage_percentage": round(score.coverage_percentage, 1),
+                    }
+                    for score in scores
+                ],
+            }
```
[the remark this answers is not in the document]
smoparth: The issue (#971) doesn't specify any JSON schema. The field names in our output (best_fit, all_collections, coverage_percentage) are intentional and match what the table displays. There's no documented contract to preserve here.

dhellmann:
> > > I'm not convinced that it makes sense to add this feature to Fromager upstream. Collections are a downstream concept, not a Fromager feature.
> > 
> > 
> > I want this feature here. Let's think of a better name for it.
> 
> How about `find-best-fit` as a name?

I think that's moving us in the right direction. What do you think, @smoparth ?

smoparth:
> > > > I'm not convinced that it makes sense to add this feature to Fromager upstream. Collections are a downstream concept, not a Fromager feature.
> > > 
> > > 
> > > I want this feature here. Let's think of a better name for it.
> > 
> > 
> > How about `find-best-fit` as a name?
> 
> I think that's moving us in the right direction. What do you think, @smoparth ?

@dhellmann, Thanks for confirming. I'll update the command name and also correct cli argument types to clickext.ClickPath instead of str and update the PR.

dhellmann:
> > > > > I'm not convinced that it makes sense to add this feature to Fromager upstream. Collections are a downstream concept, not a Fromager feature.
> > > > 
> > > > 
> > > > I want this feature here. Let's think of a better name for it.
> > > 
> > > 
> > > How about `find-best-fit` as a name?
> > 
> > 
> > I think that's moving us in the right direction. What do you think, @smoparth ?
> 
> @dhellmann, Thanks for confirming. I'll update the command name and also correct cli argument types to clickext.ClickPath instead of str and update the PR.

We may want to update the name in #1103 as well. Let me know what you think?

smoparth:
@coderabbitai resolve

smoparth:
> > > > > > I'm not convinced that it makes sense to add this feature to Fromager upstream. Collections are a downstream concept, not a Fromager feature.
> > > > > 
> > > > > 
> > > > > I want this feature here. Let's think of a better name for it.
> > > > 
> > > > 
> > > > How about `find-best-fit` as a name?
> > > 
> > > 
> > > I think that's moving us in the right direction. What do you think, @smoparth ?
> > 
> > 
> > @dhellmann, Thanks for confirming. I'll update the command name and also correct cli argument types to clickext.ClickPath instead of str and update the PR.
> 
> We may want to update the name in #1103 as well. Let me know what you think?

IMHO, I preferred the original names (suggest-collection and suggest-base). They were more intuitive, even though they used downstream terminology. find-best-fit does describe the action, but it lacks a clear noun and doesn’t feel immediately intuitive.

dhellmann:
> > > > > > > I'm not convinced that it makes sense to add this feature to Fromager upstream. Collections are a downstream concept, not a Fromager feature.
> > > > > > 
> > > > > > 
> > > > > > I want this feature here. Let's think of a better name for it.
> > > > > 
> > > > > 
> > > > > How about `find-best-fit` as a name?
> > > > 
> > > > 
> > > > I think that's moving us in the right direction. What do you think, @smoparth ?
> > > 
> > > 
> > > @dhellmann, Thanks for confirming. I'll update the command name and also correct cli argument types to clickext.ClickPath instead of str and update the PR.
> > 
> > 
> > We may want to update the name in #1103 as well. Let me know what you think?
> 
> IMHO, I preferred the original names (suggest-collection and suggest-base). They were more intuitive, even though they used downstream terminology. find-best-fit does describe the action, but it lacks a clear noun and doesn’t feel immediately intuitive.

The point about not having a noun is interesting, but we don't have nouns in all of the other graph-related subcommands so I'm not sure it's a requirement. I do agree that we might be able to come up with something more precise than "best fit".

I'll leave the naming up to you two. You've seen what happens when I'm allowed to name things. :-)

smoparth:
> > > > > > > > I'm not convinced that it makes sense to add this feature to Fromager upstream. Collections are a downstream concept, not a Fromager feature.
> > > > > > > 
> > > > > > > 
> > > > > > > I want this feature here. Let's think of a better name for it.
> > > > > > 
> > > > > > 
> > > > > > How about `find-best-fit` as a name?
> > > > > 
> > > > > 
> > > > > I think that's moving us in the right direction. What do you think, @smoparth ?
> > > > 
> > > > 
> > > > @dhellmann, Thanks for confirming. I'll update the command name and also correct cli argument types to clickext.ClickPath instead of str and update the PR.
> > > 
> > > 
> > > We may want to update the name in #1103 as well. Let me know what you think?
> > 
> > 
> > IMHO, I preferred the original names (suggest-collection and suggest-base). They were more intuitive, even though they used downstream terminology. find-best-fit does describe the action, but it lacks a clear noun and doesn’t feel immediately intuitive.
> 
> The point about not having a noun is interesting, but we don't have nouns in all of the other graph-related subcommands so I'm not sure it's a requirement. I do agree that we might be able to come up with something more precise than "best fit".
> 
> I'll leave the naming up to you two. You've seen what happens when I'm allowed to name things. :-)

In that case, we could go with find-best-fit as-is for #1030 and name suggest-base something like find-shared-packages or find-common-packages.

I've also addressed all the previous review feedback -- I think it's ready to merge. PTAL when you get a chance.

cc: @dhellmann, @rd4398

LalatenduMohanty:
What happens when the right solution is to create a new collection? How we are going to handle that use-case?

dhellmann:
> What happens when the right solution is to create a new collection? How we are going to handle that use-case?

This is a tool for exploring exactly that question. How it's used is up to the user.

dhellmann:
> I've also addressed all the previous review feedback -- I think it's ready to merge. PTAL when you get a chance.
> 
> cc: @dhellmann, @rd4398

I'm happy with this version, but I'll leave it open to give the other maintainers a chance to respond.

shifa-khan (dismissed):
LGTM

tests/test_suggest_collection.py
```diff
@@ -0,0 +1,454 @@
...
+            "best": {canonicalize_name("numpy"), canonicalize_name("pandas")},
+            "mid": {canonicalize_name("numpy")},
+            "worst": set(),
+        }
+        results = _analyze_suggestions(toplevel, collection_packages)
+        ranked = [c["collection"] for c in results[0]["all_collections"]]
+        new_pkg_counts = [c["new_packages"] for c in results[0]["all_collections"]]
+        assert new_pkg_counts == sorted(new_pkg_counts), (
+            f"all_collections not sorted by new_packages: {ranked}"
+        )
```
LalatenduMohanty, line 490: PR description claims duplicate collection names and bad graph files are handled but there are no tests. We should fix this before merging.

tests/test_suggest_collection.py
```diff
@@ -0,0 +1,454 @@
...
+    get_dependency_closure,
+    get_package_names,
+)
+from fromager.dependency_graph import DependencyGraph
+from fromager.requirements_file import RequirementType
+
+
+def _extract_json_from_output(output: str) -> str:
+    """Extract JSON array from output that may contain leading log lines."""
+    json_match = re.search(r"\[.*\]", output, re.DOTALL)
```
LalatenduMohanty: for --format json , why not , please parse result.stdout directly instead?

tests/test_suggest_collection.py
```diff
@@ -0,0 +1,454 @@
...
+                "find-best-fit",
+                "--format",
+                "json",
+                str(graph_dir / "onboarding.json"),
+                str(graph_dir / "data-science.json"),
+                str(graph_dir / "ml.json"),
+            ],
+        )
+        assert result.exit_code == 0, result.output
+        data = json.loads(_extract_json_from_output(result.output))
```
LalatenduMohanty: extract_json_from_output regex-parses merged result.output. The command writes JSON to stdout and warnings to stderr (graph.py 1008, 1038). 

test_commands_graph.py already uses json.loads(result.stdout) for graph JSON. These tests can pass while stdout isn’t clean JSON (e.g. with verbose logging).

tests/test_suggest_collection.py
```diff
@@ -0,0 +1,454 @@
...
+        """Missing collection-graphs argument causes a usage error."""
+        result = cli_runner.invoke(
+            fromager,
+            [
+                "graph",
+                "find-best-fit",
+                str(graph_dir / "onboarding.json"),
+            ],
+        )
+        assert result.exit_code != 0
```
LalatenduMohanty, line 361: I would suggest two more tests

- Two collection files whose stems collide → non-zero exit, message mentions Duplicate collection name.
- Invalid onboarding JSON → non-zero exit, message mentions Failed to load onboarding graph.

LalatenduMohanty (changes requested):

tests/test_suggest_collection.py
```diff
@@ -0,0 +1,454 @@
...
+                "graph",
+                "find-best-fit",
+                "--format",
+                "json",
+                str(graph_dir / "empty.json"),
+                str(graph_dir / "data-science.json"),
+            ],
+        )
+        assert result.exit_code == 0
+        assert "No top-level packages" in result.output
```
LalatenduMohanty: If you want to tighten this further (as the test checks result.output (which is merged stdout+stderr via default CliRunner) )
 use CliRunner(mix_stderr=False) in this test only, assert the message on result.stderr, and json.loads(result.stdout.strip()) == [].

tests/test_suggest_collection.py
smoparth: Added test_duplicate_collection_names_fails and test_invalid_onboarding_graph_fails to verify non-zero exit and the expected error message substring.

tests/test_suggest_collection.py
smoparth: I wasn't aware of this update. Looks like it was introduced in click 8.2.0. Removed this helper and now purely uses stdout to test the json.

tests/test_suggest_collection.py
smoparth: Agreed. All JSON tests now parse result.stdout directly, so they'll fail immediately if non-JSON content ever leaks into stdout.

tests/test_suggest_collection.py
smoparth: mix_stderr=False is no more supported. All JSON tests now parse result.stdout directly, so they'll fail immediately if non-JSON content ever leaks into stdout.

tests/test_suggest_collection.py
smoparth: Added both: duplicate collection stems → exit 1 with "Duplicate collection name", and invalid JSON → exit 1 with "Failed to load onboarding graph".

src/fromager/dependency_graph.py
```diff
@@ -158,6 +158,24 @@ def iter_build_requirements(self) -> typing.Iterable[DependencyNode]:
...
+        while stack:
+            current = stack.pop()
+            for edge in current.children:
+                child_node = edge.destination_node
+                if child_node.key in visited:
+                    continue
+                visited.add(child_node.key)
+                yield child_node
+                stack.append(child_node)
+
```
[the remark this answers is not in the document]
smoparth: Where did you see this guideline

LalatenduMohanty (approved):

PR #1047, opened 2026-04-13
docs: add docstrings and comments to complex functions

### LINKED ISSUE #1046: Add docstrings and comments to complex functions missing documentation
## Summary

Several functions across `src/fromager/` have complex logic but lack docstrings or inline comments, making it harder for new contributors to understand the codebase.

**Docstrings needed (12 functions)** — complex logic with multiple code paths or non-obvious algorithms:
- `find_sdist`, `find_wheel`, `find_source_dir` (finders.py) — multiple naming fallback strategies
- `get_project_from_pypi` (resolver.py) — multi-stage filtering pipeline
- `_resolve_version_from_git_url` (bootstrapper.py) — git URL parsing with ref extraction and clone caching
- `add_extra_metadata_to_wheels`, `extract_info_from_wheel_file` (wheels.py) — multi-step pipeline, normalization edge case
- `download_url`, `prepare_source` (sources.py) — retry/temp file handling, polymorphic plugin returns
- `get_build_backend_hook_caller` (dependencies.py) — 3-source env var merge
- `_update_build_requires` (pyproject.py) — normalized-name dedup algorithm
- `run` (external_commands.py) — expand existing minimal docstring

LalatenduMohanty:
Add docstrings to functions and inline comments to  non-obvious code patterns across various files. Only targets genuinely complex logic that would be hard for new contributors to understand.

Though my goal was to help contributors but it has also a positive impact on AI code assistants.  When an AI assistant needs to understand a function, it either reads the docstring or reads the entire function body. A good docstring lets it understand the function's purpose and behavior without consuming context window on implementation details.

Closes: #1046

### DIFF
--- a/src/fromager/dependency_graph.py
+++ b/src/fromager/dependency_graph.py
@@ -169,6 +169,8 @@ class DependencyNode:
         start_edges: list[DependencyEdge],
         visited: set[str],
     ) -> typing.Iterable[DependencyEdge]:
+        # Depth-first walk of install requirement edges, skipping nodes
+        # already visited to avoid loops in the dependency graph.
         for edge in start_edges:
             if edge.key in visited:
                 continue
@@ -220,6 +222,9 @@ class DependencyGraph:
         graph_dict: dict[str, dict[str, typing.Any]],
     ) -> DependencyGraph:
         graph = cls()
+        # Stack-based DFS to reconstruct graph from serialized dict,
+        # skipping nodes already visited to avoid processing shared
+        # dependencies twice.
         stack = [ROOT]
         visited = set()
         while stack:
--- a/src/fromager/resolver.py
+++ b/src/fromager/resolver.py
@@ -222,7 +222,13 @@ def get_project_from_pypi(
     *,
     override_download_url: str | None = None,
 ) -> Candidates:
-    """Return candidates created from the project name and extras."""
+    """Fetch and filter package candidates from a PyPI-compatible server.
+
+    Filters the project's package list by: project status, filename
+    validity, package type (sdist/wheel), yanked status, Python version
+    compatibility, and platform tags. Can substitute
+    ``override_download_url`` into each candidate's URL.
+    """
     found_candidates: set[str] = set()
     ignored_candidates: set[str] = set()
     logger.debug("%s: getting available versions from %s", project, sdist_server_url)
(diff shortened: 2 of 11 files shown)


src/fromager/dependency_graph.py
```diff
@@ -169,6 +169,8 @@ def _traverse_install_requirements(
         start_edges: list[DependencyEdge],
         visited: set[str],
     ) -> typing.Iterable[DependencyEdge]:
+        # Depth-first walk of install requirement edges, skipping nodes
+        # already visited to avoid loops in the dependency graph.
```
jskladan, line 173: Not that it does any harm, but the comment describes pretty self-explanatory code, while not adding any important context/information in the same fashion the other comments do. There is no (non-obvious) "why" explained.

src/fromager/dependency_graph.py
```diff
@@ -220,6 +222,9 @@ def from_dict(
         graph_dict: dict[str, dict[str, typing.Any]],
     ) -> DependencyGraph:
         graph = cls()
+        # Stack-based DFS to reconstruct graph from serialized dict,
+        # skipping nodes already visited to avoid processing shared
+        # dependencies twice.
```
jskladan, line 227: This feels like it could have been a docstring instead, if you want to keep it in? Othervise, also feels that it does not add much value other than describing "what" code does (and we have the code to describe what it does already :)).

src/fromager/resolver.py
```diff
@@ -806,6 +812,10 @@ def cache_key(self) -> str:
         raise NotImplementedError("GenericProvider does not implement caching")
 
     def find_candidates(self, identifier: typing.Any) -> Candidates:
+        # Accepts three input formats from _version_source:
```
jskladan: How about we make this a docstring instead?

jskladan (dismissed):
Obviously, all of these are nitpicks, but since we're talking docs, they feel appropriate.

I see nothing wrong with the MR as-is, so hitting that Approve button (FWIW), and leaving it up to you.

rd4398 (dismissed):
This looks good to me. I will wait for Lala to address comments on PR

src/fromager/dependency_graph.py
LalatenduMohanty: This is targeted for new contributors who might not understand that we are using DFS here

src/fromager/dependency_graph.py
LalatenduMohanty: We do not  add doc strings for private functions.  This is because docstrings populate __doc__ and are accessible via help(), which is intended for public API consumers, not relevant for private methods.

LalatenduMohanty: So for private/non-public functions, PEP 257 recommends a comment after the def line which what I did.

src/fromager/dependency_graph.py
LalatenduMohanty: This is also inline to PEP 257 recommends. I will suggest we keep it.

src/fromager/resolver.py
LalatenduMohanty: You are right. I will change it to a doc string.

LalatenduMohanty: Fixed it.

rd4398 (approved):
LGTM

PR #1051, opened 2026-04-13
fix(bootstrap): don't call get_resolver_provider hook in cache lookups

### LINKED ISSUE #1049: Bootstrap _download_wheel_from_cache should not call get_resolver_provider hook
The bootstrap helper `_download_wheel_from_cache` calls `resolver.resolve()`, which runs custom `get_resolver_provider` hook. In my opinion, it does not make sense to call the resolver hook. It complicates logic in user-provided hooks, too. Instead all cache lookups should directly use a PyPI resolver to locate wheels in a cache index.

https://github.com/python-wheel-build/fromager/blob/b8f4441643ea8d5cd30e8f9b85e7dc898555bf79/src/fromager/bootstrapper.py#L1094-L1109

LalatenduMohanty:
_download_wheel_from_cache() called resolver.resolve(), which triggered custom resolver hooks. Cache servers are always simple PyPI index servers and don't need hook-based resolution. Call default_resolver_provider() directly instead.

Closes: #1049

### DIFF
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -1100,13 +1100,17 @@ class Bootstrapper:
             f"checking if wheel was already uploaded to {self.cache_wheel_server_url}"
         )
         try:
-            wheel_url, _ = resolver.resolve(
-                ctx=self.ctx,
-                req=Requirement(f"{req.name}=={resolved_version}"),
+            # Use PyPIProvider directly for cache lookups, bypassing resolver
+            # hooks. Cache servers are always simple PyPI index servers.
+            pinned_req = Requirement(f"{req.name}=={resolved_version}")
+            provider = resolver.PyPIProvider(
                 sdist_server_url=self.cache_wheel_server_url,
                 include_sdists=False,
                 include_wheels=True,
+                constraints=self.ctx.constraints,
             )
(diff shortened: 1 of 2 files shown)


src/fromager/bootstrapper.py
```diff
@@ -1100,13 +1100,18 @@ def _download_wheel_from_cache(
             f"checking if wheel was already uploaded to {self.cache_wheel_server_url}"
         )
         try:
-            wheel_url, _ = resolver.resolve(
+            # Use PyPIProvider directly for cache lookups, bypassing resolver
+            # hooks. Cache servers are always simple PyPI index servers.
+            pinned_req = Requirement(f"{req.name}=={resolved_version}")
+            provider = resolver.default_resolver_provider(
```
tiran: how about you directly create a `PyPIProvider` here?

LalatenduMohanty: You are right. Hypothetically if default_resolver_provider ever gains additional logic (logging, filtering, etc.), those changes would not be for cache lookups against a known simple index server.

tiran:
Looks good to me.

I would appreciate a second opinion from @rd4398 or @dhellmann .

rd4398 (approved):
This looks good!

Issue #937, opened 2026-02-27
New resolver configuration with tag provider and git clone


tiran:
Design and implement a new resolver, download, and sdist settings that allows declarative configuration of common tasks in package settings:

- resolve with `GitLabTagProvider` or `GitHubTagProvider` with custom tag matcher function
- git clone a repo
- build an sdist with tarfile or PEP 517 sdist hook

PR #1052, opened 2026-04-13
feat: source resolver configuration

tiran:
# Pull Request Description

## What

Introduce new configuration for source resolver and downloads. The new system uses profiles for common tasks:

- `pypi-sdist`: resolve versions of sdists from PyPI, download sdist
- `pypi-prebuilt`: resolve versions of platform wheels from PyPI, download pre-built wheel from PyPI.
- `pypi-download`: resolve versions of any package from PyPI, download from external URL (with `{version}` variable)
- `pypi-git`: resolve versions for any package from PyPI, git clone (with `{version}` variable)
- `versionmap-git`: resolve with a version map, git clone
- `github-tag-download`: resolve from GitHub tag, download tarball
- `github-tag-git`: resolve from GitHub tag, git clone
- `gitlab-tag-download`: resolve from GitLab tag, download tarball
- `gitlab-tag-git`: resolve from GitLab tag, git clone
- `not-available`: block resolving
- `hook`: call Fromager hooks

The new settings will eventually replace `download_source`, `resolver_dist`, and `git_options` top-level options as well as `wheel_server_url` and `pre_built` flags for variants.

## Why

- #937
- #938 

### DIFF
--- a/src/fromager/packagesettings/_resolver.py
+++ b/src/fromager/packagesettings/_resolver.py
@@ -1,0 +1,513 @@
+from __future__ import annotations
+
+import enum
+import logging
+import re
+import typing
+
+import pydantic
+
+from .. import resolver
+from ._typedefs import MODEL_CONFIG, PackageVersion
+
+if typing.TYPE_CHECKING:
+    from .. import context, requirements_file
+
+logger = logging.getLogger(__name__)
+
+VERSION_QUOTED = "%7Bversion%7D"
+
+
+class BuildSDist(enum.StrEnum):
+    pep517 = "pep517"
+    tarball = "tarball"
+
+
+class AbstractResolver(pydantic.BaseModel):
+    model_config = MODEL_CONFIG
+
+    provider: str
+
+    def resolver_provider(
+        self, ctx: context.WorkContext, req_type: requirements_file.RequirementType
+    ) -> resolver.BaseProvider:
+        raise NotImplementedError
+
+
+class PyPISDistResolver(AbstractResolver):
+    """Resolve version with PyPI, download sdist from PyPI
+
+    The ``pypi-sdist`` provider uses :pep:`503` *Simple Repository API* or
+    :pep:`691` *JSON-based Simple API* to resolve packages on PyPI or a
+    PyPI-compatible index.
+
+    The provider downloads source distributions (tarballs) from the index.
+    It ignores releases that have only wheels and no sdist.
+
+    Example::
+
+        provider: pypi-sdist
+        index_url: https://pypi.test/simple
+    """
+
+    provider: typing.Literal["pypi-sdist"]
+
+    index_url: pydantic.HttpUrl = pydantic.Field(
+        default=pydantic.HttpUrl("https://pypi.org/simple/"),
+        description="Python Package Index URL",
+    )
+
+    # It is not safe to use PEP 517 to re-generate a source distribution.
+    # Some PEP 517 backends require VCS to generate correct sdist.
+    build_sdist: typing.ClassVar[BuildSDist | None] = BuildSDist.tarball
+
+    def resolver_provider(
+        self, ctx: context.WorkContext, req_type: requirements_file.RequirementType
+    ) -> resolver.PyPIProvider:
+        return resolver.PyPIProvider(
+            include_sdists=True,
+            include_wheels=False,
+            sdist_server_url=str(self.index_url),
+            constraints=ctx.constraints,
+            req_type=req_type,
+            ignore_platform=False,
+        )
+
+
+class PyPIPrebuiltResolver(AbstractResolver):
+    """Resolve version with PyPI, download pre-built wheel from PyPI
+
+    The ``pypi-prebuilt`` provider uses :pep:`503` *Simple Repository API* or
+    :pep:`691` *JSON-based Simple API* to resolve packages on PyPI or a
+    PyPI-compatible index.
+
+    The provider downloads pre-built wheels from the index. It ignores
+    versions that have no compatible wheels (sdist-only or incompatible
+    OS, CPU arch, or glibc version).
+
+    Example::
+
+        provider: pypi-prebuilt
+        index_url: https://pypi.test/simple
+    """
+
+    provider: typing.Literal["pypi-prebuilt"]
+
+    index_url: pydantic.HttpUrl = pydantic.Field(
+        default=pydantic.HttpUrl("https://pypi.org/simple/"),
+        description="Python Package Index URL",
+    )
+
+    build_sdist: typing.ClassVar[BuildSDist | None] = None
+
+    def resolver_provider(
+        self, ctx: context.WorkContext, req_type: requirements_file.RequirementType
+    ) -> resolver.PyPIProvider:
+        return resolver.PyPIProvider(
+            include_sdists=False,
+            include_wheels=True,
+            sdist_server_url=str(self.index_url),
+            constraints=ctx.constraints,
+            req_type=req_type,
+            ignore_platform=False,
+        )
+
+
+class PyPIDownloadResolver(AbstractResolver):
+    """Resolve version with PyPI, download sdist from arbitrary URL
+
+    The ``pypi-download`` provider uses :pep:`503` *Simple Repository API* or
+    :pep:`691` *JSON-based Simple API* to resolve packages on PyPI or a
+    PyPI-compatible index.
+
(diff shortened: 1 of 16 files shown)


tiran:
The tests are going to fail. I have not updated the test cases and test configs.

src/fromager/packagesettings/_resolver.py
```diff
@@ -0,0 +1,538 @@
...
+        self, ctx: context.WorkContext, req_type: requirements_file.RequirementType
+    ) -> resolver.GitLabTagProvider:
+        return self._gitlab_provider(
+            ctx=ctx,
+            req_type=req_type,
+            override_download_url=f"git+{self.project_url}@{{tagname}}",
+        )
+
+
+class NotAvailableResolver(pydantic.BaseModel):
```
rd4398: It implements `resolver_provider()` with the same signature but aren't part of the type hierarchy
Can we make it  inherit from `AbstractResolver` for consistency?

src/fromager/packagesettings/_resolver.py
```diff
@@ -0,0 +1,538 @@
...
+
+    provider: typing.Literal["not-available"]
+
+    def resolver_provider(
+        self, ctx: context.WorkContext, req_type: requirements_file.RequirementType
+    ) -> resolver.BaseProvider:
+        raise ValueError("package is not available")
+
+
+class HookResolver(pydantic.BaseModel):
```
rd4398: Same here,
it implements `resolver_provider()` with the same signature but aren't part of the type hierarchy
Can we make it  inherit from `AbstractResolver` for consistency?

src/fromager/packagesettings/_resolver.py
```diff
@@ -0,0 +1,538 @@
...
+        )
+
+    def _gitlab_provider(
+        self,
+        *,
+        ctx: context.WorkContext,
+        req_type: requirements_file.RequirementType,
+        override_download_url: str | None = None,
+    ) -> resolver.GitLabTagProvider:
+        assert self.project_url.path  # for type checker
```
rd4398: Can we make this an explicit `if not ... raise ValueError(...)` like the GitHub equivalent does?
Something like :
```
if not self.project_url.path or self.project_url.path.count("/") != 2:
            raise ValueError(
                f"Invalid path in {self.project_url}, expected two elements"
            )
```

src/fromager/packagesettings/_resolver.py
```diff
@@ -0,0 +1,538 @@
...
+            raise ValueError(f"invalid scheme in url {value}")
+        if not value.path:
+            raise ValueError(f"url {value} has an empty path")
+        return value
+
+    def resolver_provider(
+        self, ctx: context.WorkContext, req_type: requirements_file.RequirementType
+    ) -> resolver.BaseProvider:
+        return resolver.VersionMapProvider(
+            version_map=versionmap.VersionMap(self.version_map),  # type: ignore
```
rd4398: This suppresses what's likely a real type mismatch (dict[PackageVersion, str] vs expected VersionMap constructor arg). 
Can we add a proper cast/conversion, or at minimum use a scoped # type: ignore[arg-type] with a comment?

src/fromager/packagesettings/_resolver.py
```diff
@@ -0,0 +1,538 @@
...
+    """Call resolver_provider and download_source hook"""
+
+    model_config = MODEL_CONFIG
+
+    provider: typing.Literal["hook"]
+
+    def resolver_provider(
+        self, ctx: context.WorkContext, req_type: requirements_file.RequirementType
+    ) -> resolver.BaseProvider:
+        # TODO
```
rd4398, line 504: Users who configure provider: hook will get an unhelpful crash.  How about we add an error message something like "hook resolver is not yet implemented"?

src/fromager/packagesettings/_resolver.py
```diff
@@ -0,0 +1,538 @@
...
+    """Matcher import string (``package.module:name``)
+
+    Matcher can be a :class:`re.Pattern` object or a factory function
+    that accepts *ctx* arg and returns a :class:`~fromager.resolver.MatchFunction`.
+    """
+
+    build_sdist: BuildSDist = BuildSDist.pep517
+    """Source distribution build method"""
+
+    min_release_age: int | None = pydantic.Field(default=None, ge=0)
```
rd4398: Non blocker: The `min_release_age` field with identical definition and docstring appears on both AbstractPyPIResolver and AbstractGitSourceResolver. Since both inherit from AbstractResolver, this field could live there instead. Resolvers  that don't need it (like VersionMapResolver) would inherit it harmlessly as None.

rd4398 (commented):
This looks mostly okay. I have left a few comments / suggestions.
I will prefer if @LalatenduMohanty and / or @dhellmann takes a look as well

src/fromager/packagesettings/_resolver.py
tiran: Actually it is not a type mismatch. `PackageVersion` is a subclass of `Version` with extra stuff for Pydantic. I'm not sure why it's triggering a type error.

I have added ` type: ignore[arg-type]` and a comment.

src/fromager/packagesettings/_resolver.py
tiran: I didn't do that because I didn't want the option on VersionResolver, HookResolver, and NotAvailableResolver. I want an error when user adds `min_release_age` to a resolver provider that does not support it.

I have changed the implementation to use a mixin class instead.

src/fromager/packagesettings/_resolver.py
tiran: I have changed the code to raise a value error when path is empty.

src/fromager/packagesettings/_resolver.py
tiran: Good point, fixed!

src/fromager/packagesettings/_resolver.py
rd4398: Thank you!

src/fromager/packagesettings/_resolver.py
rd4398: Thank you!

src/fromager/packagesettings/_resolver.py
rd4398: Hmm, okay makes sense

src/fromager/packagesettings/_resolver.py
```diff
@@ -0,0 +1,530 @@
...
+    @pydantic.field_validator("clone_url", mode="after")
+    @classmethod
+    def validate_clone_url(cls, value: pydantic.AnyUrl) -> pydantic.AnyUrl:
+        if value.scheme not in {"https", "ssh"}:
+            raise ValueError(f"invalid scheme in url {value}")
+        if not value.path:
+            raise ValueError(f"url {value} has an empty path")
+        return value
+
+    def resolver_provider(
```
LalatenduMohanty: The version_map values (e.g., abad1dea, refs/tags/1.1) are passed directly as Candidate.url via VersionMapProvider.find_candidates() (resolver.py:1327). 
These are bare git refs, not fetchable URLs. The clone_url field is defined on the model but never referenced in resolver_provider().                                                        
                                                                                                            
Compare with PyPIGitResolver (line 237) which correctly combines clone_url + ref: download_url = f"git+{self.clone_url}@refs/tags/{self.tag}"                                               
                                                                                                                                                                             
Not a bug today since the resolver isn't wired into the pipeline yet, but it will break when it is. 

Should the map values be transformed into clone URLs here?

LalatenduMohanty (changes requested):

src/fromager/packagesettings/_resolver.py
```diff
@@ -0,0 +1,530 @@
...
+        *,
+        ctx: context.WorkContext,
+        req_type: requirements_file.RequirementType,
+        override_download_url: str | None = None,
+    ) -> resolver.GitLabTagProvider:
+        if not self.project_url.path:
+            raise ValueError(f"Empty path in {self.project_url}")
+        return resolver.GitLabTagProvider(
+            project_path=self.project_url.path.lstrip("/"),
+            server_url=f"https://{self.project_url.host}",
```
LalatenduMohanty: We are not passing any port here. is that fine? Looks like Pydantic v2 HttpUrl.port always returns an int, it fills in the default (443 for https, 80 for http)

LalatenduMohanty (approved):

PR #1057, opened 2026-04-13
fix(bootstrap): auto-disable constraints and extend e2e test coverage

### LINKED ISSUE #1044: --multiple-versions option of bootstrap should disable constraints automatically.
In e2e test for multiple version bootstrap flag we have

```
'tomli>=2.0,<=2.0.2' || true
```
The reason we need the `|| true` here is that it tries to generate a constraints file, but that will never work when ` --multiple-versions` is enabled.
We should make --multiple-versions disable constraints automatically.

Refer comment from @dhellmann here: https://github.com/python-wheel-build/fromager/pull/1035#discussion_r3066381152

### LINKED ISSUE #1045: extend e2e test for multiple version bootstrap by being generous with constraints
We should extend the test to cover dependencies by being a little more generous with the constraints. Allow a few versions of flit-core, setuptools, etc. and then verify that they all build.

Side Note: Refer to this comment from @dhellmann - https://github.com/python-wheel-build/fromager/pull/1035#discussion_r3066456875

rd4398:
Two related fixes for --multiple-versions flag:

1. **Auto-disable constraints with --multiple-versions**
   - When --multiple-versions flag is enabled, automatically set skip_constraints=True because constraints.txt cannot handle multiple versions of the same package
   - Log informative message: "automatically disabling constraints generation (incompatible with --multiple-versions)"

2. **Extend e2e test to verify dependency chain handling**
   - Use generous constraint range (flit-core>=3.9,<3.12) instead of pinning to single version
   - Verify that multiple versions of flit-core are built (not just tomli), confirming --multiple-versions works for entire chain
   - Check for at least 2 flit-core versions (test found 3)
   - Remove `|| true` since constraints are now auto-disabled

The e2e test now validates that --multiple-versions bootstraps multiple versions of both top-level packages AND their dependencies, providing better test coverage.

Fixes: https://github.com/python-wheel-build/fromager/issues/1044
Fixes: https://github.com/python-wheel-build/fromager/issues/1045

### DIFF
--- a/src/fromager/commands/bootstrap.py
+++ b/src/fromager/commands/bootstrap.py
@@ -159,6 +159,14 @@ def bootstrap(
         logger.info(
             "multiple versions mode enabled: will bootstrap all matching versions"
         )
+        # Automatically disable constraints when multiple versions mode is enabled
+        # because constraints.txt cannot handle multiple versions of the same package
+        if not skip_constraints:
+            logger.info(
+                "automatically disabling constraints generation "
+                "(incompatible with --multiple-versions)"
+            )
+            skip_constraints = True
 
     pre_built = wkctx.settings.list_pre_built()
     if pre_built:
--- a/tests/test_bootstrap.py
+++ b/tests/test_bootstrap.py
@@ -1,4 +1,5 @@
 import io
+import logging
 import pathlib
 import textwrap
 from unittest.mock import Mock, patch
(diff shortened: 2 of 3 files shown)


jskladan:
FWIW, LGTM @rd4398

src/fromager/commands/bootstrap.py
```diff
@@ -159,6 +159,14 @@ def bootstrap(
...
             "multiple versions mode enabled: will bootstrap all matching versions"
         )
+        # Automatically disable constraints when multiple versions mode is enabled
+        # because constraints.txt cannot handle multiple versions of the same package
+        if not skip_constraints:
+            logger.info(
+                "automatically disabling constraints generation "
+                "(incompatible with --multiple-versions)"
+            )
+            skip_constraints = True
```
LalatenduMohanty, line 169: We should add an unit test to cover for all three paths: (a) --multiple-versions alone auto-disables constraints, (b) --multiple-versions --skip-constraints together doesn't log the auto-disable message, (c) without --multiple-versions, constraints are not disabled.

LalatenduMohanty (changes requested):

src/fromager/commands/bootstrap.py
LalatenduMohanty: Unit test would be faster and it can be run offline as well.

rd4398: Good idea! I have added tests in latest commit

tests/test_bootstrap.py
```diff
@@ -541,6 +541,147 @@ def test_skip_constraints_cli_option() -> None:
...
+def test_multiple_versions_auto_disables_constraints(
+    mock_metrics: Mock,
+    mock_progress: Mock,
+    mock_server: Mock,
+    mock_bootstrapper: Mock,
+    tmp_context: context.WorkContext,
+    caplog: pytest.LogCaptureFixture,
+) -> None:
+    """Test that --multiple-versions alone auto-disables constraints and logs message"""
+    import logging
```
LalatenduMohanty: Please move import logging to the top-level imports of test_bootstrap.py

LalatenduMohanty (changes requested):

tests/test_bootstrap.py
rd4398: Ah, missed that. Done in latest commit

LalatenduMohanty (approved):

Issue #1059, opened 2026-04-14
feat: Unique wheel file names with a new `wheel_build_tag` hook


tiran:
## Problem

Fromager-built wheels are platform-specific and may depend on:

- **OS / distribution version**, e.g. Fedora 43, RHEL 9.6, RHEL 10.1
- **AI accelerator stack**, e.g. CUDA 13.1 vs CUDA 12.9, ROCm 7.1
- **Torch ABI**, which is unstable across versions; a wheel compiled for Torch 2.10.0 may have a different ABI than one compiled for Torch 2.11.0

Currently, wheel filenames carry none of this information. A rebuild for a new Torch version does not produce a new filename, making it difficult to:

1. Invalidate caches when the underlying platform or dependency stack changes.
2. Distinguish wheels built for different accelerator stacks or OS versions.
3. Replace outdated wheels with correctly-targeted rebuilds.
4. Share wheels between indexes. Downstream maintains a separate index for each accelerator version, but only a couple of dozen packages out of over 1,200 are CUDA/ROCm/Torch-specific. Wheels like `pillow` or `fromager` are identical across accelerator indexes and could be shared if their filenames clearly indicate they have no accelerator dependency. Sharing is out of scope for this proposal but is a possibility for future improvements in downstream.

## Proposal

Add a new stevedore hook point, `wheel_build_tag`, to the existing `fromager.hooks` namespace. This hook lets downstream plugin packages inject custom suffixes into the [wheel build tag](https://packaging.python.org/en/latest/specifications/binary-distribution-format/).

### Wheel spec background

The wheel filename format is:

```
{distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-{platform tag}.whl
```

The build tag is optional, must start with a digit, and must not contain `-`. It is parsed as `tuple[int, str]`. Fromager already fills the `int` part from variant + package changelog and sets the `str` suffix to `""`. The build tag is also stored in the `{name}-{version}.dist-info/WHEEL` metadata file and shown in `pip list` output.

This proposal extends that suffix with hook-provided segments (e.g. `_el9.6_rocm7.1_torch2.10.0`).

### Hook signature

```python
def run_wheel_build_tag_hooks(
    *,
    ctx: context.WorkContext,
    req: Requirement,
    version: Version,
    wheel_tags: frozenset[Tag],
) -> typing.Sequence[str]:
    ...
```

[...]

PR #1066, opened 2026-04-16
[proposal] wheel_build_tag hook for unique wheel file names

tiran:
# Pull Request Description

## What

Add proposal document for a new hook point that injects platform, accelerator, and ABI suffixes into wheel build tags.

## Why

See: #1059

### DIFF
--- a/docs/proposals/wheel-build-tag-hook.md
+++ b/docs/proposals/wheel-build-tag-hook.md
@@ -1,0 +1,221 @@
+# Unique wheel file names with a `wheel_build_tag` hook
+
+- Author: Christian Heimes
+- Created: 2026-04-16
+- Status: Open
+- GitHub issue: [#1059](https://github.com/python-wheel-build/fromager/issues/1059)
+
+## What
+
+This enhancement proposes a new stevedore hook point, `wheel_build_tag`, in
+the existing `fromager.hooks` namespace. The hook lets downstream plugin
+packages inject custom suffixes into the
+[wheel build tag](https://packaging.python.org/en/latest/specifications/binary-distribution-format/),
+producing unique wheel file names that encode platform, accelerator stack,
+and dependency ABI information.
+
+## Why
+
+Fromager-built wheels are platform-specific and may depend on:
+
+- **OS / distribution version**, e.g. Fedora 43, RHEL 9.6, RHEL 10.1
+- **AI accelerator stack**, e.g. CUDA 13.1 vs CUDA 12.9, ROCm 7.1
+- **Torch ABI**, which is unstable across versions; a wheel compiled for
+  Torch 2.10.0 may have a different ABI than one compiled for Torch 2.11.0
+
+Currently, wheel filenames carry none of this information. A rebuild for a
+new Torch version does not produce a new filename, making it difficult to:
+
+1. Invalidate caches when the underlying platform or dependency stack changes.
+2. Distinguish wheels built for different accelerator stacks or OS versions.
+3. Replace outdated wheels with correctly-targeted rebuilds.
+4. Share wheels between indexes. Downstream maintains a separate index for
+   each accelerator version, but only a couple of dozen packages out of over
+   1,200 are CUDA/ROCm/Torch-specific. Wheels like `pillow` or `fromager`
+   are identical across accelerator indexes and could be shared if their
+   filenames clearly indicate they have no accelerator dependency. Sharing is
+   out of scope for this proposal but is a possibility for future
+   improvements in downstream.
+
+## Goals
+
+- introduce a `wheel_build_tag` hook point in the `fromager.hooks` stevedore
+  namespace
+- allow hooks to contribute ordered suffix segments to the wheel build tag
+- produce unique, deterministic wheel file names that reflect the build
+  environment
+
+## Non-goals
+
+- Filtering or selecting wheels by build tag at install time. `pip install`
+  and `uv pip install` only use the build tag for sorting, not for filtering.
+- Sharing wheels across indexes. While unique file names enable this in
+  principle, the mechanics of cross-index sharing are out of scope.
+- Accessing wheel content, the build environment, or ELF dependency info from
+  within the hook. The hook must work identically whether a wheel is freshly
+  built or retrieved from cache.
+- Validation of annotations such as "depends on libtorch". A validation
+  system for `build_wheel` may be added in the future.
+- Per-package plugin hooks (`overrides.find_and_invoke()`). A per-package
+  override can be added later when the need arises.
+
+## How
+
+### Wheel spec background
+
+The wheel filename format is:
+
+```
+{distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-{platform tag}.whl
+```
+
+The build tag is optional, must start with a digit, and must not contain `-`.
+It is parsed as `tuple[int, str]`. Fromager already fills the `int` part from
+variant + package changelog and sets the `str` suffix to `""`. The build tag
+is also stored in the `{name}-{version}.dist-info/WHEEL` metadata file and
+shown in `pip list` output.
+
+This proposal extends that suffix with hook-provided segments (e.g.
+`_el9.6_rocm7.1_torch2.10.0`).
+
+### Hook signature
+
+```python
+def build_tag_hook(
+    *,
+    ctx: context.WorkContext,
+    req: Requirement,
+    version: Version,
+    wheel_tags: frozenset[Tag],
+) -> typing.Sequence[tuple[int, str]]:
+    ...
+```
+
+Each registered hook returns `typing.Sequence[tuple[int, str]]`, a sequence
+of (sort-order, suffix) pairs. This allows a single hook to contribute
+multiple suffix segments (e.g. both an OS tag and an accelerator tag). The
+runner collects all pairs from all hooks, sorts them in ascending order, and
+returns the suffix parts as a sequence. Neither `sort-order` nor `suffix`
+have to be unique. The first appearance of a `suffix` is used. Suffixes are
+are joined with `_` and appended to the existing build tag.
+
+`suffix` must only contain alphanumeric ASCII characters or dot
+(`[a-zA-Z0-9.]`). When a hook returns any other character or raises an
+exception, the build fails.
+
+Build tag hooks are registered for `fromager.hooks` entry point and name
+`wheel_build_tag`.
+
+### Example hook
+
+```python
+def example_hook(
+    *,
+    ctx: context.WorkContext,
+    req: Requirement,
+    version: Version,
+    wheel_tags: frozenset[Tag],
+) -> typing.Sequence[tuple[int, str]]:
+    result: list[tuple[int, str]] = []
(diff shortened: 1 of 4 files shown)


docs/proposals/wheel-build-tag-hook.md
```diff
@@ -0,0 +1,187 @@
...
+### Hook signature
+
+```python
+def run_wheel_build_tag_hooks(
+    *,
+    ctx: context.WorkContext,
+    req: Requirement,
+    version: Version,
+    wheel_tags: frozenset[Tag],
+) -> typing.Sequence[str]:
```
rd4398, line 89: Says it returns `Sequence[str]` but example on ;ine 93-94 returns Sequence[tuple[int, str]]
                                                                                                                                       
  `def example_hook(...) -> typing.Sequence[tuple[int, str]]:` 

                                                                                                                         
We should ensure the runner's signature matches the hook's return value, or clarify that the runner collects tuples and returns strings

docs/proposals/wheel-build-tag-hook.md
```diff
@@ -0,0 +1,187 @@
...
+    return result
+```
+
+### What hooks can access
+
+- **`ctx` + `req`**: for package configuration, annotations from `pbi`
+- **`wheel_tags`**: detect whether a wheel is purelib or platlib
+  (platform/arch-specific).
+
+Hooks can also user other information like **`os.environ`** or the
```
rd4398: Typo: "user other information" → "use other information"

docs/proposals/wheel-build-tag-hook.md
```diff
@@ -0,0 +1,187 @@
...
+    *,
+    ctx: context.WorkContext,
+    req: Requirement,
+    version: Version,
+    wheel_tags: frozenset[Tag],
+) -> typing.Sequence[str]:
+    ...
+```
+
+Each registered hook returns `typing.Sequence[tuple[int, str]]`, a sequence
```
rd4398: Clarification (Dumb) questions: What happens if:                                                                                                                                               
    - A hook raises an exception?                                                                                                   
    - Two hooks return the same sort order integer?                                                                                                                    
    - A hook returns invalid characters (e.g., - which is forbidden)?

docs/proposals/wheel-build-tag-hook.md
```diff
@@ -0,0 +1,187 @@
...
+    ctx: context.WorkContext,
+    req: Requirement,
+    version: Version,
+    wheel_tags: frozenset[Tag],
+) -> typing.Sequence[tuple[int, str]]:
+    result: list[tuple[int, str]] = []
+    platlib = any(tag.platform != "any" for tag in wheel_tags)
+    if platlib:
+        # fc43, el9.6, ...
+        result.append((1, get_distro_tag()))
```
rd4398: Can we add how `get_distro_tag( )` will look like? Perhaps a pseudocode?

docs/proposals/wheel-build-tag-hook.md
```diff
@@ -0,0 +1,187 @@
...
+    pbi = ctx.package_build_info(req)
+    if pbi.annotations.get("accelerator-specific") == "true":
```
rd4398: Can we document annotation requirements as well?

rd4398 (changes requested):
This is a solid design and will really help us downstream as well!

I left a few comments, after addressing those we can merge this.

docs/proposals/wheel-build-tag-hook.md
tiran: I have changed the proposal to describe the hook signature. The `run_wheel_build_tag_hooks` signature is an internal implementation detail and not relevant for the user.

docs/proposals/wheel-build-tag-hook.md
tiran: Good catch! I have clarified the situation:

- an exception or invalid character aborts the build
- same sort order or even same suffix is not an error. The first occurrence of a suffix wins.

docs/proposals/wheel-build-tag-hook.md
tiran: Sure!

docs/proposals/wheel-build-tag-hook.md
tiran: There is no requirement to use annotations. This is just an example how to use annotations. I'm updating the example to make it more obvious.

rd4398 (commented):
This looks good to me! I will wait for @LalatenduMohanty and @dhellmann to take a look as well and give feedback.

docs/proposals/wheel-build-tag-hook.md
```diff
@@ -0,0 +1,221 @@
...
+    ...
+```
+
+Each registered hook returns `typing.Sequence[tuple[int, str]]`, a sequence
+of (sort-order, suffix) pairs. This allows a single hook to contribute
+multiple suffix segments (e.g. both an OS tag and an accelerator tag). The
+runner collects all pairs from all hooks, sorts them in ascending order, and
+returns the suffix parts as a sequence. Neither `sort-order` nor `suffix`
+have to be unique. The first appearance of a `suffix` is used. Suffixes are
+are joined with `_` and appended to the existing build tag.
```
dhellmann: What is the use case for having multiple hooks?

dhellmann (commented):
This looks good to me. I have a one clarifying question, but nothing to really hold it up.

docs/proposals/wheel-build-tag-hook.md
tiran: - The other hooks also support multiple hook implementations.
- Downstream can define multiple hooks, e.g. one for platform tag, one for Torch, one for each accelerator stack, and so on.

dhellmann: The way the other hooks are invoked only allows for one entry point, based on the package name. Stevedore should report an error if multiples are registered. For example, it wouldn't make sense to have 2 plugins give resolver providers for the same package.

This case feels different, because the hooks might combine data usefully. I'm just wondering if that's a "maybe useful" thing or a "definitely useful" thing, because the sorting seems complex to get right.

If we have 1 hook and it returns a list, then it can understand the ordering itself without leaking implementation details between hooks. But if we allow mixing hooks from multiple sources, how would we ensure that their sort ordering makes sense? I could see us releasing a bunch of simple hooks upstream, for example, to do platform tags and change log length and maybe other values. But how would we ensure those sort properly without having them all know about each other to specify their sort order values? And how would we add the one for torch into that mix without upstream and downstream knowing about each other in some meaningful way?

If instead of supporting multiple hooks, we just provide helper functions, then the single downstream hook implementation could specify the ordering.

If we want multiple hooks, then maybe we need a config setting to manage the order, instead of having the hooks try to express it directly? Stevedore has a way to invoke plugins based on a list of names, for example, so we could directly convert a list of names in a config file into invocation of the hooks in the expected order.

tiran: It's "maybe useful". We can get by with just a single hook. It would simplify the implementation, too.

I designed the feature with multiple hooks in mind to give us more flexibility and because stevedore may return multiple hooks. We do not need that flexibility in downstream. There seems to be no option to enforce a single `wheel_build_tag` entry point for `fromager.hooks`. Multiple packages can provide a `wheel_build_tag` hook.

Maybe entry points and stevedore extensions are the wrong approach for hooks that are singletons? This feature feels less like a notification hook and more like a setting. We could take another approach like a global config setting of type [ImportString](https://pydantic.dev/docs/validation/2.1/usage/types/string_types/#importstring).

```yaml
# overrides/settings.yaml
[wheels]
build_tag_hook = "mysettings.hooks:custom_tag_hook"
```

dhellmann: It looks like for overrides [we're using a normal ExtensionManager](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/overrides.py#L24) and [relying on only having one plugin](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/overrides.py#L124) with a given name. That's a bug, we should at least register a [conflict_resolver](https://docs.openstack.org/stevedore/latest/reference/index.html#extensionmanager) function. Maybe just using the builtin [error_on_conflict](https://opendev.org/openstack/stevedore/src/branch/master/stevedore/extension.py#L130) is what we want?

For hooks, it may make sense to have multiple values registered. If we use multiple hooks, I recommend building some sort of determinism into the execution order of the hooks themselves, though, rather than relying on the data returned by the hooks to be mixable directly. For example, in this case if you sorted by (name, order-key, value) then even if 2 hooks return the same order-key the results would always come out in the same order.

If we want a singleton hook, using the NamedExtensionManager or DriverExtensionManager may fit that pattern better.

tiran: Does it even make sense to use an entry point and stevedore hook for this feature? The wheel build tag hook feels like a runtime configuration option.

dhellmann: That's fair. It breaks the pattern we use elsewhere, but maybe since it's just one it does make sense.

tiran:
I have re-designed the hook. It's now a single hook that is configured in global `settings.yaml`.

dhellmann (commented):
This direction makes sense to me. I'll leave the PR open for other reviewers to have a chance to read it before we merge it.

tiran:
@dhellmann @rd4398 @LalatenduMohanty Could you please review the design again? Thanks!

rd4398:
This looks good to me!

docs/proposals/wheel-build-tag-hook.md
```diff
@@ -0,0 +1,223 @@
...
+  and `uv pip install` only use the build tag for sorting, not for filtering.
+- Sharing wheels across indexes. While unique file names enable this in
+  principle, the mechanics of cross-index sharing are out of scope.
+- Accessing wheel content, the build environment, or ELF dependency info from
+  within the hook. The hook must work identically whether a wheel is freshly
+  built or retrieved from cache.
+- Validation of annotations such as "depends on libtorch". A validation
+  system for `build_wheel` may be added in the future.
+- Per-package plugin hooks (`overrides.find_and_invoke()`). A per-package
+  override can be added later when the need arises.
```
LalatenduMohanty: Can we add why we might need a per-package plugin hooks?

LalatenduMohanty: I guess it is about different naming conventions for the wheel which we might need to override the default and we donot need fromager changes for that. 

The hook keeps fromager generic and the naming policy pluggable.

LalatenduMohanty (changes requested):
I feel this proposal is little verbose. We can make it concise without loosing important context.

docs/proposals/wheel-build-tag-hook.md
tiran: I have rephrases the bullet point. I don't see a good use case for a per-package hook. For that reason, I have excluded a package override hook from the proposal.

tiran:
> I feel this proposal is little verbose. We can make it concise without loosing important context.

I have shortened the proposal and moved the example to a separate file. The file is included when Sphinx renders the markdown file.

LalatenduMohanty (approved):

PR #1073, opened 2026-04-20
feat(build): add --build-isolation for sandboxed builds

### LINKED ISSUE #1019: Add build isolation to sandbox build_sdist and build_wheel steps
## Problem                                                                                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                           
  During `build_sdist` and `build_wheel`, Fromager executes arbitrary code from upstream packages (their `setup.py`, build backends, custom build scripts). This code currently runs with the same privileges as Fromager itself — full filesystem access, full environment variables, and 
  (when running in a builder container) access to credentials like `.netrc`.                                                                                                                                                                                                               
                                                                                                                                                                                                                                                                                           
  A compromised or malicious build backend can:                                                                                                                                                                                                                                            
  - Read `.netrc` credentials (GitLab tokens, PyPI upload credentials)                                                                                                                                                                                                                     
  - Modify system directories (`/usr`) or Fromager's own venv                                                                                                                                                                                                                              
  - Access sensitive environment variables (`NETRC`, `TWINE_PASSWORD`, `NGC_API_KEY`)                                                                                                                                                                                                      
  - Interfere with parallel builds via shared `/tmp`, IPC, or process signaling                                                                                                                                                                                                            
  - Exfiltrate stolen data over the network (when network isolation is disabled)                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                           
  ## Proposal                                                                                                                                                                                                                                                                              
                                                                                                                                                                                                                                                                                           
  Extend the existing `unshare`-based network isolation (#477) to create mount, PID, and IPC namespaces for build steps. A new `--build-isolation` flag sandboxes PEP 517 hook calls with:                                                                                                 
                                                                                                                                                                                                                                                                                         
  - **Mount namespace:** Makes `/usr` read-only (bind + remount), hides credentials by overmounting `$HOME` with tmpfs, gives each build its own `/tmp`                                                                                                                                    
  - **PID namespace:** Build process cannot see or signal other processes                                                                                                                                                                                                                
  - **IPC namespace:** Isolated shared memory, semaphores, message queues                                                                                                                                                                                                                  
  - **Network namespace:** Inherited from existing `--network-isolation`                                                                                                                                                                                                                   
  - **Environment scrubbing:** Strips `NETRC`, `TWINE_PASSWORD`, `NGC_API_KEY` and similar sensitive variables before executing the build backend                                                                                                                                          
                                                                                                                                                                                                                                                                                           
  The sandbox applies **only** to `build_sdist` and `build_wheel` (via the PEP 517 hook runner in `dependencies.py`). Source download, dependency installation, and upload steps are unaffected since they need network access and/or credentials.                                         
                                                                                                                                                                                                                                                                                           
  ## Alternatives and Their Caveats                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                         
  **Why `unshare` and not bubblewrap:** bwrap was evaluated in #473 and rejected because it requires `--privileged` or `CAP_SYS_ADMIN` inside containers. This is still the case ([containers/bubblewrap#505](https://github.com/containers/bubblewrap/issues/505) remains open). The `unshare` approach works in unprivileged Podman/Docker containers out of the box.                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                           
  **Why not ephemeral user accounts (`useradd` per package):** Tested inside builder containers. Basic file permission blocking works (`.netrc` returns "Permission denied"), but `/tmp`, IPC, and process list are still shared across builds. Parallel `useradd` contends on `/etc/passwd` locks. Once you mitigate all the gaps, you end up reimplementing `unshare`.
                                                                                                                                                                                                                                                                                           
  **Why not systemd `DynamicUser=yes`:** Would be ideal — it provides exactly the isolation we need. But it requires systemd as PID 1 inside the container, which build containers don't have. Tested: `systemd-run` fails with "System has not been booted with systemd as init system."  Design influenced by systemd's `DynamicUser=` ([reference](https://0pointer.net/blog/dynamic-users-with-systemd.html)) — same kernel primitives, applied without the systemd dependency.

pavank63:
## Summary                                                
            
  - Add `--build-isolation` flag that sandboxes build backend subprocesses using ephemeral Unix users and Linux namespaces (network, PID, IPC, UTS)                      
  - Ephemeral user cannot read credential files like `.netrc` (root:root 600), network namespace blocks exfiltration, PID/IPC namespaces prevent cross-process           
  interference                                                                                                                                                           
  - Works in unprivileged containers (Podman/Docker) without `--privileged` or `--cap-add SYS_ADMIN`                                                                     
  - Add `FROMAGER_SCRUB_ENV_VARS` environment variable for downstream build systems to strip sensitive env vars from build subprocesses
                                                                                                                                                                         
  ## Test plan                                                                                                                                                           
                                                                                                                                                                         
  - [ ] All 432 unit tests pass                                                                                                                                          
  - [ ] Verified `.netrc` is unreadable by build subprocess (Permission denied)                                                                                          
  - [ ] Verified network is unreachable from build subprocess                  
  - [ ] Verified PID isolation (only build process visible)                                                                                                              
  - [ ] Verified `useradd`/`userdel` overhead is ~10ms each                                                                                                              
  - [ ] Verified `uv pip install` for build deps correctly opts out of isolation                                                                                         
                                                                                                                                                                         
  Closes: #1019

Related proposal https://github.com/python-wheel-build/fromager/pull/1077                                                                                                                                                          
                                                                                                                                                                         
### DIFF
--- a/src/fromager/__main__.py
+++ b/src/fromager/__main__.py
@@ -30,6 +30,15 @@ else:
     SUPPORTS_NETWORK_ISOLATION = True
     NETWORK_ISOLATION_ERROR = None
 
+try:
+    external_commands.detect_build_isolation()
+except Exception as e:
+    SUPPORTS_BUILD_ISOLATION: bool = False
+    BUILD_ISOLATION_ERROR: str | None = str(e)
+else:
+    SUPPORTS_BUILD_ISOLATION = True
+    BUILD_ISOLATION_ERROR = None
+
 
 @click.group()
 @click.version_option(
@@ -143,6 +152,14 @@ else:
     help="Build sdist and when with network isolation (unshare -cn)",
     show_default=True,
 )
+@click.option(
+    "--build-isolation/--no-build-isolation",
+    default=False,
+    help="Sandbox build steps with mount, PID, IPC, and network namespace isolation. "
+    "Hides credentials, makes system directories read-only, and isolates /tmp. "
+    "Supersedes --network-isolation for build steps.",
+    show_default=True,
+)
 @click.pass_context
 def main(
     ctx: click.Context,
@@ -163,6 +180,7 @@ def main(
     variant: str,
     jobs: int | None,
     network_isolation: bool,
+    build_isolation: bool,
 ) -> None:
     # Save the debug flag so invoke_main() can use it.
     global _DEBUG
@@ -220,6 +238,7 @@ def main(
             logger.info(f"maximum concurrent jobs: {jobs}")
             logger.info(f"constraints file: {constraints_file}")
             logger.info(f"network isolation: {network_isolation}")
+            logger.info(f"build isolation: {build_isolation}")
             if build_wheel_server_url:
                 logger.info(f"external build wheel server: {build_wheel_server_url}")
             else:
@@ -229,6 +248,9 @@ def main(
 
     if network_isolation and not SUPPORTS_NETWORK_ISOLATION:
         ctx.fail(f"network isolation is not available: {NETWORK_ISOLATION_ERROR}")
+
+    if build_isolation and not SUPPORTS_BUILD_ISOLATION:
+        ctx.fail(f"build isolation is not available: {BUILD_ISOLATION_ERROR}")
 
     wkctx = context.WorkContext(
         active_settings=packagesettings.Settings.from_files(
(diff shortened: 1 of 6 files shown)


tiran (changes requested):
I would prefer to start with a design proposal first and come to a mutual agreement how we want to address build isolation.

pavank63:
I have created this design proposal https://github.com/python-wheel-build/fromager/pull/1077. PTAL

PR #1077, opened 2026-04-21
docs(proposal): add build isolation design for sandboxed builds

pavank63:
Add design proposal for --build-isolation flag that sandboxes PEP 517 build backend subprocesses using ephemeral Unix users and Linux namespaces. Includes security findings from proof-of-concept testing with build-attack-test package.

See: #1019

### DIFF
--- a/docs/proposals/build-isolation.md
+++ b/docs/proposals/build-isolation.md
@@ -1,0 +1,159 @@
+# Build isolation for sandboxing build backends
+
+- Author: Pavan Kalyan Reddy Cherupally
+- Created: 2026-04-21
+- Status: Open
+- Issue: [#1019](https://github.com/python-wheel-build/fromager/issues/1019)
+
+## What
+
+A `--build-isolation` flag that sandboxes PEP 517 build backend
+subprocesses (`build_sdist`, `build_wheel`) so they cannot read
+credentials, access the network, or interfere with the host system.
+
+## Why
+
+Fromager executes upstream-controlled code (setup.py, build backends)
+during wheel builds. A compromised or malicious package can:
+
+- Read credential files like `$HOME/.netrc` and exfiltrate tokens
+- Reach the network to upload stolen data or download payloads
+- Signal or inspect other processes via `/proc` or shared IPC
+- Interfere with parallel builds through shared `/tmp`
+- Leave persistent backdoors: `.pth` files that run on every Python
+  startup, shell profile entries that run on every login, or
+  background daemons that survive the build
+
+The existing `--network-isolation` flag blocks network access but does
+not protect against credential theft, process/IPC visibility, or
+persistent backdoors.
+
+Build isolation wraps each build backend invocation in a sandbox that
+combines file-level credential protection with OS-level namespace
+isolation. Only the PEP 517 hook calls are sandboxed; download,
+installation, and upload steps run normally.
+
+## Goals
+
+- A `--build-isolation/--no-build-isolation` CLI flag (default off)
+  that supersedes `--network-isolation` for build steps
+- Credential protection: build processes cannot read `.netrc` or
+  other root-owned credential files
+- Network isolation: no routing in the build namespace
+- Process and IPC isolation: build cannot see other processes or
+  access shared memory and semaphores
+- Persistence protection: build cannot drop `.pth` backdoors, modify
+  shell profiles, or leave background daemons running after the build
+- Works in unprivileged containers (Podman/Docker) without
+  `--privileged` or `--cap-add SYS_ADMIN`
+
+## Non-goals
+
+- **Mount namespace isolation.** Breaks `pyproject_hooks` IPC, which
+  exchanges `input.json`/`output.json` through `/tmp`.
+- **macOS / Windows support.** Linux-only; flag is unavailable on
+  other platforms.
+
+## How
+
+Build isolation combines an ephemeral Unix user with Linux namespace
+isolation. Before each build, a short-lived system user (`fmr_<random>`)
+is created with `useradd` and removed on exit via `trap EXIT`. The user
+has no home directory and no login shell, so it cannot read root-owned
+credential files like `.netrc` (mode 600). After dropping to the
+ephemeral user with `setpriv`, the script enters new namespaces with
+`unshare`:
+
+| Namespace | Flag | Purpose |
+| -- | -- | -- |
+| Network | `--net` | No routing; blocks all network access |
+| PID | `--pid --fork` | Build sees only its own processes |
+| IPC | `--ipc` | Isolated shared memory and semaphores |
+| UTS | `--uts` | Separate hostname |
+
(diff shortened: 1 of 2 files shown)


docs/proposals/build-isolation.md
```diff
@@ -0,0 +1,300 @@
...
+## How
+
+### Isolation mechanism
+
+Build isolation combines two complementary techniques:
+
+#### 1. Ephemeral Unix user
+
+Before each build invocation, the isolation script creates a
+short-lived system user with `useradd` and removes it with `userdel`
```
rd4398: `useradd / userdel`, can modify `/etc/passwd` and `/etc/shadow` from what I know. This means fromager (or the isolation script) must run as root inside the container. That's a major assumption that isn't mentioned  anywhere.  What happens if fromager runs as a non-root user?

docs/proposals/build-isolation.md
```diff
@@ -0,0 +1,300 @@
...
+on exit (via `trap EXIT`). The user has:
```
rd4398: The `userdel` runs in a trap EXIT handler, but `SIGKILL` cannot be trapped. If a build gets `OOM-killed` or force-killed, the ephemeral user leaks. Over a long bootstrap run with hundreds of packages, this could accumulate orphaned  `fmr_* users in /etc/passwd`.

Can we add that as a limitation?

docs/proposals/build-isolation.md
```diff
@@ -0,0 +1,300 @@
...
+
+Build isolation wraps each build backend invocation in a sandbox that
+combines file-level credential protection with OS-level namespace
+isolation. Only the PEP 517 hook calls are sandboxed; download,
+installation, and upload steps run normally.
+
+## Goals
+
+- A `--build-isolation/--no-build-isolation` CLI flag (default off)
+  that supersedes `--network-isolation` for build steps
```
rd4398: Clarification question: What happens with these combinations?

1. - --build-isolation --network-isolation — redundant? Does build isolation absorb network isolation for build steps while network isolation still applies to non-build steps?                                                      
2. - --build-isolation --no-network-isolation — does the user get network isolation for builds anyway (since build isolation includes it)?
3. - --no-build-isolation --network-isolation — today's behavior?

Looking at the current code, network_isolation is passed to `_run_hook_with_extra_environ` for build hooks but also to ` _createenv` for venv creation. Does build isolation apply to venv creation too, or 
only PEP 517 hooks?

docs/proposals/build-isolation.md
```diff
@@ -0,0 +1,300 @@
...
+- Fails early with a clear message if the platform does not support
+  build isolation
+
+#### WorkContext (`context.py`)
+
+- New `build_isolation: bool` field (default `False`)
+
+#### BuildEnvironment (`build_environment.py`)
+
+- `run()` method accepts `build_isolation` parameter, defaults to
```
rd4398: This doesn't look right. Looking at the actual code in dependencies.py:547-553, `_run_hook_with_extra_environ` calls `external_commands.run()` directly — it doesn't go   
  through `BuildEnvironment.run()`. This matters because `BuildEnvironment.run()` is where env var setup (like CARGO_NET_OFFLINE) happens. The proposal needs to either change the hook runner to go through `BuildEnvironment.run()`, or duplicate that logic.

rd4398 (commented):
The proposal is strong on threat modeling but needs tightening on the integration details. I have added few comments / questions

docs/proposals/build-isolation.md
pavank63: Addressed — the updated proposal now documents this in the How section, including the fallback behavior when running as non-root.

docs/proposals/build-isolation.md
pavank63: Added as a documented limitation under How > Limitations.

docs/proposals/build-isolation.md
pavank63: Both flags apply to the same build steps (PEP 517 hooks, venv creation, wheel packing). `--build-isolation` supersedes `--network-isolation` — using both is redundant but harmless. `--no-build-isolation --network-isolation` is today's behavior. Clarified in the updated Goals section.

docs/proposals/build-isolation.md
pavank63: The build isolation wrapper will be prepended at the `external_commands.run()` level, same as `run_network_isolation.sh` today, so `_run_hook_with_extra_environ` is covered. Env vars like `CARGO_NET_OFFLINE` will be set inside the wrapper script. Implementation details removed from the proposal — will be addressed in the implementation PR.

tiran:
Hi @pavank63,

Thank you for this proposal. Your threat analysis is thorough -- the attack surface catalog (credential theft, process visibility, IPC, persistence via `.pth` files and shell profiles) and the proof-of-concept with `build-attack-test` are valuable contributions. The findings table makes a clear case for why `--network-isolation` alone is insufficient.

That said, I have some architectural concerns about the approach that I think we should discuss before moving forward.

**We should not build our own sandboxing tool.** Sandboxing is notoriously hard to get right, and we are not sandboxing experts. There are mature, well-tested tools out there (bubblewrap, firejail, systemd-run with its sandboxing options, etc.) that have been hardened against edge cases we would likely miss. The `useradd`/`unshare` wrapper approach works for a proof-of-concept, but maintaining a custom sandboxing solution long-term is a burden I would rather avoid. We should let users plug in the tool that fits their environment.

**The design assumes a specific deployment model.** The proposal is tailored to how we run fromager downstream: as root inside a disposable container. But fromager is a general-purpose tool. Many users run it as a regular user on their workstation, in CI without root, or in environments where `useradd` is not available. The proposal acknowledges this ("when running as non-root, namespace isolation still applies but credential and persistence protection are not available"), but I think we need a design that works well across deployment models rather than one that degrades in the common case.

**Sandboxing needs to be stateful.** More complex packages run several commands during a build (multiple PEP 517 hooks, custom build steps). The sandbox needs a lifecycle: set up the environment once, run multiple commands inside it, then tear down and clean up. A per-invocation wrapper script does not capture this well.

I have been working on a draft proposal for a pluggable hooks-based system that addresses these points -- it would let users configure `setup_sandbox`, `teardown_sandbox`, `run_sandboxed`, and `run_unconfined` hooks through the settings file. The default implementation would be our current network isolation, but users could plug in bubblewrap, firejail, or whatever fits their setup. I will push that proposal soon so we can compare approaches.

Your threat modeling and the `build-attack-test` package would be great assets regardless of which implementation approach we choose -- they give us a solid test matrix for validating any sandboxing solution.

We also need to address some prerequisites on the fromager side -- for example, sdist and wheel builds should each use a dedicated directory per package/version, which would make it much easier for any sandboxing solution to restrict filesystem access. Once we have agreed on a design for the plugin API, I would like you to explore concrete sandbox backends using tools like Landlock, firejail, or bubblewrap.

*This reply was co-written with Claude.*

Issue #1083, opened 2026-04-23
Add environment variable filtering for external commands


tiran:
Fromager passes the full parent environment to every subprocess. This leaks sensitive variables (credentials, CI tokens, cloud keys) into build scripts and PEP 517 hooks. Add configuration to control which variables from `os.environ` are inherited by subprocesses. Variables injected by fromager itself (`extra_environ`, `build_environment`, package-level `env`) are not subject to filtering.

LalatenduMohanty:
https://github.com/python-wheel-build/fromager/pull/1084 is the proposal for the feature.

smoparth:
@tiran @rd4398 @LalatenduMohanty Now that the data model PR is up, we need clarification with below questions for wiring into external_commands.run():

1. Scope: Should ALL subprocesses be filtered, or only build-related ones?
gitutils (clone, checkout) and sources (patch) don't run untrusted code. Should they also get filtered envs? 

2. Wiring approach preference for `external_commands.run()`?
-  parameter approach: Adding an `env_filter` parameter to run() (**explicit, but callers without ctx object won't be able filter**)
- global state: Module-level state configured at startup (**all calls get filter**, zero signature changes). **But that behavior is not conveyed by run() as explicit parameter**

3. Should filtering be opt-in per call site, or global?
If parameter: git/patch calls won't filter unless someone threads settings (ctx object) through. 
If module-level: everything filters from the moment settings load. Which is the intended behavior.

We can't pass ctx directly to external_commands.run() that creates a circular import since context.py already calls external_commands.run() in uv_clean_cache(). Would need refactoring to break the cycle.

rd4398:
Yes, I think all sub processes should filter. I am inclined to the parameter approach with `ExternalCommands` (not ctx), defaulting to `None`  so like `env_filter: ExternalCommands | None = None` to `run()` as it is the smallest possible change without circular dependency and helps us achieve our goal.  It will also be backward compatible.

@dhellmann any thoughts here?

LalatenduMohanty:
IMO we should  not pass ExternalCommands directly to run(). Instead, we should apply the filter at the env-construction step inside run()

LalatenduMohanty:
Also adding env_filter: ExternalCommands | None = None parameter to external_commands.run() sounds right to me

PR #1084, opened 2026-04-23
docs: add environment variable filtering proposal

tiran:
# Pull Request Description

## What

Add proposal for configurable filtering of sensitive environment variables from subprocesses, modeled after sudo's env_keep/env_delete.

## Why

See: #1083

### DIFF
--- a/docs/proposals/filter-env.md
+++ b/docs/proposals/filter-env.md
@@ -1,0 +1,121 @@
+# Filter environment variables
+
+- Author: Christian Heimes
+- Created: 2026-04-23
+- Status: Open
+- Issue: [#1083](https://github.com/python-wheel-build/fromager/issues/1083)
+
+## What
+
+A configurable filter for environment variables inherited by external
+commands. Fromager strips sensitive variables from `os.environ` before
+they reach build scripts, hooks, or other subprocesses.
+
+## Why
+
+Fromager runs arbitrary third-party code during builds -- PEP 517 hooks,
+setup scripts, patching commands, git operations. Every subprocess
+today inherits the full parent environment, including credentials,
+CI tokens, and registry passwords that have no business inside a build.
+
+This is a supply-chain risk. A compromised or careless build script can
+read and exfiltrate any variable it inherits. The risk is higher in CI
+environments where runners routinely carry cloud credentials, API tokens,
+and service account keys.
+
+Controlling which variables pass through also improves reproducibility.
+Stray environment can silently alter build behavior (compiler flags,
+proxy settings, locale overrides). Filtering makes builds more
+deterministic.
+
+`sudo` solves the same problem with `env_keep` and `env_delete` in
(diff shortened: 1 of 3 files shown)


docs/proposals/filter-env.md
```diff
@@ -0,0 +1,120 @@
...
+
+Entries without `*` are exact matches. Entries ending with `*` are
+prefix matches: `AWS_*` matches `AWS_REGION`, `AWS_SECRET_ACCESS_KEY`,
+etc. A `*` in any other position is a validation error. A bare `*` is
+only valid in `delete_env` (catch-all); in `keep_env` it is a
+validation error.
+
+### Evaluation order
+
+`keep_env` is evaluated before `delete_env`. A variable matching both
```
smoparth: The Evaluation order says "A variable matching both is kept," but the numbered steps don't enforce that. Step 3 (delete) would sweep variable of step 2 (keep) when variable is matching  in both.

Suggest adding "skip remaining steps" to steps 1 and 2 to make it an explicit first-match-wins chain:

If in always-keep set -- keep, stop.
If any keep_env matches -- keep, stop.
If any delete_env matches -- delete.
Otherwise -- keep.
This makes the keep-over-delete priority unambiguous. or Maybe I'm seeing the keep or delete notation from different PoV that its used for.

docs/proposals/filter-env.md
```diff
@@ -0,0 +1,120 @@
...
+is kept.
+
+For each variable in `os.environ`:
+
+1. If it is in the always-keep set -- **keep**, regardless of
```
smoparth: Where would this list live? In code or data file like settings.yaml? 
Can we add it to examples to demonstrate how the always-keep is used.

smoparth (commented):
The proposal looks good overall. Just left a couple of comments requesting additional clarity on the evaluation order and the always-keep set.

docs/proposals/filter-env.md
tiran: This document is a high-level proposal that focuses on UX and scope. The internal representation of the always-keep set is an implementation detail. The important point is that some env vars like PATH and HOME are always present.

docs/proposals/filter-env.md
tiran: I have rephrased the section and explained that evaluation is case-insensitive and short-circuit.

docs/proposals/filter-env.md
smoparth: Agreed. Since there's already an examples section, thought it'd be a natural place to write but understood.

smoparth (approved):

rd4398 (approved):
LGTM! We should implement this asap

rd4398:
@mergify rebase

rd4398:
@tiran can you please address the feedback so that we can land this?

tiran:
@mergifyio rebase

PR #1086, opened 2026-04-23
fix(sources): skip broken symlinks in scan_compiled_extensions (#1082)

### LINKED ISSUE #1082: `scan_compiled_extensions` crashes on broken symlinks since 0.80.0
### Description

`scan_compiled_extensions` in `src/fromager/sources.py` raises an unhandled `FileNotFoundError` when it encounters a broken symlink in a source tree, causing the entire bootstrap to fail.

### Root cause

Commit `400d3bd8` (`refactor(sources): use EAFP pattern instead of is_file() check`) replaced:

```python
if not filepath.is_file():
    continue
with filepath.open("rb") as f:
    ...
```

with

```python
try:
    with filepath.open("rb") as f:
        ...
except IsADirectoryError:
    continue
```

The original `is_file()` check returned `False` for broken symlinks (symlinks whose target doesn't exist), safely skipping them. The new code only catches `IsADirectoryError`, so broken symlinks — which raise `FileNotFoundError` on `open()` — propagate up and crash the bootstrap.

### Reproducer

PyTorch release tarballs (all versions tested: 2.7.0, 2.8.0, 2.9.0, 2.9.1, 2.10.0) contain `.dockerignore` as a symlink to `.gitignore`, but `.gitignore` is not included in the tarball. Bootstrapping any package that depends on torch fails with:

```
ERROR <package>: [Errno 2] No such file or directory: '/path/to/work-dir/torch-2.10.0/torch-2.10.0/.dockerignore'
```

### Versions Affected

- 0.80.0
- 0.81.0

voidborne-d:
## Summary

Fixes #1082 — `scan_compiled_extensions` crashed on broken symlinks since 0.80.0.

Commit 400d3bd8 moved to an EAFP pattern catching only `IsADirectoryError` (for symlinks-to-directories returned as filenames by `Path.walk()`). Broken symlinks — where the target does not exist — raise `FileNotFoundError` on `open()` and were not caught, propagating out and crashing the entire bootstrap.

The fix adds `FileNotFoundError` to the except tuple and a comment noting why.

## Changes

- `src/fromager/sources.py`: catch `(IsADirectoryError, FileNotFoundError)` in the binary-header read path.
- `tests/test_sources.py`: two regression tests
  - `test_scan_compiled_extensions_skips_broken_symlinks` — standalone dangling symlink does not crash.
  - `test_scan_compiled_extensions_skips_dangling_symlink_alongside_real_file` — dangling symlink does not prevent detection of a real `.so` elsewhere in the tree.

## Test plan

- [x] `pytest tests/test_sources.py -k scan_compiled_extensions` → 10 passed (8 existing + 2 new)
- [x] `ruff check` / `ruff format --check` on both modified files → clean
- [x] Verified both new tests fail without the `FileNotFoundError` addition (original bug reproduces)

### DIFF
--- a/src/fromager/sources.py
+++ b/src/fromager/sources.py
@@ -883,10 +883,11 @@ def scan_compiled_extensions(
             elif suffix not in ignore_suffixes:
                 # Path.walk() lists symlinks to directories as filenames
                 # rather than dirnames, causing IsADirectoryError on open().
+                # Broken symlinks (target does not exist) raise FileNotFoundError.
                 try:
                     with filepath.open("rb") as f:
                         header = f.read(_MAGIC_HEADERS_READ)
-                except IsADirectoryError:
+                except (IsADirectoryError, FileNotFoundError):
                     continue
                 if header.startswith(magic_headers):
                     relpath = filepath.relative_to(root_dir)
--- a/tests/test_sources.py
+++ b/tests/test_sources.py
@@ -285,3 +285,24 @@ def test_scan_compiled_extensions(
         assert matches == [pathlib.Path(filename)]
     else:
         assert matches == []
+
+
+def test_scan_compiled_extensions_skips_broken_symlinks(
+    tmp_path: pathlib.Path,
+) -> None:
+    # Symlink whose target does not exist: open() raises FileNotFoundError.
+    # Must be skipped, not propagated.
+    broken = tmp_path / "broken_link"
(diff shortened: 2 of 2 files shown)


LalatenduMohanty:
Thank you for the pull request,  however we do not accept PR/changes from unknown AI bots as it might have security implications in future.

PR #1087, opened 2026-04-23
fix(sources): catch FileNotFoundError for broken symlinks

### LINKED ISSUE #1082: `scan_compiled_extensions` crashes on broken symlinks since 0.80.0

andre-motta:
Closes: #1082

# Pull Request Description

## What

Add FileNotFoundError to the except clause in scan_compiled_extensions and add a regression test.

## Why

Broken symlinks (e.g. .dockerignore → missing .gitignore in PyTorch tarballs) crash the bootstrap with FileNotFoundError since 0.80.0.

### DIFF
--- a/src/fromager/sources.py
+++ b/src/fromager/sources.py
@@ -883,10 +883,11 @@ def scan_compiled_extensions(
             elif suffix not in ignore_suffixes:
                 # Path.walk() lists symlinks to directories as filenames
                 # rather than dirnames, causing IsADirectoryError on open().
+                # Broken symlinks raise FileNotFoundError on open().
                 try:
                     with filepath.open("rb") as f:
                         header = f.read(_MAGIC_HEADERS_READ)
-                except IsADirectoryError:
+                except (IsADirectoryError, FileNotFoundError):
                     continue
                 if header.startswith(magic_headers):
                     relpath = filepath.relative_to(root_dir)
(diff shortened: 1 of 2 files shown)


LalatenduMohanty (approved):

LalatenduMohanty:
@mergify rebase

tiran (approved):

PR #1089, opened 2026-04-23
feat(bootstrap): add --max-release-age for multiple-versions mode

rd4398:
Add --max-release-age flag to bootstrap command to filter out package versions older than N days. This flag is required when --multiple-versions is enabled, preventing builds of hundreds of old releases. Together with --min-release-age (cooldown), it creates a sliding time window for version selection in the multiple-versions bootstrap feature.

Closes https://github.com/python-wheel-build/fromager/issues/1037

### DIFF
--- a/src/fromager/commands/bootstrap.py
+++ b/src/fromager/commands/bootstrap.py
@@ -110,6 +110,13 @@ def _get_requirements_from_args(
     default=False,
     help="Bootstrap all matching versions instead of only the highest version",
 )
+@click.option(
+    "--max-release-age",
+    "max_release_age",
+    type=click.IntRange(min=1),
+    default=None,
+    help="Reject package versions published more than this many days ago.",
+)
 @click.argument("toplevel", nargs=-1)
 @click.pass_obj
 def bootstrap(
@@ -121,6 +128,7 @@ def bootstrap(
     skip_constraints: bool,
     test_mode: bool,
     multiple_versions: bool,
+    max_release_age: int | None,
     toplevel: list[str],
 ) -> None:
     """Compute and build the dependencies of a set of requirements recursively
@@ -167,6 +175,13 @@ def bootstrap(
                 "(incompatible with --multiple-versions)"
             )
             skip_constraints = True
+
+    if max_release_age is not None:
+        wkctx.set_max_release_age(max_release_age)
+        logger.info(
+            "max release age: rejecting versions older than %d days",
+            max_release_age,
+        )
 
     pre_built = wkctx.settings.list_pre_built()
     if pre_built:
@@ -492,6 +507,13 @@ bootstrap._fromager_show_build_settings = True  # type: ignore
     default=False,
     help="Bootstrap all matching versions instead of only the highest version",
 )
+@click.option(
+    "--max-release-age",
+    "max_release_age",
+    type=click.IntRange(min=1),
+    default=None,
+    help="Reject package versions published more than this many days ago.",
+)
 @click.argument("toplevel", nargs=-1)
 @click.pass_obj
 @click.pass_context
@@ -506,6 +528,7 @@ def bootstrap_parallel(
     force: bool,
     max_workers: int | None,
     multiple_versions: bool,
+    max_release_age: int | None,
     toplevel: list[str],
 ) -> None:
     """Bootstrap and build-parallel
@@ -533,6 +556,7 @@ def bootstrap_parallel(
         sdist_only=True,
         skip_constraints=skip_constraints,
         multiple_versions=multiple_versions,
+        max_release_age=max_release_age,
         toplevel=toplevel,
     )
 
--- a/src/fromager/context.py
+++ b/src/fromager/context.py
@@ -63,6 +63,7 @@ class WorkContext:
         settings_dir: pathlib.Path | None = None,
         wheel_server_url: str = "",
         cooldown: Cooldown | None = None,
+        max_release_age: datetime.timedelta | None = None,
     ):
         if active_settings is None:
             active_settings = packagesettings.Settings(
@@ -113,6 +114,17 @@ class WorkContext:
         self._parallel_builds = False
 
         self.cooldown: Cooldown | None = cooldown
+        self._max_release_age: datetime.timedelta | None = max_release_age
+
+    @property
+    def max_release_age(self) -> datetime.timedelta | None:
+        return self._max_release_age
+
+    def set_max_release_age(self, days: int) -> None:
+        """Set the maximum release age in days."""
+        if days < 1:
+            raise ValueError(f"max_release_age must be positive, got {days}")
+        self._max_release_age = datetime.timedelta(days=days)
 
     def enable_parallel_builds(self) -> None:
         self._parallel_builds = True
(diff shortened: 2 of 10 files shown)


rd4398:
This adds --max-release-age to the bootstrap (and bootstrap-parallel) commands as part of the multiple-versions bootstrap feature. It complements the existing --min-release-age (cooldown from PR #1018) to create a sliding time window for version selection.

What it does:
  - --max-release-age N rejects package versions published more than N days ago
  - Required when --multiple-versions is enabled; ignored otherwise
  - Filtering happens in find_all_matching_from_provider() on the resolved candidates list — candidates without upload_time (e.g. git URLs) pass through
  - If all candidates are older than the cutoff, all are kept with a warning to avoid empty resolution
  - Uses the cooldown's bootstrap_time when available for consistent cutoffs across a single run

EmilienM (dismissed):

src/fromager/commands/bootstrap.py
```diff
@@ -156,6 +164,10 @@ def bootstrap(
         )
 
     if multiple_versions:
+        if max_release_age <= 0:
+            raise click.UsageError(
+                "--max-release-age is required when using --multiple-versions"
```
dhellmann: Why is the age value required?

src/fromager/commands/bootstrap.py
```diff
@@ -168,6 +180,13 @@ def bootstrap(
             )
             skip_constraints = True
 
+    if max_release_age > 0:
+        wkctx.max_release_age = timedelta(days=max_release_age)
```
dhellmann: It would be better here to have a method on wkctx like `set_max_release_age()` and let it change the representation of the value internally. Otherwise nothing prevents a caller from setting `max_release_age()` to a string or whatever they want.

src/fromager/resolver.py
```diff
@@ -242,10 +270,37 @@ def find_all_matching_from_provider(
...
+    # Materialize candidates so we can iterate more than once if filtering
+    candidates_list = list(candidates)
+
+    if max_age_cutoff is not None:
+        filtered = [
+            c
+            for c in candidates_list
+            if c.upload_time is None or c.upload_time >= max_age_cutoff
+        ]
+        if filtered:
```
dhellmann, line 300: We're going to want some structure to the logging as we add more filtering features. So maybe we log the full number of versions found, then for each filter step how many remain and the criteria used to do the filtering. Something like this, for example:

```
INFO have 100 versions of foo>=1.0.0
INFO have 10 version of foo>=1.0.0 older than 30 days
INFO have 1 version of foo>=1.0.0 at least 14 days old
```

src/fromager/commands/bootstrap.py
rd4398: Without `--max-release-age`, `--multiple-versions` would attempt to build every version that matches the requirement specifier. For a package like setuptools>=60, that could mean dozens or hundreds of versions going back years. The age limit acts as a guardrail to keep the build scope manageable.

Let me know if you think I should make it optional

src/fromager/resolver.py
rd4398: makes sense. I have attempted to add some structure to logging in my latest push

src/fromager/commands/bootstrap.py
rd4398: I agree. I have fixed this in latest push

src/fromager/commands/bootstrap.py
dhellmann: It's definitely advisable. I'm not sure I want to make it required. We could require a user to provide some value, but allow 0 as an explicit choice, I guess. Let's see what others think.

LalatenduMohanty: I would suggest that we make it optional. That way it fixes our concern for the  downstream.

LalatenduMohanty (changes requested):

src/fromager/commands/bootstrap.py
LalatenduMohanty: Users with tight specifiers like tomli>=2.0,<=2.0.2 don't need the guardrail, for example in e2e test, which has to dynamically compute an age back to 2021 just to include tomli 2.0.0.     
          
Either way, we should change default=0 to default=None with IntRange(min=1) right now --max-release-age 0 tells the user "is required" even though they provided it.

LalatenduMohanty (changes requested):

src/fromager/commands/bootstrap.py
rd4398: Okay, I have made changes in latest commit.

src/fromager/context.py
```diff
@@ -113,6 +114,15 @@ def __init__(
...
         self.cooldown: Cooldown | None = cooldown
+        self._max_release_age: datetime.timedelta | None = max_release_age
+
+    @property
+    def max_release_age(self) -> datetime.timedelta | None:
+        return self._max_release_age
+
+    def set_max_release_age(self, days: int) -> None:
+        """Set the maximum release age in days."""
+        self._max_release_age = datetime.timedelta(days=days)
```
dhellmann, line 127: We need validation here to ensure the value has the right sign.

src/fromager/resolver.py
```diff
@@ -242,10 +270,37 @@ def find_all_matching_from_provider(
...
+                    len(candidates_list),
+                )
+            candidates_list = filtered
+        else:
+            logger.warning(
+                "%s: all %d versions are older than max release age cutoff, "
+                "keeping all versions to avoid empty resolution",
+                req.name,
+                len(candidates_list),
+            )
```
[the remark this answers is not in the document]
dhellmann: I was surprised by this, too.

src/fromager/resolver.py
```diff
@@ -242,10 +270,45 @@ def find_all_matching_from_provider(
...
         ) from err
 
+    # Materialize candidates so we can iterate more than once if filtering
+    candidates_list = list(candidates)
+    logger.info(
+        "%s: found %d candidate(s) matching %s",
+        req.name,
+        len(candidates_list),
+        req,
+    )
```
LalatenduMohanty: Nitpick: This log will fire on every find_all_matching_from_provider call, even when --max-release-age is not active. In a large bootstrap this could add hundreds of extra INFO lines.

LalatenduMohanty (changes requested):

src/fromager/resolver.py
LalatenduMohanty: I would suggest that we move this to debug.

src/fromager/resolver.py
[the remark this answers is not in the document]
rd4398: Hmm, I somehow missed it. I have fixed it in latest push.

src/fromager/context.py
rd4398: Agreed, done in latest push.

src/fromager/resolver.py
rd4398: okay, fixed in latest commit but not as debug. Let me know if you prefer debug

LalatenduMohanty (dismissed):

LalatenduMohanty:
@mergify rebase

LalatenduMohanty (approved):

PR #1090, opened 2026-04-24
feat(sbom): use annotated data types in Pydantic config

### LINKED ISSUE #1072: SBOM: use annotated data types in Pydantic config
PR https://github.com/python-wheel-build/fromager/pull/1032 added new Pydantic models that use generic `str` like `upstream_purl` and `repository_url`. Define more specialized, annotated string types with validators to make the schema self-documenting.

Example:

```python
BuildDirectory = typing.Annotated[
    pathlib.Path,
    pydantic.BeforeValidator(_before_builddirectory),
]
```

smoparth:
Replace generic `str` fields in SbomSettings and PurlConfig with
Pydantic's built-in types annotated validators to validate input and make the schema
self-documenting.

Added pydantic>=2.12 minimum version for url_preserve_empty_path support.

Changes:
- PurlType: StringConstraints with strip, lowercase, min_length
- UpstreamPurl: AfterValidator with PackageURL.from_string()
- URL fields (namespace, repository_url): use pydantic.AnyUrl directly
- supplier/creators: remain plain str per SPDX 2.3 JSON schema

The @field_validator on PurlConfig.upstream is replaced by the
UpstreamPurl annotated type, moving validation into the type itself.

Closes: #1072

# Pull Request Description

### DIFF
--- a/src/fromager/packagesettings/_models.py
+++ b/src/fromager/packagesettings/_models.py
@@ -10,10 +10,9 @@ from collections.abc import Mapping
 
 import pydantic
 import yaml
-from packageurl import PackageURL
 from packaging.requirements import Requirement
 from packaging.utils import canonicalize_name
-from pydantic import Field
+from pydantic import AnyUrl, Field
 from pydantic_core import core_schema
 
 from ._typedefs import (
@@ -21,8 +20,10 @@ from ._typedefs import (
     BuildDirectory,
     EnvVars,
     Package,
+    PurlType,
     RawAnnotations,
     Template,
+    UpstreamPurl,
     Variant,
     VariantChangelog,
 )
@@ -49,7 +50,7 @@ class SbomSettings(pydantic.BaseModel):
     supplier: str = "NOASSERTION"
     """SPDX supplier field for the wheel package (e.g. ``Organization: ExampleCo``)"""
 
-    namespace: str = "https://spdx.org/spdxdocs"
+    namespace: AnyUrl = AnyUrl("https://spdx.org/spdxdocs")
     """Base URL for the SPDX documentNamespace"""
 
     creators: list[str] = Field(default_factory=list)
@@ -58,10 +59,10 @@ class SbomSettings(pydantic.BaseModel):
     The fromager tool creator entry is always added automatically.
     """
 
-    purl_type: str = "pypi"
+    purl_type: PurlType = "pypi"
     """Default purl type for all packages (e.g. ``pypi``, ``generic``)"""
 
-    repository_url: str | None = None
+    repository_url: AnyUrl | None = None
     """Default purl ``repository_url`` qualifier for all packages
 
     When set, this URL is added to every purl as a qualifier
@@ -89,7 +90,7 @@ class PurlConfig(pydantic.BaseModel):
 
     model_config = MODEL_CONFIG
 
-    type: str | None = None
+    type: PurlType | None = None
     """Override the purl type (e.g. ``generic`` instead of ``pypi``)"""
 
     namespace: str | None = None
@@ -101,13 +102,13 @@ class PurlConfig(pydantic.BaseModel):
     version: str | None = None
     """Override the purl version component (defaults to the resolved version)"""
 
-    repository_url: str | None = None
+    repository_url: AnyUrl | None = None
     """Per-package override for the purl ``repository_url`` qualifier.
 
     Overrides the global ``sbom.repository_url`` setting for this package.
     """
 
-    upstream: str | None = None
+    upstream: UpstreamPurl | None = None
     """Full purl string identifying the upstream source package.
 
     When set, this is used as the upstream identity in the SBOM's
@@ -117,18 +118,6 @@ class PurlConfig(pydantic.BaseModel):
     When absent, the upstream purl is auto-derived from the downstream
     purl without the ``repository_url`` qualifier.
     """
-
-    @pydantic.field_validator("upstream")
-    @classmethod
-    def validate_upstream_purl(cls, v: str | None) -> str | None:
-        """Validate that upstream is a valid purl string."""
-        if v is None:
-            return v
-        try:
-            PackageURL.from_string(v)
-        except ValueError as err:
-            raise ValueError(f"invalid upstream purl {v!r}") from err
-        return v
 
 
 class ResolverDist(pydantic.BaseModel):
--- a/src/fromager/packagesettings/_typedefs.py
+++ b/src/fromager/packagesettings/_typedefs.py
@@ -7,8 +7,10 @@ import typing
 from collections.abc import Mapping
 
 import pydantic
+from packageurl import PackageURL
 from packaging.utils import NormalizedName, canonicalize_name
 from packaging.version import Version
+from pydantic import StringConstraints
 from pydantic_core import CoreSchema, core_schema
 
 # common settings
@@ -19,6 +21,8 @@ MODEL_CONFIG = pydantic.ConfigDict(
     frozen=True,
     # read inline doc strings
     use_attribute_docstrings=True,
+    # preserve URLs as-is without trailing slash normalization
+    url_preserve_empty_path=True,
 )
 
 
@@ -97,6 +101,31 @@ Variant = typing.NewType("Variant", str)
 # Changelog
 GlobalChangelog = Mapping[Variant, list[str]]
 VariantChangelog = Mapping[PackageVersion, list[str]]
+
+
+# purl type (e.g. "pypi", "generic", "github")
+PurlType = typing.Annotated[
+    str,
+    StringConstraints(strip_whitespace=True, to_lower=True, min_length=1),
+]
+
+
+# full purl string identifying an upstream source package
+def _validate_upstream_purl(v: str) -> str:
+    """Validate that *v* is a well-formed purl string."""
+    try:
+        PackageURL.from_string(v)
+    except ValueError as err:
+        raise ValueError(f"invalid upstream purl {v!r}: {err}") from err
+    return v
+
+
+UpstreamPurl = typing.Annotated[
+    str,
+    StringConstraints(strip_whitespace=True, min_length=1),
+    pydantic.AfterValidator(_validate_upstream_purl),
+]
+
 
 # Annotations
 RawAnnotations = Mapping[str, str]
--- a/tests/test_sbom.py
+++ b/tests/test_sbom.py
@@ -61,7 +61,9 @@ def test_generate_sbom_default_purls(tmp_path: pathlib.Path) -> None:
 
 def test_generate_sbom_repository_url_qualifier(tmp_path: pathlib.Path) -> None:
     """Verify global repository_url adds qualifier to downstream but not upstream."""
-    settings = SbomSettings(repository_url="https://packages.redhat.com")
+    settings = SbomSettings(
+        repository_url="https://packages.redhat.com",  # type: ignore[arg-type]
+    )
     ctx = make_sbom_ctx(tmp_path, sbom_settings=settings)
     doc = sbom.generate_sbom(
         ctx=ctx,
@@ -82,7 +84,7 @@ def test_generate_sbom_custom_settings(tmp_path: pathlib.Path) -> None:
     """Verify custom supplier, namespace, and creators are used."""
     settings = SbomSettings(
         supplier="Organization: ExampleCo",
-        namespace="https://www.example.com",
+        namespace="https://www.example.com",  # type: ignore[arg-type]
         creators=["Organization: ExampleCo"],
     )
     ctx = make_sbom_ctx(tmp_path, sbom_settings=settings)
@@ -132,7 +134,9 @@ def test_generate_sbom_package_repository_url_override(tmp_path: pathlib.Path) -
     """Verify per-package repository_url overrides the global value."""
     ctx = make_sbom_ctx(
         tmp_path,
-        sbom_settings=SbomSettings(repository_url="https://packages.redhat.com"),
+        sbom_settings=SbomSettings(
+            repository_url="https://packages.redhat.com",  # type: ignore[arg-type]
+        ),
         package_overrides={
             "purl": {"repository_url": "https://mirror.example.com/simple"},
         },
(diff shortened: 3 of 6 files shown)


src/fromager/packagesettings/_models.py
```diff
@@ -46,22 +50,22 @@ class SbomSettings(pydantic.BaseModel):
...
 
     The fromager tool creator entry is always added automatically.
     """
 
-    purl_type: str = "pypi"
+    purl_type: PurlType = "pypi"
     """Default purl type for all packages (e.g. ``pypi``, ``generic``)"""
 
-    repository_url: str | None = None
+    repository_url: RepositoryUrl | None = None
```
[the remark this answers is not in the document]
smoparth: Good observation! but we haven't started the downstream integration yet.

src/fromager/packagesettings/_typedefs.py
```diff
@@ -98,6 +101,99 @@ def _validate_envkey(v: typing.Any) -> str:
 GlobalChangelog = Mapping[Variant, list[str]]
 VariantChangelog = Mapping[PackageVersion, list[str]]
 
+_SPDX_ACTOR_RE = re.compile(r"^(Organization|Person|Tool):\s+\S.*$", flags=re.DOTALL)
```
mprpic: The official SPDX 2.3 schema lists these as examples but doesn't actually enforce it at the JSON schema level, it only expects a string:

https://github.com/spdx/tools-java/blob/master/resources/spdx-schema-v2.3.json#L58

So I think we're being overly defensive here.

src/fromager/packagesettings/_typedefs.py
```diff
@@ -98,6 +101,99 @@ def _validate_envkey(v: typing.Any) -> str:
...
+        raise ValueError(f"URL is missing a host/netloc component: {v!r}")
+    return v
+
+
+# purl type (e.g. "pypi", "generic", "github")
+def _validate_purl_type(v: str) -> str:
+    """Strip, lowercase, and reject empty purl type strings."""
+    if not isinstance(v, str):
+        raise TypeError(f"expected str, got {type(v)}: {v!r}")
+    v = v.strip().lower()
```
mprpic: The purl library normalizes strings to lower-case, so this too is being overly pedantic (or shall I say "pydantic") as far as validation goes.

src/fromager/packagesettings/_typedefs.py
```diff
@@ -98,6 +101,99 @@ def _validate_envkey(v: typing.Any) -> str:
...
+    if not v:
+        raise ValueError("purl type must not be empty")
```
mprpic: Do we care about catching this here? When constructing the purl, the purl library would yell about this. Or are we trying to just fail early (or rather earlier)?

src/fromager/packagesettings/_typedefs.py
```diff
@@ -98,6 +101,99 @@ def _validate_envkey(v: typing.Any) -> str:
...
+
+# full purl string identifying an upstream source package
+def _validate_upstream_purl(v: str) -> str:
+    """Validate that *v* is a well-formed purl string."""
+    if not isinstance(v, str):
+        raise TypeError(f"expected str, got {type(v)}: {v!r}")
+    v = v.strip()
+    try:
+        PackageURL.from_string(v)
+    except ValueError as err:
```
mprpic, line 118: This same exception would be caught in when this runs in the purl method. Same question as above, is the intent here to catch things earlier?

mprpic (commented):
This looks fine overall, but I do have to question the level of validation we're going into. If this is the intent and other parts of the Fromager code base expect this type of detailed validation, then fine with me.

src/fromager/packagesettings/_typedefs.py
```diff
@@ -98,6 +101,99 @@ def _validate_envkey(v: typing.Any) -> str:
...
+    """Validate that *v* has an ``http`` or ``https`` scheme and a host."""
+    if not isinstance(v, str):
+        raise TypeError(f"expected str, got {type(v)}: {v!r}")
+    v = v.strip()
+    parsed = urlparse(v)
+    if parsed.scheme not in ("http", "https"):
+        raise ValueError(f"URL must use http or https scheme, got {v!r}")
+    if not parsed.netloc:
+        raise ValueError(f"URL is missing a host/netloc component: {v!r}")
+    return v
```
tiran: Pydantic has a dedicated validator and constraint for network types.

src/fromager/packagesettings/_typedefs.py
tiran: On top of that, Pydantic has a StringConstraints that can archive all these checks and conversations.

src/fromager/packagesettings/_typedefs.py
```diff
@@ -98,6 +101,99 @@ def _validate_envkey(v: typing.Any) -> str:
...
+    str,
+    pydantic.BeforeValidator(_validate_url),
+]
+
+
+# full purl string identifying an upstream source package
+def _validate_upstream_purl(v: str) -> str:
+    """Validate that *v* is a well-formed purl string."""
+    if not isinstance(v, str):
+        raise TypeError(f"expected str, got {type(v)}: {v!r}")
```
tiran: You can get rid of this check by using an after validator instead of a before validator.

tiran (changes requested):
@smoparth You can get rid of most helper functions and type checks by using Pydantic's built-in constraints, dedicated data types, and after validator instead of before validator. Pydantic has great documentation that explains its validator concept very well.

@mprpic The checks are going to detect misconfiguration at startup, e.g. in `fromager lint`.

src/fromager/packagesettings/_typedefs.py
smoparth: Thanks @tiran ! I used Annotated[AnyHttpUrl, AfterValidator(str)] initially, it works at runtime since AfterValidator(str) converts the Url object back to a plain string. 
However, mypy sees the static type as AnyHttpUrl, not str, which causes type errors on default values (namespace: SpdxNamespace = "https://...") and dict assignments in sbom.py requiring type: ignore comments across multiple files. 
To avoid that, I found a work-around using TypeAdapter, which keeps the declared type as str. Pushed in the latest commit. Let me know if there's a better approach.

src/fromager/packagesettings/_typedefs.py
smoparth: Updated in the latest commit. I originally went a different direction with these, appreciate you pointing it out.

src/fromager/packagesettings/_typedefs.py
smoparth: Updated in the latest commit.

src/fromager/packagesettings/_typedefs.py
smoparth: Based on project standards to validate the user inputs I added those checks but I'm open to adjusting the strictness, happy to widen or tighten the checks based on what is preferred.

src/fromager/packagesettings/_typedefs.py
smoparth: As Christian's pointed out we are trying to detect misconfiguration at startup.

src/fromager/packagesettings/_typedefs.py
smoparth: We are trying to detect misconfiguration at startup.

src/fromager/packagesettings/_models.py
tiran: Would it makse sense to use `AnyHttpUrl` instead of defining a custom type `RepositoryUrl`?

src/fromager/packagesettings/_models.py
```diff
@@ -46,22 +50,22 @@ class SbomSettings(pydantic.BaseModel):
 
     model_config = MODEL_CONFIG
 
-    supplier: str = "NOASSERTION"
+    supplier: SpdxActor = "NOASSERTION"
     """SPDX supplier field for the wheel package (e.g. ``Organization: ExampleCo``)"""
 
-    namespace: str = "https://spdx.org/spdxdocs"
+    namespace: SpdxNamespace = "https://spdx.org/spdxdocs"
```
tiran: Same here, why not use `AnyHttpUrl` type?

src/fromager/packagesettings/_typedefs.py
```diff
@@ -98,6 +101,73 @@ def _validate_envkey(v: typing.Any) -> str:
...
+            f"SPDX actor must be 'NOASSERTION' or "
+            f"'Organization: <name>' / 'Person: <name>' / 'Tool: <name>', "
+            f"got {v!r}"
+        )
+    return v
+
+
+SpdxActor = typing.Annotated[
+    str,
+    StringConstraints(strip_whitespace=True, min_length=1),
```
tiran: Is `StringConstraints` necessary? `_validate_spdx_actor` will fail with a better error message when the input is misformed.

tiran (changes requested):

src/fromager/packagesettings/_models.py
smoparth: I added the challenge in https://github.com/python-wheel-build/fromager/pull/1090#discussion_r3147688936.

The issue is that AnyHttpUrl resolves to a pydantic.Url object, not str. The consumer of this model (sbom.py) uses these values in f-strings (f"{namespace}/{name}-{version}.spdx.json") and assigns them into dict[str, str] (qualifiers["repository_url"] = repo_url). With AnyHttpUrl as the field type, **mypy flags all of these as type errors since it expects str, not Url**. We'd either **need to update sbom.py to call str() on every access (which expands the scope of this PR), or add # type: ignore comments across multiple files**. 

The RepositoryUrl / SpdxNamespace wrappers exist purely to keep the field type as str while delegating validation to AnyHttpUrl internally. If you'd prefer, I can switch the fields to AnyHttpUrl directly and **update sbom.py to handle the Url type** -- happy to go either way.

src/fromager/packagesettings/_typedefs.py
smoparth: No, I wil remove it.

src/fromager/packagesettings/_typedefs.py
mprpic: I would relax it to any string, not strict to Org/Person/Tool. Follow the SPDX2.3 JSON schema, if that allows any string, so should we.

src/fromager/packagesettings/__init__.py
tiran, line 1: Do we need to export these types? An exported symbol becomes part of the stable API.

src/fromager/packagesettings/_models.py
tiran: I would prefer to use `AnyUrl` and convert it to `str` in a few places. @mprpic what would you prefer?

If we decide to use a custom validation type, then we don't need need separate RepositoryUrl and SpdxNamespace types. A single StringUrl type would do.

mprpic: `AnyUrl` sounds good to me!

src/fromager/packagesettings/_typedefs.py
smoparth: Done. supplier and creators are back to plain str with no validation.

src/fromager/packagesettings/_models.py
smoparth: Done. Thanks for the suggestions! Switched to AnyUrl directly on both repository_url and namespace fields. Added str() calls in sbom.py where the values enter string contexts and removed RepositoryUrl and SpdxNamespace types entirely.

src/fromager/packagesettings/_models.py
smoparth: Updated.

src/fromager/packagesettings/_typedefs.py
smoparth: Updated.

src/fromager/packagesettings/__init__.py
smoparth: Good point. removed all new types from __init__.py exports. PurlType and UpstreamPurl stay internal to the package.

src/fromager/packagesettings/_models.py
```diff
@@ -58,10 +59,10 @@ class SbomSettings(pydantic.BaseModel):
     The fromager tool creator entry is always added automatically.
     """
 
-    purl_type: str = "pypi"
+    purl_type: PurlType = "pypi"
     """Default purl type for all packages (e.g. ``pypi``, ``generic``)"""
 
-    repository_url: str | None = None
+    repository_url: AnyUrl | None = None
     """Default purl ``repository_url`` qualifier for all packages
```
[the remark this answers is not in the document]
smoparth: We agreed to AnyURL based on reviewers feedback.

tests/test_sbom.py
```diff
@@ -61,7 +61,9 @@ def test_generate_sbom_default_purls(tmp_path: pathlib.Path) -> None:
 
 def test_generate_sbom_repository_url_qualifier(tmp_path: pathlib.Path) -> None:
     """Verify global repository_url adds qualifier to downstream but not upstream."""
-    settings = SbomSettings(repository_url="https://packages.redhat.com")
+    settings = SbomSettings(
+        repository_url="https://packages.redhat.com/simple",  # type: ignore[arg-type]
```
mprpic: Why switch this to add `/simple` at the end? The repo URL should be generic at this point and is more of a namespace that defines our packages, not an actual pinpointed location.

smoparth: @mprpic You're right. I only added /simple to work around AnyUrl normalizing bare hostnames with a trailing slash. I'll revert the URLs back to "https://packages.redhat.com" but **update the assertions to expect the normalized form** ("https://packages.redhat.com/")??

mprpic: Where does it normalize to add the trailing slash? If I assign `https://packages.redhat.com` to the repo URL variable, does pydantic overwrite that to `https://packages.redhat.com/`?

smoparth: Yes, when the user writes "https://packages.redhat.com" in their settings.yaml YAML, the value that ends up in the SBOM JSON is "https://packages.redhat.com/". Pydantic's AnyUrl does this normalization.

smoparth: @mprpic Here's the reference for that behavior: [pydantic#7186](https://github.com/pydantic/pydantic/issues/7186). AnyUrl adds a trailing / to bare hostnames without a path component.

A workaround from that thread strips it via AfterValidator:

Annotated[AnyUrl, AfterValidator(lambda x: str(x).rstrip("/"))]
This preserves AnyUrl validation without altering the original URL. Should I apply this, or is the trailing slash acceptable?

mprpic: Ok, we don't want that then. The value should be left as is specified in the settings. Can we just use the config that disables this behavior: https://pydantic.dev/articles/pydantic-v2-12-release#preserve-empty-url-paths ?

smoparth: Yes updated the model config and pinned pydantic>=2.12 in pyproject.toml. URLs are now preserved as-is. PTAL.

@tiran I have added a pin pydantic>=2.12 in pyproject.toml due to this thread.

mprpic (approved):

tiran (approved):

tiran:
@mergifyio rebase

PR #1093, opened 2026-04-26
test(unit): mark PyPI-backed tests as network

booxter:
# Pull Request Description

## What

Mark PyPI-backed tests as network.

## Why

These tests install build requirements from https://pypi.org/simple via
uv, so they require real network access and should be skipped unless
--with-network is enabled.

### DIFF
--- a/tests/test_dependencies.py
+++ b/tests/test_dependencies.py
@@ -181,6 +181,7 @@ def test_get_build_system_dependencies_cached(
 
 @patch("fromager.dependencies._write_requirements_file")
 @_clean_build_artifacts
+@pytest.mark.network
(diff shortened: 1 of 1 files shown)


LalatenduMohanty:
These tests are already running and passing with network access in CI. They aren't failing today. It also passes local dev environments which has internet access. 

Without @pytest.mark.network, the tests silently make network calls they pass if you have internet, but they're slower and would fail in an air-gapped environment. 
With the marker, they're skipped by default and only run when you explicitly opt in

LalatenduMohanty (approved):

LalatenduMohanty:
@mergify rebase

PR #1094, opened 2026-04-26
fix(deps): drop requests-mock from runtime deps

booxter:
# Pull Request Description

## What

Drop requests-mock from runtime deps.

## Why

requests-mock is only imported by tests and does not belong in runtime
dependencies.

It was originally added as a test-only dependency and was accidentally
promoted into project.dependencies during the hatchling migration in
5e47f8b9254a7231bce627b9a7e5336d8707e711.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,7 +44,6 @@ dependencies = [
     "rich",
     "requests",
     "resolvelib",
-    "requests-mock",
     "starlette",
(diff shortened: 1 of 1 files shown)


LalatenduMohanty (approved):

LalatenduMohanty:
@booxter Thank you for the fix.

PR #1095, opened 2026-04-29
fix(resolver): make resolution cache values immutable to prevent corr…

### LINKED ISSUE #1021: Resolution cache values should be immutable to prevent accidental corruption
`BootstrapRequirementResolver._resolved_requirements` (introduced in #975) uses mutable `list` values. Since `get_cached_resolution` returns the cached list by reference, any caller that appends, removes, or reorders items would silently corrupt the shared cache, affecting all subsequent lookups for that requirement. 

Making the values immutable prevents this class of bug at the type level.

smoparth:
fix(resolver): make resolution cache values immutable to prevent corruption

Store session-level cache values as tuples instead of lists so that callers cannot accidentally mutate shared cache state. Return defensive list copies from resolve() and resolve_versions() to keep the public API unchanged.

Add tests verifying that cached values are immutable and that repeated resolve() calls return independent lists.

Closes: #1021

# Pull Request Description

## Why

### DIFF
--- a/src/fromager/bootstrap_requirement_resolver.py
+++ b/src/fromager/bootstrap_requirement_resolver.py
@@ -51,9 +51,11 @@ class BootstrapRequirementResolver:
         self.prev_graph = prev_graph
         # Session-level resolution cache to avoid re-resolving same requirements
         # Key: (requirement_string, pre_built) to distinguish source vs prebuilt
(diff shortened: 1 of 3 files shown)


rd4398 (approved):
LGTM

PR #1099, opened 2026-04-30
feat(cli): add option for a single output-dir and hide options

### LINKED ISSUE #721: Simplify output directory handling

jlarkin09:
Add a new -O/--output-dir option that sets the base directory for sdists-repo, wheels-repo, and work-dir. The individual directory options are hidden but still functional for backwards compatibility. When --output-dir is "." (the default), behavior is identical to before.

Simplifies the CLI by consolidating three separate directory options into one. Addresses review feedback from #905.

Closes: #721

### DIFF
--- a/src/fromager/__main__.py
+++ b/src/fromager/__main__.py
@@ -68,18 +68,27 @@ else:
     help="save error messages to a file",
 )
 @click.option(
+    "-O",
+    "--output-dir",
+    default=pathlib.Path("."),
+    type=clickext.ClickPath(),
+    help="base directory for sdists-repo, wheels-repo, and work-dir (default: current directory)",
+)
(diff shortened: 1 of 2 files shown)


smoparth (approved):
LGTM

rd4398 (approved):
This looks good! Thanks

shifa-khan:
LGTM

shifa-khan (approved):

tiran:
@mergifyio rebase

PR #1102, opened 2026-05-01
feat(resolver): enforce PyPI quarantine check for all resolver types

### LINKED ISSUE #1101: Enforce PyPI quarantine check for non-PyPI resolvers (GitHub/GitLab)
## Problem                                                

The PEP 792 quarantine check currently only runs inside `get_project_from_pypi()` in `resolver.py`, which is only called by `PyPIProvider.find_candidates()`. When a package is configured to resolve from GitHub or GitLab (via `GitHubTagProvider` or `GitLabTagProvider` through override plugins), the quarantine status on PyPI is never checked.                                                                                                                                                         
                                                            
This means a quarantined package can still be resolved and built if it uses a non-PyPI resolver, bypassing a safety mechanism.                                         
  
## Expected behavior                                                                                                                                                   
                                                            
Regardless of which resolver provider is used, fromager should check the package's quarantine status on PyPI before proceeding with resolution. If the package is quarantined on PyPI, resolution should fail with a clear error.
                                                                                                                                                                         
## Proposed solution                                                                                                                                                   

### Option A: Unconditional check at resolution entry points

Add a standalone `check_pypi_quarantine_status()` function and call it unconditionally from `resolve()` and `resolve_source()` for all resolver types. Remove the existing quarantine check from `get_project_from_pypi()`.
                                                                                                                                                                         
- **Pros:** Simpler, single responsibility, no conditional logic, also fixes the custom-index case                                                                     
- **Cons:** For `PyPIProvider` resolving from pypi.org, the project page is fetched twice — once for the quarantine check and once for candidates
                                                                                                                                                                         
### Option B: Conditional check only for non-PyPI providers                                                                                                            
                                                                                                                                                                         
Same standalone function, but only called when the provider is not a `PyPIProvider`. Keep the existing quarantine check inside `get_project_from_pypi()`.              
                                                            
- **Pros:** No duplicate HTTP requests                                                                                                                                 
- **Cons:** Quarantine logic split across two locations, conditional branching, does not fix the custom-index case

pavank63:
The PEP 792 quarantine check previously only ran inside get_project_from_pypi(), which is only called by PyPIProvider. Packages resolved via GitHubTagProvider, GitLabTagProvider, or PyPIProvider with a custom index URL bypassed the check entirely.

Add a standalone check_pypi_quarantine_status() function that queries pypi.org for quarantine status and call it unconditionally from both resolve() and resolve_source() entry points. Remove the quarantine check from get_project_from_pypi() to avoid split responsibility.

Closes: https://github.com/python-wheel-build/fromager/issues/1101

### DIFF
--- a/src/fromager/packagesettings/_models.py
+++ b/src/fromager/packagesettings/_models.py
@@ -170,6 +170,15 @@ class ResolverDist(pydantic.BaseModel):
     .. versionadded:: 0.82
     """
 
+    skip_pypi_quarantine: bool = False
+    """Skip PyPI quarantine status check for this package? (default: no)
+
+    Use for packages resolved from non-PyPI sources (GitHub, GitLab)
+    where the PyPI name is a different, unrelated project.
+
+    .. versionadded:: 0.86
+    """
+
     @pydantic.model_validator(mode="after")
     def validate_ignore_platform(self) -> typing.Self:
         if self.ignore_platform and not self.include_wheels:
--- a/src/fromager/packagesettings/_pbi.py
+++ b/src/fromager/packagesettings/_pbi.py
@@ -262,6 +262,14 @@ class PackageBuildInfo:
         return self._ps.resolver_dist.min_release_age
 
     @property
+    def resolver_skip_pypi_quarantine(self) -> bool:
+        """Skip PyPI quarantine status check for this package?
+
+        .. versionadded:: 0.86
+        """
+        return self._ps.resolver_dist.skip_pypi_quarantine
+
+    @property
     def use_pypi_org_metadata(self) -> bool:
         """Can use metadata from pypi.org JSON / Simple API?
 
--- a/src/fromager/resolver.py
+++ b/src/fromager/resolver.py
@@ -76,6 +76,30 @@ def match_py_req(py_req: str, *, python_version: Version = PYTHON_VERSION) -> bo
     return python_version in SpecifierSet(py_req)
 
 
+def check_pypi_quarantine_status(project: str) -> None:
+    """Check if a project is quarantined on PyPI (PEP 792).
+
+    Raises ValueError if the project is quarantined.
+    """
+    client = pypi_simple.PyPISimple(
+        endpoint=PYPI_SERVER_URL,
+        session=session,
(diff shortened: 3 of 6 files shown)


tiran:
I see two potential issues with this approach:

- some downstream packages are not on PyPI (e.g. aotriton)
- in some cases, a downstream package is typo-squatted by somebody else (e.g. deep-gemm) or may be a completely different package than on PyPI.

pavank63:
@tiran thanks for flagging these. 

> some downstream packages are not on PyPI (e.g. aotriton)

This one should be handled already. The except Exception block in `check_pypi_quarantine_status()` catches the 404 and silently returns, so packages that don't exist on PyPI won't be blocked.

> in some cases, a downstream package is typo-squatted by somebody else (e.g. deep-gemm) or may be a completely different package than on PyPI.

I can think of two options to handle this.
                                                                                                                                                                                                                                        
1. Warn instead of block for non-PyPI resolvers — For non-PyPI resolvers (GitHub/GitLab), log a warning instead of raising an error when a package with the same name is quarantined on PyPI. This gives visibility without incorrectly blocking legitimate packages. If the package doesn't exist on PyPI at all, the check is silently skipped. 

2. Per-package opt-out setting — Add a package-level setting (e.g. skip_pypi_quarantine_check: true) that allows maintainers to disable the quarantine check for specific packages where the PyPI name is known to be a different project or doesn't exist.

pavank63:
Added a per-package opt-out: `resolver_dist.skip_pypi_quarantine: true` in the package settings YAML. Packages not on PyPI are already handled (404 is silently skipped). The opt-out covers the typo-squatting case where the PyPI name is an unrelated project.

tiran:
There is a similar option `use_pypi_org_metadata`. How does the new option relate to the existing option?

pavank63:
I think they serve different purposes. `use_pypi_org_metadata` auto-defaults to `false` for any package that has a settings file, patches, or a plugin. That's a very broad set and reusing it for quarantine would silently skip the security check for most packages in a typical deployment, which is the opposite of what this PR is trying to achieve. `skip_pypi_quarantine` is an explicit opt-out for the narrow case where the PyPI name is a different project (e.g. typo-squatting).

PR #1109, opened 2026-05-04
refactor(bootstrapper): replace recursion with trampoline for iterative execution

### LINKED ISSUE #1068: Convert Bootstrapper from Recursive to Iterative
The Bootstrapper class in `src/fromager/bootstrapper.py` currently uses recursive calls to process dependency trees. @dhellmann discovered that when bootstrapping packages with very deep or wide dependency graphs (especially with --multiple-versions mode enabled), this hits Python's recursion depth limit and causes stack overflow errors.

The recursion occurs at two critical points:

Line 521: Install dependencies (processed AFTER building the package)
Line 708: Build dependencies (BUILD_SYSTEM, BUILD_BACKEND, BUILD_SDIST - processed BEFORE building)

We should convert Bootstrapper to iterative approach

rd4398:
This commit converts recursive bootstrap methods to generators using the trampoline library, eliminating Python's recursion depth limit for deep/wide dependency graphs.

trampoline source: https://gitlab.com/ferreum/trampoline

Closes: https://github.com/python-wheel-build/fromager/issues/1068

### DIFF
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -19,6 +19,7 @@ from packaging.requirements import Requirement
 from packaging.utils import NormalizedName, canonicalize_name
 from packaging.version import Version
 from resolvelib.resolvers import ResolverException
+from trampoline import trampoline
 
 from . import (
     bootstrap_requirement_resolver,
@@ -274,6 +275,16 @@ class Bootstrapper:
     def bootstrap(self, req: Requirement, req_type: RequirementType) -> None:
         """Bootstrap a package and its dependencies.
 
+        Public entry point that drives the generator-based implementation
+        via trampoline() for iterative (non-recursive) execution.
+        """
+        trampoline(self._bootstrap_gen(req, req_type))
+
+    def _bootstrap_gen(
+        self, req: Requirement, req_type: RequirementType
+    ) -> typing.Generator[typing.Any, typing.Any, None]:
+        """Generator that bootstraps a package and its dependencies.
+
         Handles setup, validation, and error handling. Delegates actual build
         work to _bootstrap_impl().
 
@@ -307,7 +318,9 @@ class Bootstrapper:
 
         # Bootstrap each resolved version
         for source_url, resolved_version in resolved_versions:
-            self._bootstrap_single_version(req, req_type, source_url, resolved_version)
+            yield self._bootstrap_single_version(
+                req, req_type, source_url, resolved_version
+            )
 
         # In multiple versions mode, report any failures for this requirement
         if self.multiple_versions and self._failed_versions:
@@ -329,7 +342,7 @@ class Bootstrapper:
         req_type: RequirementType,
         source_url: str,
         resolved_version: Version,
-    ) -> None:
+    ) -> typing.Generator[typing.Any, typing.Any, None]:
         """Bootstrap a single version of a package.
 
         Extracted from bootstrap() to handle both single and multiple version modes.
@@ -365,7 +378,7 @@ class Bootstrapper:
         # Track dependency chain - context manager ensures cleanup even on exception
         with self._track_why(req_type, req, resolved_version):
             try:
-                self._bootstrap_impl(
+                yield self._bootstrap_impl(
                     req, req_type, source_url, resolved_version, build_sdist_only
                 )
             except Exception as err:
@@ -400,7 +413,7 @@ class Bootstrapper:
         source_url: str,
         resolved_version: Version,
         build_sdist_only: bool,
-    ) -> None:
+    ) -> typing.Generator[typing.Any, typing.Any, None]:
         """Internal implementation - performs the actual bootstrap work.
 
         Called by bootstrap() after setup, validation, and seen-checking.
@@ -453,7 +466,7 @@ class Bootstrapper:
             )
 
             # Build from source (handles test-mode fallback internally)
-            build_result = self._build_from_source(
+            build_result = yield self._build_from_source(
                 req=req,
                 resolved_version=resolved_version,
                 source_url=source_url,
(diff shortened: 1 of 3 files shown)


src/fromager/bootstrapper.py
```diff
@@ -274,6 +275,16 @@ def _processing_build_requirement(self, current_req_type: RequirementType) -> bo
...
         """Bootstrap a package and its dependencies.
 
+        Public entry point that drives the generator-based implementation
+        via trampoline() for iterative (non-recursive) execution.
+        """
+        trampoline(self._bootstrap_gen(req, req_type))
+
+    def _bootstrap_gen(
+        self, req: Requirement, req_type: RequirementType
+    ) -> typing.Generator[typing.Any, typing.Any, None]:
```
dhellmann, line 285: Maybe we could be more specific with types here, instead of using typing.Any?

src/fromager/bootstrapper.py
```diff
@@ -1334,11 +1341,13 @@ def _get_version_from_package_metadata(
             ctx=self.ctx,
             parent_dir=source_dir.parent,
         )
-        build_dependencies = self._prepare_build_dependencies(
-            req=req,
-            resolved_version=None,
-            sdist_root_dir=source_dir,
-            build_env=build_env,
+        build_dependencies = trampoline(
```
dhellmann, line 1344: Why call `trampoline()` again here instead of yielding?

dhellmann (commented):
I have to admit, this is a small clean change but I really had to study the `trampoline` code to make sense of how it works. I'm not sure how I feel about hiding the cleverness in another library vs. having a non-recursive depth-first traversal algorithm here in fromager.

rd4398:
> I have to admit, this is a small clean change but I really had to study the `trampoline` code to make sense of how it works. I'm not sure how I feel about hiding the cleverness in another library vs. having a non-recursive depth-first traversal algorithm here in fromager.

Yeah, I agree. I wonder whether we want to move back to implementing our plan for iterative bootstrap even though it is a significantly big change. @LalatenduMohanty what do you think? cc @tiran

src/fromager/bootstrapper.py
rd4398: `_get_version_from_package_metadata` is not part of the recursive chain — it's a regular method called from `_resolve_version_from_git_url` → resolve_versions, which are also regular methods. If we used yield here, we'd have to  convert _get_version_from_package_metadata, then _resolve_version_from_git_url, then resolve_versions, and then every caller of resolve_versions into generators too — cascading the conversion well beyond the recursive chain.
                                                                                                                                                                                                                                    
The local `trampoline()` call keeps the change contained: it runs `_prepare_build_dependencies` (now a generator) synchronously to completion and returns the set[Requirement] result, without forcing the resolution call chain to  become generators.

src/fromager/bootstrapper.py
rd4398: I think we cannot be more specific because of the way the way `trampoline` works

```
- YieldType — we yield child generators, but each yield site yields a different generator type, so a single precise type isn't possible within one method 
                                                                 
- SendType — the value received from yield depends on which child generator was yielded (e.g., SourceBuildResult from _build_from_source, None from _bootstrap_gen)

```

src/fromager/bootstrapper.py
dhellmann: That all makes sense, thanks for explaining.

src/fromager/bootstrapper.py
dhellmann: We can still declare our type accurately, even if trampoline doesn't care, can't we?

dhellmann:
> > I have to admit, this is a small clean change but I really had to study the `trampoline` code to make sense of how it works. I'm not sure how I feel about hiding the cleverness in another library vs. having a non-recursive depth-first traversal algorithm here in fromager.
> 
> Yeah, I agree. I wonder whether we want to move back to implementing our plan for iterative bootstrap even though it is a significantly big change. @LalatenduMohanty what do you think? cc @tiran

I'd be interested in seeing the change to compare them for readability.

rd4398:
> > > I have to admit, this is a small clean change but I really had to study the `trampoline` code to make sense of how it works. I'm not sure how I feel about hiding the cleverness in another library vs. having a non-recursive depth-first traversal algorithm here in fromager.
> > 
> > 
> > Yeah, I agree. I wonder whether we want to move back to implementing our plan for iterative bootstrap even though it is a significantly big change. @LalatenduMohanty what do you think? cc @tiran
> 
> I'd be interested in seeing the change to compare them for readability.

Ack, will submit the PR today with that implementation.

rd4398:
> > > > I have to admit, this is a small clean change but I really had to study the `trampoline` code to make sense of how it works. I'm not sure how I feel about hiding the cleverness in another library vs. having a non-recursive depth-first traversal algorithm here in fromager.
> > > 
> > > 
> > > Yeah, I agree. I wonder whether we want to move back to implementing our plan for iterative bootstrap even though it is a significantly big change. @LalatenduMohanty what do you think? cc @tiran
> > 
> > 
> > I'd be interested in seeing the change to compare them for readability.
> 
> Ack, will submit the PR today with that implementation.

Done. @dhellmann @LalatenduMohanty https://github.com/python-wheel-build/fromager/pull/1116 is the alternative implementation. Please take a look when you have time

LalatenduMohanty:
@rd4398 did we try with inlining trampoline logic directly into bootstrap() without depending on the external lib?

rd4398:
> @rd4398 did we try with inlining trampoline logic directly into bootstrap() without depending on the external lib?

@LalatenduMohanty no, we did not. We wanted to compare both the implementations so I drafted https://github.com/python-wheel-build/fromager/pull/1116 and looks like that is the one we want to go with

LalatenduMohanty:
@dhellmann @tiran do you have any preference around what implementation we should use i.e. `inlining trampoline logic directly into bootstrap() without depending on the external lib` VS https://github.com/python-wheel-build/fromager/pull/1116

dhellmann:
> @dhellmann @tiran do you have any preference around what implementation we should use i.e. `inlining trampoline logic directly into bootstrap() without depending on the external lib` VS #1116

The issue I had with trampoline wasn't that it was a dependency, it was that you _really_ have to understand iterators in order to make any sense of the actual implementation. When @rd4398 put together the other PR, I spent an hour or so looking at it and the trampoline code, and there was just too much reliance on Python language internals for me to be comfortable maintaining it long term. So while the code change was smaller, the result was less readable.

This version has all of the code you need right here in the repo, doesn't rely on any special behaviors of python, and relies on logic I learned in my first year college course (which was long enough ago that I hope it's no longer even college-level material these days).

rd4398:
I am closing this PR in favor of https://github.com/python-wheel-build/fromager/pull/1116

PR #1116, opened 2026-05-05
refactor(bootstrapper): convert recursive bootstrap to iterative proessing

### LINKED ISSUE #1068: Convert Bootstrapper from Recursive to Iterative

rd4398:
Replace the recursive DFS in Bootstrapper.bootstrap() with an explicit LIFO stack and a six-phase state machine (START, PREPARE_SOURCE, PREPARE_BUILD, BUILD, PROCESS_INSTALL_DEPS, COMPLETE). This eliminates Python stack overflow on deep/wide dependency graphs, especially with --multiple-versions enabled.

Key changes:
- Add BootstrapPhase enum and WorkItem dataclass for phase-based processing
- Add six phase handler methods that replace the monolithic _bootstrap_impl
- Rewrite bootstrap() to use an iterative DFS loop with explicit work stack
- Remove _bootstrap_single_version, _bootstrap_impl, _build_from_source
- Preserve all three error handling modes (normal, test, multiple-versions)
- Keep _prepare_build_dependencies for git URL resolution path only
- Update tests to match new internal structure

Closes: https://github.com/python-wheel-build/fromager/issues/1068

### DIFF
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -12,6 +12,7 @@ import shutil
 import tempfile
 import typing
 import zipfile
+from enum import StrEnum
 from urllib.parse import urlparse
 
 import requests.exceptions
@@ -47,9 +48,10 @@ SeenKey = tuple[NormalizedName, tuple[str, ...], str, typing.Literal["sdist", "w
 
 @dataclasses.dataclass
 class SourceBuildResult:
-    """Result of building a package from source.
+    """Result of building or downloading a package.
 
-    Used to return multiple values from _build_from_source().
+    Captures the output artifacts from either a source build or
+    prebuilt wheel download, used across bootstrap phases.
     """
 
     wheel_filename: pathlib.Path | None
@@ -80,6 +82,66 @@ class FailureRecord(typing.TypedDict):
     exception_type: str
     exception_message: str
     failure_type: FailureType
+
+
+class BootstrapPhase(StrEnum):
+    """Processing phases for iterative bootstrap.
+
+    All packages: RESOLVE -> START -> ...
+    Source packages: ... -> PREPARE_SOURCE -> PREPARE_BUILD -> BUILD
+                     -> PROCESS_INSTALL_DEPS -> COMPLETE.
+    Prebuilt packages: ... -> PREPARE_SOURCE -> PROCESS_INSTALL_DEPS -> COMPLETE.
+    """
+
+    RESOLVE = "resolve"
+    START = "start"
+    PREPARE_SOURCE = "prepare-source"
+    PREPARE_BUILD = "prepare-build"
+    BUILD = "build"
+    PROCESS_INSTALL_DEPS = "process-install-deps"
+    COMPLETE = "complete"
+
+    @property
+    def tracks_why(self) -> bool:
+        """Whether this phase pushes onto the dependency-chain (why) stack."""
+        return self not in (BootstrapPhase.RESOLVE, BootstrapPhase.START)
+
+
+@dataclasses.dataclass
+class WorkItem:
+    """A unit of work in the iterative bootstrap loop.
+
+    Carries identity fields set at creation time and accumulated state
+    populated across phases as processing advances.
+
+    Items enter at the RESOLVE phase with only req and req_type set.
+    The RESOLVE phase populates source_url and resolved_version, then
+    creates new items at the START phase for each resolved version.
+    """
+
+    # Identity (set at creation)
+    req: Requirement
+    req_type: RequirementType
+    phase: BootstrapPhase
+    why_snapshot: list[tuple[RequirementType, Requirement, Version]]
+    parent: tuple[Requirement, Version] | None = None
+
+    # Populated by RESOLVE phase (None until then)
+    source_url: str | None = None
+    resolved_version: Version | None = None
+
+    build_sdist_only: bool = False
+
+    # Accumulated state (populated during phases)
+    build_env: build_environment.BuildEnvironment | None = None
+    sdist_root_dir: pathlib.Path | None = None
+    unpack_dir: pathlib.Path | None = None
+    cached_wheel_filename: pathlib.Path | None = None
+    build_result: SourceBuildResult | None = None
+    pbi_pre_built: bool = False
+    build_system_deps: set[Requirement] = dataclasses.field(default_factory=set)
+    build_backend_deps: set[Requirement] = dataclasses.field(default_factory=set)
+    build_sdist_deps: set[Requirement] = dataclasses.field(default_factory=set)
 
 
 class Bootstrapper:
@@ -272,10 +334,10 @@ class Bootstrapper:
         return False
 
     def bootstrap(self, req: Requirement, req_type: RequirementType) -> None:
-        """Bootstrap a package and its dependencies.
+        """Bootstrap a package and its dependencies using an iterative loop.
 
-        Handles setup, validation, and error handling. Delegates actual build
-        work to _bootstrap_impl().
+        Uses an explicit LIFO stack instead of recursion to handle arbitrarily
+        deep dependency graphs without hitting Python's recursion limit.
 
         In test mode, catches build exceptions, records package name, and continues.
         In normal mode, raises exceptions immediately (fail-fast).
@@ -285,29 +347,54 @@ class Bootstrapper:
         """
         logger.info(f"bootstrapping {req} as {req_type} dependency of {self.why[-1:]}")
 
-        # Resolve versions - get all if multiple_versions mode is enabled, else get highest
-        # In test mode, record resolution failures and continue.
-        try:
-            resolved_versions = self.resolve_versions(
+        # Capture parent from current why stack before creating work items
+        parent: tuple[Requirement, Version] | None = None
+        if self.why:
+            _, parent_req, parent_version = self.why[-1]
+            parent = (parent_req, parent_version)
+
+        # Save the why stack so we can restore it after the iterative loop
+        # (the loop modifies self.why for each work item)
+        saved_why = list(self.why)
+
+        # Single RESOLVE item — resolution, version expansion, and error
+        # handling all happen inside the loop via _phase_resolve.
+        stack: list[WorkItem] = [
+            WorkItem(
                 req=req,
                 req_type=req_type,
-                return_all_versions=self.multiple_versions,
+                phase=BootstrapPhase.RESOLVE,
+                why_snapshot=list(self.why),
+                parent=parent,
             )
-            if self.multiple_versions:
-                logger.info(f"resolved {len(resolved_versions)} version(s) for {req}")
-        except Exception as err:
-            if not self.test_mode:
-                raise
-            self._record_test_mode_failure(req, None, err, "resolution")
-            return
+        ]
 
-        # Check if resolution returned no versions
-        if not resolved_versions:
-            raise RuntimeError(f"Could not resolve any versions for {req}")
+        # Main iterative DFS loop
+        while stack:
+            item = stack.pop()
+            self.why = list(item.why_snapshot)
 
-        # Bootstrap each resolved version
-        for source_url, resolved_version in resolved_versions:
-            self._bootstrap_single_version(req, req_type, source_url, resolved_version)
+            with req_ctxvar_context(item.req), self._track_why(item):
+                try:
+                    new_items = self._dispatch_phase(item)
+                except Exception as err:
+                    new_items = self._handle_phase_error(item, err)
+
+            # Progress bar: count new RESOLVE-phase items as new dependencies
+            new_dep_count = sum(
+                1 for it in new_items if it.phase == BootstrapPhase.RESOLVE
+            )
+            if new_dep_count > 0:
+                self.progressbar.update_total(new_dep_count)
+            if not new_items:
+                self.progressbar.update()
+
+            # Phase handlers return [continuation, *new_deps] so extend()
+            # naturally puts new deps on top of the stack (processed first).
+            stack.extend(new_items)
+
+        # Restore why stack for the caller
+        self.why = saved_why
 
         # In multiple versions mode, report any failures for this requirement
         if self.multiple_versions and self._failed_versions:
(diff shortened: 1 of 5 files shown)


rd4398:
We have to choose between https://github.com/python-wheel-build/fromager/pull/1109 and this PR to proceed forward. Please take a look and let me know what you think.

src/fromager/bootstrapper.py
```diff
@@ -1473,6 +1297,388 @@ def _add_to_build_order(
...
+            sdist_root_dir=item.sdist_root_dir,
+            build_env=item.build_env,
+        )
+
+        dep_items = self._create_dep_work_items(
+            item.build_backend_deps | item.build_sdist_deps,
+            RequirementType.BUILD_BACKEND,
+            item.req,
+            item.resolved_version,
+        )
```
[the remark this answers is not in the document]
rd4398: fixed in latest commit

src/fromager/bootstrapper.py
```diff
@@ -1473,6 +1297,388 @@ def _add_to_build_order(
...
+        if self.multiple_versions:
+            pkg_name = canonicalize_name(item.req.name)
+            self._failed_versions.append((pkg_name, str(item.resolved_version), err))
+            logger.warning(
+                f"{item.req.name}=={item.resolved_version}: "
+                f"failed to bootstrap: {type(err).__name__}: {err}"
+            )
+            self.ctx.dependency_graph.remove_dependency(pkg_name, item.resolved_version)
+            self.ctx.write_to_graph_to_file()
+            return []
```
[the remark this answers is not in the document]
rd4398: fixed in latest commit

src/fromager/bootstrapper.py
```diff
@@ -82,6 +84,52 @@ class FailureRecord(typing.TypedDict):
...
+
+@dataclasses.dataclass
+class WorkItem:
+    """A unit of work in the iterative bootstrap loop.
+
+    Carries identity fields set at creation time and accumulated state
+    populated across phases as processing advances.
+    """
+
+    # Identity (set at creation)
```
dhellmann, line 122: How important is it that these "identity" values don't change? I wonder if we want to put them in their own data class so we can mark it as frozen, for example. Maybe that's overkill?

src/fromager/bootstrapper.py
```diff
@@ -305,9 +353,75 @@ def bootstrap(self, req: Requirement, req_type: RequirementType) -> None:
...
+        parent: tuple[Requirement, Version] | None = None
+        if self.why:
+            _, parent_req, parent_version = self.why[-1]
+            parent = (parent_req, parent_version)
+
+        # Save the why stack so we can restore it after the iterative loop
+        # (the loop modifies self.why for each work item)
+        saved_why = list(self.why)
+
+        # Create initial work items (reversed so first version is on top of stack)
```
dhellmann: What does "first" mean in this context? Lowest version number?

src/fromager/bootstrapper.py
```diff
@@ -305,9 +353,75 @@ def bootstrap(self, req: Requirement, req_type: RequirementType) -> None:
...
+                    phase=BootstrapPhase.START,
+                    why_snapshot=list(self.why),
+                    parent=parent,
+                )
+            )
+
+        # Main iterative DFS loop
+        while stack:
+            item = stack.pop()
+            self.why = list(item.why_snapshot)
```
dhellmann, line 375: I wonder if instead of resetting `self.why` we could use `item.why_snapshot` in all of those places? Do we have the `item` value everywhere we would need to do that?

src/fromager/bootstrapper.py
```diff
@@ -305,9 +353,75 @@ def bootstrap(self, req: Requirement, req_type: RequirementType) -> None:
...
+
+            with req_ctxvar_context(item.req):
+                try:
+                    if item.phase == BootstrapPhase.START:
+                        # START phase runs outside _track_why to match
+                        # original behavior where seen-check precedes
+                        # pushing onto the why stack.
```
dhellmann: What would be the effect of eliminating that special case?

src/fromager/bootstrapper.py
```diff
@@ -305,9 +353,75 @@ def bootstrap(self, req: Requirement, req_type: RequirementType) -> None:
...
+            # Progress bar: count new START-phase items as new dependencies
+            new_dep_count = sum(
+                1 for it in new_items if it.phase == BootstrapPhase.START
+            )
+            if new_dep_count > 0:
+                self.progressbar.update_total(new_dep_count)
+            if not new_items:
+                self.progressbar.update()
+
+            # Push items so first in list ends up on top (processed first)
```
dhellmann: I think we need to reverse the list here because we're not treating the list in the phase functions as a stack. I would have expected the phase to return something like `[item] + new_items` and that resulting list would then need to have `new_items` processed before trying `item` again. That way `extend()` here just adds to the stack in the same natural order.

src/fromager/bootstrapper.py
```diff
@@ -1473,6 +1297,401 @@ def _add_to_build_order(
...
+                    )
+                )
+        return items
+
+    def _phase_start(self, item: WorkItem) -> list[WorkItem]:
+        """START phase: add to graph, check if already seen.
+
+        Called outside _track_why context to match original behavior
+        where graph addition and seen-check happen before pushing
+        the current item onto the why stack.
```
dhellmann: It would help to explicitly explain what the return list is for each phase function. It's here in this paragraph, but calling it out as something like this would help make the flow more obvious:

```
Returns the current item to push it back onto the stack for the prepare-source phase.
```

src/fromager/bootstrapper.py
```diff
@@ -1473,6 +1297,401 @@ def _add_to_build_order(
...
+
+        dep_items = self._create_dep_work_items(
+            item.build_system_deps,
+            RequirementType.BUILD_SYSTEM,
+            item.req,
+            item.resolved_version,
+        )
+
+        item.phase = BootstrapPhase.PREPARE_BUILD
+        return dep_items + [item]
```
dhellmann: Treat the return value as a stack. We want to process `dep_items` before looking at `item` again.

```suggestion
        return [item] + dep_items
```

src/fromager/bootstrapper.py
```diff
@@ -1473,6 +1297,401 @@ def _add_to_build_order(
...
+        """COMPLETE phase: clean up build directories."""
+        if item.build_result is not None:
+            self.ctx.clean_build_dirs(
+                item.build_result.sdist_root_dir,
+                item.build_result.build_env,
+            )
+        return []
+
+    def _dispatch_phase(self, item: WorkItem) -> list[WorkItem]:
+        """Route a work item to the appropriate phase handler."""
```
dhellmann, line 1654: Do we log the phase anywhere?

src/fromager/bootstrapper.py
```diff
@@ -305,9 +353,75 @@ def bootstrap(self, req: Requirement, req_type: RequirementType) -> None:
...
+            _, parent_req, parent_version = self.why[-1]
+            parent = (parent_req, parent_version)
+
+        # Save the why stack so we can restore it after the iterative loop
+        # (the loop modifies self.why for each work item)
+        saved_why = list(self.why)
+
+        # Create initial work items (reversed so first version is on top of stack)
+        stack: list[WorkItem] = []
+        for source_url, resolved_version in reversed(resolved_versions):
```
dhellmann: What do you think about making resolving the versions a phase, too?

src/fromager/bootstrapper.py
```diff
@@ -1473,6 +1297,401 @@ def _add_to_build_order(
...
+    ) -> list[WorkItem]:
+        """Resolve and create work items for dependencies.
+
+        Called inside a parent's _track_why context so that why_snapshot
+        captures the parent's dependency chain.
+        """
+        items: list[WorkItem] = []
+        for dep in self._sort_requirements(deps):
+            try:
+                resolved = self.resolve_versions(
```
dhellmann: Making the version resolution a phase would mean we would only have this logic in 1 place.

src/fromager/bootstrapper.py
rd4398: I think it's overkill. The truly immutable fields (req, req_type, source_url, resolved_version, parent, why_snapshot) are set once at creation in `_create_dep_work_items` and `bootstrap()`, and no phase handler ever reassigns them. Meanwhile phase and  `build_sdist_only`  look like identity but actually mutate during processing, so they'd need to stay on the outer mutable dataclass anyway.                                                                                                                             

A frozen sub-dataclass would add indirection (item.identity.req everywhere) and a second class to maintain, for protection against a bug that doesn't exist today in a type that's internal to the module.

Adding a frozen class is relatively small change. Let me know if you want me to do it

src/fromager/bootstrapper.py
rd4398: "First" here means highest version — `resolve_versions` returns results sorted highest-first (line 248)
I will add a comment to explicitly state that

src/fromager/bootstrapper.py
rd4398: ~~No, even I thought of that initially fwiw.~~
No, we don't have the item value everywhere we would need to do that.  I looked at it but:

`self.why` is read by too many methods that don't have access to the item:                                                                                                                                                                         
                                                                                                                                                                                                                                                                       
  - `_processing_build_requirement()` iterates self.why (called in _phase_start)                                                                                                                                                                                         
  - `_explain` property reads `self.why` (called in _phase_start)
  - `resolve_versions()` reads `self.why[-1]` for parent context (called from _create_dep_work_items inside phase handlers)                                                                                                                                                
  - `_handle_test_mode_failure()` reads `self.why[-1]` (called from _handle_phase_error)                                                                                                                                                                                   
  - `_track_why()` appends/pops on `self.why`                                                                                                                                                                                                                              
  - `_create_dep_work_items()` captures `list(self.why)` for new items' snapshots                                                                                                                                                                                          
                                                                                                                                                                                                                                                                       
  To use `item.why_snapshot` we would need to change the signatures of `_processing_build_requirement`, `_explain`, `resolve_versions`,` _handle_test_mode_failure`,` _track_why`, and `_create_dep_work_items`. Several of those (resolve_versions, _track_why) are also used outside the iterative loop (e.g. resolve_and_add_top_level, the git URL path).

It is a more complicated refactor. Let me know if you think we should do it

src/fromager/bootstrapper.py
rd4398: Two things would change if START ran inside _track_why like every other phase:                                                                                                                                                                                       
                                                     
  1.` _explain` would include the current item in the "redundant dependency" log message. Currently when a dep is already seen, it logs "redundant ... for {self._explain}" where _explain shows the parent chain. With `_track_why`, the current item would be on the stack too, so the message would include itself — slightly odd but not incorrect.                                                                                                                                                                                     
  2. `_processing_build_requirement` would see the current item's `req_type` in `self.why`. It walks self.why in reverse looking for build requirement types. If the current item is itself a BUILD_SYSTEM dep, it already returns True from the direct  `current_req_type.is_build_requirement` check before consulting self.why, so no behavior change there. But if the current item is an INSTALL dep whose parent is a build dep, the extra entry on the stack wouldn't affect the result since the method checks reversed(self.why) and the current item's INSTALL type would just be skipped.

src/fromager/bootstrapper.py
dhellmann: OK, it sounds like you're confident it's not needed and that makes sense.

src/fromager/bootstrapper.py
dhellmann: OK, it sounds like it would be pretty invasive, so let's not worry about it.

src/fromager/bootstrapper.py
dhellmann: OK, that sounds like it might be tricky to get right, so let's leave it like this. Maybe we can refine it further in the future.

src/fromager/bootstrapper.py
rd4398: I considered this when claude was making changes. The main issue is that resolution is one-to-many in multiple_versions mode: one unresolved requirement can produce N resolved versions, each needing its own work item. Every other phase is one-to-one (item advances) or  one-to-one-plus-deps (item advances and spawns children). A RESOLVE phase would be the only one that replaces itself with multiple peer items — a different pattern that complicates the state machine.                                                              
                                                                                                                                                                                                                                                                       
It would also make source_url and resolved_version Optional on WorkItem, adding None-guard noise in every subsequent phase. Right now those fields are always populated at creation time, which is a nice invariant.                                                 
                                                                                                                                                                                                                                                                       
The current split feels natural: bootstrap() resolves the top-level entry point, and _create_dep_work_items resolves each dependency inline when the parent discovers it. Resolution stays close to where deps are discovered rather than being deferred to a later loop iteration. 

I am inclined to leave it as it is unless you prefer making it a separate state

src/fromager/bootstrapper.py
rd4398: Yeah, agreed. answered in above comment

src/fromager/bootstrapper.py
rd4398: We were not logging it. I have made changes in latest commit

src/fromager/bootstrapper.py
rd4398: Fixed in latest commit

src/fromager/bootstrapper.py
rd4398: Sure, I have made changes in latest commit.

src/fromager/bootstrapper.py
rd4398: Hmm, I agree. I have made changes in latest commit

src/fromager/bootstrapper.py
dhellmann: We would just be moving the creation of the work items that's happening here inside a phase.

What I'm thinking is that a work item would start in the RESOLVE phase and that phase returns a stack containing the original work item and new work items for all of the resolved versions. The original work item would have its next phase set to COMPLETE, because we don't need to put it through all of the other stages but we want it to stay in the stack until its versions have been processed. The individually resolved items would have their next phase set to what is currently called START because those don't need to go through the resolution process. So, all of the existing steps would still receive resolved_version and source_url values. 

The benefit of having that logic in one place means that all new dependencies we encounter get "expanded" the same way and that means when we add the logic to filter out versions that already exist in the package index or on disk or whatever, we can do it _there_ one time.

dhellmann: Another benefit of doing it on the stack is when we eventually start trying to parallelize some of these things we can have code look at all of the RESOLVE tasks on the stack and run them in threads, feeding the results back to the right spots within the stack.

rd4398: Okay, I like this idea. I will spend some time on it and push changes

rd4398: Done, I have pushed changes in latest commit. We now have a RESOLVE phase and WorkItem starts in that phase.

src/fromager/bootstrapper.py
```diff
@@ -285,29 +342,85 @@ def bootstrap(self, req: Requirement, req_type: RequirementType) -> None:
...
-        if not resolved_versions:
-            raise RuntimeError(f"Could not resolve any versions for {req}")
+        # Main iterative DFS loop
+        while stack:
+            item = stack.pop()
+            self.why = list(item.why_snapshot)
+
+            with req_ctxvar_context(item.req):
+                try:
+                    if item.phase in (
```
dhellmann: nit: Maybe we could do something like [this property](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/requirements_file.py#L22) to encode the logic in the enum.

Alternatively, we could pass item directly to `_track_why()` and it could just not make any changes for those types. That would simplify the logic here.

Neither of those changes should hold this up.

src/fromager/bootstrapper.py
```diff
@@ -285,29 +342,85 @@ def bootstrap(self, req: Requirement, req_type: RequirementType) -> None:
...
+                except Exception as err:
+                    # For phases that run inside _track_why, restore the
+                    # context for _handle_phase_error (needed by
+                    # _handle_test_mode_failure).
+                    if item.phase not in (
+                        BootstrapPhase.RESOLVE,
+                        BootstrapPhase.START,
+                    ):
+                        assert item.resolved_version is not None
+                        self.why.append(
```
dhellmann: I wonder if this should be a `with` statement, too?

src/fromager/bootstrapper.py
```diff
@@ -1473,6 +1296,470 @@ def _add_to_build_order(
...
+            raise RuntimeError(f"Could not resolve any versions for {item.req}")
+
+        if self.multiple_versions:
+            logger.info(f"resolved {len(resolved_versions)} version(s) for {item.req}")
+
+        # Build list so highest version ends up on top of the stack
+        # (last element after extend) and is processed first.
+        items: list[WorkItem] = []
+        for source_url, version in reversed(resolved_versions):
+            items.append(
```
dhellmann, line 1327: Could we use `_create_unresolved_work_items` here, too?

dhellmann (commented):
I have a couple of nits/cleanups but nothing to hold this up.

I like this approach because it doesn't rely on "magic". Let's get some other reviewers to look at it.

src/fromager/bootstrapper.py
rd4398: I don't think we can.

`_create_unresolved_work_items` creates RESOLVE-phase items from a list of Requirement objects. `_phase_resolve` creates START-phase items from a list of           
  (source_url, Version) tuples. Merging them would require overloading the helper with optional parameters that defeats its purpose. 

Let me know if I should do it anyways

src/fromager/bootstrapper.py
rd4398: Done in latest push

src/fromager/bootstrapper.py
rd4398: I like the idea of using enum. I have done it in latest push

rd4398:
> I have a couple of nits/cleanups but nothing to hold this up.
> 
> I like this approach because it doesn't rely on "magic". Let's get some other reviewers to look at it.

ack, sounds good. I have added @LalatenduMohanty, @ryanpetrello, @EmilienM and @tiran as well for reviews

LalatenduMohanty (changes requested):
This PR introduces 14 new methods/classes (BootstrapPhase, WorkItem, _phase_resolve, _phase_start, _phase_prepare_source, _phase_prepare_build,
  _phase_process_install_deps, _phase_complete, _dispatch_phase, _handle_phase_error, _create_unresolved_work_items, _track_why refactor, iterative bootstrap() loop). Only 2
   existing tests were updated. No new tests were added.

We should add more tests as this is a significant change. 

At minimum, add tests for:
  1. _handle_phase_error — each branch (RESOLVE in test_mode, build-phase test_mode with fallback success/failure, multiple_versions mode with seen cleanup, normal mode
  raise)
  2. _phase_start — already-seen returns [], new work advances phase
  3. _track_why — no-op for RESOLVE/START, pushes/pops for other phases
  4. A small integration test that drives bootstrap() through multiple phases with mocked externals to verify the LIFO ordering produces correct results

LalatenduMohanty:
@rd4398 Incase you want to all good test coverage, here are some pointers

  - Phase transitions: No test verifies the full lifecycle RESOLVE -> START -> PREPARE_SOURCE -> PREPARE_BUILD -> BUILD -> PROCESS_INSTALL_DEPS -> COMPLETE
  - _phase_resolve: Untested (single version, multiple versions, empty resolution)
  - _phase_start: Untested (graph addition, seen-check short-circuit, sdist_only logic)
  - _phase_prepare_source: Untested (prebuilt path, source path, cached wheel path)
  - _phase_prepare_build: Untested (build system dep installation, backend/sdist dep extraction)
  - _phase_process_install_deps: Untested (hook execution, dependency extraction, build-order writing)
  - _phase_complete: Untested (cleanup)
  - _dispatch_phase: Untested (routing, unknown phase error)
  - _handle_phase_error: Untested — this is the most critical gap; it contains logic for test-mode fallback, multiple-versions cleanup, and normal-mode fail-fast across
  different phases
  - _create_unresolved_work_items: Untested (why_snapshot capture, parent propagation)
  - _track_why: Only exercised indirectly inside test_phase_build; no direct test of the no-op path (RESOLVE/START) or the push/pop behavior
  - BootstrapPhase.tracks_why: Untested
  - WorkItem dataclass: No tests for field defaults or state accumulation

rd4398:
> @rd4398 Incase you want to all good test coverage, here are some pointers
> 
> * Phase transitions: No test verifies the full lifecycle RESOLVE -> START -> PREPARE_SOURCE -> PREPARE_BUILD -> BUILD -> PROCESS_INSTALL_DEPS -> COMPLETE
> * _phase_resolve: Untested (single version, multiple versions, empty resolution)
> * _phase_start: Untested (graph addition, seen-check short-circuit, sdist_only logic)
> * _phase_prepare_source: Untested (prebuilt path, source path, cached wheel path)
> * _phase_prepare_build: Untested (build system dep installation, backend/sdist dep extraction)
> * _phase_process_install_deps: Untested (hook execution, dependency extraction, build-order writing)
> * _phase_complete: Untested (cleanup)
> * _dispatch_phase: Untested (routing, unknown phase error)
> * _handle_phase_error: Untested — this is the most critical gap; it contains logic for test-mode fallback, multiple-versions cleanup, and normal-mode fail-fast across
>   different phases
> * _create_unresolved_work_items: Untested (why_snapshot capture, parent propagation)
> * _track_why: Only exercised indirectly inside test_phase_build; no direct test of the no-op path (RESOLVE/START) or the push/pop behavior
> * BootstrapPhase.tracks_why: Untested
> * WorkItem dataclass: No tests for field defaults or state accumulation

@LalatenduMohanty thank you! I have added tests in latest commit

src/fromager/bootstrapper.py
rd4398: Resolving since it was done

src/fromager/bootstrapper.py
rd4398: Resolving since it is done

src/fromager/bootstrapper.py
rd4398: Resolving since it is done

src/fromager/bootstrapper.py
rd4398: Resolving as is is done

src/fromager/bootstrapper.py
rd4398: I am resolving this for now. I will open a follow-up if required.

tests/test_bootstrapper_iterative.py
```diff
@@ -0,0 +1,823 @@
...
+            BootstrapPhase.PREPARE_SOURCE,
+            BootstrapPhase.PREPARE_BUILD,
+            BootstrapPhase.BUILD,
+            BootstrapPhase.PROCESS_INSTALL_DEPS,
+            BootstrapPhase.COMPLETE,
+        ):
+            assert phase.tracks_why is True, f"{phase} should track why"
+
+
+# ---- WorkItem ----
```
LalatenduMohanty: nit picks: Do we need these separators?

rd4398: Not required, but they help to separate tests. Let me remove them

rd4398: Done

LalatenduMohanty (approved):

LalatenduMohanty:
Created https://github.com/python-wheel-build/fromager/issues/1137 as a follow up ticket to further improve the tests.

PR #1121, opened 2026-05-06
fix(resolver): make GitHubTagProvider server URL configurable

### LINKED ISSUE #1119: Make GitHubTagProvider server URL configurable to enable mock testing
## Problem

`GitHubTagProvider` hardcodes its API endpoint as a class variable:

```python
host = "github.com:443"
api_url = "https://api.{self.host}/repos/{self.organization}/{self.repo}/tags"
```

This can't be overridden per-instance, so every use hits the real GitHub API. In CI, fork PRs can't use `GITHUB_TOKEN` (GitHub security restriction), leaving tests subject to the shared 60 req/hr unauthenticated rate limit.

`GitLabTagProvider` already supports this — it accepts a `server_url` constructor parameter.

## Proposed Solution

1. Add a `server_url` parameter to `GitHubTagProvider.__init__()`, defaulting to `https://api.github.com`. Derive `api_url` from it, matching the GitLab pattern.

2. Update the `stevedore_github.py` e2e plugin to accept a mock URL (e.g. via env var).

3. Update `test_bootstrap_cooldown_github.sh` to start a local HTTP server serving static tag JSON, using the existing local server pattern from `e2e/common.sh`.

This eliminates ~16 GitHub API calls per CI run (4 paginated pages × 4 parallel `ci_bootstrap_suite` jobs) and makes the test deterministic regardless of GitHub API availability.

## Context

- Rate limiting: 
  - #1118 
  - PR #1117
- Fork PRs cannot use `GITHUB_TOKEN`, so mocking is the only reliable fix for CI
- `GitLabTagProvider` already has `server_url` parameter — this brings parity

## Files

- `src/fromager/resolver.py` — `GitHubTagProvider`
- `e2e/github_override_example/src/package_plugins/stevedore_github.py`
- `e2e/test_bootstrap_cooldown_github.sh`

LalatenduMohanty:
Add a server_url parameter to GitHubTagProvider, matching the existing GitLabTagProvider pattern. The e2e test now uses a local mock server instead of hitting the real GitHub API, avoiding rate-limit failures in CI (especially for fork PRs without GITHUB_TOKEN).

Closes: #1119

### DIFF
--- a/e2e/common.sh
+++ b/e2e/common.sh
@@ -55,42 +55,45 @@ export NETWORK_ISOLATION
 export WHEEL_PLATFORM_TAG
 export HAS_ELFDEP
 
-# Local web server management
-HTTP_SERVER_PID=""
+# Background server management
+_BACKGROUND_PIDS=()
 on_exit() {
-    if [ -n "$HTTP_SERVER_PID" ]; then
-        echo "Stopping wheel server"
-        kill "$HTTP_SERVER_PID"
-    fi
+    # ${arr[@]+...} avoids "unbound variable" under set -u when empty
+    for pid in "${_BACKGROUND_PIDS[@]+"${_BACKGROUND_PIDS[@]}"}"; do
+        kill "$pid" 2>/dev/null || true
+    done
 }
 trap on_exit EXIT SIGINT SIGTERM
 
-start_local_wheel_server() {
-    local serve_dir="${1:-$OUTDIR/wheels-repo}"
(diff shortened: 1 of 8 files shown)


LalatenduMohanty:
Closing — fork PRs do get a read-only GITHUB_TOKEN (confirmed in #1122), which is sufficient for the GitHub API tag reads (5,000 req/hr). The rate-limiting concern is not a real problem, so the mock server approach adds complexity without enough benefit.

Issue #1096, opened 2026-04-29
Support multiple constraints and constraint files


tiran:
Fromager 0.82 and earlier only supports a single constraints file with unique entries. `packaging` 26.1 added a new feature to check if a [SpecifierSet is satisfiable](https://packaging.pypa.io/en/stable/specifiers.html). This feature finally allows us to support multiple constraints and constraints files safely.

```python
>>> from packaging.specifiers import SpecifierSet
>>> SpecifierSet("<1").is_unsatisfiable()
False
>>> SpecifierSet("<1,>2").is_unsatisfiable()
True
```

1. extend `fromager.constraints.Constraints.add_constraint` to combine constraints and verify that the combined constraints are satisfiable.
2. update the global `-c` / `--constraints-file` option to support multiple constraints files

LalatenduMohanty:
@smoparth I would like to see a proposal on this before you implement the actual feature. I will help us agree on the benefit it will bring to downstream usage of fromager.

LalatenduMohanty:
For example how fromager support of multiple constraint will look, is it going to be something like this  `fromager -c global.txt -c collection.txt -c variant.txt bootstrap` 

Lookslike `SpecifierSet` doesn't just look for string mismatches; it understands the numerical logic of [PEP 440](https://peps.python.org/pep-0440/) versions. It detects:

- Disjoint ranges: e.g., >5.0 and <3.0.
- Exact match conflicts: e.g., ==1.0 and ==2.0.
- Exclusion conflicts: e.g., ==1.2 and !=1.2.

tiran:
> `fromager -c global.txt -c collection.txt -c variant.txt`

Yes, it is as simple as adding `multiple=True` to the click parameter and change the parameter from `str` to `tuple[str]`.

smoparth:
Makes sense, happy to write a proposal. I'll cover:

The downstream benefit I see -- multiple constraint sources over manual merging.
Key design decisions that need agreement: error type change (KeyError → ValueError), packaging>=26.1 pin, and whether to pass multiple --constraint files to pip vs a single merged file but I thought these were more of implementation decisions than architecture.

I don't see a proposal template in the repo, not a need of the hour but we can have one to standardize the proposals. I'll follow existing proposal formats and draft one.

tiran:
IMHO we don't need a design proposal for this feature. It's a simple improvement and we discuss the details in this ticket.

[...]

PR #1125, opened 2026-05-07
feat: combine and validate constraints

tiran:
# Pull Request Description

## What

Fromager's `Constraints` class now supports combining multiple constraints and performs stricter validation of constraints. The class now refuses constraints without a version specifier, extras, or an URL. It packaging's new `SpecifierSet.is_unsatisfiable()` feature to detect constraints that are not satisfiable.

Example:
The constraints `foo>=1.0`, `foo<2.0`, and `foo!=1.1.0` are combined into to `foo<2.0,>=1.0,!=1.1.0`. A combination of `foo<1.0` and `foo>=2.0` would result in a `InvalidConstraintError` instead.

## Why

#1096

### DIFF
--- a/src/fromager/constraints.py
+++ b/src/fromager/constraints.py
@@ -11,6 +11,10 @@ from . import requirements_file
 logger = logging.getLogger(__name__)
 
 
+class InvalidConstraintError(ValueError):
+    pass
+
+
 class Constraints:
     def __init__(self) -> None:
         # mapping of canonical names to requirements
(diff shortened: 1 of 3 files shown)


tests/test_constraints.py
```diff
@@ -36,28 +36,24 @@ def test_constraint_not_is_satisfied_by() -> None:
...
         c.add_constraint("flit_core>2.0.0")
 
     # Normalized name conflict should raise error
-    with pytest.raises(KeyError):
+    with pytest.raises(InvalidConstraintError):
         c.add_constraint("flit-core>2.0.0")
 
     # Different, but equivalent markers should raise KeyError
-    with pytest.raises(KeyError):
+    with pytest.raises(InvalidConstraintError):
```
smoparth, line 77: Pre-existing minor issue: test_add_constraint_conflict is not portable across platforms

This test will fail on platforms like ppc64le or s390x (or Python < 3.11), because marker evaluation happens before the combining logic and if none of the three architectures match the test runner's platform_machine, every constraint is silently dropped and the conflict is never reached.

We can add a skip guard to the test.
@pytest.mark.skipif(
    platform.machine() not in ("x86_64", "arm64", "aarch64") or sys.version_info < (3, 11),
    reason="Marker-based conflict test requires x86_64/arm64/aarch64 and Python >= 3.11"
)

smoparth: No need to address here, but curious if others think a pytest.mark.skipif guard would be worth adding. I know it's rare to have test runners other than x86_64/arm64/aarch64.

smoparth (dismissed):
LGTM!

src/fromager/constraints.py
```diff
@@ -11,6 +11,10 @@
 logger = logging.getLogger(__name__)
 
 
+class InvalidConstraintError(ValueError):
+    pass
```
LalatenduMohanty, line 15: We should add doc string for this class. It would give better context around why we are adding it.

tests/test_constraints.py
tiran: There is a simpler and better solution: `mock.patch`. I'll update the PR.

tiran: I have modified the test case. It now mocks `platform.machine` so it returns a fixed value on all machines.

smoparth (dismissed):

src/fromager/constraints.py
```diff
@@ -20,23 +24,50 @@ def __init__(self) -> None:
...
             logger.debug(f"Constraint {req} does not match environment")
             return
 
         if previous is not None:
-            raise KeyError(
-                f"{canon_name}: new constraint '{req}' conflicts with '{previous}'"
-            )
-        if requirements_file.evaluate_marker(req, req):
+            logger.debug("combining constraints %s and %s", previous, req)
+            new_specifier = req.specifier & previous.specifier
```
rd4398, line 60: When combining constraints, req (the new one) is stored, which means the previous constraint's marker is silently dropped. So something like:

```
# If previous had: foo>=1.0; python_version >= '3.11'                                                                       
  # And new is:      foo<2.0                                                                                                  
  # Result stored:   foo<2.0,>=1.0     (marker lost)
```
Is this intentional? Or am I missing something?

rd4398 (commented):
Looks overall good! I had a clarification question about markers

src/fromager/constraints.py
tiran: Does it matter? The `add_constraints` function evaluates markers and ignores markers that don't match the environment.

rd4398: Right, I missed that part. Thank you for clarifying!

rd4398 (approved):
This looks good!

Also, the old code had a redundant second `evaluate_marker` which this PR correctly removes. This is a subtle bugfix — the old code had a dead if branch.

PR #1126, opened 2026-05-07
feat: add protocol for override hooks

tiran:
# Pull Request Description

## What

OverrideHookProtocol documents the interface for per-package override hooks and provides runtime validation of hook signatures via check_signature(). The lint command now checks override hook signatures.

## Why

### DIFF
--- a/src/fromager/overrides.py
+++ b/src/fromager/overrides.py
@@ -1,3 +1,6 @@
+from __future__ import annotations
+
+import importlib
 import inspect
 import logging
 import pathlib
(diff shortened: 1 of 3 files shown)


src/fromager/overrides.py
dhellmann, line 1: Another way to do this would be to have fromager define a type so we could pass fewer individual arguments to hooks. For example, the build hooks both take the same arguments. We could define a type and pass 1 value to the hook. Then when we add fields, the hooks don't have to change if they don't need the value.

Maybe the methods that manipulate the source (download, prepare, etc.) could be standardized to take a single type, too?

How many different types would we need?

PR #1127, opened 2026-05-07
feat(list-versions): show cooldown status and upload timestamps

### LINKED ISSUE #1078: list-versions: show cooldown status and upload timestamps
## Summary

When release-age cooldown is active (`--min-release-age`), `list-versions` should display upload timestamps and indicate which candidates are blocked by the cooldown policy.

## Design

See Phase 2 in the [release cooldown proposal](https://github.com/python-wheel-build/fromager/blob/main/docs/proposals/release-cooldown.md).

## Requirements

- Show `upload_time` and age (days) for each candidate
- Mark candidates blocked by the cooldown
- Add `--ignore-per-package-overrides` flag to show what the cooldown would block without per-package exemptions

Ref: #877

shifa-khan:
package list-versions now shows upload_time, age in days, and cooldown status for each version.

--format controls the output: versions (default) and requirements print a filtered list excluding blocked entries; table, csv, and json include the full detail columns. This replaces the previous --format-as-requirements flag.

--ignore-per-package-overrides shows what the global --min-release-age would block without per-package exemptions.

Closes: #1078

### DIFF
--- a/src/fromager/commands/package.py
+++ b/src/fromager/commands/package.py
@@ -1,17 +1,31 @@
+import csv
 import datetime
 import enum
+import json
 import logging
+import pathlib
 import sys
 import typing
+from collections import defaultdict
 
 import click
 import pypi_simple
+import rich
 from packaging.requirements import Requirement
 from packaging.version import Version
 from resolvelib.resolvers import ResolverException
+from rich.table import Table
 
-from .. import context, log, overrides, packagesettings, request_session, resolver
-from ..candidate import Candidate
+from .. import (
+    clickext,
+    context,
+    log,
+    overrides,
+    packagesettings,
+    request_session,
+    resolver,
+)
+from ..candidate import Candidate, Cooldown
 
 logger = logging.getLogger(__name__)
 
@@ -94,9 +108,29 @@ def package() -> None:
     help="Do not treat missing versions as an error",
 )
 @click.option(
-    "--format-as-requirements/--no-format-as-requirements",
+    "--format",
+    "output_format",
+    type=click.Choice(
+        ["versions", "requirements", "table", "csv", "json"],
+        case_sensitive=False,
+    ),
+    default="versions",
+    help="Output format (default: versions)",
+)
+@click.option(
+    "-o",
+    "--output",
+    type=clickext.ClickPath(),
+    help="Output file (default: stdout)",
+)
+@click.option(
+    "--ignore-per-package-overrides",
+    is_flag=True,
     default=False,
-    help="Format output as requirement specifiers (name==version) instead of just version numbers",
+    help=(
+        "Ignore per-package min_release_age overrides when computing cooldown "
+        "status; uses only the global --min-release-age value."
+    ),
 )
 @click.argument("requirement_spec", required=True)
 @click.pass_obj
@@ -106,7 +140,9 @@ def list_versions(
     distribution_type: str,
     sdist_server_url: str,
     ignore_no_versions: bool,
-    format_as_requirements: bool,
+    output_format: str,
+    output: pathlib.Path | None,
+    ignore_per_package_overrides: bool,
 ) -> None:
     """List all available versions for a package requirement specifier.
 
@@ -123,7 +159,18 @@ def list_versions(
     - "sdist": Only include source distributions
     - "wheel": Only include wheels
     - "both": Include both source distributions and wheels
+
+    Output formats:
+    - "versions": one version per line (default)
+    - "requirements": name==version per line (pip-installable pins)
+    - "table": Rich table with upload timestamps, age, and cooldown status
+    - "csv": CSV with the same detail columns
+    - "json": JSON array with the same detail columns
+
+    Use --ignore-per-package-overrides to see what the global cooldown
+    policy would block without per-package exemptions.
     """
+
     try:
         req = Requirement(requirement_spec)
     except Exception as e:
@@ -154,7 +201,8 @@ def list_versions(
         sdist_server_url=override_sdist_server_url,
     )
 
-    # Get all available candidates from the provider
+    # Get all available candidates from the provider (cooldown is NOT set on
+    # the provider so we receive every version that matches the specifier).
     candidates = list(
         provider.find_matches(
             identifier=req.name,
@@ -170,14 +218,213 @@ def list_versions(
         else:
             raise click.ClickException(f"No versions found for {req.name}")
 
-    versions: list[Version] = sorted(set(candidate.version for candidate in candidates))
-    logger.info(f"Found {len(versions)} version(s)")
+    logger.info(f"Found {len(set(c.version for c in candidates))} version(s)")
 
-    for version in versions:
-        if format_as_requirements:
-            print(f"{req.name}=={version}")
+    cooldown = _resolve_list_versions_cooldown(wkctx, req, ignore_per_package_overrides)
+    version_rows = _compute_version_details(
+        req.name,
+        candidates,
+        cooldown,
+        provider.supports_upload_time,
+    )
+
+    if output is not None and output_format in ("versions", "requirements"):
+        click.echo(
+            "Warning: --output option is ignored for 'versions' and 'requirements' formats",
+            err=True,
+        )
+
+    match output_format:
+        case "versions":
+            _export_versions_plain(version_rows, req.name, cooldown)
+        case "requirements":
+            _export_versions_plain(
+                version_rows, req.name, cooldown, as_requirements=True
+            )
+        case "table":
+            _export_versions_table(version_rows, req.name, cooldown, output)
+        case "csv":
+            _export_versions_csv(version_rows, output)
+        case "json":
+            _export_versions_json(version_rows, output)
+        case _:
+            raise ValueError(f"Invalid output format: {output_format}")
+
+
+def _resolve_list_versions_cooldown(
+    wkctx: context.WorkContext,
+    req: Requirement,
+    ignore_per_package_overrides: bool,
+) -> Cooldown | None:
+    """Determine the effective cooldown for the list-versions detail view.
+
+    When *ignore_per_package_overrides* is ``True``, only the global
+    ``--min-release-age`` value is used so the caller can audit what the
+    policy would block without per-package exemptions.
+    """
+    if ignore_per_package_overrides:
+        return wkctx.cooldown
+    return resolver.resolve_package_cooldown(wkctx, req)
+
+
+def _compute_version_details(
+    package_name: str,
+    candidates: list[Candidate],
+    cooldown: Cooldown | None,
+    supports_upload_time: bool,
+) -> list[dict[str, str]]:
+    """Group candidates by version and compute cooldown status.
+
+    Returns one row per version, sorted ascending.  Each row is a dict with
+    keys: ``package``, ``version``, ``upload_time``, ``age_days``,
+    ``cooldown``.
+    """
+    by_version: dict[Version, list[Candidate]] = defaultdict(list)
+    for c in candidates:
+        by_version[c.version].append(c)
+
+    reference_time = (
+        cooldown.bootstrap_time
+        if cooldown is not None
+        else datetime.datetime.now(datetime.UTC)
+    )
+
+    rows: list[dict[str, str]] = []
+    for version in sorted(by_version):
+        version_candidates = by_version[version]
+
+        upload_times = [
+            c.upload_time for c in version_candidates if c.upload_time is not None
+        ]
+        upload_time = max(upload_times) if upload_times else None
+
+        if upload_time is not None:
+            age_days = (reference_time - upload_time).days
         else:
-            print(version)
+            age_days = None
+
+        status = _cooldown_status(upload_time, cooldown, supports_upload_time)
+
+        rows.append(
+            {
+                "package": package_name,
+                "version": str(version),
+                "upload_time": upload_time.strftime("%Y-%m-%d %H:%M")
+                if upload_time
+                else "",
+                "age_days": str(age_days) if age_days is not None else "",
+                "cooldown": status,
+            }
+        )
+    return rows
+
+
+def _cooldown_status(
+    upload_time: datetime.datetime | None,
+    cooldown: Cooldown | None,
+    supports_upload_time: bool,
+) -> str:
+    """Classify cooldown status for a single version.
+
+    Returns one of ``"blocked"``, ``"available"``, ``"skipped"``, or ``""``
+    (no cooldown configured).
+    """
+    if cooldown is None:
+        return ""
+    if upload_time is None:
+        if not supports_upload_time:
+            return "skipped"
+        return "blocked"
+    cutoff = cooldown.bootstrap_time - cooldown.min_age
+    if upload_time > cutoff:
+        return "blocked"
+    return "available"
+
+
+# -- export helpers for list-versions -------------------------------------------
+
+
+def _export_versions_plain(
+    data: list[dict[str, str]],
+    package_name: str,
+    cooldown: Cooldown | None,
+    *,
+    as_requirements: bool = False,
+) -> None:
+    """Export versions as a plain list, filtering out cooldown-blocked entries."""
+    for row in data:
+        if cooldown is not None and row["cooldown"] == "blocked":
+            continue
+        if as_requirements:
+            print(f"{package_name}=={row['version']}")
+        else:
+            print(row["version"])
+
+
+def _export_versions_json(
+    data: list[dict[str, str]], output: pathlib.Path | None
+) -> None:
+    """Export version details as JSON."""
+    if output:
+        with open(output, "w") as outfile:
+            json.dump(data, outfile, indent=2)
+            print(file=outfile)
+    else:
+        json.dump(data, sys.stdout, indent=2)
+        print()
+
+
+_VERSIONS_CSV_FIELDS = ["package", "version", "upload_time", "age_days", "cooldown"]
+
+
+def _export_versions_csv(
+    data: list[dict[str, str]], output: pathlib.Path | None
+) -> None:
+    """Export version details as CSV."""
+    if output:
+        with open(output, "w", newline="") as outfile:
+            writer = csv.DictWriter(
+                outfile,
+                fieldnames=_VERSIONS_CSV_FIELDS,
+                quoting=csv.QUOTE_NONNUMERIC,
+            )
+            writer.writeheader()
+            writer.writerows(data)
+    else:
+        writer = csv.DictWriter(
+            sys.stdout,
+            fieldnames=_VERSIONS_CSV_FIELDS,
+            quoting=csv.QUOTE_NONNUMERIC,
+        )
+        writer.writeheader()
+        writer.writerows(data)
+
+
+def _export_versions_table(
+    data: list[dict[str, str]],
+    package_name: str,
+    cooldown: Cooldown | None,
+    output: pathlib.Path | None = None,
+) -> None:
+    """Export version details as a Rich table."""
+    table = Table(title=f"Versions for {package_name}")
+    table.add_column("Version", justify="left", no_wrap=True)
+    table.add_column("Upload Time", justify="left", no_wrap=True)
+    table.add_column("Age (days)", justify="right", no_wrap=True)
+    if cooldown is not None:
+        table.add_column("Cooldown", justify="left", no_wrap=True)
+
+    for row in data:
+        cells = [row["version"], row["upload_time"], row["age_days"]]
+        if cooldown is not None:
+            cells.append(row["cooldown"])
+        table.add_row(*cells)
+
+    if output:
+        with open(output, "w") as fh:
+            rich.console.Console(file=fh, width=120).print(table)
+    else:
+        rich.get_console().print(table)
 
 
 def _versions_string(versions: typing.Iterable[Version]) -> str:
--- a/tests/test_list_versions.py
+++ b/tests/test_list_versions.py
@@ -1,0 +1,502 @@
+"""Tests for the ``list-versions`` command cooldown enhancements (issue #1078).
+
+Verifies that ``package list-versions`` shows upload timestamps, age in days,
+and cooldown status via ``--format`` choices, and that
+``--ignore-per-package-overrides`` correctly bypasses per-package overrides.
+"""
+
+import datetime
+import json
+import pathlib
+import re
+import typing
+
+import pytest
+import requests_mock
+from click.testing import CliRunner
+
+from fromager import candidate, resolver
+from fromager.__main__ import main as fromager
+from fromager.commands.package import _cooldown_status
+
+_BOOTSTRAP_TIME = datetime.datetime(2026, 3, 26, 0, 0, 0, tzinfo=datetime.UTC)
+_COOLDOWN_7_DAYS = 7
+_PYPI_SIMPLE_JSON_CONTENT_TYPE = "application/vnd.pypi.simple.v1+json"
+
+# Three versions at known ages relative to _BOOTSTRAP_TIME:
+#   2.0.0  uploaded 2026-03-24 →  2 days old (within 7-day cooldown)
+#   1.3.2  uploaded 2026-03-15 → 11 days old (outside cooldown)
+#   1.2.2  uploaded 2026-01-01 → 84 days old (outside cooldown)
+_PYPI_JSON_RESPONSE = {
+    "meta": {"api-version": "1.1"},
+    "name": "test-pkg",
+    "files": [
+        {
+            "filename": "test_pkg-2.0.0.tar.gz",
+            "url": "https://files.pythonhosted.org/packages/test_pkg-2.0.0.tar.gz",
+            "hashes": {"sha256": "aaa"},
+            "upload-time": "2026-03-24T00:00:00+00:00",
+        },
+        {
+            "filename": "test_pkg-1.3.2.tar.gz",
+            "url": "https://files.pythonhosted.org/packages/test_pkg-1.3.2.tar.gz",
+            "hashes": {"sha256": "bbb"},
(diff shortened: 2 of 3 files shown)


shifa-khan:
Tested the changes locally: 
```bash
$ fromager --min-release-age 20 package list-versions --details --format json click
13:52:27 INFO Looking up versions for click
13:52:27 INFO Found 57 version(s)
    },
    {
        "package": "click",
        "version": "8.3.0",
        "upload_time": "2025-09-18T17:32:23.696258+00:00",
        "age_days": "231",
        "cooldown": "allowed"
    },
    {
        "package": "click",
        "version": "8.3.1",
        "upload_time": "2025-11-15T20:45:42.706404+00:00",
        "age_days": "172",
        "cooldown": "allowed"
    },
    {
        "package": "click",
        "version": "8.3.2",
        "upload_time": "2026-04-03T19:14:45.118083+00:00",
        "age_days": "33",
        "cooldown": "allowed"
    },
    {
        "package": "click",
        "version": "8.3.3",
        "upload_time": "2026-04-22T15:11:27.506141+00:00",
        "age_days": "15",
        "cooldown": "blocked"
    }
]

```
With a 20-day cooldown, click 8.3.3 (15 days old) is correctly marked as blocked while 8.3.2 (33 days old) is allowed.

src/fromager/commands/package.py
```diff
@@ -170,14 +240,185 @@ def list_versions(
...
+    )
+
+    match output_format:
+        case "json":
+            _export_versions_json(version_rows, output)
+        case "csv":
+            _export_versions_csv(version_rows, output)
+        case "table":
+            _export_versions_table(version_rows, req.name, cooldown)
+        case _:
```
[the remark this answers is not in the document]
shifa-khan: @coderabbitai fixed it. Thanks

src/fromager/commands/package.py
```diff
@@ -170,14 +240,190 @@ def list_versions(
...
             raise click.ClickException(f"No versions found for {req.name}")
 
-    versions: list[Version] = sorted(set(candidate.version for candidate in candidates))
+    versions: list[Version] = sorted(set(c.version for c in candidates))
     logger.info(f"Found {len(versions)} version(s)")
 
-    for version in versions:
-        if format_as_requirements:
-            print(f"{req.name}=={version}")
+    if not details:
```
dhellmann: If I'm not listing details and I do provide a cooldown period, then I would only expect to see versions that meet the cooldown requirements.

shifa-khan: fixed. The plain output (without --details) now respects --min-release-age and filters out blocked versions, so only versions that pass the cooldown check are listed. I also made --format and --output implicitly enable --details, so --format json produces structured output directly without needing the extra flag.

dhellmann: I think this could be further simplified by having export functions that take the detail rows as input and print either just the version or the requirements.txt format output. Then you wouldn't need this conditional block at all.

shifa-khan: Thanks for the feedback. Fixed it in the latest version

src/fromager/commands/package.py
```diff
@@ -98,6 +112,34 @@ def package() -> None:
...
+    "--details",
+    is_flag=True,
+    default=False,
+    help="Show upload timestamps, age, and cooldown status for each version.",
+)
+@click.option(
+    "--format",
+    "output_format",
+    type=click.Choice(["table", "csv", "json"], case_sensitive=False),
+    default="table",
```
dhellmann: What do you think about adding "requirements" to this list of options, and eliminating the `--format-as-requirements` flag? Then we could also add a format of "versions" and that would be the default, to maintain consistency. Then I think we can skip the `--details` flag entirely, because each format would tell us whether we should be including those details or not.

shifa-khan: Yeah, that makes sense. I was thinking about it too but since --format-as-requirements preexisted, I didn't want to make that change as part of this PR. I'll go ahead and fold it in.

dhellmann: My first instinct would have been to preserve backwards compatibility, too, but I think in this case it's OK to fix things without doing that.

shifa-khan: yes, the deprecated list-versions shim still accepts --format-as-requirements and maps it to --format requirements

dhellmann:
We need to add some e2e tests, it looks like. The version-only listing works properly, but the table formatter does not filter by age.

```
$ .tox/cli/bin/fromager --min-release-age 800 package list-versions torch
18:12:55 INFO loading settings from /home/dhellmann/devel/aipcc/builder/overrides/settings.yaml
18:12:55 INFO Looking up versions for torch
18:12:55 INFO Found 17 version(s)
2.2.0
2.2.1

$ .tox/cli/bin/fromager --min-release-age 800 package list-versions --format table torch
18:13:07 INFO loading settings from /home/dhellmann/devel/aipcc/builder/overrides/settings.yaml
18:13:07 INFO Looking up versions for torch
18:13:07 INFO Found 17 version(s)
                          Versions for torch
┏━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━┓
┃ Version ┃ Upload Time                      ┃ Age (days) ┃ Cooldown ┃
┡━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━┩
│ 2.2.0   │ 2024-01-30T17:31:54.836329+00:00 │        829 │ allowed  │
│ 2.2.1   │ 2024-02-22T19:17:39.398581+00:00 │        806 │ allowed  │
│ 2.2.2   │ 2024-03-27T21:09:07.247742+00:00 │        772 │ blocked  │
│ 2.3.0   │ 2024-04-24T15:47:07.772402+00:00 │        744 │ blocked  │
│ 2.3.1   │ 2024-06-05T16:42:05.146656+00:00 │        702 │ blocked  │
│ 2.4.0   │ 2024-07-24T15:29:48.218784+00:00 │        653 │ blocked  │
│ 2.4.1   │ 2024-09-04T19:13:19.273720+00:00 │        611 │ blocked  │
│ 2.5.0   │ 2024-10-17T14:47:22.380657+00:00 │        568 │ blocked  │
│ 2.5.1   │ 2024-10-29T17:32:42.789246+00:00 │        556 │ blocked  │
│ 2.6.0   │ 2025-01-29T16:25:55.649425+00:00 │        464 │ blocked  │
│ 2.7.0   │ 2025-04-23T14:35:15.589323+00:00 │        380 │ blocked  │
│ 2.7.1   │ 2025-06-04T17:39:12.852763+00:00 │        338 │ blocked  │
│ 2.8.0   │ 2025-08-06T14:53:52.631799+00:00 │        275 │ blocked  │
│ 2.9.0   │ 2025-10-15T15:46:20.883880+00:00 │        205 │ blocked  │
│ 2.9.1   │ 2025-11-12T15:20:41.620358+00:00 │        177 │ blocked  │
│ 2.10.0  │ 2026-01-21T16:24:44.171477+00:00 │        107 │ blocked  │
│ 2.11.0  │ 2026-03-23T18:11:06.944421+00:00 │         46 │ blocked  │
└─────────┴──────────────────────────────────┴────────────┴──────────┘
```

tiran:
- Can you please get rid microseconds and TZ offset? They are noise.
- The words `allowed` and `blocked` are hard to distinguish. They have the same character count, same `ed` suffix, and look kinda similar. How about different words or no entry when a package is not blocked by cooldown?
- does the table show yanked release?

src/fromager/commands/package.py
```diff
@@ -170,14 +218,203 @@ def list_versions(
         else:
             raise click.ClickException(f"No versions found for {req.name}")
 
-    versions: list[Version] = sorted(set(candidate.version for candidate in candidates))
-    logger.info(f"Found {len(versions)} version(s)")
+    logger.info(f"Found {len(sorted(set(c.version for c in candidates)))} version(s)")
```
smoparth: Since we are just capturing length, can we avoid sorting

src/fromager/commands/package.py
```diff
@@ -170,14 +218,203 @@ def list_versions(
...
+
+_VERSIONS_CSV_FIELDS = ["package", "version", "upload_time", "age_days", "cooldown"]
+
+
+def _export_versions_csv(
+    data: list[dict[str, str]], output: pathlib.Path | None
+) -> None:
+    """Export version details as CSV."""
+    if output:
+        with open(output, "w", newline="") as outfile:
```
smoparth, line 385: This duplicates the full DictWriter block across if - else branches. 

If we determine the output stream earlier we can avoid the duplication.

```
outfile = open(output, "w") if output else sys.stdout
try:
    writer = csv.DictWriter(.....)
    writer.writeheader()
    writer.writerows(data)
finally:
    if output:
        outfile.close()
```

smoparth (changes requested):

src/fromager/commands/package.py
tiran: Or simpler: use `click.File` in the argument and pass `io.TextIO` type around.

shifa-khan:
> We need to add some e2e tests, it looks like. The version-only listing works properly, but the table formatter does not filter by age.
> 
> ```
> $ .tox/cli/bin/fromager --min-release-age 800 package list-versions torch
> 18:12:55 INFO loading settings from /home/dhellmann/devel/aipcc/builder/overrides/settings.yaml
> 18:12:55 INFO Looking up versions for torch
> 18:12:55 INFO Found 17 version(s)
> 2.2.0
> 2.2.1
> 
> $ .tox/cli/bin/fromager --min-release-age 800 package list-versions --format table torch
> 18:13:07 INFO loading settings from /home/dhellmann/devel/aipcc/builder/overrides/settings.yaml
> 18:13:07 INFO Looking up versions for torch
> 18:13:07 INFO Found 17 version(s)
>                           Versions for torch
> ┏━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━┓
> ┃ Version ┃ Upload Time                      ┃ Age (days) ┃ Cooldown ┃
> ┡━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━┩
> │ 2.2.0   │ 2024-01-30T17:31:54.836329+00:00 │        829 │ allowed  │
> │ 2.2.1   │ 2024-02-22T19:17:39.398581+00:00 │        806 │ allowed  │
> │ 2.2.2   │ 2024-03-27T21:09:07.247742+00:00 │        772 │ blocked  │
> │ 2.3.0   │ 2024-04-24T15:47:07.772402+00:00 │        744 │ blocked  │
> │ 2.3.1   │ 2024-06-05T16:42:05.146656+00:00 │        702 │ blocked  │
> │ 2.4.0   │ 2024-07-24T15:29:48.218784+00:00 │        653 │ blocked  │
> │ 2.4.1   │ 2024-09-04T19:13:19.273720+00:00 │        611 │ blocked  │
> │ 2.5.0   │ 2024-10-17T14:47:22.380657+00:00 │        568 │ blocked  │
> │ 2.5.1   │ 2024-10-29T17:32:42.789246+00:00 │        556 │ blocked  │
> │ 2.6.0   │ 2025-01-29T16:25:55.649425+00:00 │        464 │ blocked  │
> │ 2.7.0   │ 2025-04-23T14:35:15.589323+00:00 │        380 │ blocked  │
> │ 2.7.1   │ 2025-06-04T17:39:12.852763+00:00 │        338 │ blocked  │
> │ 2.8.0   │ 2025-08-06T14:53:52.631799+00:00 │        275 │ blocked  │
> │ 2.9.0   │ 2025-10-15T15:46:20.883880+00:00 │        205 │ blocked  │
> │ 2.9.1   │ 2025-11-12T15:20:41.620358+00:00 │        177 │ blocked  │
> │ 2.10.0  │ 2026-01-21T16:24:44.171477+00:00 │        107 │ blocked  │
> │ 2.11.0  │ 2026-03-23T18:11:06.944421+00:00 │         46 │ blocked  │
> └─────────┴──────────────────────────────────┴────────────┴──────────┘
> ```

@dhellmann Thanks for pointing that out. Question on scope - the issue says "list-versions should display upload timestamps and indicate which candidates are blocked by the cooldown policy." Should the detail formats (table/csv/json) still show blocked versions with a "blocked" label so the command can serve as a diagnostic tool? With a short cooldown like 3-7 days there could be versions available on PyPI but blocked by the policy, and it'd be useful to see those rather than have them silently filtered.

dhellmann:
> > We need to add some e2e tests, it looks like. The version-only listing works properly, but the table formatter does not filter by age.
> > ```
> > $ .tox/cli/bin/fromager --min-release-age 800 package list-versions torch
> > 18:12:55 INFO loading settings from /home/dhellmann/devel/aipcc/builder/overrides/settings.yaml
> > 18:12:55 INFO Looking up versions for torch
> > 18:12:55 INFO Found 17 version(s)
> > 2.2.0
> > 2.2.1
> > 
> > $ .tox/cli/bin/fromager --min-release-age 800 package list-versions --format table torch
> > 18:13:07 INFO loading settings from /home/dhellmann/devel/aipcc/builder/overrides/settings.yaml
> > 18:13:07 INFO Looking up versions for torch
> > 18:13:07 INFO Found 17 version(s)
> >                           Versions for torch
> > ┏━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━┳━━━━━━━━━━┓
> > ┃ Version ┃ Upload Time                      ┃ Age (days) ┃ Cooldown ┃
> > ┡━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━╇━━━━━━━━━━┩
> > │ 2.2.0   │ 2024-01-30T17:31:54.836329+00:00 │        829 │ allowed  │
> > │ 2.2.1   │ 2024-02-22T19:17:39.398581+00:00 │        806 │ allowed  │
> > │ 2.2.2   │ 2024-03-27T21:09:07.247742+00:00 │        772 │ blocked  │
> > │ 2.3.0   │ 2024-04-24T15:47:07.772402+00:00 │        744 │ blocked  │
> > │ 2.3.1   │ 2024-06-05T16:42:05.146656+00:00 │        702 │ blocked  │
> > │ 2.4.0   │ 2024-07-24T15:29:48.218784+00:00 │        653 │ blocked  │
> > │ 2.4.1   │ 2024-09-04T19:13:19.273720+00:00 │        611 │ blocked  │
> > │ 2.5.0   │ 2024-10-17T14:47:22.380657+00:00 │        568 │ blocked  │
> > │ 2.5.1   │ 2024-10-29T17:32:42.789246+00:00 │        556 │ blocked  │
> > │ 2.6.0   │ 2025-01-29T16:25:55.649425+00:00 │        464 │ blocked  │
> > │ 2.7.0   │ 2025-04-23T14:35:15.589323+00:00 │        380 │ blocked  │
> > │ 2.7.1   │ 2025-06-04T17:39:12.852763+00:00 │        338 │ blocked  │
> > │ 2.8.0   │ 2025-08-06T14:53:52.631799+00:00 │        275 │ blocked  │
> > │ 2.9.0   │ 2025-10-15T15:46:20.883880+00:00 │        205 │ blocked  │
> > │ 2.9.1   │ 2025-11-12T15:20:41.620358+00:00 │        177 │ blocked  │
> > │ 2.10.0  │ 2026-01-21T16:24:44.171477+00:00 │        107 │ blocked  │
> > │ 2.11.0  │ 2026-03-23T18:11:06.944421+00:00 │         46 │ blocked  │
> > └─────────┴──────────────────────────────────┴────────────┴──────────┘
> > ```
> 
> @dhellmann Thanks for pointing that out. Question on scope - the issue says "list-versions should display upload timestamps and indicate which candidates are blocked by the cooldown policy." Should the detail formats (table/csv/json) still show blocked versions with a "blocked" label so the command can serve as a diagnostic tool? With a short cooldown like 3-7 days there could be versions available on PyPI but blocked by the policy, and it'd be useful to see those rather than have them silently filtered.

Ah, that's a great point. I was focused on making it work the same as the default view, but you're right, it shows that the package would not be included so that's actually a better UX. Ignore my comment.

shifa-khan:
> * Can you please get rid microseconds and TZ offset? They are noise.
> * The words `allowed` and `blocked` are hard to distinguish. They have the same character count, same `ed` suffix, and look kinda similar. How about different words or no entry when a package is not blocked by cooldown?
> * does the table show yanked release?


@tiran 
Agreed on trimming microseconds and TZ offset — will clean that up.

How about "blocked" for blocked entries and "available" for allowed? Leaving the column blank could confuse users.

Yanked versions are already filtered out by the resolver. I verified this with the click package — version 8.2.2 is yanked on PyPI and doesn't appear in the output.

shifa-khan:
@dhellmann 
So to confirm the current behavior, the plain format (fromager --min-release-age 77 package list-versions click) filters out blocked versions, while detail formats (--format table/csv/json) show all versions with a "blocked"/"available" label. Also, the default resolver searches for sdists only, meaning the results from the list-versions command depend on whether sdists are available on PyPI?

src/fromager/commands/package.py
shifa-khan: @smoparth @tiran The initial version matches the pattern in list_overrides._export_csv to stay consistent. If both commands need further cleanup (e.g., click.File or try/finally), I can address that in a follow-up PR. Note that click.File has drawbacks, such as not supporting the newline="" requirement for CSV files.

dhellmann:
> @dhellmann So to confirm the current behavior, the plain format (fromager --min-release-age 77 package list-versions click) filters out blocked versions, while detail formats (--format table/csv/json) show all versions with a "blocked"/"available" label. 

Yes, I think the plain version list and the requirements.txt formatter should filter and the others that provide more details should not filter.

> Also, the default resolver searches for sdists only, meaning the results from the list-versions command depend on whether sdists are available on PyPI?

There's an option to look for both types: https://github.com/python-wheel-build/fromager/blob/main/src/fromager/commands/package.py#L37

And the command uses the plugin to get a resolver, if there is a custom one: https://github.com/python-wheel-build/fromager/blob/main/src/fromager/commands/package.py#L146

So I think it's possible to do the full range of querying we expect?

shifa-khan:
> > @dhellmann So to confirm the current behavior, the plain format (fromager --min-release-age 77 package list-versions click) filters out blocked versions, while detail formats (--format table/csv/json) show all versions with a "blocked"/"available" label.
> 
> Yes, I think the plain version list and the requirements.txt formatter should filter and the others that provide more details should not filter.
> 
> > Also, the default resolver searches for sdists only, meaning the results from the list-versions command depend on whether sdists are available on PyPI?
> 
> There's an option to look for both types: https://github.com/python-wheel-build/fromager/blob/main/src/fromager/commands/package.py#L37
> 
> And the command uses the plugin to get a resolver, if there is a custom one: https://github.com/python-wheel-build/fromager/blob/main/src/fromager/commands/package.py#L146
> 
> So I think it's possible to do the full range of querying we expect?

Right, that's what I figured. list-versions already has --distribution-type and goes through get_resolver_provider, so the full querying range is covered. I ran into this with torch (which only publishes wheels) and --distribution-type wheel handled it as expected.

src/fromager/commands/package.py
shifa-khan: Thanks for pointing out. Fixed in the latest version.

src/fromager/commands/package.py
```diff
@@ -123,7 +159,18 @@ def list_versions(
...
+    Output formats:
+    - "versions": one version per line (default)
+    - "requirements": name==version per line (pip-installable pins)
+    - "table": Rich table with upload timestamps, age, and cooldown status
+    - "csv": CSV with the same detail columns
+    - "json": JSON array with the same detail columns
+
+    Use --ignore-per-package-overrides to see what the global cooldown
+    policy would block without per-package exemptions.
     """
```
shifa-khan, line 172: @dhellmann your thoughts on including the above ^ info?

dhellmann: That looks helpful and it's consistent with the way the distribution types option is documented. +1

src/fromager/commands/package.py
smoparth: Yes, I noticed the other commands have that duplication logic as well, but I figured we could start cleaning it up with this one. However, doing it as a follow-up works too, could you please create a ticket so we don't lose track of it? Also, I’m not sure I fully understand the comment about

> not supporting the newline="" requirement for CSV files

.  Could you clarify why we need that?

shifa-khan: I can create a follow-up ticket for cleanup. 
On newline=""  when writing CSV, Python's csv module handles line endings itself. If the file is opened without newline="", Python's text mode also translates line endings which can produce double carriage returns (\r\r\n instead of \r\n) on Windows. click.File("w") doesn't expose a way to pass newline="" to the underlying open() call.
Likely a non-issue for fromager in practice, we can go with the click.File approach in the follow-up if that's even simpler like how tiran suggested.

smoparth: Considering Windows is not a target platform for us, this shouldn't be a problem. Please include context of these options in the follow up issue. I really liked the Click.File option.

smoparth (dismissed):

src/fromager/commands/package.py
shifa-khan: @smoparth opened a [follow-up issue](https://github.com/python-wheel-build/fromager/issues/1165). Feel free to drop in comments if needed.

smoparth: Thank you!

dhellmann (commented):
This looks good to me. @rd4398 or @LalatenduMohanty could one of you give it a final review, please?

LalatenduMohanty:
Yes, It is in my review queue. I will get it to soon.

tests/test_list_versions.py
```diff
@@ -0,0 +1,494 @@
...
+
+
+# ---------------------------------------------------------------------------
+# Unit tests for helper functions
+# ---------------------------------------------------------------------------
+
+
+def test_cooldown_status_no_cooldown() -> None:
+    """No cooldown configured returns empty string."""
+    from fromager.commands.package import _cooldown_status
```
rd4398: Nit: Can we move this import to the top?

src/fromager/commands/package.py
```diff
@@ -170,14 +218,205 @@ def list_versions(
...
+            print(row["version"])
+
+
+def _export_versions_json(
+    data: list[dict[str, str]], output: pathlib.Path | None
+) -> None:
+    """Export version details as JSON."""
+    if output:
+        with open(output, "w") as outfile:
+            json.dump(data, outfile, indent=2)
```
rd4398, line 370: I think we need a new line after this dump. Without that, when writing to a terminal, the shell prompt will appear on the same line as the closing `]`

src/fromager/commands/package.py
```diff
@@ -170,14 +218,205 @@ def list_versions(
...
+    cooldown: Cooldown | None,
+    *,
+    as_requirements: bool = False,
+) -> None:
+    """Export versions as a plain list, filtering out cooldown-blocked entries."""
+    for row in data:
+        if cooldown is not None and row["cooldown"] == "blocked":
+            continue
+        if as_requirements:
+            print(f"{package_name}=={row['version']}")
```
rd4398, line 359: The `versions` and `requirements` format branches don't pass through the output parameter at all — they always go to `stdout`. The table, csv, and json branches all respect `--output`. This is inconsistent. If a user runs `--format` versions `--output versions.txt`, the file is never written. Either `_export_versions_plain` should accept an output parameter, or the `--output` option should be restricted to formats that use it (via a callback or validation)

rd4398 (commented):
I added a few comments, but overall it looks good

src/fromager/commands/package.py
shifa-khan: Thanks for flagging this. I've updated "versions" and "requirements" to emit a warning when `--output` is used with those formats - `Warning: --output option is ignored for 'versions' and 'requirements' formats`

tests/test_list_versions.py
shifa-khan: Thanks, fixed in the latest version

src/fromager/commands/package.py
shifa-khan: Thanks, fixed in the latest version

src/fromager/commands/package.py
```diff
@@ -170,14 +218,213 @@ def list_versions(
...
+        candidates,
+        cooldown,
+        provider.supports_upload_time,
+    )
+
+    if output is not None and output_format in ("versions", "requirements"):
+        click.echo(
+            "Warning: --output option is ignored for 'versions' and 'requirements' formats",
+            err=True,
+        )
```
LalatenduMohanty, line 235: This is not a standard practice in the existing code except list-overrides.py

LalatenduMohanty: I would perfer similar to find-updates: pass output into _export_versions_plain() and write plain lines to the file,  no warning needed.

LalatenduMohanty: But this is a nit pick.

LalatenduMohanty (approved):

src/fromager/commands/package.py
LalatenduMohanty: Will create a follow up issue.

src/fromager/commands/package.py
LalatenduMohanty: I have created https://github.com/python-wheel-build/fromager/issues/1177 as this UX should be consistent across the code base. and resolving this thread.

src/fromager/commands/package.py
LalatenduMohanty: Created https://github.com/python-wheel-build/fromager/issues/1177

PR #1132, opened 2026-05-11
feat(constraints): support <0 specifier to block packages

tiran:
# Pull Request Description

## What

The convention `<0` (also `<0.0` and `<0.0.0`) is intentionally unsatisfiable and used to exclude a package from builds. Skip the `is_unsatisfiable()` rejection for these specifiers and add a `Constraints.is_blocked()` query method. Combining a blocked and a non-blocked constraint for the same package raises an error.

## Why

See: #1096

## Summary by CodeRabbit

* **New Features**
  * Added support for blocking packages from version resolution using the `<0` constraint syntax (or equivalent forms).
  * Blocked constraints cannot be combined with regular constraints for the same package; mixing them results in an error.

* **Documentation**
  * New guide explaining how to exclude packages using blocking constraints.

[Review Change Stack](https://app.coderabbit.ai/change-stack/python-wheel-build/fromager/pull/1132)

### DIFF
--- a/src/fromager/constraints.py
+++ b/src/fromager/constraints.py
@@ -4,12 +4,27 @@ import typing
 from collections.abc import Generator
 
 from packaging.requirements import Requirement
+from packaging.specifiers import SpecifierSet
 from packaging.utils import NormalizedName, canonicalize_name
 from packaging.version import Version
 
 from . import requirements_file
 
 logger = logging.getLogger(__name__)
+
+
+def _is_blocked_specifier(specifier: SpecifierSet) -> bool:
+    """Return True if specifier blocks a package entirely.
+
+    The convention ``<0``, ``<0.0``, or ``<0.0.0`` is used to mark a
+    package as blocked so that no version can satisfy the constraint.
+    """
+    specs = list(specifier)
+    return (
+        len(specs) == 1
+        and specs[0].operator == "<"
+        and Version(specs[0].version) == Version("0")
+    )
 
 
 class InvalidConstraintError(ValueError):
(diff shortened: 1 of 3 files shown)


src/fromager/constraints.py
```diff
@@ -3,14 +3,28 @@
...
 from packaging.requirements import Requirement
+from packaging.specifiers import SpecifierSet
 from packaging.utils import NormalizedName, canonicalize_name
 from packaging.version import Version
 
 from . import requirements_file
 
+#: Specifiers that block a package entirely (no version can satisfy them).
+#: Used as a convention to exclude packages from builds.
+_BLOCKED_SPECIFIERS: frozenset[str] = frozenset({"<0", "<0.0", "<0.0.0"})
```
rd4398: Should we worry about `0.0.0.0` as well?

Suggestion:

```
def _is_blocked_specifier(specifier: SpecifierSet) -> bool:
      specs = list(specifier)
      return (
          len(specs) == 1
          and specs[0].operator == "<"
          and Version(specs[0].version) == Version("0")
      )
```
This handles <0, <0.0, <0.0.0, and even <0.0.0.0 uniformly, without relying on string representation.

rd4398 (commented):
Overall looks good! I added a suggestion

src/fromager/constraints.py
tiran: good idea

tiran: Fixed!

rd4398 (approved):
Looks good!

PR #1065, opened 2026-04-16
[proposal] Multiple cache wheel servers

tiran:
# Pull Request Description

## What

Propose support for multiple --cache-wheel-server-url values so fromager can pull wheels from several caches in order, e.g. a variant-wide Pulp cache followed by an arch-specific GitLab PyPI registry.

## Why

See: #1064

- [X] PR follows [CONTRIBUTING.md](https://github.com/python-wheel-build/fromager/blob/main/CONTRIBUTING.md) guidelines

PR #1141, opened 2026-05-12
fix(wheels): disable cooldown for local and cache wheel servers by having separate cache wheel resolution from pre-built path

### LINKED ISSUE #1140: disable cooldown for local and cache wheel server
When a cache wheel server is passed via --cache-wheel-server-url, Fromager creates a PyPIProvider for it and applies the package cooldown. If the server supports upload_time, cooldown is enforced. Otherwise, it logs warnings:

> release-age cooldown cannot be enforced — upload timestamp support is not yet implemented for PyPI resolver (searching at <CACHE_WHEEL_SERVER_URL>); cooldown check skipped

This produces hundreds of warnings per run, obscuring real errors.

**Why cooldown is wrong here**
The cache wheel server only contains wheels that fromager itself previously built and uploaded. The cooldown check was already enforced during the original sdist resolution against the upstream source. Re-checking cooldown on wheels we built ourselves is semantically incorrect. The upload timestamp on the cache reflects when our build uploaded it, not when the package was published upstream.

The same logic applies to the local wheel server.

smoparth:
Use PyPICacheProvider (introduced in https://github.com/python-wheel-build/fromager/pull/1143) for local and cache wheel server lookups in the build and download-sequence paths.  This separates trusted internal sources (no cooldown, no upload-time checks) from upstream pre-built sources.

- Add resolve_cached_wheel() using PyPICacheProvider
- Simplify _is_wheel_built() to only check cache servers (local + cache_wheel_server) via resolve_cached_wheel
- Route download_sequence local server through resolve_cached_wheel
- Remove supports_upload_time workaround from resolve_all_prebuilt_wheels
- Add tests for cache/local cooldown bypass and upstream cooldown retention

Follows up on https://github.com/python-wheel-build/fromager/pull/1143 and closes https://github.com/python-wheel-build/fromager/issues/1140.

# Pull Request Description

## Why

### DIFF
--- a/src/fromager/commands/build.py
+++ b/src/fromager/commands/build.py
@@ -346,17 +346,13 @@ def _build(
     pbi = wkctx.package_build_info(req)
     prebuilt = pbi.pre_built
 
-    wheel_server_urls = wheels.get_wheel_server_urls(
-        wkctx, req, cache_wheel_server_url=cache_wheel_server_url
-    )
-
     # See if we can reuse an existing wheel.
     if not force:
         wheel_filename = _is_wheel_built(
             wkctx,
             req.name,
             resolved_version,
-            wheel_server_urls,
+            cache_wheel_server_url=cache_wheel_server_url,
         )
         if wheel_filename:
             logger.info("using existing wheel from %s", wheel_filename)
@@ -469,21 +465,40 @@ def _is_wheel_built(
     wkctx: context.WorkContext,
     dist_name: str,
     resolved_version: Version,
-    wheel_server_urls: list[str],
+    *,
+    cache_wheel_server_url: str | None = None,
 ) -> pathlib.Path | None:
     req = Requirement(f"{dist_name}=={resolved_version}")
 
+    cache_servers: list[str] = list(
+        filter(None, [wkctx.wheel_server_url, cache_wheel_server_url])
+    )
+    if not cache_servers:
+        return None
+
+    logger.info(
+        "checking if a suitable wheel for %s was already built on %s",
+        req,
+        cache_servers,
+    )
+
     try:
-        logger.info(
-            "checking if a suitable wheel for %s was already built on %s",
-            req,
-            wheel_server_urls,
-        )
-        url, _ = wheels.resolve_prebuilt_wheel(
-            ctx=wkctx,
-            req=req,
-            wheel_server_urls=wheel_server_urls,
-        )
+        url: str | None = None
+        for server_url in cache_servers:
+            try:
+                url, _ = wheels.resolve_cached_wheel(
+                    ctx=wkctx,
+                    req=req,
+                    cache_server_url=server_url,
+                )
+                break
+            except Exception:
+                logger.debug("wheel not found on %s", server_url, exc_info=True)
+
+        if url is None:
+            logger.info("could not locate existing wheel")
+            return None
+
         logger.info("found candidate wheel %s", url)
         pbi = wkctx.package_build_info(req)
         build_tag_from_settings = pbi.build_tag(resolved_version)
@@ -505,7 +520,7 @@ def _is_wheel_built(
             return None
 
         wheel_filename: pathlib.Path | None = None
-        if url.startswith(wkctx.wheel_server_url):
+        if wkctx.wheel_server_url and url.startswith(wkctx.wheel_server_url):
             logging.debug("found wheel on local server")
             wheel_filename = wkctx.wheels_downloads / wheel_basename
             if not wheel_filename.exists():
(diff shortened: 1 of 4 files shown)


smoparth:
Considered storing cache_wheel_server_url on WorkContext directly, but PR #1065 already proposes that (as cache_wheel_server_urls on ctx). Kept this fix minimal by threading it as a function parameter; once #1065 lands, the parameter can be replaced with ctx.cache_wheel_server_urls.

src/fromager/wheels.py
```diff
@@ -552,11 +553,16 @@ def resolve_all_prebuilt_wheels(
...
+            # Local wheel server: wheels built this run were already vetted
+            # during sdist resolution. Cooldown is not applicable.
             if ctx.wheel_server_url and url == ctx.wheel_server_url:
-                provider.supports_upload_time = False
+                provider.cooldown = None
+
+            # Cache wheel server: wheels from previous runs were already
+            # vetted when originally built. Cooldown is not applicable.
+            if cache_wheel_server_url and url == cache_wheel_server_url:
+                provider.cooldown = None
```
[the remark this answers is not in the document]
smoparth: There's no normalization or transformation step in the code path but it's a reasonable defensive measure. Happy to add URL normalization based on reviewers feedback.

tiran:
This PR feels like a hack on top of a hack. The code is becoming increasingly confusing and hard to understand.

I think the root cause of the issue is the fact that code paths are mixing cache wheel server and PyPI server for pre-built wheels. Cache lookup is using the same code path as pre-built download. At least that's what I think is happening...

LalatenduMohanty (changes requested):
We should change this PR to use PyPICacheProvider for both  the local and cache wheel server URLs inside resolve_all_prebuilt_wheels, rather than conditionally nulling out cooldown after the fact. refer #1143

smoparth:
@coderabbitai review

src/fromager/commands/build.py
```diff
@@ -476,21 +472,21 @@ def _is_wheel_built(
...
     req = Requirement(f"{dist_name}=={resolved_version}")
 
     try:
-        logger.info(
-            "checking if a suitable wheel for %s was already built on %s",
-            req,
-            wheel_server_urls,
-        )
-        url, _ = wheels.resolve_prebuilt_wheel(
+        url = wheels.resolve_existing_wheel(
```
rd4398: Can we move this call inside the existing try block, or wrap it in its own exception handler that returns None?

Is there a reason why we moved it out?

smoparth: It is inside the try block. Can you check again

rd4398: Right, I missed it somehow while looking at the code.

rd4398 (commented):
This looks good! I will wait for @LalatenduMohanty to approve since he requested changes before

tiran:
The function `resolve_existing_wheel` is conflating two different concepts. Let's treat pre-built wheels and cached wheels as separate entities.

- pre-built wheels come from PyPI or 3rd party wheel index. They should use cooldown.
- cached wheels are built by us and stored in a cache index that contains only our wheels + previously cached pre-built wheels.

smoparth:
> The function `resolve_existing_wheel` is conflating two different concepts. Let's treat pre-built wheels and cached wheels as separate entities.
> 
> * pre-built wheels come from PyPI or 3rd party wheel index. They should use cooldown.
> * cached wheels are built by us and stored in a cache index that contains only our wheels + previously cached pre-built wheels.

@tiran, fixed. _is_wheel_built now only checks cache servers (local + cache_wheel_server) via resolve_cached_wheel using PyPICacheProvider. 
Pre-built wheel resolution is no longer part of this function -- it stays as a separate concern handled in _build.

smoparth:
@Mergifyio rebase

smoparth:
@Mergifyio rebase

LalatenduMohanty (changes requested):
@smoparth  Any package configured with a per-package wheel_server_url in its settings YAML (e.g.):

### settings/some-pkg.yaml
  wheel_server_url: https://custom-wheel.test/simple/

will no longer have that custom server checked for existing wheels in _is_wheel_built(). The function will only check the local and cache servers, won't find the wheel there, and proceed to rebuild the package from source unnecessarily.

Because the new _is_wheel_built() at src/fromager/commands/build.py:473-476 builds its server list like this:
```
  cache_servers: list[str] = list(
      filter(None, [wkctx.wheel_server_url, cache_wheel_server_url])
  )
```
It only considers the global local server and the cache server. pbi.wheel_server_url is never consulted. The call to get_wheel_server_urls() was removed from _build()  entirely

LalatenduMohanty:
Also we should add a test test for the pbi.wheel_server_url path.

PR #1143, opened 2026-05-12
feat(finders): add PyPICacheProvider for cache server lookups

tiran:
# Pull Request Description

## What

Add a dedicated PyPICacheProvider that wraps PyPIProvider with a simplified interface for Fromager's local PyPI-compatible cache server. This makes it explicit that packages are coming from a cache server rather than an upstream index, and eliminates cache-irrelevant options like ignore_platform, override_download_url, cooldown, and supports_upload_time.

The provider enforces that include_wheels and include_sdists are mutually exclusive and defaults to wheels only.

Use PyPICacheProvider in _download_wheel_from_cache to replace the direct PyPIProvider instantiation.

## Why

Avoid confusion between pre-built, upstream PyPI and internal cache indexes.

### DIFF
--- a/src/fromager/finders.py
+++ b/src/fromager/finders.py
@@ -8,12 +8,62 @@ import typing
 from packaging.requirements import Requirement
 from packaging.utils import BuildTag, canonicalize_name
 
-from . import overrides
+from . import overrides, resolver
+from .constraints import Constraints
+from .requirements_file import RequirementType
 
 if typing.TYPE_CHECKING:
     from . import context
 
 logger = logging.getLogger(__name__)
+
+
+class PyPICacheProvider(resolver.PyPIProvider):
+    """Provider for Fromager's PyPI-compatible cache server.
+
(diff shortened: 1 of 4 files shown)


src/fromager/finders.py
```diff
@@ -8,14 +8,60 @@
...
+            include_sdists=include_sdists,
+            include_wheels=include_wheels,
+            sdist_server_url=cache_server_url,
+            constraints=constraints,
+            req_type=req_type,
+            ignore_platform=False,
+            use_resolver_cache=use_resolver_cache,
+            override_download_url=None,
+            cooldown=None,
+            supports_upload_time=False,
```
ryanpetrello, line 65: 👍

ryanpetrello (approved):

andre-motta (approved):

smoparth (approved):

rd4398 (approved):
LGTM!

PR #1148, opened 2026-05-12
feat(bootstrapper): skip already-cached versions in multiple version mode

rd4398:
In the RESOLVE phase, check the wheel cache before creating work items for each resolved version. Versions with a cached wheel are filtered out entirely, avoiding all downstream phases and transitive dependency processing. Only applies in --multiple-versions mode.

Fixes: https://github.com/python-wheel-build/fromager/issues/1038

### DIFF
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -1299,11 +1299,13 @@ class Bootstrapper:
         """RESOLVE phase: resolve versions and expand into START-phase items.
 
         Centralizes version resolution so all dependencies are expanded
-        uniformly. Future filtering (e.g. versions already on disk) and
-        parallelization can be added here in one place.
+        uniformly. In multiple_versions mode, filters out versions whose
+        wheels are already cached to avoid redundant builds and
+        transitive dependency processing.
 
         Returns:
-            One START-phase item per resolved version.
+            One START-phase item per resolved version that needs building.
+            Empty list if all versions are already cached.
         """
         resolved_versions = self.resolve_versions(
             item.req,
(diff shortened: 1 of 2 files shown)


src/fromager/bootstrapper.py
```diff
@@ -1315,6 +1317,22 @@ def _phase_resolve(self, item: WorkItem) -> list[WorkItem]:
...
+                        f"{item.req.name}=={version}: wheel already cached "
+                        f"at {cached_wheel.name}, skipping"
+                    )
+                else:
+                    filtered.append((source_url, version))
+            if not filtered:
+                logger.info(
+                    f"all versions of {item.req.name} already cached, nothing to do"
+                )
+                return []
```
dhellmann: We will want to process at least one version (probably the most recent), otherwise we won't pick up any new transitive dependencies at all.

src/fromager/bootstrapper.py
```diff
@@ -1315,6 +1317,22 @@ def _phase_resolve(self, item: WorkItem) -> list[WorkItem]:
 
         if self.multiple_versions:
             logger.info(f"resolved {len(resolved_versions)} version(s) for {item.req}")
+            filtered: list[tuple[str, Version]] = []
```
smoparth, line 1320: We already check the cache in **PREPARE SOURCE** before any build happens. By short-circuiting earlier at RESOLVE, the extra operations we skip are: graph addition, metadata unpacking, and install dependency extraction which are fast, local operations.

What we lose is correctness: cached versions never enter the dependency graph, never appear in the build order, and their install dependencies are never discovered. If a cached package is the sole path to a transitive dep, it silently drops from the output.

**Suggestion**:
The actually expensive work for cached versions is in PREPARE BUILD creating a venv, installing build system deps, resolving backend/sdist deps. A potentially better approach would be to fast-path from PREPARE SOURCE directly to PROCESS INSTALL DEPS when a cached wheel is found (similar to how we already handle prebuilt wheels), skipping PREPARE BUILD and BUILD entirely. That saves the real I/O cost without losing dependency discovery or graph completeness.

smoparth: Still getting familiar with this part of the codebase so I may be missing context, happy to be corrected.

src/fromager/bootstrapper.py
rd4398: Good catch, fixed in latest commit

src/fromager/bootstrapper.py
LalatenduMohanty: Also filtering at RESOLVE has a side effect, using _find_cached_wheel() to do it is too heavy because it downloads wheels from the remote cache and re-indexes the local mirror just to answer a yes/no question.

LalatenduMohanty: I need to look in to this little more to decide which one is a better design. But at this point I agree that PREPARE_SOURCE (where the cache check already lives) is a better place to do this , not RESOLVE

dhellmann: What benefit do you see to re-processing the dependencies of older cached wheels?

LalatenduMohanty: Thats a valid point. As the goal  is to do incremental rebuilds, then re-processing dependencies of older cached wheels is wasted work as the prior run already discovered and built them.

LalatenduMohanty (approved):

src/fromager/bootstrapper.py
LalatenduMohanty: Also the cache server is the local localhost server that fromager itself runs. In that case
  it's a local HTTP request. So calling _find_cached_wheel() at RESOLVE is not expensive, it's local filesystem globs plus a localhost HTTP request

PR #1150, opened 2026-05-13
feat(bootstrapper): run phase items in parallel when possible

### LINKED ISSUE #1149: investigate running some phases of bootstrap in parallel
Now that we are using a mostly iterative approach to managing bootstrapping, instead of pure recursion, it should be possible for us to run some operations in parallel. For example, if the stack has 3 work items in the resolve phase, we can run all of those at the same time.

dhellmann:
## Summary

Restructure the `bootstrap-parallel` command's main loop to run independent work items concurrently using a `ThreadPoolExecutor`.

**How it works**

The bootstrap loop maintains a LIFO stack of `WorkItem` objects. Each iteration collects a batch from the top of the stack and submits it to the executor. `BootstrapPhase.can_parallelize` controls batch size: parallel phases collect all consecutive same-phase items (running them concurrently); serial phases pop only the single top item (allowing other phases to interleave while still processing one at a time). All items — serial or parallel — go through the same executor.

**Phase breakdown after this change**

| Phase | Parallelism | Reason |
|---|---|---|
| `RESOLVE` | parallel | resolver is thread-safe; independent per-package |
| `START` | parallel | graph writes are locked |
| `PREPARE_SOURCE` | parallel | downloads/unpacking are independent |
| `PREPARE_BUILD` | **serial** | installs build system deps; must complete before backend queries |
| `GET_BUILD_DEPS` | parallel | backend queries are independent (each package has its own build env) |
| `INSTALL_BUILD_DEPS` | **serial** | installs build backend/sdist deps; dep's wheel must exist in mirror first |
| `BUILD` | parallel | compilation is independent per package |
| `UPDATE_BUILD_SEQUENCE` | **serial** | serialises writes to `build-order.json` |
| `PROCESS_INSTALL_DEPS` | parallel | hook + dep extraction are independent |
| `COMPLETE` | parallel | cleanup is independent |

**Thread-safety changes**

- `why` dependency-chain stack converted to a `contextvars.ContextVar`; `ThreadPoolExecutor` copies the caller's context automatically for each task
- `_state_lock` guards all check-then-set write sites (`_seen_requirements`, `_build_requirements`, `_build_stack`, `failed_packages`, `_failed_versions`, dependency graph, progress bar)
- `_wheel_dir_lock` guards wheel directory operations to prevent TOCTOU races between `_find_cached_wheel` and `update_wheel_mirror`
- Progress bar updates via `add_done_callback` so items tick off as they complete rather than at batch boundaries

**Testing**

New e2e test `test_bootstrap_parallel_multiple_versions` bootstraps flit-core 3.0.0--3.12.0 with 5 workers, verifying that all buildable versions are produced and that each parallel phase logged at least one parallel batch.

Closes: #1149

### DIFF
--- a/e2e/test_bootstrap_parallel_multiple_versions.sh
+++ b/e2e/test_bootstrap_parallel_multiple_versions.sh
@@ -1,0 +1,97 @@
+#!/bin/bash
+# -*- indent-tabs-mode: nil; tab-width: 2; sh-indentation: 2; -*-
+
+# Test bootstrap-parallel with --multiple-versions flag.
+# Bootstraps flit-core 3.0.0 through 3.12.0 inclusive (many versions) with
+# only 5 parallel workers to verify that batches larger than the worker pool
+# are handled correctly.
+#
+# Note: flit-core <3.10.0 uses ast.Str which was removed in Python 3.14 and
+# will fail during GET_BUILD_DEPS; those failures are expected and handled
+# gracefully by the bootstrapper.  The test only asserts that the versions
+# that *can* build (>=3.10.0) are present in the output.
+
+SCRIPTDIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+source "$SCRIPTDIR/common.sh"
+
+# Constrain flit-core (its own build backend) to the same range so the
+# resolver never pulls in a version outside 3.0.0–3.12.0 as a build dep.
+constraints_file=$(mktemp)
+trap 'rm -f "$constraints_file"; on_exit' EXIT
+cat > "$constraints_file" <<EOF
+flit-core>=3.0.0,<=3.12.0
+setuptools==82.0.1
+poetry-core==2.4.0
+tomli==2.4.1
+toml==0.10.2
+EOF
+
+fromager \
+  --log-file="$OUTDIR/bootstrap.log" \
+  --error-log-file="$OUTDIR/fromager-errors.log" \
+  --sdists-repo="$OUTDIR/sdists-repo" \
+  --wheels-repo="$OUTDIR/wheels-repo" \
+  --work-dir="$OUTDIR/work-dir" \
+  --constraints-file="$constraints_file" \
+  bootstrap-parallel \
+  --multiple-versions \
+  --max-workers 5 \
+  'flit-core>=3.0.0,<=3.12.0'
+
+echo "Checking for flit-core wheels..."
+find "$OUTDIR/wheels-repo/downloads/" -name 'flit_core-*.whl' | sort
+
+# Verify that a representative sample of versions that can build on this
+# Python (>=3.10.0, which dropped ast.Str) were bootstrapped.
+# flit-core 3.0.0 through 3.12.0 includes many releases; 3.10.0–3.12.0
+# are the ones compatible with Python 3.14+.
+EXPECTED_VERSIONS="3.10.0 3.11.0 3.12.0"
+MISSING_VERSIONS=""
+
+for version in $EXPECTED_VERSIONS; do
+  if find "$OUTDIR/wheels-repo/downloads/" -name "flit_core-${version}-*.whl" | grep -q .; then
+    echo "found wheel for flit-core $version"
+  else
+    echo "missing wheel for flit-core $version" 1>&2
+    MISSING_VERSIONS="$MISSING_VERSIONS $version"
+  fi
+done
+
+if [ -n "$MISSING_VERSIONS" ]; then
+  echo "" 1>&2
+  echo "ERROR: Missing expected versions:$MISSING_VERSIONS" 1>&2
+  echo "The --multiple-versions flag should have bootstrapped all matching versions" 1>&2
+  echo "" 1>&2
+  echo "Found flit-core wheels:" 1>&2
+  find "$OUTDIR/wheels-repo/downloads/" -name 'flit_core-*.whl' 1>&2
+  exit 1
+fi
+
+# Verify that the phases expected to parallelize each logged at least one
+# "starting parallel batch" line, confirming the parallel code path ran.
+#
+# Note: "resolve" is intentionally excluded — with a single top-level
+# requirement there is only one resolve item so that phase always runs serially.
+PARALLEL_PHASES="start prepare-source get-build-deps build process-install-deps complete"
+PHASES_WITHOUT_PARALLEL_BATCH=""
+
+echo ""
+echo "Checking that parallel phases ran at least one parallel batch..."
+for phase in $PARALLEL_PHASES; do
+  if grep -q "starting parallel batch: phase=${phase} " "$OUTDIR/bootstrap.log" 2>/dev/null; then
+    echo "phase=${phase}: OK (parallel batch confirmed)"
+  else
+    echo "phase=${phase}: MISSING parallel batch" 1>&2
+    PHASES_WITHOUT_PARALLEL_BATCH="${PHASES_WITHOUT_PARALLEL_BATCH} ${phase}"
+  fi
+done
+
+if [ -n "$PHASES_WITHOUT_PARALLEL_BATCH" ]; then
+  echo "" 1>&2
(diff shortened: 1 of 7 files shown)


src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -57,6 +58,7 @@ def __init__(
         self._resolved_requirements: dict[
             tuple[str, bool], tuple[tuple[str, Version], ...]
         ] = {}
+        self._lock = threading.Lock()
```
tiran: This lock is not necessary. `dict.__setitem__()` and `dict.get` are atomic operations and thread safe.

src/fromager/bootstrapper.py
```diff
@@ -154,20 +171,24 @@ def __init__(
...
+        self.ctx.enable_parallel_builds()
         self.progressbar = progressbar or progress.Progressbar(None)
         self.prev_graph = prev_graph
         self.cache_wheel_server_url = cache_wheel_server_url or ctx.wheel_server_url
         self.sdist_only = sdist_only
         self.test_mode = test_mode
         self.multiple_versions = multiple_versions
-        self.why: list[tuple[RequirementType, Requirement, Version]] = []
+        self._max_workers = max_workers
+        self._thread_local = threading.local()
```
tiran: Can we use contextvars here? It has some benefits, e.g. it's concurrency-safe between threads and asyncio contexts.

src/fromager/bootstrapper.py
```diff
@@ -192,10 +213,38 @@ def __init__(
...
+        # _failed_versions, ctx.dependency_graph, and progress bar calls that
+        # originate from worker threads (e.g. via _handle_build_requirements).
+        self._state_lock = threading.Lock()
+
+        # Lock protecting wheel directory operations (wheels_build / wheels_downloads).
+        # Prevents TOCTOU races between _find_cached_wheel (which inspects wheels_build)
+        # and _download_prebuilt (which calls update_wheel_mirror, moving files out of
+        # wheels_build).
+        self._wheel_dir_lock = threading.Lock()
+
```
tiran, line 254: The wheel caching code has more threading issues. For example remote cache downloads write downloads directly into the download directory. The download and index directories may contain incomplete wheels.

I think it's time that we overhaul `wheels-dir`.

src/fromager/bootstrapper.py
```diff
@@ -369,29 +418,67 @@ def bootstrap(self, req: Requirement, req_type: RequirementType) -> None:
...
+                "starting batch: phase=%s items=%d",
+                top_phase.value,
+                len(batch),
+            )
+
+            # Run batch through a thread pool. Parallelizable phases use
+            # self._max_workers threads; serial phases use 1 thread so items
+            # execute in submission (FIFO) order.
+            workers = self._max_workers if top_phase.can_parallelize else 1
+            with concurrent.futures.ThreadPoolExecutor(max_workers=workers) as executor:
```
tiran: We should not create and shut down a pool for every loop. It's more efficient to re-use the same executor.

src/fromager/bootstrapper.py
```diff
@@ -369,29 +418,67 @@ def bootstrap(self, req: Requirement, req_type: RequirementType) -> None:
...
                 self.progressbar.update_total(new_dep_count)
-            if not new_items:
-                self.progressbar.update()
-
-            # Phase handlers return [continuation, *new_deps] so extend()
-            # naturally puts new deps on top of the stack (processed first).
-            stack.extend(new_items)
+            for result_items in all_new_items:
+                if not result_items:
+                    self.progressbar.update()
```
tiran: Use a futures `add_done_callback` to update the progress bar.

src/fromager/bootstrapper.py
```diff
@@ -369,29 +418,67 @@ def bootstrap(self, req: Requirement, req_type: RequirementType) -> None:
...
+            workers = self._max_workers if top_phase.can_parallelize else 1
+            with concurrent.futures.ThreadPoolExecutor(max_workers=workers) as executor:
+                futures = [executor.submit(run_one, item) for item in batch]
+                all_new_items: list[list[WorkItem]] = []
+                try:
+                    all_new_items = [f.result() for f in futures]
+                except Exception:
+                    for f in futures:
+                        f.cancel()
+                    raise
```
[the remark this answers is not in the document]
dhellmann: The results of partial writes to disk aren't bad enough to warrant this level of complexity.

src/fromager/bootstrapper.py
dhellmann: fixed in the latest version

src/fromager/bootstrapper.py
dhellmann: fixed in the latest version

src/fromager/bootstrapper.py
dhellmann: fixed in the latest version

src/fromager/bootstrapper.py
dhellmann: fixed in the latest version

src/fromager/bootstrap_requirement_resolver.py
dhellmann: fixed in the latest version

tests/test_bootstrapper.py
```diff
@@ -549,3 +589,302 @@ def test_cache_lookup_no_cache_url_returns_none(tmp_context: WorkContext) -> Non
...
+        return [("https://pkg.test/pkg-b-1.0.tar.gz", Version("1.0"))]
+
+    with patch.object(bt._resolver, "resolve", side_effect=mock_resolve):
+        with patch.object(bt, "_has_been_seen", return_value=True):
+            bt.bootstrap(req=Requirement("pkg-a"), req_type=RequirementType.TOP_LEVEL)
+
+    # One failure recorded (the first call that raised)
+    assert len(bt.failed_packages) == 1
+    assert bt.failed_packages[0]["package"] == "pkg-a"
+
```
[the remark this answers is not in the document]
dhellmann: fixed in the latest version

src/fromager/bootstrapper.py
```diff
@@ -623,8 +713,11 @@ def _download_prebuilt(
         wheel_filename = wheels.download_wheel(req, wheel_url, self.ctx.wheels_prebuilt)
         unpack_dir = self._create_unpack_dir(req, resolved_version)
         # Update the wheel mirror so pre-built wheels are indexed
-        # and available to subsequent builds that need them as dependencies
-        server.update_wheel_mirror(self.ctx)
+        # and available to subsequent builds that need them as dependencies.
+        # Hold _wheel_dir_lock around update_wheel_mirror, which moves files out of
+        # wheels_build, to prevent TOCTOU races with _find_cached_wheel.
+        with self._wheel_dir_lock:
+            server.update_wheel_mirror(self.ctx)
```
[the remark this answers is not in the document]
dhellmann: fixed in the latest version

e2e/test_bootstrap_parallel_multiple_versions.sh
```diff
@@ -0,0 +1,114 @@
...
+  if [ "$max_items" -le 1 ]; then
+    PHASES_WITHOUT_PARALLEL_BATCH="${PHASES_WITHOUT_PARALLEL_BATCH} ${phase}"
+  fi
+done
+
+if [ -n "$PHASES_WITHOUT_PARALLEL_BATCH" ]; then
+  echo "" 1>&2
+  echo "ERROR: these phases never ran a multi-item batch (parallelism not exercised):${PHASES_WITHOUT_PARALLEL_BATCH}" 1>&2
+  exit 1
+fi
```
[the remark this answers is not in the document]
dhellmann: fixed in the latest version

dhellmann:
After experimenting further, this approach isn't going to work. It's too hard to combine a stack with a topology-aware prioritization system. I keep running into issues because we're marking something seen when we start it, even though it's not built.

I'm going to salvage an idea or two and experiment with other approaches.

PR #1154, opened 2026-05-14
fix(resolver): bypass cooldown for transitive deps when top-level pin exists

### LINKED ISSUE #1153: Cooldown check blocks transitive dependency that matches top-level exact pin
When a package appears both as a top-level exact-pinned requirement (`foo==1.0`) and as a transitive dependency of another top-level requirement (`bar` depends on `foo>=0.9`), the cooldown check can block or downgrade the transitive resolution even though the user explicitly approved that version.

## Steps to reproduce

1. Requirements file with multiple top-level pins:
   ```
   foo==1.0
   bar==2.0
   ```
2. `bar` depends on `foo>=0.9`
3. `foo` 1.0 was published within the cooldown window (e.g., 2 days ago, cooldown is 7 days)
4. Run bootstrap with `--min-release-age 7`

## What happens

- `foo==1.0` resolves successfully (top-level exact pin bypasses cooldown)
- `bar==2.0` resolves successfully
- When `bar`'s transitive dependency `foo>=0.9` is resolved, it goes through a separate resolution path with `req_type=INSTALL`. Cooldown is enforced, blocking version 1.0
- Result: either an older version is selected (contradicting the pin) or resolution fails

## Root cause

`resolve_package_cooldown()` in `resolver.py` only bypasses cooldown for `req_type == TOP_LEVEL` with an exact `==` pin. The transitive resolution uses a different requirement string (`foo>=0.9` vs `foo==1.0`), so the session cache is also a miss. The resolver has no way to know that this package+version was already approved as a top-level pin.

## Expected behavior

If a package already has a top-level exact-pinned entry in the dependency graph, cooldown should be bypassed when the same package is resolved as a transitive dependency.

## Suggested fix

In `resolve_package_cooldown()`, check the dependency graph for an existing top-level exact-pinned entry for the package. If one exists, return `None` (bypass cooldown) for transitive resolution of the same package.

LalatenduMohanty:
When a package is both a top-level exact-pinned requirement (e.g., `foo==1.0`) and a transitive dependency of another package (e.g., `bar` depends on `foo>=0.9`), the cooldown check was blocking the transitive resolution even though the user explicitly approved that version via the pin. This could cause version downgrades or resolution failures.

`resolve_package_cooldown()` now checks the dependency graph for an existing top-level exact pin before enforcing cooldown on transitive dependencies.

Closes: #1153

### DIFF
--- a/src/fromager/resolver.py
+++ b/src/fromager/resolver.py
@@ -167,6 +167,17 @@ def resolve_package_cooldown(
             logger.info("cooldown bypassed as the top-level requirement uses == pin")
         return None
 
+    if req_type != RequirementType.TOP_LEVEL:
+        root = ctx.dependency_graph.get_root_node()
+        top_level_edges = root.get_outgoing_edges(req.name, RequirementType.TOP_LEVEL)
+        if any(_has_equality_pin(edge.req) for edge in top_level_edges):
+            if ctx.cooldown is not None:
+                logger.info(
+                    "cooldown bypassed — package has a top-level == pin "
+                    "in the dependency graph"
+                )
+            return None
+
     per_package_days = ctx.package_build_info(req).resolver_min_release_age
     global_cooldown = ctx.cooldown
     if per_package_days is None:
--- a/tests/test_cooldown.py
+++ b/tests/test_cooldown.py
@@ -941,3 +941,75 @@ def test_resolve_package_cooldown_toplevel_compound_specifier_not_exempt(
         ctx, Requirement("test-pkg==1.0,>0.9"), req_type=RequirementType.TOP_LEVEL
     )
     assert result is _COOLDOWN
+
+
+def test_transitive_dep_bypasses_cooldown_when_toplevel_pin_exists(
+    tmp_path: pathlib.Path,
+) -> None:
+    """Transitive dep should bypass cooldown when the same package has a top-level exact pin.
+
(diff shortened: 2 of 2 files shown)


rd4398 (approved):
LGTM!

ryanpetrello:
I'm late to party on review here, but just thinking out loud.

@LalatenduMohanty @rd4398 

I *think* it makes sense to *always* do this, but I'm wondering about edge cases?

```
  - top-level: foo==1.5, transitive: foo — bypass lets 1.5 be a candidate. Resolver picks it. Correct.
  - top-level: foo==1.5, transitive: foo>=1.0 — same thing.
  - top-level: foo==1.5, transitive: foo>=1.5 — same thing.
  - top-level: foo==1.5, transitive: foo>=2.0 — dependency conflict, cooldown is irrelevant
```

src/fromager/resolver.py
```diff
@@ -167,6 +167,17 @@ def resolve_package_cooldown(
             logger.info("cooldown bypassed as the top-level requirement uses == pin")
         return None
 
+    if req_type != RequirementType.TOP_LEVEL:
```
smoparth, line 170: Nit: this graph lookup (get_root_node + get_outgoing_edges) runs for every non-top-level resolution even when ctx.cooldown is None. We could add an early if ctx.cooldown is None: return None guard at the top of the function to skip both this check and the existing top-level pin check when cooldown isn't active.

Issue #1155, opened 2026-05-14
Add edge case tests for cooldown bypass via top-level pin in dependency graph


LalatenduMohanty:
Follow-up to #1153. The fix in `resolve_package_cooldown()` bypasses cooldown for transitive dependencies when a top-level exact pin exists in the dependency graph. The current tests cover the happy path but several edge cases need coverage to prevent regressions.

## Test cases to add

1. **Non-exact top-level entry does not bypass cooldown**
   Graph has `test-pkg>=1.0` as TOP_LEVEL (not an exact `==` pin). The `_has_equality_pin` check on the graph edge should prevent bypass, and cooldown should be enforced normally.

2. **Wildcard equality pin does not bypass cooldown**
   Graph has `test-pkg==1.*` as TOP_LEVEL. This is not a true exact pin — `_has_equality_pin` rejects wildcards. Cooldown should be enforced.

3. **Name normalization across requirement and graph**
   Transitive requirement uses `Test_Pkg>=1.0` while graph has the entry under `test-pkg`. `get_outgoing_edges` canonicalizes both sides, so the bypass should still work.

4. **Top-level pin takes precedence over per-package `min_release_age` override**
   Package has `resolver_dist.min_release_age: 30` in its settings, but also has a top-level `==` pin in the graph. The graph bypass runs before the per-package check, so the pin wins and cooldown is bypassed entirely. This is the correct precedence — an explicit pin is a stronger trust signal than a per-package cooldown override.

LalatenduMohanty:
Adding these tests to https://github.com/python-wheel-build/fromager/pull/1157

PR #1157, opened 2026-05-14
fix(resolver): exempt top-level pinned versions from transitive cooldown

### LINKED ISSUE #1153: Cooldown check blocks transitive dependency that matches top-level exact pin

LalatenduMohanty:
Add `exempt_versions` field to `Cooldown` so that only the specific version(s) approved via a top-level `==` pin bypass the cooldown age check when resolved as transitive dependencies. Other versions remain subject to cooldown filtering.

Fixes: #1153 #1155

### DIFF
--- a/src/fromager/resolver.py
+++ b/src/fromager/resolver.py
@@ -164,68 +164,57 @@ def _get_toplevel_pinned_versions(
 def _resolve_cooldown_params(
     ctx: context.WorkContext,
     req: Requirement,
-) -> tuple[datetime.timedelta, datetime.datetime] | None:
+) -> tuple[datetime.timedelta, datetime.datetime]:
     """Resolve min_age and bootstrap_time for a package's cooldown.
 
-    Returns ``None`` when cooldown is disabled (no global/per-package
-    configuration, or a per-package override of 0).  Otherwise returns
-    the effective ``(min_age, bootstrap_time)`` pair.
+    Always returns a ``(min_age, bootstrap_time)`` pair.  When cooldown is
+    disabled (per-package override of 0, or no global/per-package config),
+    ``min_age`` will be zero — every package age exceeds that threshold.
     """
     min_age_override = ctx.package_build_info(req).resolver_min_release_age
 
-    if ctx.cooldown is None and min_age_override is None:
-        return None
     if min_age_override == 0:
-        return None
-
-    if min_age_override is not None:
+        min_age = datetime.timedelta(0)
+    elif min_age_override is not None:
         min_age = datetime.timedelta(days=min_age_override)
-    elif ctx.cooldown is not None:
-        min_age = ctx.cooldown.min_age
     else:
-        return None
+        min_age = ctx.cooldown.min_age
 
-    bootstrap_time = (
-        ctx.cooldown.bootstrap_time
-        if ctx.cooldown is not None
-        else datetime.datetime.now(datetime.UTC)
-    )
-    return min_age, bootstrap_time
+    return min_age, ctx.cooldown.bootstrap_time
 
 
 def resolve_package_cooldown(
     ctx: context.WorkContext,
     req: Requirement,
     req_type: RequirementType | None = None,
-) -> Cooldown | None:
+) -> Cooldown:
     """Compute the effective cooldown for a single package.
 
-    Returns ``None`` (cooldown disabled) when:
+    Always returns a ``Cooldown`` instance.  A ``min_age`` of zero means
+    cooldown is effectively disabled — every package age exceeds zero.
+
+    Returns a *disabled* cooldown (min_age=0) when:
 
     * The requirement is a top-level exact ``==`` pin — the user explicitly
       approved that version.
     * A per-package ``min_release_age=0`` override disables cooldown.
     * No global cooldown is configured and no per-package override enables one.
 
-    Otherwise a ``Cooldown`` is returned with:
+    Otherwise returns an *active* cooldown with:
 
     * *min_age* from the per-package override (if set) or the global cooldown.
     * *bootstrap_time* inherited from the global cooldown (for a consistent
-      cutoff across the entire run) or the current time.
+      cutoff across the entire run).
     * *exempt_versions* populated from top-level exact-pinned entries in the
       dependency graph, so transitive resolutions of the same package honour
       the user's explicit pin.
     """
     if req_type == RequirementType.TOP_LEVEL and _has_equality_pin(req):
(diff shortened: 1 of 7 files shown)


src/fromager/resolver.py
```diff
@@ -147,6 +147,20 @@ def _has_equality_pin(req: Requirement) -> bool:
...
+        req.name, RequirementType.TOP_LEVEL
+    )
+    return frozenset(
+        edge.destination_node.version
+        for edge in top_level_edges
+        if _has_equality_pin(edge.req)
+    )
+
+
 def resolve_package_cooldown(
```
dhellmann, line 186: I'm finding the logic in this function really hard to follow. There are a lot of exceptions to rules applying the days or deciding which number of days to use.

Why not make it always return a new valid Cooldown instance, even if the number of days is 0 in some cases or the values match the global settings? That would simplify the logic to just needing to decide which value to use for each argument to that class.

src/fromager/resolver.py
```diff
@@ -167,10 +181,20 @@ def resolve_package_cooldown(
             logger.info("cooldown bypassed as the top-level requirement uses == pin")
         return None
 
+    exempt = _get_toplevel_pinned_versions(ctx, req)
+
     per_package_days = ctx.package_build_info(req).resolver_min_release_age
     global_cooldown = ctx.cooldown
     if per_package_days is None:
-        return global_cooldown
+        if global_cooldown is None:
```
dhellmann: If the global cooldown is None, does that mean cooldown is disabled?

LalatenduMohanty: Yes, global_cooldown is None (i.e. ctx.cooldown is None) means cooldown is disabled globally. It means the user did not pass --min-release-age on the CLI, so no cooldown policy is in effect for the run.

dhellmann: OK, so it seems like that check could be done earlier, outside of this branch, because that decision overrides any other settings, right?

LalatenduMohanty: ~~ctx.cooldown does not override everything. For example. It gets the value from individual package package override. A per-package min_release_age setting in the package's YAML config would lead to ctx.cooldown have the value even when no global cooldown is configured~~

Sorry I was wrong, the right answer is  ctx.cooldown is none alone doesn't mean cooldown is disabled for all packages. A per-package min_release_age setting in the package's YAML config can enable cooldown for that package as well. That's why the check is ctx.cooldown is None and per_package_days is None , we must check both.

LalatenduMohanty: @dhellmann does that answer your question. So IMO we should check value of ctx.cooldown here.

src/fromager/resolver.py
LalatenduMohanty: I have tried to make the code more easy to follow. PTAL

src/fromager/resolver.py
```diff
@@ -147,42 +147,89 @@ def _has_equality_pin(req: Requirement) -> bool:
...
+        return None
+    if min_age_override == 0:
+        return None
+
+    if min_age_override is not None:
+        min_age = datetime.timedelta(days=min_age_override)
+    elif ctx.cooldown is not None:
+        min_age = ctx.cooldown.min_age
+    else:
+        return None
```
rd4398: Non blocker: I think this is dead code and can be removed. The final `else: return None` is unreachable: if `min_age_override` is `None` and `ctx.cooldown` is `None`, we already returned None in the first guard.

rd4398 (commented):
I have a nit but it looks overall okay. I see Doug has left comments as well, so I will wait for him to take a look again

src/fromager/resolver.py
```diff
@@ -147,42 +147,89 @@ def _has_equality_pin(req: Requirement) -> bool:
...
+        req.name, RequirementType.TOP_LEVEL
+    )
+    return frozenset(
+        edge.destination_node.version
+        for edge in top_level_edges
+        if _has_equality_pin(edge.req)
+    )
+
+
+def _resolve_cooldown_params(
```
dhellmann, line 164: This is still doing lots of conditional checks to try to return None when we're not going to apply a cooldown.

Why not just say that the cooldown period is 0 (or -1), every age of every package will be more than that, and you return a real value. Then everywhere else in the code you don't have to check for None at all.

Why is ctx.cooldown ever None? Why don't we enforce that that thing has a value when the process starts?

LalatenduMohanty: Agree the code can be simplified further.

src/fromager/resolver.py
LalatenduMohanty: Good catch, fixed it in a followup commit.

src/fromager/resolver.py
LalatenduMohanty: @dhellmann I have added another commit for the refactor. PTAL

LalatenduMohanty:
I have added some tests which will fail unless #1188 is not merged.

PR #1159, opened 2026-05-14
feat(bootstrapper): record bootstrap stack state on each loop iteration

dhellmann:
## Summary

Writes `bootstrap-stack.json` to the work directory before each iteration of the DFS loop in `Bootstrapper.bootstrap()`. The file is a JSON array where index 0 is the next item to be processed (`stack[-1]`), overwritten on each iteration. Each entry captures identity and accumulated state fields (`req`, `req_type`, `phase`, `resolved_version`, `source_url`, `why`, `parent`, dep sets); heavy build objects (`build_env`, paths, `build_result`) are intentionally excluded.

Enables debugging, progress inspection, and post-mortem analysis of interrupted runs.

### DIFF
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -188,6 +188,7 @@ class Bootstrapper:
         self._seen_requirements: set[SeenKey] = set()
 
         self._build_order_filename = self.ctx.work_dir / "build-order.json"
+        self._stack_filename = self.ctx.work_dir / "bootstrap-stack.json"
 
         # Track failed packages in test mode (list of typed dicts for JSON export)
         self.failed_packages: list[FailureRecord] = []
@@ -371,6 +372,7 @@ class Bootstrapper:
 
         # Main iterative DFS loop
         while stack:
+            self._record_stack_state(stack)
             item = stack.pop()
             self.why = list(item.why_snapshot)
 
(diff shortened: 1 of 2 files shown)


src/fromager/bootstrapper.py
```diff
@@ -188,6 +188,7 @@ def __init__(
         self._seen_requirements: set[SeenKey] = set()
 
         self._build_order_filename = self.ctx.work_dir / "build-order.json"
+        self._stack_filename = self.ctx.work_dir / "bootstrap-stack.json"
```
LalatenduMohanty, line 191: We should add a log line conmmunicating users that this file exists. That way AI agents can also will find the file easily.

LalatenduMohanty (changes requested):

src/fromager/bootstrapper.py
```diff
@@ -371,6 +372,7 @@ def bootstrap(self, req: Requirement, req_type: RequirementType) -> None:
 
         # Main iterative DFS loop
         while stack:
+            self._record_stack_state(stack)
```
LalatenduMohanty, line 376: Rather than directly calling it , should we call it from a try/exception block. The idea is to write  exceptions that escape the loop. Because it would be efficient writing when required as compared to writing the entire stack as JSON to disk on every iteration of the DFS loop

```
try:
    while stack:
        item = stack.pop()
        self.why = list(item.why_snapshot)

        with req_ctxvar_context(item.req), self._track_why(item):
            try:
                new_items = self._dispatch_phase(item)
            except Exception as err:
                new_items = self._handle_phase_error(item, err)
        # ... rest of loop ...
except Exception:
    self._record_stack_state(stack)
    raise
```

LalatenduMohanty (changes requested):

src/fromager/bootstrapper.py
dhellmann: It takes a lot less time to write than you might think, and having it available as work is progressing gives you the option of watching what the tool is doing as it works.

src/fromager/bootstrapper.py
dhellmann: Good idea, done

LalatenduMohanty (approved):

PR #1163, opened 2026-05-19
refactor(resolver): move age filter fallback to resolver with cache server lookup

### LINKED ISSUE #1162: [Bug]: fromager resolves all versions if age filtering returns 0 candidates
### Environment

- Fromager version: 0.85.0
- Python version: 3.12
- OS / architecture: NA
- Container or local: Both

### Bug description

**Expected behavior:**
fromager should return the one in cache and return zero new versions resolved. It returns all the versions currently because there was no setuptools release in last 30 days. This fallback behavior is implemented in resolver https://github.com/python-wheel-build/fromager/blob/029327a412aa2da00f387ad414be5193b31250a7/src/fromager/resolver.py#L318 

**Actual behavior:**
fromager resolves all  829 version(s) for setuptools

**Steps to reproduce / logs:**
Run bootstrap command in multiple version mode with setuptools as requirement and set --max-release-age as 30 days

rd4398:
Move the two-step fallback logic from `_phase_resolve()` into
`BootstrapRequirementResolver._resolve()`. When age filtering removes
all candidates in multi-version mode, the resolver now queries the
remote wheel cache server for the newest available version and returns
only that single version. The resolver warns on zero results but does
not error — `_phase_resolve()` raises the exception.

Fixes: https://github.com/python-wheel-build/fromager/issues/1162

### DIFF
--- a/src/fromager/bootstrap_requirement_resolver.py
+++ b/src/fromager/bootstrap_requirement_resolver.py
@@ -12,7 +12,7 @@ import typing
 from packaging.requirements import Requirement
 from packaging.version import Version
 
-from . import resolver, sources, wheels
+from . import finders, resolver, sources, wheels
 from .dependency_graph import DependencyGraph
 from .requirements_file import RequirementType
 
@@ -40,15 +40,25 @@ class BootstrapRequirementResolver:
         self,
         ctx: context.WorkContext,
         prev_graph: DependencyGraph | None = None,
+        multiple_versions: bool = False,
+        cache_wheel_server_url: str = "",
     ) -> None:
         """Initialize requirement resolver.
 
         Args:
             ctx: Work context with constraints and settings
             prev_graph: Optional previous dependency graph for caching
+            multiple_versions: If ``True`` and no results are found through
+                any other approach, takes the latest candidate from the
+                cache server, ignoring the age filters.  In all other
+                cases, returns an empty list when no candidates are found.
+            cache_wheel_server_url: URL of the remote wheel cache server.
+                Used as a fallback when age filtering produces no candidates.
         """
         self.ctx = ctx
         self.prev_graph = prev_graph
+        self.multiple_versions = multiple_versions
+        self.cache_wheel_server_url = cache_wheel_server_url
         # Session-level resolution cache to avoid re-resolving same requirements
         # Key: (requirement_string, pre_built) to distinguish source vs prebuilt
         # Value: tuple of (url, version) tuples sorted by version (highest first)
@@ -72,6 +82,8 @@ class BootstrapRequirementResolver:
         1. Session cache (if previously resolved)
         2. Previous dependency graph
         3. PyPI resolution (source or prebuilt based on package build info)
+        4. Remote wheel cache server (multi-version mode only, when age
+           filtering produced no candidates)
 
         Args:
             req: Package requirement
@@ -110,36 +122,7 @@ class BootstrapRequirementResolver:
             logger.debug(f"resolved {req} from cache")
             return list(cached_result) if return_all_versions else [cached_result[0]]
 
-        # Resolve using strategies
-        results = self._resolve(req, req_type, parent_req, pre_built)
-
-        # Cache the result
-        self.cache_resolution(req, pre_built, results)
-        return results if return_all_versions else [results[0]]
-
-    def _resolve(
-        self,
-        req: Requirement,
-        req_type: RequirementType,
-        parent_req: Requirement | None,
-        pre_built: bool,
-    ) -> list[tuple[str, Version]]:
-        """Internal resolution logic without caching.
-
-        Tries resolution strategies in order:
-        1. Previous dependency graph
-        2. PyPI resolution (source or prebuilt)
-
-        Args:
-            req: Package requirement
-            req_type: Type of requirement
-            parent_req: Parent requirement from dependency chain
-            pre_built: Whether to resolve prebuilt (True) or source (False)
-
-        Returns:
-            List of (url, version) tuples sorted by version (highest first)
-        """
-        # Try graph
+        # Try previous dependency graph
         cached_resolution = self._resolve_from_graph(
             req=req,
             req_type=req_type,
@@ -151,17 +134,13 @@ class BootstrapRequirementResolver:
             logger.debug(
                 f"resolved from previous bootstrap: {len(cached_resolution)} version(s)"
             )
-            return cached_resolution
-
-        # Fallback to PyPI using provider pattern
-        if pre_built:
+            results = cached_resolution
+        elif pre_built:
             # Resolve prebuilt wheel
-            # Get wheel server URLs (use PyPI as cache/fallback server)
             wheel_server_urls = wheels.get_wheel_server_urls(
                 self.ctx, req, cache_wheel_server_url=resolver.PYPI_SERVER_URL
             )
-            # Use shared retry loop logic from wheels module
-            return wheels.resolve_all_prebuilt_wheels(
+            results = wheels.resolve_all_prebuilt_wheels(
                 ctx=self.ctx,
                 req=req,
                 wheel_server_urls=wheel_server_urls,
@@ -176,9 +155,75 @@ class BootstrapRequirementResolver:
                 req_type=req_type,
             )
             max_age_cutoff = resolver._compute_max_age_cutoff(self.ctx)
-            return resolver.find_all_matching_from_provider(
-                provider, req, max_age_cutoff=max_age_cutoff
+            results = resolver.find_all_matching_from_provider(
+                provider,
+                req,
+                max_age_cutoff=max_age_cutoff,
+                fallback_on_empty_age_filter=not self.multiple_versions,
             )
+
+            if not results and self.multiple_versions and self.cache_wheel_server_url:
+                logger.info(
+                    "no results found with normal resolution, "
+                    "falling back to the cache server %s",
+                    self.cache_wheel_server_url,
+                )
+                results = self._resolve_from_cache_server(req)
+
+            if not results:
+                logger.warning(
+                    "resolver returned no results "
+                    "(wheel server URL %s, %s version mode)",
+                    self.cache_wheel_server_url or "(none)",
+                    "multiple" if self.multiple_versions else "single",
+                )
+
+        # Only cache non-empty results.
+        if results:
+            self.cache_resolution(req, pre_built, results)
+
+        if not results:
+            return []
+        if return_all_versions:
+            return results
+        return [results[0]]
+
+    def _resolve_from_cache_server(self, req: Requirement) -> list[tuple[str, Version]]:
+        """Fall back to the remote wheel cache server for a cached version.
+
+        When age filtering removes all candidates in multi-version mode,
+        queries the remote cache server for the newest available wheel.
+        Returns at most one version so that transitive dependencies are
+        re-processed without rebuilding every old version.
+        """
+        logger.info(
+            "checking cache server %s for existing build",
+            self.cache_wheel_server_url,
+        )
+        best: tuple[str, Version] | None = None
+        for include_sdists, include_wheels in [(False, True), (True, False)]:
+            try:
+                provider = finders.PyPICacheProvider(
+                    cache_server_url=self.cache_wheel_server_url,
+                    constraints=self.ctx.constraints,
+                    include_sdists=include_sdists,
+                    include_wheels=include_wheels,
+                )
+                results = resolver.find_all_matching_from_provider(provider, req)
+                if results:
+                    url, version = results[0]
+                    if best is None or version > best[1]:
+                        best = (url, version)
+            except Exception as err:
+                logger.warning(
+                    "error checking cache server %s: %s",
+                    self.cache_wheel_server_url,
+                    err,
+                )
+        if best is not None:
+            logger.info("found version %s on cache server", best[1])
+            return [best]
+        return []
 
     def get_cached_resolution(
         self,
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -173,6 +173,8 @@ class Bootstrapper:
         self._resolver = bootstrap_requirement_resolver.BootstrapRequirementResolver(
             ctx=ctx,
             prev_graph=prev_graph,
+            multiple_versions=multiple_versions,
+            cache_wheel_server_url=self.cache_wheel_server_url,
         )
         # Push items onto the stack as we start to resolve their
         # dependencies so at the end we have a list of items that need to
--- a/src/fromager/resolver.py
+++ b/src/fromager/resolver.py
@@ -243,6 +243,7 @@ def find_all_matching_from_provider(
     provider: BaseProvider,
     req: Requirement,
     max_age_cutoff: datetime.datetime | None = None,
+    fallback_on_empty_age_filter: bool = True,
 ) -> list[tuple[str, Version]]:
     """Find all matching candidates from provider without full dependency resolution.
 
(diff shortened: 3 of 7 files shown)


rd4398:
I considered two approaches:

  1. Move age filtering to the bootstrapper — Remove age checks from the resolver entirely and filter in the bootstrapper where we have cache access.
  
  2. Two-step resolution in the bootstrapper (chosen) — Let the resolver signal "all candidates too old" by returning empty, then let the bootstrapper decide what to do based on cache state.

  This PR implements approach 2 because

  - The age filter is used in three call sites (resolver.resolve(), bootstrap_requirement_resolver._resolve(), sources.resolve_source()). Moving it out would require threading upload_time metadata through all return paths — today the resolver returns
  list[tuple[str, Version]], not full Candidate objects, so the bootstrapper has no way to inspect upload times without duplicating PyPI metadata lookups.
  - Single-version mode still needs the fallback. The resolver.resolve() and sources.resolve_source() paths return results[0] directly — an empty list there would crash. The fallback_on_empty_age_filter flag (default True) preserves this safety for all existing
  callers while letting multi-version mode opt out.
  - The bootstrapper already has the right context. _phase_resolve() already checks the wheel cache per-version. Adding a "do we have any cached wheel?" check before falling back is a natural extension of that existing pattern, not a new abstraction.

  In short: the resolver stays a pure "find candidates" layer, and the bootstrapper — which owns the build/cache decisions — gets to make the smarter choice when no fresh candidates exist.

src/fromager/bootstrap_requirement_resolver.py
dhellmann: Why don't we include the age check here for pre-built wheels?

src/fromager/bootstrapper.py
```diff
@@ -669,6 +674,26 @@ def _find_cached_wheel(
 
         return None, None
 
+    def _has_any_cached_wheel(self, req: Requirement) -> bool:
```
dhellmann: I think we're going to want to look at the remote cache server, not the local one. The first time we bootstrap a package that has not had a recent release we won't have the file locally but it might already have been built and put in the wheel server from a previous job.

dhellmann:
Rohan and I spoke directly about some other changes to this PR to update the contract for the bootstrap resolver to be "tell me the list of versions to build for this rule" and to therefore move some of the new logic from this change out of the phase method and into the bootstrap resolver instead. 

We agreed that if the "normal" resolution using the existing steps results in no versions, the resolver should fall back to looking at the remote wheel server (if any) and pick the newest version of the package available on that server that matches the rule, but to ignore the age and to only return 1 version, even in the mode where we are building multiple versions. That will ensure that the bootstrapped re-processes _one_ version of the package to get its transitive dependencies again and keep those up to date, without re-processing _all_ versions of the package when we know they are not new.

We also agreed that the bootstrap resolver should warn if it gets 0 results, but not report an error, and the phase method should be responsible for raising an exception because that is where we know that we will eventually fail the current build because we could not find any versions of a dependency anywhere.

@rd4398 please let me know if I forgot any details about what we said.

rd4398:
> Rohan and I spoke directly about some other changes to this PR to update the contract for the bootstrap resolver to be "tell me the list of versions to build for this rule" and to therefore move some of the new logic from this change out of the phase method and into the bootstrap resolver instead.
> 
> We agreed that if the "normal" resolution using the existing steps results in no versions, the resolver should fall back to looking at the remote wheel server (if any) and pick the newest version of the package available on that server that matches the rule, but to ignore the age and to only return 1 version, even in the mode where we are building multiple versions. That will ensure that the bootstrapped re-processes _one_ version of the package to get its transitive dependencies again and keep those up to date, without re-processing _all_ versions of the package when we know they are not new.
> 
> We also agreed that the bootstrap resolver should warn if it gets 0 results, but not report an error, and the phase method should be responsible for raising an exception because that is where we know that we will eventually fail the current build because we could not find any versions of a dependency anywhere.
> 
> @rd4398 please let me know if I forgot any details about what we said.

@dhellmann yeah that summarizes it! I have pushed changes in latest commit based on our discussion

src/fromager/bootstrapper.py
rd4398: Yeah done in latest commit

src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -113,8 +124,10 @@ def resolve(
         # Resolve using strategies
         results = self._resolve(req, req_type, parent_req, pre_built)
 
-        # Cache the result
-        self.cache_resolution(req, pre_built, results)
+        # Only cache non-empty results.
+        if results:
+            self.cache_resolution(req, pre_built, results)
+
         return results if return_all_versions else [results[0]]
```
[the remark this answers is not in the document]
rd4398: fixed in latest commit

src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -113,8 +124,12 @@ def resolve(
         # Resolve using strategies
         results = self._resolve(req, req_type, parent_req, pre_built)
```
dhellmann: Why is `_resolve()` its own method?

src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -167,18 +184,70 @@ def _resolve(
...
+            fallback_on_empty_age_filter=not self.multiple_versions,
+        )
+
+        if not results and self.multiple_versions and self.cache_wheel_server_url:
+            results = self._resolve_from_cache_server(req)
+
+        if not results:
+            logger.warning(
+                "%s: could not find any versions (all filtered by "
+                "max-release-age and no cached wheel on server)",
```
dhellmann: it would help someone debugging this error if the log message included the max-release-age value.

src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -167,18 +184,70 @@ def _resolve(
...
+        )
+        max_age_cutoff = resolver._compute_max_age_cutoff(self.ctx)
+        results = resolver.find_all_matching_from_provider(
+            provider,
+            req,
+            max_age_cutoff=max_age_cutoff,
+            fallback_on_empty_age_filter=not self.multiple_versions,
+        )
+
+        if not results and self.multiple_versions and self.cache_wheel_server_url:
```
dhellmann: This section deserves a comment explaining why we look at the cache server.

dhellmann: You might also consider if it makes sense to reverse the logic and break up each reason for returning to make the logic clearer.

```
if results:
    return results

if not self.multiple_versions:
    # without multiple-versions enabled we don't use the age filter so a failure to resolve a version means it is definitely not there
    logger.warning("resolver returned no results")
    return []

if not self.cache_wheel_server_url:
    # if we do not have a cache wheel server URL we cannot look for a previous build
    logger.warning("resolver returned no results and we have no wheel cache server URL to fall back on")
    return []

logger.info("no results found with normal resolution, falling back to the cache server %s", self.cache_wheel_server_url)
return self._resolve_from_cache_server(req)
```

src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -167,18 +184,70 @@ def _resolve(
...
+
+    def _resolve_from_cache_server(self, req: Requirement) -> list[tuple[str, Version]]:
+        """Fall back to the remote wheel cache server for a cached version.
+
+        When age filtering removes all candidates in multi-version mode,
+        queries the remote cache server for the newest available wheel.
+        Returns at most one version so that transitive dependencies are
+        re-processed without rebuilding every old version.
+        """
+        logger.info(
```
dhellmann, line 199: This log message assumes an execution order of the methods that might change in the future. We should log the detail about the max-release-age filtering before calling this function. See my suggested changes in the earlier comment.

rd4398: Yeah, I agree. I have made changes in latest commit.

src/fromager/bootstrap_requirement_resolver.py
rd4398: Good idea, I have changed it in latest commit

src/fromager/bootstrap_requirement_resolver.py
rd4398: Okay, added in latest push

src/fromager/bootstrap_requirement_resolver.py
rd4398: I think it was added when we refactored the bootstrapper to create `BootstrapRequirementResolver`. I don't see though why it should be separate as it just adds indirection. Let me move it inline and remove it as independent method in latest commit

src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -176,9 +154,79 @@ def _resolve(
...
+            )
+
+            if results:
+                pass
+            elif not self.multiple_versions:
+                logger.warning(
+                    "%s: resolver returned no results",
+                    req.name,
+                )
+            elif not self.cache_wheel_server_url:
```
dhellmann: It's not clear that the 2 checks here are mutually exclusive. Another way to log the message with details would be to build the log message from the variables. Something like

```
logger.warning(
  "resolver returned no results with fallback wheel server URL %s and %s version mode",
  self.cache_wheel_server_url,
  "multiple" if self.multiple_versions else "single",
)
```

I _think_ the message will automatically have the requirements details added as a prefix because the logger looks at the thread context, but I'm not sure.

src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -176,9 +154,79 @@ def _resolve(
...
+                logger.warning(
+                    "%s: resolver returned no results and we have no "
+                    "wheel cache server URL to fall back on",
+                    req.name,
+                )
+            else:
+                logger.info(
```
dhellmann, line 166: And this block, unlike the others, does additional work so really needs to be highlighted more than the logging differences.

src/fromager/bootstrap_requirement_resolver.py
dhellmann: I think here you want to leave out the "filtered by max-release-age" part. That should already have been logged closer to where that happened, and here you want to log what is about to happen here.

src/fromager/bootstrap_requirement_resolver.py
rd4398: Hmm, okay. I have made changes in latest commit

src/fromager/bootstrap_requirement_resolver.py
rd4398: Agreed. Fixed in latest commit

src/fromager/bootstrap_requirement_resolver.py
rd4398: Makes sense. Changed in latest commit

src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -176,9 +154,74 @@ def _resolve(
...
+                    req.name,
+                    self.cache_wheel_server_url,
+                )
+                results = self._resolve_from_cache_server(req)
+
+            if not results:
+                logger.warning(
+                    "%s: resolver returned no results with "
+                    "fallback wheel server URL %s and %s version mode",
+                    req.name,
```
dhellmann: nit: I really think you don't need to pass the requirement name to the logger explicitly because the code in https://github.com/python-wheel-build/fromager/blob/main/src/fromager/log.py should be setting that up automatically. Is that not happening?

src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -176,9 +154,74 @@ def _resolve(
...
+                    "multiple" if self.multiple_versions else "single",
+                )
+
+        # Only cache non-empty results.
+        if results:
+            self.cache_resolution(req, pre_built, results)
+
+        if not results:
+            return []
+        return results if return_all_versions else [results[0]]
```
dhellmann: nit: I think I would have split this line up to be consistent with the style on line 186, but that's not worth holding this up over.

src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -176,9 +154,74 @@ def _resolve(
...
+        Returns at most one version so that transitive dependencies are
+        re-processed without rebuilding every old version.
+        """
+        logger.info(
+            "%s: checking cache server %s for existing wheel",
+            req.name,
+            self.cache_wheel_server_url,
+        )
+        try:
+            provider = finders.PyPICacheProvider(
```
dhellmann: I don't remember if the default settings look for both sdist and wheels, but I think we want both here.

rd4398: `PyPICacheProvider` enforces that include_sdists and include_wheels are mutually exclusive. We are now making two separate queries (one for wheels, one for sdists) and picking the newest result across both. Let me know if this is not something we want

More context:
```
The default is include_sdists=False, include_wheels=True — and they're mutually exclusive (line 50-53 raises if both are the same). So we can't pass both True. But the cache server could have sdists too. We need two lookups
```

src/fromager/bootstrap_requirement_resolver.py
rd4398: Agreed, done in latest commit

src/fromager/bootstrap_requirement_resolver.py
rd4398: Yeah, you are right. I missed that part. Fixed in latest push

rd4398:
@dhellmann I think addressed all the comments. Can you please take a look when time permits?

Also, @LalatenduMohanty would appreciate any feedback that you might have as well

src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -176,10 +154,76 @@ def _resolve(
...
+            if not results and self.multiple_versions and self.cache_wheel_server_url:
+                logger.info(
+                    "no results found with normal resolution, "
+                    "falling back to the cache server %s",
+                    self.cache_wheel_server_url,
+                )
+                results = self._resolve_from_cache_server(req)
+
+            if not results:
+                logger.warning(
```
dhellmann, line 174: We can end up here without trying the fallback server, so this log message might be misleading in cases where multiple_versions is false and there are no results.

src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -40,15 +40,24 @@ def __init__(
...
         prev_graph: DependencyGraph | None = None,
+        multiple_versions: bool = False,
+        cache_wheel_server_url: str = "",
     ) -> None:
         """Initialize requirement resolver.
 
         Args:
             ctx: Work context with constraints and settings
             prev_graph: Optional previous dependency graph for caching
+            multiple_versions: If ``True``, age filtering returns an empty
```
dhellmann: I think the code does the right thing, but this comment describes it incorrectly.

If multiple_versions is True and we find no results through any other approach we take the latest candidate in the cache server, ignoring the age filters. In all other cases, if we find no candidates we return an empty list.

dhellmann (commented):
Two small changes about the accuracy of comments and log messages, but otherwise this looks ready.

src/fromager/bootstrap_requirement_resolver.py
rd4398: fixed in latest commit

src/fromager/bootstrap_requirement_resolver.py
rd4398: Good catch, I have made changes in latest push

dhellmann (approved):

rd4398:
@mergify rebase

PR #1164, opened 2026-05-19
test(bootstrapper): fix caplog assertion that depends on global pytes…

### LINKED ISSUE #1092: tests: caplog assertion depends on global pytest --log-level
`tests/test_bootstrapper.py::test_cache_lookup_resolver_exception_logs_info` currently assumes that `INFO` logs are being captured globally by pytest.

The test calls `_download_wheel_from_cache()` and then asserts:

```python
assert "did not find wheel for" in caplog.text
```

But the code under test logs that message at `INFO` level:

```python
except ResolverException:
    logger.info(
        f"did not find wheel for {resolved_version} in {self.cache_wheel_server_url}"
    )
```

The test does not call `caplog.set_level(...)` or `caplog.at_level(...)`, so it only passes when the test runner is configured with a sufficiently low global log level.

Upstream CI currently masks this because the Hatch test script runs pytest with:

```bash
pytest --log-level DEBUG
```

Downstream runners that invoke pytest with default logging behavior can fail on this test. For example, nixpkgs' `pytestCheckHook` reproduces:

```text
FAILED tests/test_bootstrapper.py::test_cache_lookup_resolver_exception_logs_info
E   AssertionError: assert 'did not find wheel for' in ''
```

I think the test should request the log level it needs explicitly rather than depending on a runner-wide pytest flag, e.g. with something like:

```python
with caplog.at_level(logging.INFO, logger="fromager.bootstrapper"):
    ...
```

That would make the test self-contained and avoid CI-specific behavior differences.

shifa-khan:
Use caplog.at_level(logging.INFO) for the bootstrapper logger so the test doesn't rely on --log-level DEBUG being passed by the runner.

Closes: #1092

### DIFF
--- a/tests/test_bootstrapper.py
+++ b/tests/test_bootstrapper.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import pathlib
 import typing
 from unittest.mock import Mock, patch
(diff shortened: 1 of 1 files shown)


rd4398 (approved):
LGTM

PR #1169, opened 2026-05-26
feat(sdist): add helpers to normalize git clones and tarballs into va…

### LINKED ISSUE #554: Add helpers to turn git checkout / tar ball into valid sdist
A Python sdist has a specific naming convention, directory structure, and an additional file. A git checkout or a tar ball from a git commit / GitHub release are not valid Python sdists. We have a few cases where we download a tag or commit directly from GitHub or `tar` a local git clone. These tar balls work as sdists in some cases but break in other cases.

## Suggestion

Add a helper method that turns a directory into a (mostly) valid sdist. Perhaps add a second function that automates unpacking a tar ball and creating a new tar ball.

- output file name must be `{name}-{version}.tar.gz` with correct normalization of `name` and `version`
- tar ball must contain a single top-level directory with name `{name}-{version}`
- tar ball must contain `{name}-{version}/PKG-INFO` file with at least `Metadata-Version`, `Name`, and `Version` entries. Metadata version should be 2.2. Later versions require more fields.
- **SPECIAL CASE** if the project needs `build_dir`, then also place a PKG-INFO file in `{name}-{version}/{build_dir}/PKG-INFO`, so `setuptools-scm` can find it.
- perhaps deprecate `download_source.destination_filename`?

```
Metadata-Version: 2.2
Name: {req.name}
Version: {version}
Summary: Dummy PKG-INFO from Fromager
```

## References

- https://packaging.python.org/en/latest/specifications/source-distribution-format/
- https://setuptools-scm.readthedocs.io/en/latest/usage/#builtin-mechanisms-for-obtaining-version-numbers
- https://github.com/pypa/setuptools-scm/issues/806

jlarkin09:
Add sdist.make_sdist_directory() and sdist.repack_as_sdist() helpers that normalize git clones and non-standard tarballs into valid sdist layouts. Bump stub PKG-INFO from Metadata-Version 1.0 to 2.2 and integrate into the git clone prepare_source path and default_build_sdist.

Git clones and GitHub release tarballs lack PKG-INFO, causing setuptools-scm failures when PEP 517 hooks run during PREPARE_BUILD. Today ensure_pkg_info only runs during build_sdist -- one step too late. These helpers create a valid-enough sdist layout earlier in the pipeline using only (name, version).

Closes #554

### DIFF
--- a/src/fromager/sdist.py
+++ b/src/fromager/sdist.py
@@ -1,0 +1,157 @@
+"""Helpers to normalize arbitrary source directories into valid sdist layout.
+
+Git clones and non-standard tarballs (e.g. GitHub release assets) lack the
+``PKG-INFO`` file and standardized directory naming that PEP 517 build
+backends expect.  These helpers bridge that gap *before* the build backend
+is available -- only the package name and version are required.
+"""
+
+from __future__ import annotations
+
+import logging
+import pathlib
+import shutil
+import tarfile
+
+from packaging.version import Version
+
+from . import dependencies, overrides, tarballs
+
+logger = logging.getLogger(__name__)
+
+PKG_INFO_TEMPLATE = """\
+Metadata-Version: 2.2
+Name: {name}
+Version: {version}
+Summary: {summary}
+"""
+
+
+def _write_pkg_info(
+    directory: pathlib.Path,
+    name: str,
+    version: Version,
+) -> pathlib.Path:
+    """Write a stub ``PKG-INFO`` into *directory* if one does not exist.
+
+    Returns the path to the ``PKG-INFO`` file.
+    """
+    pkg_info_file = directory / "PKG-INFO"
+    if not pkg_info_file.is_file():
+        logger.info("writing stub PKG-INFO in %s", directory)
+        pkg_info_file.write_text(
+            PKG_INFO_TEMPLATE.format(
+                name=name,
+                version=str(version),
+                summary=dependencies.STUB_PKG_INFO_SUMMARY,
+            )
+        )
+    return pkg_info_file
+
+
+def make_sdist_directory(
+    source_dir: pathlib.Path,
+    name: str,
+    version: Version,
+    *,
+    build_dir: pathlib.Path | None = None,
+) -> pathlib.Path:
+    """Normalize *source_dir* into a valid sdist directory layout.
+
+    The directory is renamed to ``{normalized_name}-{version}`` (using
+    :func:`~fromager.overrides.pkgname_to_override_module`) and a stub
+    ``PKG-INFO`` is written if missing.  When *build_dir* differs from
+    the source root a second ``PKG-INFO`` is placed there for
+    ``setuptools-scm`` compatibility.
+
+    Args:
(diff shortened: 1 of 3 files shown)


src/fromager/sdist.py
```diff
@@ -0,0 +1,147 @@
...
+import tarfile
+
+from packaging.version import Version
+
+from . import dependencies, overrides, tarballs
+
+logger = logging.getLogger(__name__)
+
+PKG_INFO_TEMPLATE = """\
+Metadata-Version: 2.2
```
tiran, line 23: Is there any reason to bump the metadata version?

tiran (changes requested):
Unfortunately this approach is not going to work. We have several packages that have a non-standard package layout and do not support Python sdist format.

src/fromager/sdist.py
jlarkin09: @LalatenduMohanty  mentioned it in the issue https://github.com/python-wheel-build/fromager/issues/554#issuecomment-4509510321

jlarkin09: Which packages are affected and do they use custom build_sdist plugin overrides?

LalatenduMohanty: @jlarkin09 you should be able to run fromager+ with this PR against downstream bootstrap and build. We should atleast run an analysis to see if this will break some packages.

LalatenduMohanty: @tiran You asked to bump to 2.2 meta data version in you original issue description.

src/fromager/sources.py
```diff
@@ -698,31 +708,27 @@ def default_build_sdist(
...
+    :func:`pep517_build_sdist` instead.  Many packages assume the sdist
+    is built from a repository checkout and error out when the PEP 517
+    interface is used, so this function simply tars the tree back up.
+    """
+    sdist.make_sdist_directory(
+        sdist_root_dir,
+        req.name,
+        version,
+        build_dir=build_dir,
+    )
```
LalatenduMohanty, line 724: We are discarding return value. If the directory is renamed (source directory name doesn't match
  {normalized_name}-{version}), both sdist_root_dir and build_dir still point to the old, now-nonexistent paths.

LalatenduMohanty (changes requested):

src/fromager/sdist.py
jlarkin09: @LalatenduMohanty @tiran ran the non-accelerated/cpu-ubi9 collection on walkerpass against thisbranch. All packages bootstrapped and built successfully

jlarkin09:
@tiran should we scrap this PR?

jlarkin09:
The helpers only operate on the top-level source directory (rename + PKG-INFO injection). They don't modify the internal layout that build_dir packages rely on. We tested this against the downstream builder on walkerpass -- packages passed including packages with build_dir settings (triton, etc.). I also was able to point fromager at this branch for CI validation https://gitlab.com/redhat/rhel-ai/wheels/builder/-/pipelines/2577168867

Issue #1168, opened 2026-05-26
Support GitLab API archive download URLs with CI_JOB_TOKEN auth


tiran:
GitLab's web UI archive URLs (`/-/archive/{tag}/{name}-{tag}.tar.gz`) are not accessible with `CI_JOB_TOKEN` authentication. The API endpoint (`/api/v4/projects/:id/repository/archive.tar.gz?sha=<tag>`)
supports token-based auth but differs in behavior:

- The top-level directory includes the commit SHA (e.g. `project-v1.0.0-c0ec65db...`)
- The `+` character in tags (e.g. `v0.19.1+rhaiv.4`) must be URL-encoded as `%2B`

## Example

- old: `https://gitlab.com/my/example/project/-/tree/v0.1.0+local`
- new: `https://gitlab.com/api/v4/projects/my%2Fexample%2Fproject/repository/archive.tar.gz?sha=v0.1.0%2Blocal`

## Authentication

Fromager should also support the standard `python-gitlab` environment variables:

| Variable | Auth method |
|---|---|
| `GITLAB_PRIVATE_TOKEN` | `PRIVATE-TOKEN` header |
| `GITLAB_OAUTH_TOKEN` | `Authorization: Bearer` header |
| `CI_JOB_TOKEN` | `JOB-TOKEN` header |

smoparth:
Just wanted to mention a workaround we implemented for this. In Downstream, we recently added a CLI entrypoint with a side effect to support authentication for GitLab API calls when fetching the constraints file, based on your suggestion.

LalatenduMohanty:
> Just wanted to mention a workaround we implemented for this. In Downstream, we recently added a CLI entrypoint with a side effect to support authentication for GitLab API calls when fetching the constraints file, based on your suggestion.

@smoparth Can you please add some more context? It is not clear what workaround we added and why the workaround was necessary?

rd4398:
@tiran I see a related PR merged. Can we close this?

tiran:
The PR is related to this problem, but does not solve this problem.

jskladan:
[...]

PR #1173, opened 2026-05-27
feat(auth): add URL-dispatched session auth for GitHub and GitLab

tiran:
# Pull Request Description

## What

Add `SessionAuth` class that dispatches authentication by (scheme, hostname). Lazy callbacks resolve credentials from netrc or environment variables on first request and cache the result. `create_session()` returns the session and auth handler so callers can register additional hosts.

Move authentication documentation to docs/how-tos/authentication.md and update http-retry.md to reference the new guide.

## Why

see #1168

### DIFF
--- a/docs/http-retry.md
+++ b/docs/http-retry.md
@@ -20,7 +20,7 @@ The retry system provides:
 
 - **GitHub API rate limit handling** with proper reset time detection
 
-- **GitHub authentication** automatically applied for GitHub API requests via `GITHUB_TOKEN` environment variable
+- **Automatic authentication** for GitHub and GitLab APIs (see {doc}`how-tos/authentication`)
 
 - **Temporary file handling** to prevent partial downloads
 
@@ -37,10 +37,10 @@ export FROMAGER_HTTP_BACKOFF_FACTOR=2.0
 
 # Request timeout in seconds (default: 120)
 export FROMAGER_HTTP_TIMEOUT=180
-
-# Token for GitHub API authentication (prevents rate limiting)
-export GITHUB_TOKEN=your_github_token_here
 ```
+
+Authentication credentials (`GITHUB_TOKEN`, `GITLAB_PRIVATE_TOKEN`,
+etc.) are documented in {doc}`how-tos/authentication`.
 
 ## Error Types Handled
 
@@ -73,35 +73,27 @@ The retry functionality is automatically enabled for all HTTP operations in From
 
 ### For Plugin Developers
 
-If you're writing plugins that need HTTP functionality, you can use the retry session:
+If you're writing plugins that need HTTP functionality, use the
+shared session from `request_session`. It includes retry handling
+and automatic authentication for GitHub and GitLab:
 
 ```python
-from fromager.http_retry import get_retry_session
-
-# Get a session with retry capabilities
-session = get_retry_session()
+from fromager.request_session import session
 
 # Use it like a normal requests session
-response = session.get("https://example.com/api/data")
+response = session.get("https://pkg.test/api/data")
 response.raise_for_status()
 ```
 
-For more advanced retry configuration:
+To register authentication for additional hosts:
 
 ```python
-from fromager.http_retry import create_retry_session
+from fromager.request_session import session_auth
 
-# Custom retry configuration
-retry_config = {
-    "total": 5,
-    "backoff_factor": 2.0,
-    "status_forcelist": [429, 502, 503, 504],
-}
+def _resolve_my_auth(scheme: str, hostname: str) -> dict[str, str]:
+    return {"Authorization": "Bearer my-token"}
 
-session = create_retry_session(
-    retry_config=retry_config,
-    timeout=60.0
-)
+session_auth.add("https://my-registry.test", _resolve_my_auth)
 ```
 
 ### Decorating Functions with Retry Logic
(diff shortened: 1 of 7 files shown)


tiran:
The new code generalized authentication for GitHub and GitLab. We can drop a bunch of custom code from downstream.

ryanpetrello (commented):
The caching/lazy mechanism makes sense to me as I read it, but I'm wondering if the complexity is worth it in terms of what we're saving/avoiding here.  I had to step through this and think about it to make sure I understood - for example - the cache invalidation.

Without the caching, we're basically re-reading a `.netrc` on every request.  Is that it?

If so, would it be simpler to just make that file read a singleton and change our callback signature so that we pass in the parsed netrc as a parameter (psuedo-code ahead)

```python
class SessionAuth(requests.auth.AuthBase):

    def __init__(self) -> None:
        self._netrc = get_netrc_auth()

    def get(self, url: str) -> dict[str, str]:
        ...
        callback = self._callbacks.get(key)
        return callback(*key, self._netrc) if callback else {}
```

ryanpetrello:
I should also say, @tiran I don't feel particularly strongly about https://github.com/python-wheel-build/fromager/pull/1173#pullrequestreview-4372946735

If you're happy with the implementation as-is, I'm happy to give a thumbs up 👍 (though my approval is non-voting on merge).

This in particular:

> We can drop a bunch of custom code from downstream.

...makes me inclined to say "ship it".

src/fromager/request_session.py
```diff
@@ -11,8 +21,128 @@
...
+    token = os.environ.get("CI_JOB_TOKEN")
+    if token:
+        logger.debug("GitLab auth: using CI_JOB_TOKEN environment variable")
+        return {"JOB-TOKEN": token}
+
+    token = os.environ.get("GITLAB_PRIVATE_TOKEN")
+    if token:
+        logger.debug("GitLab auth: using GITLAB_PRIVATE_TOKEN environment variable")
+        return {"PRIVATE-TOKEN": token}
+    return {}
```
rd4398, line 118: I think we should also add `GITLAB_OAUTH_TOKEN`  here right? Did we leave it intentionally?

docs/http-retry.md
```diff
@@ -151,13 +143,13 @@ INFO saved /path/to/package.tar.gz
...
 
-- Setting `GITHUB_TOKEN` for GitHub API calls (automatically applied to GitHub requests)
+- Configuring authentication credentials (see {doc}`how-tos/authentication`) to avoid API rate limits
 - Increasing timeout values for slow connections
 - Checking network connectivity and DNS resolution
 
 ### API Rate Limiting
 
-- Use `GITHUB_TOKEN` for GitHub repositories
+- Configure credentials via netrc or environment variables (see {doc}`how-tos/authentication`)
```
rd4398, line 152: Can we also update the `http_retry.py` file here https://github.com/python-wheel-build/fromager/blob/33420105c1322deb48f1fd3fa7f9acdfa764b102/src/fromager/http_retry.py#L249

Since this PR centralizes auth documentation, this error message could point users to the authentication guide or at least mention netrc as an alternative, to stay consistent with the new approach.

rd4398 (commented):
Looks overall good. I have left a couple of comments

tiran:
> The caching/lazy mechanism makes sense to me as I read it, but I'm wondering if the complexity is worth it in terms of what we're saving/avoiding here. I had to step through this and think about it to make sure I understood - for example - the cache invalidation.
> 
> Without the caching, we're basically re-reading a `.netrc` on every request. Is that it?

Kind of. The two callbacks use requests' utils to locate, load, and parse a netrc file, then find a matching entry for an URL. In the future, callbacks may do other things that do not involve a netrc file.

rd4398:
> The caching/lazy mechanism makes sense to me as I read it, but I'm wondering if the complexity is worth it in terms of what we're saving/avoiding here. I had to step through this and think about it to make sure I understood - for example - the cache invalidation.
> 
> Without the caching, we're basically re-reading a `.netrc` on every request. Is that it?
> 
> If so, would it be simpler to just make that file read a singleton and change our callback signature so that we pass in the parsed netrc as a parameter (psuedo-code ahead)
> 
> ```python
> class SessionAuth(requests.auth.AuthBase):
> 
>     def __init__(self) -> None:
>         self._netrc = get_netrc_auth()
> 
>     def get(self, url: str) -> dict[str, str]:
>         ...
>         callback = self._callbacks.get(key)
>         return callback(*key, self._netrc) if callback else {}
> ```

+1 to what @ryanpetrello commented. 

Alternate suggestion:
```
drop ` _cache`  entirely and call the callback on every request. If profiling later shows netrc parsing is consuming resources, we could add caching then.
```

tiran:
What profiling? We don't do performance profiling regularly.

docs/http-retry.md
tiran: None of our exception points the user to Fromager's documentation. In my opinion it is fair to assume that a user of Fromager can search our documentation without an explicit link to the docs.

src/fromager/request_session.py
tiran: I have omitted oauth because we have neither a use case for the feature nor a way to verify that it works. Our GitLab accounts only support PAT and job token.

rd4398: Right, makes sense

rd4398:
> What profiling? We don't do performance profiling regularly.

Yeah, I missed that part.

docs/http-retry.md
rd4398: okay, fair enough

rd4398 (approved):
LGTM!

tiran:
@Mergifyio rebase

PR #1176, opened 2026-06-01
Adding proposal for test-mode retrospectively

LalatenduMohanty:
As the proposal docs are are useful to understand the architecture behind the big features we are doing. So having a doc which explain the motivation and architecture behind the feature would be useful for future contributors.


rd4398 (approved):
LGTM!

PR #1185, opened 2026-06-02
fix(resolver): resolve sdist URL via source provider in cache fallback

### LINKED ISSUE #1184: Cache server fallback returns wheel URL as sdist source_url
## Bug

When multi-version bootstrap mode is active and age filtering removes all candidates, `_resolve_from_cache_server` in `BootstrapRequirementResolver` can return a wheel URL as the `source_url`. The URL gets written to `build-order.json` with `source_url_type: "sdist"` (determined by package config, not the actual URL content). Downstream consumers then fail on a `.whl` basename where they expect `.tar.gz`/`.zip`.

## Root Cause

`_resolve_from_cache_server` queries the cache server for both wheels and sdists, picking whichever has the higher version. When only a wheel exists on the cache server (no sdist), it returns the wheel URL. The caller uses this URL as `source_url` without checking whether it's actually an sdist — `source_url_type` is set by `sources.get_source_type()` which checks package config, not the URL.

## Reproducer

Run bootstrap with `ENABLE_MULTI_VERSION_BOOTSTRAP=true`. Any package with all candidates older than the age threshold (e.g., `flit-core`, last released >365 days ago) will trigger the fallback.

## Expected Behavior

Cache server fallback should only return sdist URLs. If the cache server has only wheels for the package, it's already built and doesn't need to appear in the build order.

andre-motta:
## Summary

- `_resolve_from_cache_server` now queries the cache server for both wheels and sdists to find the newest version, then re-resolves the sdist URL through `sources.get_source_provider()`
- This uses the normal source resolution path including overrides (`get_resolver_provider` hook), custom download URLs, and `resolver_sdist_server_url` settings
- If the source provider cannot find an sdist for the pinned version, returns empty and logs a warning

Previously the method could return a wheel URL as `source_url`, which was written to `build-order.json` with `source_url_type: "sdist"`, causing downstream consumers to fail on a `.whl` basename.

Closes: #1184

## Test plan

- [x] `test_resolves_sdist_url_via_source_provider` — finds version on cache, resolves sdist via source provider
- [x] `test_returns_empty_when_source_provider_has_no_sdist` — cache has version but no sdist available
- [x] `test_returns_empty_when_cache_has_no_match` — cache server has nothing
- [x] `test_returns_empty_on_cache_exception` — cache server query fails
- [x] Integration tests for `resolve()` fallback path unchanged
- [x] All 28 resolver tests pass, mypy and lint clean

### DIFF
--- a/src/fromager/bootstrap_requirement_resolver.py
+++ b/src/fromager/bootstrap_requirement_resolver.py
@@ -168,7 +168,7 @@ class BootstrapRequirementResolver:
                     "falling back to the cache server %s",
                     self.cache_wheel_server_url,
                 )
-                results = self._resolve_from_cache_server(req)
+                results = self._resolve_from_cache_server(req, req_type)
 
             if not results:
                 logger.warning(
@@ -188,11 +188,16 @@ class BootstrapRequirementResolver:
             return results
         return [results[0]]
 
-    def _resolve_from_cache_server(self, req: Requirement) -> list[tuple[str, Version]]:
+    def _resolve_from_cache_server(
+        self, req: Requirement, req_type: RequirementType
+    ) -> list[tuple[str, Version]]:
         """Fall back to the remote wheel cache server for a cached version.
 
         When age filtering removes all candidates in multi-version mode,
-        queries the remote cache server for the newest available wheel.
+        queries the remote cache server (both wheels and sdists) to find
+        the newest available version, then re-resolves the sdist URL
+        through the normal source provider (including overrides).
+
         Returns at most one version so that transitive dependencies are
         re-processed without rebuilding every old version.
         """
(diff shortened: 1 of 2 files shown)


src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -189,40 +189,45 @@ def resolve(
...
 
         When age filtering removes all candidates in multi-version mode,
-        queries the remote cache server for the newest available wheel.
+        queries the remote cache server for the newest available sdist.
         Returns at most one version so that transitive dependencies are
         re-processed without rebuilding every old version.
+
+        Only sdists are returned.  If the cache server has only wheels for
+        the package, an empty list is returned because the package is
+        already built and does not need to appear in the build order.
```
dhellmann: I think we need more discussion of this change.

The existing implementation was meant to find a wheel that was already built so that during bootstrapping we can re-process its dependencies in case there are newer versions of those. So we treat it as not needing to be built, but we do look at the dependencies. Failing to do that would mean we fail to fully re-process the dependency graph and miss some updates that would be available to upstream users.

How would this change modify that behavior?

andre-motta: This came from a downstream issue on builder. 
When age filtering removes all candidates in multi-version mode, the fallback returns a wheel URL as `source_url`, but `source_url_type` in `build-order.json` is set to `"sdist"`
  (determined by package config, not the actual URL). 

The onboarder then fails trying to treat a `.whl` basename as a source archive. 
  
You're right that just returning empty would skip dependency reprocessing, which defeats the purpose. 

A better approach might be to still return the wheel URL when no sdist exists, but signal to the caller
  that it's a wheel so `source_url_type` gets set correctly: 

-  either by returning a richer type from `_resolve_from_cache_server` (e.g. including whether the result is a wheel or sdist) 
- or by having the caller inspect the URL extension before writing to `build-order.json`. 

That way dependency reprocessing still happens and downstream consumers get the correct `source_url_type`.

Since we already have a workaround in builder currently I can rework the PR to take this approach.

andre-motta: Updated the approach based on discussion with @rd4398 and @LalatenduMohanty. The method now queries the cache server for both wheels and sdists to find the version, then re-resolves the sdist URL through `sources.get_source_provider()`

this preserves dependency reprocessing while ensuring `source_url` always points to a real sdist.

See the discussion summary in #1184.

LalatenduMohanty: This looks good to me. taking a closer look again. Lets see what others has to say.

PR #1188, opened 2026-06-04
fix(resolver): cache known versions by package name

### LINKED ISSUE #1187: Release-age cooldown bypass ignored when package is first resolved as transitive dependency
## Summary

When a package is declared as a top-level requirement with an `==` pin (to bypass the release-age cooldown), the bypass is ignored if the same package was already resolved as a transitive dependency earlier in the bootstrap process. This is due to the session-level cache in `BootstrapRequirementResolver` not including `req_type` in its cache key.

## Root Cause

In `src/fromager/bootstrap_requirement_resolver.py`, the resolution cache key is `(str(req), pre_built)`:

```python
self._resolved_requirements: dict[
    tuple[str, bool], tuple[tuple[str, Version], ...]
] = {}
```

The `req_type` (TOP_LEVEL vs INSTALL) is not part of the cache key. When a package is first encountered as a transitive dependency (`RequirementType.INSTALL`), the cooldown is enforced and the result is cached. When the same package is later encountered as a top-level requirement (`RequirementType.TOP_LEVEL`) with an `==` pin, the cached (cooldown-blocked) result is returned without re-evaluating `resolve_package_cooldown()`.

## Reproduction

1. Package `B` depends on `A>=1.0`
2. Package `A==2.0.0` was published within the last X days (within cooldown window)
3. Declare both `A==2.0.0` and `B` as top-level requirements
4. If `B` is processed first, `A>=1.0` resolves as transitive → cooldown blocks version `2.0.0` → result cached
5. When `A==2.0.0` is later processed as top-level, cache returns the blocked result → bypass never applies

**Expected:** Top-level `A==2.0.0` should bypass the cooldown regardless of processing order.

**Actual:** Bypass only works if `A==2.0.0` is resolved before any transitive dependency pulls in `A`.

## Suggested Fix

Include `req_type` in the cache key:

```python
cache_key = (str(req), pre_built, req_type)
```

This ensures transitive and top-level resolutions are cached separately, preserving the `req_type` context needed for cooldown bypass decisions.

## Workaround

Move the cooldown-bypassing top-level requirement to the very first position in the requirements file so it is resolved before any transitive dependency can pull it in.

rd4398:
Replace the per-requirement-string resolution cache with a
package-level known-versions cache, as discussed in issue https://github.com/python-wheel-build/fromager/issues/1187.

The resolver now:
1. Tracks which requirement rules have been resolved to avoid
   redundant network calls.
2. Accumulates all discovered versions into a package-level cache
   keyed by (normalized_name, pre_built).
3. Filters the accumulated versions by the current specifier,
   returning the widest available set.

This ensures that versions discovered in one context (e.g., top-level
with cooldown bypass finding v2.0) are visible to later lookups with
different specifiers (e.g., transitive A>=1.0 will see v2.0 even
though its own network call only found v1.5).

Closes: https://github.com/python-wheel-build/fromager/issues/1187

### DIFF
--- a/src/fromager/bootstrap_requirement_resolver.py
+++ b/src/fromager/bootstrap_requirement_resolver.py
@@ -7,9 +7,11 @@ Git URL resolution stays in Bootstrapper (orchestration concern).
 from __future__ import annotations
 
 import logging
+import threading
 import typing
 
 from packaging.requirements import Requirement
+from packaging.utils import NormalizedName, canonicalize_name
 from packaging.version import Version
 
 from . import finders, resolver, sources, wheels
@@ -59,14 +61,18 @@ class BootstrapRequirementResolver:
         self.prev_graph = prev_graph
         self.multiple_versions = multiple_versions
         self.cache_wheel_server_url = cache_wheel_server_url
-        # Session-level resolution cache to avoid re-resolving same requirements
-        # Key: (requirement_string, pre_built) to distinguish source vs prebuilt
-        # Value: tuple of (url, version) tuples sorted by version (highest first)
-        # Values are stored as immutable tuples to prevent accidental corruption
-        # when callers modify the returned reference.
-        self._resolved_requirements: dict[
-            tuple[str, bool], tuple[tuple[str, Version], ...]
-        ] = {}
+        # All known versions for a package, accumulated across resolution
+        # contexts.  Versions discovered via different specifiers or req_types
+        # are merged so that later lookups see the widest set.
+        # Key: (normalized_name, pre_built)
+        # Value: {version: url}
+        self._known_versions: dict[tuple[NormalizedName, bool], dict[Version, str]] = {}
+        # Requirement rules already resolved from network/graph.
+        # Key: (str(req), pre_built)
+        # Prevents redundant network calls for the same specifier.
+        self._resolved_rules: set[tuple[str, bool]] = set()
+        # Protects _known_versions and _resolved_rules for thread safety.
+        self._lock = threading.Lock()
 
     def resolve(
         self,
@@ -78,12 +84,16 @@ class BootstrapRequirementResolver:
     ) -> list[tuple[str, Version]]:
         """Resolve package requirement to matching version(s).
 
-        Tries resolution strategies in order:
-        1. Session cache (if previously resolved)
-        2. Previous dependency graph
-        3. PyPI resolution (source or prebuilt based on package build info)
-        4. Remote wheel cache server (multi-version mode only, when age
-           filtering produced no candidates)
+        Uses a two-step strategy:
+        1. If this requirement rule has not been resolved before, fetch
+           versions from the network (or previous graph) and extend the
+           package-level known-versions cache.
+        2. Filter all known versions for the package by the current
+           requirement specifier.
+
+        This ensures that versions discovered in one context (e.g.,
+        top-level with cooldown bypass) are visible to later lookups
+        with different specifiers for the same package.
 
         Args:
             req: Package requirement
@@ -116,12 +126,34 @@ class BootstrapRequirementResolver:
                 f"Git URL requirements must be handled by Bootstrapper: {req}"
             )
 
-        # Check session cache (keyed by requirement + pre_built)
-        cached_result = self.get_cached_resolution(req, pre_built)
-        if cached_result is not None:
-            logger.debug(f"resolved {req} from cache")
-            return list(cached_result) if return_all_versions else [cached_result[0]]
+        rule_key = (str(req), pre_built)
 
+        with self._lock:
+            if rule_key not in self._resolved_rules:
+                # Rule not seen before — resolve from graph or network and
+                # extend the package-level known-versions cache.
+                self._resolve_and_extend(req, req_type, pre_built, parent_req)
+                self._resolved_rules.add(rule_key)
+            else:
+                logger.debug(f"rule already resolved: {req}")
+
+            # Filter all known versions by the current requirement specifier.
+            matching = self._get_matching_versions(req, pre_built)
+
+        if not matching:
+            return []
+        if return_all_versions:
+            return matching
+        return [matching[0]]
+
+    def _resolve_and_extend(
+        self,
+        req: Requirement,
+        req_type: RequirementType,
+        pre_built: bool,
+        parent_req: Requirement | None,
+    ) -> None:
+        """Resolve versions from graph/network and extend known versions cache."""
         # Try previous dependency graph
         cached_resolution = self._resolve_from_graph(
             req=req,
@@ -178,15 +210,8 @@ class BootstrapRequirementResolver:
                     "multiple" if self.multiple_versions else "single",
                 )
 
-        # Only cache non-empty results.
         if results:
-            self.cache_resolution(req, pre_built, results)
-
-        if not results:
-            return []
-        if return_all_versions:
-            return results
-        return [results[0]]
+            self._extend_known_versions(req, pre_built, results)
 
     def _resolve_from_cache_server(self, req: Requirement) -> list[tuple[str, Version]]:
         """Fall back to the remote wheel cache server for a cached version.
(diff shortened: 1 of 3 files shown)


src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -60,12 +60,13 @@ def __init__(
         self.multiple_versions = multiple_versions
         self.cache_wheel_server_url = cache_wheel_server_url
         # Session-level resolution cache to avoid re-resolving same requirements
-        # Key: (requirement_string, pre_built) to distinguish source vs prebuilt
+        # Key: (requirement_string, pre_built, req_type) to distinguish source
+        # vs prebuilt and top-level vs transitive (needed for cooldown bypass)
```
dhellmann: If all we need is top-level vs. other then maybe we just want a boolean? As it is now, if something shows up as different types of dependencies across different types of relationships we may resolve the rule several times.

rd4398: Yeah you are right. A boolean makes more sense here.  I have pushed the change in latest commit

dhellmann (dismissed):

LalatenduMohanty:
@dhellmann This alone might not fix the issue `A>=1.0 blocking A==2.0.0 top-level`. We need #1157 as well. But we should merge this first then merge #1157

tiran (dismissed):
IMO the suggestion from 1187 is not the best solution. Breadth-first resolution of top-level requirements should solve the issue, too.

rd4398:
> IMO the suggestion from 1187 is not the best solution. Breadth-first resolution of top-level requirements should solve the issue, too.

@tiran I agree that 1187 is not the best solution however, I think the BFS approach will make things more complicated. I see the following challenges:

1. Build dependency ordering constraint

```
DFS exists for a reason: to build packages in dependency order. Before you can
build package B, you need B's build dependencies (setuptools, wheel, etc.) to
be available as wheels. The DFS loop ensures that when `_phase_prepare_build`
installs build deps into the build environment, those wheels have already been
built.

A pure BFS approach would try to build all top-level packages simultaneously,
but their build dependencies might not be ready yet. This is the fundamental
constraint.

Impact: A naive BFS would require a two-pass approach:
- Pass 1: Resolve all top-level versions (already done)
- Pass 2: Build in topological order based on dependency graph

This is essentially what `bootstrap_parallel` already does (sdist-only serial
pass, then parallel wheel build).
```
2. Scope of changes

```
Changing from DFS to BFS in the iterative bootstrap loop would be a
significant refactor:

- The `WorkItem` phase machine and LIFO stack are designed for DFS
- Changing to a FIFO queue (BFS) would break build ordering guarantees
- The `_track_why` dependency chain tracking assumes DFS order
- The `_seen_requirements` deduplication assumes DFS order
- Progress tracking and logging assume DFS order

I wonder this will complicate the already existing bootstrap logic. A more targeted approach would be to add a resolution-only BFS pass before the DFS build pass, but this is essentially what pre-resolution already does.
```
3. BFS doesn't fully solve the problem

```
Even with BFS for top-level resolution, transitive dependencies of B can
still conflict with transitive dependencies of C. If B depends on `X>=1.0`
and C depends on `X>=2.0`, whichever is resolved first sets the cache. This
is a broader issue than just top-level vs transitive.
```
4. Multiple version mode

```
In `--multiple-versions` mode, the resolver returns ALL matching versions, not
just the highest. BFS resolution would need to handle this correctly — all
versions of all top-level deps resolved before any transitive deps.

This is already handled by pre-resolution (`resolve_and_add_top_level` with
`return_all_versions=self.multiple_versions`)
```

Some points to consider:

1. Is building two versions actually wrong? -- In single-version mode, both
   versions will be built, but downstream packages will use whichever version
   the dependency graph and constraints file specify. If A==2.0.0 satisfies
   `A>=1.0`, the constraints file should pick 2.0.0

2. The pre-resolution gap -- re-resolution caches `("A==2.0.0", False)`.
   The transitive dep is `("A>=1.0", False)` — a different cache key. These
   don't collide. The bug requires identical requirement strings. How often we will hit this scenario?

3. To avoid redundant builds, I think the fix could be at the
   `_phase_start` level: when a transitive resolution produces version X,
   check if a top-level resolution for the same package already produced
   version Y > X that satisfies the transitive specifier, and use Y instead.
   This is version reconciliation, not resolution order change.

4. If we all still want to do a BFS approach, the lowest refactor and less risky solution will be: resolve all top-level deps in pre-resolution (already done), then in the DFS build
 loop, have `_phase_resolve` check if the package was already resolved as top-level in pre-resolution and reuse that version when compatible. This avoids changing DFS→BFS entirely

Let me know what you think / whether I am missing anything. cc @LalatenduMohanty @dhellmann

rd4398:
> > IMO the suggestion from 1187 is not the best solution. Breadth-first resolution of top-level requirements should solve the issue, too.
> 
> @tiran I agree that 1187 is not the best solution however, I think the BFS approach will make things more complicated. I see the following challenges:
> 
> 1. Build dependency ordering constraint
> 
> ```
> DFS exists for a reason: to build packages in dependency order. Before you can
> build package B, you need B's build dependencies (setuptools, wheel, etc.) to
> be available as wheels. The DFS loop ensures that when `_phase_prepare_build`
> installs build deps into the build environment, those wheels have already been
> built.
> 
> A pure BFS approach would try to build all top-level packages simultaneously,
> but their build dependencies might not be ready yet. This is the fundamental
> constraint.
> 
> Impact: A naive BFS would require a two-pass approach:
> - Pass 1: Resolve all top-level versions (already done)
> - Pass 2: Build in topological order based on dependency graph
> 
> This is essentially what `bootstrap_parallel` already does (sdist-only serial
> pass, then parallel wheel build).
> ```
> 
> 2. Scope of changes
> 
> ```
> Changing from DFS to BFS in the iterative bootstrap loop would be a
> significant refactor:
> 
> - The `WorkItem` phase machine and LIFO stack are designed for DFS
> - Changing to a FIFO queue (BFS) would break build ordering guarantees
> - The `_track_why` dependency chain tracking assumes DFS order
> - The `_seen_requirements` deduplication assumes DFS order
> - Progress tracking and logging assume DFS order
> 
> I wonder this will complicate the already existing bootstrap logic. A more targeted approach would be to add a resolution-only BFS pass before the DFS build pass, but this is essentially what pre-resolution already does.
> ```
> 
> 3. BFS doesn't fully solve the problem
> 
> ```
> Even with BFS for top-level resolution, transitive dependencies of B can
> still conflict with transitive dependencies of C. If B depends on `X>=1.0`
> and C depends on `X>=2.0`, whichever is resolved first sets the cache. This
> is a broader issue than just top-level vs transitive.
> ```
> 
> 4. Multiple version mode
> 
> ```
> In `--multiple-versions` mode, the resolver returns ALL matching versions, not
> just the highest. BFS resolution would need to handle this correctly — all
> versions of all top-level deps resolved before any transitive deps.
> 
> This is already handled by pre-resolution (`resolve_and_add_top_level` with
> `return_all_versions=self.multiple_versions`)
> ```
> 
> Some points to consider:
> 
> 1. Is building two versions actually wrong? -- In single-version mode, both
>    versions will be built, but downstream packages will use whichever version
>    the dependency graph and constraints file specify. If A==2.0.0 satisfies
>    `A>=1.0`, the constraints file should pick 2.0.0
> 2. The pre-resolution gap -- re-resolution caches `("A==2.0.0", False)`.
>    The transitive dep is `("A>=1.0", False)` — a different cache key. These
>    don't collide. The bug requires identical requirement strings. How often we will hit this scenario?
> 3. To avoid redundant builds, I think the fix could be at the
>    `_phase_start` level: when a transitive resolution produces version X,
>    check if a top-level resolution for the same package already produced
>    version Y > X that satisfies the transitive specifier, and use Y instead.
>    This is version reconciliation, not resolution order change.
> 4. If we all still want to do a BFS approach, the lowest refactor and less risky solution will be: resolve all top-level deps in pre-resolution (already done), then in the DFS build
>    loop, have `_phase_resolve` check if the package was already resolved as top-level in pre-resolution and reuse that version when compatible. This avoids changing DFS→BFS entirely
> 
> Let me know what you think / whether I am missing anything. cc @LalatenduMohanty @dhellmann

UPDATE: I posted this before reading Doug's comment here: https://github.com/python-wheel-build/fromager/issues/1187#issuecomment-4632435708
I am okay with implementing what he suggested and will make changes to this PR

rd4398:
@dhellmann @LalatenduMohanty @tiran I have implemented Doug's suggestion which was discussed [here](https://github.com/python-wheel-build/fromager/issues/1187#issuecomment-4632435708) in my latest commit

src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -116,12 +123,33 @@ def resolve(
...
-        cached_result = self.get_cached_resolution(req, pre_built)
-        if cached_result is not None:
-            logger.debug(f"resolved {req} from cache")
-            return list(cached_result) if return_all_versions else [cached_result[0]]
+        rule_key = (str(req), pre_built)
+
+        if rule_key not in self._resolved_rules:
+            # Rule not seen before — resolve from graph or network and
+            # extend the package-level known-versions cache.
+            self._resolve_and_extend(req, req_type, pre_built, parent_req)
```
dhellmann: Since we're updating 2 data structures now, we need a lock to make sure it's thread safe.

rd4398: Good catch, I did not think about it. I have pushed this implementation in latest commit:

- `resolve()` holds `self._lock` around the check + extend + filter block
- `_resolve_and_extend()` and `resolve()` call the unlocked `_extend_known_versions()` / `_get_matching_versions()` (no nested
  locking)
- The public `extend_known_versions()` / `get_matching_versions()` acquire the lock themselves, for external callers in `bootstrapper.py`

Does this make sense?

src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -116,12 +126,34 @@ def resolve(
...
-        if cached_result is not None:
-            logger.debug(f"resolved {req} from cache")
-            return list(cached_result) if return_all_versions else [cached_result[0]]
+        rule_key = (str(req), pre_built)
+
+        with self._lock:
+            if rule_key not in self._resolved_rules:
+                # Rule not seen before — resolve from graph or network and
+                # extend the package-level known-versions cache.
+                self._resolve_and_extend(req, req_type, pre_built, parent_req)
```
dhellmann, line 135: This looks safe. If we were worried about the resolver running in multiple threads then we would probably want to not hold the lock for so long, because that would make this part of the code a bottleneck. We would want to try to organize the order of operations so that the resolution happened outside of the locked region and only operations to save state happen within the locked region. I don't think it's worth reorganizing the code to make that easy right now, though, because we are not using multiple threads to resolve rules.

src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -225,46 +250,78 @@ def _resolve_from_cache_server(self, req: Requirement) -> list[tuple[str, Versio
...
-            Tuple of (url, version) tuples if cached, None otherwise
+            List of (url, version) tuples matching the specifier.
         """
-        cache_key = (str(req), pre_built)
-        return self._resolved_requirements.get(cache_key)
+        with self._lock:
+            return self._get_matching_versions(req, pre_built)
 
-    def cache_resolution(
+    def _get_matching_versions(
```
dhellmann, line 273: nit: This doesn't really need to be a separate method.

dhellmann (commented):
This looks good to me but I'm reluctant to approve and merge it late on a Friday.

src/fromager/bootstrap_requirement_resolver.py
dhellmann: Nevermind, I see that you're using this method inside a lock context, so it does need to be separate.

src/fromager/bootstrap_requirement_resolver.py
rd4398: ack, makes sense.

rd4398:
> This looks good to me but I'm reluctant to approve and merge it late on a Friday.

@dhellmann can we merge this if possible?

dhellmann:
> > This looks good to me but I'm reluctant to approve and merge it late on a Friday.
> 
> @dhellmann can we merge this if possible?

Let's get @tiran to remove his requested change flag.

tiran:
> Let's get @tiran to remove his requested change flag.

I have dismissed my own change request.

dhellmann (approved):

rd4398:
@mergify rebase

PR #1189, opened 2026-06-05
feat(constraints): add per-specifier provenance tracking

### LINKED ISSUE #1186: Track per-constraint file provenance in merged output
## Problem

When multiple constraint files are passed via `-c` / `--constraints-file`, the merged output (`merged-constraints.txt`) only lists the source files in a header block:

```
# auto-generated constraints file
# /path/to/constraints-base.txt
# /path/to/constraints-override.txt

foo!=2.1.1,>=2.0
bar==3.2.0
```

When a build fails due to a constraint (e.g., resolver cannot find a version satisfying `foo!=2.1.1,>=2.0`), engineers have no easy way to determine **which input file(s)** contributed the offending specifier. They must manually search through all constraint files, which becomes increasingly difficult as teams scale to many files (per-platform, per-product, shared base, etc.).

## Proposed Solution

Enhance the `Constraints` class to track the originating file for each constraint entry, and surface that provenance in the merged output as inline comments.

### 1. Add provenance tracking to `Constraints`

Store the source file alongside each constraint. When specifiers from different files are merged (via `&`), accumulate all contributing files:

```python
# Sketch — exact implementation TBD
self._sources: dict[NormalizedName, list[str]] = {}
```

Update `add_constraint()` to accept an optional `source: str | None` parameter and record it. Update `load_constraints_file()` to pass the file path/URL as the source.

### 2. Annotate merged output with per-line provenance

Update `dump_constraints()` (or `write_constraints()` in `context.py`) to emit the origin as an inline comment:

```
# auto-generated constraints file
# /path/to/constraints-base.txt
# /path/to/constraints-override.txt

foo!=2.1.1,>=2.0  # from constraints-base.txt, constraints-override.txt
bar==3.2.0  # from constraints-base.txt
```

This keeps the file pip-compatible (inline `#` comments are valid in constraints/requirements files) while giving engineers immediate visibility into where each constraint originated.

### 3. Improve error messages

When `InvalidConstraintError` is raised due to a conflict, include the source files in the error message:

```
InvalidConstraintError: Combined specifier '<empty>' is not satisfiable
  (existing: foo<=1.1 from constraints-base.txt, new: foo>1.1 from constraints-override.txt)
```

This eliminates the need to search through files when diagnosing conflicts.

## Acceptance Criteria

- [ ] `Constraints` tracks which file(s) each constraint originated from
- [ ] `merged-constraints.txt` includes inline `# from <file>` comments per line
- [ ] Conflict error messages include source file names for both sides
- [ ] Tests and docs updated

## Notes

- Inline `#` comments are supported by both pip and uv, so this is backward-compatible.
- Provenance in `graph.json` (`DependencyNode.constraint`) could be a follow-up.

smoparth:
Track which constraint files contributed each package constraint so that
engineers can quickly identify the source of a conflicting specifier when
a build fails.

- Store provenance alongside constraints in a single `_data` dict as
  `dict[NormalizedName, tuple[Requirement, set[str]]]`
- Add optional `provenance` parameter to `add_constraint()`
- `load_constraints_file()` passes the file path as provenance
- `dump_constraints()` writes source files as comments above each line
- Add `get_constraint_with_provenance()` and `format_provenance()` methods
- Enrich `InvalidConstraintError` and resolver error messages with
  source file info

Closes: #1186

# Pull Request Description

## Why


src/fromager/constraints.py
```diff
@@ -81,43 +108,76 @@ def add_constraint(self, unparsed: str) -> None:
...
+    def get_provenance(self, name: str) -> dict[str, list[str]]:
+        """Return provenance info for *name*.
+
+        Returns:
+            Mapping of ``{source_file: [original_constraint_lines]}``,
+            or an empty dict if the package has no constraints.
+
+        .. versionadded:: 0.84.0
+        """
+        return dict(self._provenance.get(canonicalize_name(name), {}))
```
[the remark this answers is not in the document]
smoparth: Fixed!

tests/test_constraints.py
```diff
@@ -187,18 +191,117 @@ def test_blocked_package(specifier: str) -> None:
...
+        in result
+    )
+
+
+def test_conflict_error_includes_provenance() -> None:
+    """InvalidConstraintError message includes source file provenance."""
+    c = Constraints()
+    c.add_constraint("foo>=2.0", source="/constraints/base.txt")
+    with pytest.raises(InvalidConstraintError, match=r"base\.txt"):
+        c.add_constraint("foo<1.0", source="/constraints/override.txt")
```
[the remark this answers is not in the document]
smoparth: FIxed!

tests/test_constraints.py
```diff
@@ -187,18 +191,117 @@ def test_blocked_package(specifier: str) -> None:
...
+    assert c.get_provenance("bar") == {str(base): ["bar==1.0"]}
+
+
+def test_provenance_returns_copy() -> None:
+    """get_provenance returns a copy, not the internal dict."""
+    c = Constraints()
+    c.add_constraint("foo>=1.0", source="a.txt")
+    prov = c.get_provenance("foo")
+    prov["injected.txt"] = ["foo<99"]
+    assert "injected.txt" not in c.get_provenance("foo")
```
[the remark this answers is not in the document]
smoparth: Fixed!

src/fromager/constraints.py
```diff
@@ -46,13 +65,21 @@ def __bool__(self) -> bool:
     def __len__(self) -> int:
         return len(self._data)
 
-    def add_constraint(self, unparsed: str) -> None:
-        """Add new constraint, must not conflict with any existing constraints
+    def add_constraint(self, unparsed: str, *, source: str) -> None:
```
tiran: - Do we need to break the API here?
- `source` is a general, overloaded term. Why not use `provenance`?

```python
def add_constraint(self, unparsed: str, *, provenance: str | None = None) -> None:
```

src/fromager/constraints.py
```diff
@@ -36,6 +53,8 @@ def __init__(self) -> None:
         # mapping of canonical names to requirements
         # NOTE: Requirement.name is not normalized
         self._data: dict[NormalizedName, Requirement] = {}
+        # per-package provenance: {canonical_name: {source_file: [original_lines]}}
+        self._provenance: dict[NormalizedName, dict[str, list[str]]] = {}
```
tiran: How useful is it to track individual lines? In downstream, we are mostly interested in whether a constraints come from security, global constraints, or local constraints. Additional information is coming from comment lines near the constraint.

I think `dict[NormalizedName, (Requirement, set[str]]` is enough, where `set[str]` is a set of files.

src/fromager/constraints.py
```diff
@@ -81,43 +108,77 @@ def add_constraint(self, unparsed: str) -> None:
...
+           Output now includes per-line provenance comments.
+        """
+        # sort by normalized name, write requirement without markers.
+        # They have been evaluated in add_constraint()
+        for name, req in sorted(self._data.items()):
+            line = f"{req.name}{req.specifier}"
+            prov = self._provenance.get(name, {})
+            if prov:
+                line = f"{line}  # {_format_provenance(prov)}"
+            output.write(f"{line}\n")
```
tiran: I would rather have the provenance information in comments before the line:

```
# global.txt
# https://server.example/security.txt
# local.txt
egg>1.0,!=1.1
```

src/fromager/resolver.py
```diff
@@ -32,7 +32,7 @@
 
 from . import overrides
 from .candidate import Candidate, Cooldown
-from .constraints import Constraints
+from .constraints import Constraints, _format_provenance
```
tiran: I consider it bad style to leak private names between modules and packages.

tiran (changes requested):

src/fromager/constraints.py
smoparth: Good catch, I missed other teams using Fromager.

smoparth: Changed the param to provenance.

src/fromager/constraints.py
smoparth: I wrote it to track which particular version specifier came from which file but I have simplified it as you suggested. We can improve it later if there is a need.

src/fromager/constraints.py
smoparth: Fixed in latest commit

src/fromager/resolver.py
smoparth: Fixed it!

iangelak:
A minor thing. You should add some additional tests that are missing:

No test for provenance when `source/provenance` is None (relevant once the parameter becomes optional)
No test for provenance with marker-filtered constraints (constraint that doesn't match the current environment should not be tracked)
No test for provenance in the resolver error path (`find_all_matching_from_provider` with provenance in the exception)
No test for the `is_satisfied_by` debug logging path in `resolver.py`

src/fromager/constraints.py
```diff
@@ -46,13 +61,22 @@ def __bool__(self) -> bool:
...
-        """Add new constraint, must not conflict with any existing constraints
+    def add_constraint(self, unparsed: str, *, provenance: str | None = None) -> None:
+        """Add new constraint, must not conflict with any existing constraints.
 
-        .. versionchanged: 0.83.0
+        Args:
+            unparsed: Raw constraint string, e.g. ``"foo>=2.0"``.
+            provenance: Path or URL of the file that contains this
+                constraint. Used for provenance tracking in error messages
+                and merged output.
```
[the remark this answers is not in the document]
smoparth: The input link we pass is just the URI we pass the token later as part of request headers.

smoparth:
> A minor thing. You should add some additional tests that are missing:
> 
> No test for provenance when `source/provenance` is None (relevant once the parameter becomes optional) No test for provenance with marker-filtered constraints (constraint that doesn't match the current environment should not be tracked) No test for provenance in the resolver error path (`find_all_matching_from_provider` with provenance in the exception) No test for the `is_satisfied_by` debug logging path in `resolver.py`

The earlier design required provenance as a mandatory parameter. With the latest changes making it optional (None default), I've added test_add_constraint_without_provenance and test_conflict_error_without_provenance to cover that path.

For the resolver error and debug logging paths, format_provenance() is already unit-tested directly in test_format_provenance. Testing consumers of that method would be integration-level scope rather than provenance unit tests.

src/fromager/constraints.py
tiran: The code is still tracking the requirement and provenance information as two separate dicts.

- change `_data` to `dict[NormalizedName, (Requirement, set[str]]`
- replace `get_provenance` with `get_constraint_with_provenance(name: str) -> tuple[Requirement, set[str]] | tuple[None, None]`

tests/test_resolver.py
```diff
@@ -306,7 +306,7 @@ def test_provider_choose_sdist() -> None:
 
 def test_provider_choose_either_with_constraint() -> None:
     constraint = constraints.Constraints()
-    constraint.add_constraint("hydra-core==1.3.2")
+    constraint.add_constraint("hydra-core==1.3.2", provenance="test")
```
tiran: Could you please remove these unnecessary changes?

tiran (changes requested):

tests/test_resolver.py
smoparth: Yes, updated in the latest commit

src/fromager/constraints.py
smoparth: My bad, when I skimmed the comment I just observed the individual lines part and removed it. Now I have merged both into same data structure. Fixed!

smoparth:
@Mergifyio rebase

src/fromager/constraints.py
```diff
@@ -78,46 +100,106 @@ def add_constraint(self, unparsed: str) -> None:
...
+        self, name: str
+    ) -> tuple[Requirement, set[str]] | tuple[None, None]:
+        """Return the constraint and its provenance sources for *name*.
+
+        Returns:
+            ``(requirement, source_files)`` if constrained, or
+            ``(None, None)`` if the package has no constraints.
+            The returned set is a copy.
+
+        .. versionadded:: 0.84.0
```
LalatenduMohanty: We need to fix this to 0.93.0

src/fromager/constraints.py
```diff
@@ -78,46 +100,106 @@ def add_constraint(self, unparsed: str) -> None:
...
+
+    def format_provenance(self, name: str) -> str:
+        """Return a human-readable provenance string for *name*.
+
+        Returns:
+            Comma-separated list of source files, e.g.
+            ``"/path/to/base.txt, /path/to/override.txt"``,
+            or an empty string if the package has no constraints.
+
+        .. versionadded:: 0.84.0
```
LalatenduMohanty: Same here.

src/fromager/constraints.py
```diff
@@ -78,46 +100,106 @@ def add_constraint(self, unparsed: str) -> None:
...
-            # in add_constraint()
+        """Dump combined constraints to a text stream.
+
+        Source files that contributed each constraint are listed as comment
+        lines above the constraint line.
+
+        Args:
+            output: Writable text stream.
+
+        .. versionchanged:: 0.84.0
```
LalatenduMohanty: The version need to be fixed.

src/fromager/constraints.py
```diff
@@ -46,17 +59,26 @@ def __bool__(self) -> bool:
...
+                and merged output.
 
-        .. versionchanged: 0.83.0
+        .. versionchanged:: 0.83.0
            Non-conflicting constraints are now combined. Constraints with
            conflicts now raise :exc:`InvalidConstraintError`. Inputs without a
-           version specifier or with extras are also refused.
+           version specifier or with extras or url are also refused.
+
+        .. versionchanged:: 0.84.0
```
LalatenduMohanty: We need to fix this version

LalatenduMohanty (changes requested):
With provenance now tracked per-constraint, the header block in write_constraints() (context.py:193-196) is redundant with the per-line provenance from dump_constraints() (constraints.py:156-158). They write the same paths from two different data sources (self.input_constraints_files vs Constraints._data), 

Since the per-constraint provenance now subsumes the header's purpose, consider simplifying write_constraints() to just:
```                                                 
    f.write("# auto-generated constraints file\n\n")                                                                                                                         
    self.constraints.dump_constraints(f)  
```

smoparth:
> With provenance now tracked per-constraint, the header block in write_constraints() (context.py:193-196) is redundant with the per-line provenance from dump_constraints() (constraints.py:156-158). They write the same paths from two different data sources (self.input_constraints_files vs Constraints._data),
> 
> Since the per-constraint provenance now subsumes the header's purpose, consider simplifying write_constraints() to just:
> 
> ```
>     f.write("# auto-generated constraints file\n\n")                                                                                                                         
>     self.constraints.dump_constraints(f)  
> ```

The header serves as a quick summary of all input files in one place. Without it, we have to scan the entire file to know which files were merged. Happy to remove it if you feel strongly, but I think having both serves different purposes -- header for "what went in" and per-constraint comments for "where did this specific constraint come from."
cc: @LalatenduMohanty

src/fromager/resolver.py
```diff
@@ -293,10 +293,13 @@ def find_all_matching_from_provider(
         )
     except resolvelib.resolvers.ResolverException as err:
         constraint = provider.constraints.get_constraint(req.name)
+        prov_str = provider.constraints.format_provenance(req.name)
         provider_desc = provider.get_provider_description()
         original_msg = str(err)
+        prov_msg = f" (from {prov_str})" if prov_str else ""
         raise resolvelib.resolvers.ResolverException(
-            f"Unable to resolve requirement specifier {req} with constraint {constraint} using {provider_desc}: {original_msg}"
+            f"Unable to resolve requirement specifier {req} with constraint "
```
rd4398, line 301: Nit: Can we add a test in test_resolver.py (or test_constraints.py) that exercises the resolver err or path with a Constraints object containing provenance, and asserts the error message includes the source file info

rd4398 (dismissed):
Overall looks good! I added a non blocker nit which can be a follow up as well.

src/fromager/resolver.py
smoparth: I was avoiding integration tests in this PR but considering this was asked before too I have added the test for it.

rd4398 (approved):
LGTM

PR #1191, opened 2026-06-05
feat(resolver): report cooldown-skipped versions to work_dir

### LINKED ISSUE #1190: feat: report cooldown-skipped package versions to work_dir
## Summary

When `--min-release-age` (cooldown) is active, Fromager silently skips package versions that are "too new" during resolution. Currently this information is only visible in INFO/DEBUG logs — there is no persistent, structured summary of what was skipped.

Operators need a way to audit which versions of which packages were skipped due to cooldown, both for debugging resolution outcomes and for compliance/audit trails.

## Proposed Change

Add a `cooldown-skipped-versions.json` report written to `work_dir` after resolution completes, without changing cooldown filtering behavior.

### What gets collected

During resolution, each version blocked by cooldown in `BaseProvider.find_matches()` gets recorded with:
- Package name
- Version
- Upload time (ISO 8601 or null)
- Upload age in days
- Effective cooldown window in days
- Provider description

### Output format

```json
{
  "generated_at": "2026-06-05T14:30:00+00:00",
  "total_blocked": 3,
  "packages": {
    "requests": [
      {
        "version": "2.32.0",
        "upload_time": "2026-06-03T00:00:00+00:00",
        "upload_age_days": 2,
        "cooldown_min_age_days": 7,
        "provider": "PyPI (https://pypi.org/simple/)"
      }
    ],
    "urllib3": [...]
  }
}
```

### Implementation approach

1. **`src/fromager/resolver.py`** — Add a `CooldownReport` module-level collector (thread-safe, follows pattern of `BaseProvider._cooldown_unsupported_warned`). Modify `BaseProvider.find_matches()` to record blocked candidates and include the package identifier in the log message.

2. **`src/fromager/commands/bootstrap.py`** — Clear report at start, write after `bt.finalize()`.

3. **`src/fromager/commands/build.py`** — Clear report at start, write after resolution.

4. **`tests/test_cooldown.py`** — Tests for recording, deduplication, JSON output, and empty-report handling.

The file is only written when there are blocked versions — no clutter when cooldown is disabled.

### Secondary improvement

Enhance the existing log message from:
```
cooldown blocked 1 version(s): 2.0.0
```
to:
```
requests: cooldown blocked 1 version(s): 2.0.0
```

## Motivation

This came up while investigating build resolution outcomes — there was no easy way to answer "which packages had versions skipped due to cooldown, and which versions were they?" without parsing logs.

andre-motta:
## Summary

- Add `CooldownReport` collector and `CooldownBlockedEntry` dataclass to `resolver.py` to accumulate blocked versions during resolution
- Write `cooldown-skipped-versions.json` to `work_dir` after resolution in `bootstrap` and `build` commands (file only created when versions were actually blocked)
- Improve cooldown log message to include the package identifier
- Add 6 new tests covering recording, deduplication, JSON output, empty-report handling, and clear

## Motivation

When `--min-release-age` (cooldown) is active, skipped versions are only visible in INFO/DEBUG logs with no persistent summary. Operators need a structured way to audit which versions of which packages were skipped due to cooldown.

Closes #1190

## Test plan

- [x] All 42 cooldown tests pass (`hatch run test:test tests/test_cooldown.py`)
- [x] mypy clean (`hatch run mypy:check`)
- [x] Lint clean (`hatch run lint:fix && hatch run lint:check`)

### DIFF
--- a/src/fromager/resolver.py
+++ b/src/fromager/resolver.py
@@ -5,11 +5,15 @@
 #
 from __future__ import annotations
 
+import dataclasses
 import datetime
 import functools
+import json
 import logging
 import os
+import pathlib
 import re
+import threading
 import typing
 from collections.abc import Iterable
 from operator import attrgetter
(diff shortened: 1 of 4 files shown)


src/fromager/resolver.py
```diff
@@ -59,6 +63,77 @@
...
+    def clear(self) -> None:
+        with self._lock:
+            self._entries.clear()
+            self._seen.clear()
+
+    def write_to(self, path: pathlib.Path) -> None:
+        """Write JSON report to *path*. No-op when there are nothing to report."""
+        entries = self.entries()
+        if not entries:
+            return
```
dhellmann: The file wouldn't be created here. Should we throw an error or somehow otherwise indicate that? Or maybe even just create the file with an empty JSON list?

src/fromager/resolver.py
```diff
@@ -59,6 +63,77 @@
...
+            json.dump(report, f, indent=2)
+        logger.info(
+            "cooldown skipped %d version(s) across %d package(s); report written to %s",
+            len(entries),
+            len(packages),
+            path,
+        )
+
+
+cooldown_report = CooldownReport()
```
dhellmann, line 139: Using a global singleton like this makes me pause. I think it's probably safe, but I don't think we use that type of pattern anywhere else in the code so far (I could be wrong). Is there a way to have the bootstrapper manage the lifecycle of this object and pass it to the resolver? Or maybe something else can create it?

src/fromager/resolver.py
```diff
@@ -59,6 +63,77 @@
...
 
 
+@dataclasses.dataclass(frozen=True, slots=True)
+class CooldownBlockedEntry:
+    """Record of a single package version blocked by the cooldown policy."""
+
+    package: str
+    version: str
+    upload_time: str | None
+    upload_age_days: int | None
```
dhellmann: Why save both upload_time and upload_age_days? The display system can compute upload_age_days from the timestamp, right?

src/fromager/resolver.py
```diff
@@ -59,6 +63,77 @@
...
+        """Record a blocked entry, deduplicating by (package, version)."""
+        key = (entry.package, entry.version)
+        with self._lock:
+            if key not in self._seen:
+                self._seen.add(key)
+                self._entries.append(entry)
+
+    def entries(self) -> list[CooldownBlockedEntry]:
+        with self._lock:
+            return list(self._entries)
```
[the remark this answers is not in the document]
andre-motta: fixed

src/fromager/resolver.py
```diff
@@ -59,6 +63,77 @@
...
+
+    def clear(self) -> None:
+        with self._lock:
+            self._entries.clear()
+            self._seen.clear()
```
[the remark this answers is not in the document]
andre-motta: fixed

src/fromager/resolver.py
andre-motta: Agree created an empty file so that the behavior is consistent.

src/fromager/resolver.py
andre-motta: Agree, reduced the saved information to have any display system just compute it from timestamp!

src/fromager/resolver.py
andre-motta: If we do not want a singleton like this we could also do it via a few options:

1.  ClassVar on BaseProvider. Similar to what `resolver_cache` does
2. Add it on workcontext, but this would require a heavier change
3. Pass it directly through resolve but its also a big churn.

I think option 1 here is the simplest after the singleton.

dhellmann: OK, I don't have a strong objection to the singleton, it's just one of those patterns that makes me stop and look harder at the code, so I thought I'd bring it up. Let's see what the rest of the team thinks.

rd4398:
We decided on the issue that this will be done in upcoming log aggregation plan and will be defined more properly.
cc @andre-motta

PR #1193, opened 2026-06-09
fix(output): standardize --output to work for all formats

### LINKED ISSUE #1177: Standardize --output to work for all formats
### Problem

Commands that accept `--format` and `-o` handle file output inconsistently:

| Command | Plain format | `-o` with plain format |
|---------|--------------|------------------------|
| `find-updates` | `requirements` | Works |
| `package list-versions` (#1127) | `versions`, `requirements` | Warns and ignores |
| `list-overrides` | default (no `--details`) | Warns and ignores |

Raised in #1127 review. Only `list-overrides` uses warn-and-ignore today; `find-updates` is the reference.

### Proposed standard

Commands already share `--format` + `-o`. Users should not need to know which formats honor `-o` and which do not.

**`-o` means "write output here." Always.**

- `-o` set → write to file; stdout empty (logs on stderr only)
- `-o` omitted → write to stdout
- No silent or warn-and-ignore behavior

Reference: `src/fromager/commands/find_updates.py`

### Scope

1. **`package list-versions`** — wire `-o` into plain formats; remove #1127 warning
2. **`list-overrides`** — write plain name list to `-o`; remove warnings
3. **Tests** — `-o` creates file with expected content; no `-o` prints to stdout
4. **Docs** — one line in help or CONTRIBUTING stating the rule above

**Out of scope:** cooldown filtering behavior (#1078/#1127), commands without a format split (`build-order`, `graph`, `minimize`).

**Related:** #1165 covers output plumbing (`click.File`, dedupe writers); this issue defines the behavioral contract.

### Acceptance criteria

- [ ] Plain formats honor `-o` on `list-versions` and `list-overrides`
- [ ] No "`--output` option is ignored…" warnings
- [ ] `find-updates` unchanged
- [ ] Tests for plain-format file output

### Feedback welcome

This proposes one UX rule, but alternatives exist — please comment if you prefer a different approach:

- **Option A (proposed):** `-o` works for every format, matching `find-updates`
- **Option B:** keep plain formats stdout-only, but warn when `-o` is passed (current #1127 approach for `list-versions`)
- **Option C:** restrict `-o` to detail formats only (hide or reject `-o` for plain formats via CLI validation)

Happy to adjust before implementation. #1127 can land with a temporary compromise if needed; this issue tracks the longer-term standard.

### References

#1078 · #1127 · #1165

shifa-khan:
Wire -o into plain formats for `list-versions` (versions, requirements) and `list-overrides` (no --details), removing inconsistent warn-and-ignore behavior. Matches the find-updates reference implementation.

Closes: #1177

### DIFF
--- a/src/fromager/commands/package.py
+++ b/src/fromager/commands/package.py
@@ -121,7 +121,7 @@ def package() -> None:
     "-o",
     "--output",
     type=clickext.ClickPath(),
-    help="Output file (default: stdout)",
+    help="Write output to file instead of stdout",
 )
 @click.option(
     "--ignore-per-package-overrides",
@@ -228,18 +228,12 @@ def list_versions(
         provider.supports_upload_time,
     )
 
-    if output is not None and output_format in ("versions", "requirements"):
-        click.echo(
-            "Warning: --output option is ignored for 'versions' and 'requirements' formats",
-            err=True,
-        )
-
     match output_format:
         case "versions":
-            _export_versions_plain(version_rows, req.name, cooldown)
+            _export_versions_plain(version_rows, req.name, cooldown, output=output)
         case "requirements":
             _export_versions_plain(
-                version_rows, req.name, cooldown, as_requirements=True
+                version_rows, req.name, cooldown, as_requirements=True, output=output
             )
         case "table":
             _export_versions_table(version_rows, req.name, cooldown, output)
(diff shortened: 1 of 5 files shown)


src/fromager/commands/package.py
```diff
@@ -350,15 +344,26 @@ def _export_versions_plain(
...
     """Export versions as a plain list, filtering out cooldown-blocked entries."""
-    for row in data:
-        if cooldown is not None and row["cooldown"] == "blocked":
-            continue
-        if as_requirements:
-            print(f"{package_name}=={row['version']}")
-        else:
-            print(row["version"])
+    fh: typing.TextIO
+    if output:
```
dhellmann: I like the pattern used here, which avoids duplicating the logic to produce the output. Some of the other changes duplicate logic that is simple today but that could change over time.

I recommend you go 1 step further and create a [context manager function](https://pymotw.com/3/contextlib/index.html) that takes as input the path or None value and manages the file handle for you. That would eliminate another source of duplication and allow you to do something like

```
with get_file_handle_or_console(output) as fh:
    # logic to produce the output and send it to fh
```

shifa-khan: @dhellmann Thanks for the feedback. Updated in latest version.

dhellmann (approved):

rd4398:
@mergify rebase

PR #1196, opened 2026-06-11
fix(bootstrap): handle RESOLVE-phase errors in multiple versions mode

### LINKED ISSUE #1195: Multiple version bootstrap crashes on RESOLVE-phase failures instead of continuing
## Summary

In multiple version bootstrap mode (`--multiple-versions`), if version resolution fails for a dependency (e.g. all candidates filtered by `--max-release-age` and cache server fallback also finds no match), the entire bootstrap crashes instead of recording the failure and continuing with remaining packages.

## Root Cause

In `Bootstrapper._handle_phase_error()` (`src/fromager/bootstrapper.py`), RESOLVE-phase errors always re-raise unless in test mode. The `self.multiple_versions` error handler (which records failures and continues) is only reachable for build-phase errors — not for RESOLVE failures:

```python
# Resolution failures: only recoverable in test mode
if item.phase == BootstrapPhase.RESOLVE:
    if self.test_mode:
        self._record_test_mode_failure(item.req, None, err, "resolution")
        return []
    raise  # <-- crashes in multiple_versions mode too
```

The `if self.multiple_versions` block further down is never reached for RESOLVE-phase errors because the `raise` fires first.

## How to Reproduce

1. Run multiple version bootstrap with `--max-release-age` against a collection where a dependency (e.g. `importlib-metadata`) has all candidates older than the age cutoff
2. Ensure the cache server does not have a version matching the requirement specifier

Example log:
```
INFO importlib-metadata: found 32 candidate(s) matching importlib-metadata<8.8.0,>=6.0
INFO importlib-metadata: have 0 candidate(s) published within 90 days
INFO importlib-metadata: all 32 candidate(s) are older than 90 days
INFO importlib-metadata: no results found with normal resolution, falling back to the cache server
WARNING importlib-metadata: error checking cache server: found no match for importlib-metadata<8.8.0,>=6.0
WARNING importlib-metadata: resolver returned no results (multiple version mode)
ERROR its-hub: Could not resolve any versions for importlib-metadata<8.8.0,>=6.0
```

The bootstrap crashes at the ERROR line instead of recording the failure and continuing to process other packages.

## Expected Behavior

Multiple version bootstrap mode should handle RESOLVE-phase failures the same way it handles build-phase failures: record the failure, log a warning, and continue processing remaining packages. A summary of all failed versions should be displayed at the end.

rd4398:
In multiple version bootstrap mode, resolution failures (e.g. all candidates filtered by `--max-release-age` with no cache fallback) crashed the entire bootstrap instead of continuing with remaining packages.

The bug was in `_handle_phase_error()` where RESOLVE-phase errors always re-raised unless in test mode, preventing the `multiple_versions` handler from being reached. A second crash path existed in `resolve_and_add_top_level()` where empty results caused an `IndexError`.

Both paths now record failures in `_failed_versions`, log warnings, and continue processing. A rich table summary of all failed versions is displayed at the end via `finalize()`.

Closes: #1195

### DIFF
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -247,9 +247,25 @@ class Bootstrapper:
                     constraint=self.ctx.constraints.get_constraint(req.name),
                 )
 
+            if not results:
+                if self.multiple_versions:
+                    err = RuntimeError(f"no versions found for {req}")
+                    self._record_failed_version(
+                        req, "unresolved", err, "no versions resolved, skipping"
+                    )
+                    if self.test_mode:
+                        self._record_test_mode_failure(req, None, err, "resolution")
+                    return None
+                raise RuntimeError(f"Could not resolve any versions for {req}")
+
             # Return first result for backward compatibility
             return results[0]
         except Exception as err:
+            if self.multiple_versions:
+                self._record_failed_version(req, "unresolved", err, "failed to resolve")
+                if self.test_mode:
+                    self._record_test_mode_failure(req, None, err, "resolution")
+                return None
             if not self.test_mode:
                 raise
             self._record_test_mode_failure(req, None, err, "resolution")
@@ -1739,10 +1755,22 @@ class Bootstrapper:
         Returns work items to continue processing (e.g. prebuilt fallback),
         or empty list to skip this item. Raises in normal mode (fail-fast).
         """
-        # Resolution failures: only recoverable in test mode
+        # Resolution failures: recoverable in test mode and multiple versions mode
         if item.phase == BootstrapPhase.RESOLVE:
             if self.test_mode:
                 self._record_test_mode_failure(item.req, None, err, "resolution")
+                if self.multiple_versions:
+                    self._record_failed_version(
+                        item.req,
+                        "unresolved",
+                        err,
+                        f"failed during {item.phase} phase",
+                    )
+                return []
+            if self.multiple_versions:
+                self._record_failed_version(
+                    item.req, "unresolved", err, f"failed during {item.phase} phase"
+                )
                 return []
             raise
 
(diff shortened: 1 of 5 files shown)


src/fromager/bootstrapper.py
```diff
@@ -1796,15 +1831,39 @@ def _handle_phase_error(
         # Normal mode: fail-fast
         raise
 
+    def _log_failed_versions_table(self) -> None:
+        """Log a summary table of all failed versions."""
+        import rich
+        import rich.box
+        from rich.table import Table
```
dhellmann: Instead of using rich to print a table to the console, we should just log the information using the logger. That ensures it is captured with the rest of the log messages.

src/fromager/bootstrapper.py
```diff
@@ -247,9 +247,33 @@ def resolve_and_add_top_level(
                     constraint=self.ctx.constraints.get_constraint(req.name),
                 )
 
+            if not results:
+                if self.multiple_versions:
+                    pkg_name = canonicalize_name(req.name)
```
dhellmann: I feel like this error handling pattern repeats a few times. We could clean that up separately, maybe by adding a method that takes a requirement and optional exception object and does the name cleanup, appends the error, then logs a message.

src/fromager/bootstrapper.py
```diff
@@ -247,9 +247,33 @@ def resolve_and_add_top_level(
...
+                pkg_name = canonicalize_name(req.name)
+                self._failed_versions.append((pkg_name, "unresolved", err))
+                logger.warning(
+                    "%s: failed to resolve: %s: %s",
+                    req.name,
+                    type(err).__name__,
+                    err,
+                )
+                return None
             if not self.test_mode:
```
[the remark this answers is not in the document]
rd4398: fixed

src/fromager/bootstrapper.py
rd4398: Good idea, I have done it in my latest commit

src/fromager/bootstrapper.py
rd4398: okay, makes sense. I have done it in latest commit

dhellmann (commented):
This looks good to me but I'm reluctant to merge a logic change like this late on a Friday.

rd4398:
> This looks good to me but I'm reluctant to merge a logic change like this late on a Friday.

sounds good, can we merge this today please? cc @LalatenduMohanty

dhellmann (approved):

rd4398:
@mergify rebase

PR #1197, opened 2026-06-17
feat(resolver): add `versionmap-git` source resolver

### LINKED ISSUE #1180: VersionMap values should be git refs, not full URLs
## Summary

The current `VersionMap` implementation (`src/fromager/versionmap.py`) stores values as **full download URLs** (e.g. `https://example.com/pkg-1.2.tar.gz`), but the [new-resolver-config proposal](https://fromager.readthedocs.io/en/latest/proposals/new-resolver-config.html) specifies that `versionmap-git` values should be **git refs** — commit SHAs or ref paths — used alongside a separate `clone_url`.

## Proposal example (`docs/proposals/new-resolver-config.md`)

```yaml
source:
  provider: versionmap-git
  clone_url: https://git.example/viking/viking.git
  build_sdist: pep517
  versionmap:
    '1.0': abad1dea
    '1.1': refs/tags/1.1
```

Here the values are git refs (`abad1dea`, `refs/tags/1.1`), not URLs.

## Current behavior

`VersionMapProvider` in `src/fromager/resolver.py` wraps `VersionMap` and passes each value directly as a `url` to `Candidate`. The tests in `tests/test_resolver.py` confirm this — they use full URLs as values:

```python
version_map = VersionMap({
    "1.2": "https://example.com/pkg-1.2.tar.gz",
    "1.3": "https://example.com/pkg-1.3.tar.gz",
})
```

## What needs to change

Update `VersionMap` and `VersionMapProvider` so that the values represent **git refs** (commit IDs or tag refs) rather than full URLs, consistent with the `versionmap-git` provider design in the proposal. The `clone_url` should be a separate field on the provider, and the ref should be used to check out the correct commit after cloning.

Update existing tests to reflect the new value format.

jskladan:
Add `VersionMapGitResolver` config model for the `versionmap-git` provider from the new-resolver-config proposal. Maps version numbers to git refs (commit SHAs or ref paths) with a separate `clone_url`, replacing the need for handwritten Python plugins.

- Add `VersionMapGitResolver` pydantic model with `clone_url`, `build_sdist`, and `versionmap` fields
- Register in `SourceResolver` discriminated union
- Add to config reference docs
- Add tests for config parsing, provider output, and validation

Closes: #1180

### DIFF
--- a/docs/reference/config-reference.rst
+++ b/docs/reference/config-reference.rst
@@ -41,6 +41,8 @@ Source Resolver
 .. autopydantic_model:: PyPIGitResolver
    :inherited-members: AbstractPyPIResolver, CooldownMixin
 
+.. autopydantic_model:: VersionMapGitResolver
+
 .. autopydantic_model:: GitHubTagDownloadResolver
    :inherited-members: AbstractGitSourceResolver, CooldownMixin
 
--- a/src/fromager/packagesettings/_resolver.py
+++ b/src/fromager/packagesettings/_resolver.py
@@ -11,6 +11,7 @@ import pydantic
 
 from .. import resolver
 from ..candidate import Cooldown
+from ..versionmap import VersionMap
 from ._typedefs import MODEL_CONFIG
 
 if typing.TYPE_CHECKING:
@@ -233,7 +234,7 @@ class PyPIGitResolver(AbstractPyPIResolver):
     def validate_clone_url(cls, value: pydantic.AnyUrl) -> pydantic.AnyUrl:
         if value.scheme not in {"https", "ssh"}:
             raise ValueError(f"invalid scheme in url {value}")
-        if not value.path:
+        if not value.path or value.path == "/":
             raise ValueError(f"url {value} has an empty path")
         return value
 
@@ -482,6 +483,61 @@ class GitLabTagCloneResolver(AbstractGitSourceResolver):
         )
 
 
+class VersionMapGitResolver(AbstractResolver):
+    """Resolve version from a version map, build sdist from git clone.
+
+    The ``versionmap-git`` provider maps known version numbers to known git
+    refs (commit SHAs or ref paths such as ``refs/tags/1.1``).  It clones a
+    git repo at the configured ref and builds an sdist with PEP 517.
+
+    .. versionadded:: 0.79.0
+
+    Example::
+
+       provider: versionmap-git
+       clone_url: https://git.test/project/repo.git
+       build_sdist: pep517
+       versionmap:
+         '1.0': abad1dea
+         '1.1': refs/tags/1.1
+    """
+
+    provider: typing.Literal["versionmap-git"]
+
+    clone_url: pydantic.AnyUrl
+    """Git clone URL (``https`` or ``ssh`` scheme)."""
+
+    build_sdist: BuildSDist = BuildSDist.pep517
+    """Source distribution build method."""
+
+    versionmap: dict[str, str]
+    """Mapping of version strings to git refs."""
+
+    @pydantic.field_validator("clone_url", mode="after")
+    @classmethod
+    def validate_clone_url(cls, value: pydantic.AnyUrl) -> pydantic.AnyUrl:
+        if value.scheme not in {"https", "ssh"}:
+            raise ValueError(f"invalid scheme in url {value}")
+        if not value.path or value.path == "/":
+            raise ValueError(f"url {value} has an empty path")
+        return value
+
+    def resolver_provider(
+        self, ctx: context.WorkContext, req_type: requirements_file.RequirementType
+    ) -> resolver.VersionMapProvider:
+        clone_url = str(self.clone_url)
+        url_map = {
+            ver: f"git+{clone_url}@{ref}" for ver, ref in self.versionmap.items()
+        }
+        version_map = VersionMap(url_map)  # type: ignore[arg-type]
+        return resolver.VersionMapProvider(
+            version_map=version_map,
+            package_name=None,
+            constraints=ctx.constraints,
+            req_type=req_type,
+        )
+
+
 class NotAvailableResolver(AbstractResolver):
     """Prevent resolve and download"""
 
@@ -510,6 +566,7 @@ SourceResolver = typing.Annotated[
     | PyPIPrebuiltResolver
     | PyPIDownloadResolver
     | PyPIGitResolver
+    | VersionMapGitResolver
     | GitHubTagCloneResolver
     | GitHubTagDownloadResolver
     | GitLabTagCloneResolver
--- a/tests/test_packagesettings_resolver.py
+++ b/tests/test_packagesettings_resolver.py
@@ -25,6 +25,7 @@ from fromager.packagesettings._resolver import (
     PyPIPrebuiltResolver,
     PyPISDistResolver,
     SourceResolver,
+    VersionMapGitResolver,
 )
 from fromager.packagesettings._typedefs import MODEL_CONFIG
 from fromager.requirements_file import RequirementType
@@ -424,6 +425,50 @@ class TestGitLabTagCloneResolver:
 
 
 # -- Special resolvers --------------------------------------------------------
+
+
+class TestVersionMapGitResolver:
+    YAML = """\
+        source:
+          provider: versionmap-git
+          clone_url: https://git.test/project/repo.git
+          build_sdist: pep517
+          versionmap:
+            '1.0': abad1dea
+            '1.1': refs/tags/1.1
+    """
+
+    def test_parse(self) -> None:
+        r = _parse(self.YAML)
+        assert isinstance(r, VersionMapGitResolver)
+        assert r.provider == "versionmap-git"
+        assert str(r.clone_url) == "https://git.test/project/repo.git"
+        assert r.build_sdist == BuildSDist.pep517
+        assert r.versionmap == {"1.0": "abad1dea", "1.1": "refs/tags/1.1"}
+
+    def test_resolver_provider(self, tmp_context: WorkContext) -> None:
+        r = _parse(self.YAML)
+        p = r.resolver_provider(tmp_context, _REQ_TYPE)
+        assert isinstance(p, resolver.VersionMapProvider)
+        clone_url = "https://git.test/project/repo.git"
+        assert p.version_map["1.0"] == f"git+{clone_url}@abad1dea"
+        assert p.version_map["1.1"] == f"git+{clone_url}@refs/tags/1.1"
+
+    def test_clone_url_rejects_http(self) -> None:
+        with pytest.raises(pydantic.ValidationError):
+            VersionMapGitResolver(
+                provider="versionmap-git",
+                clone_url="http://git.test/project/repo.git",  # type: ignore[arg-type]
+                versionmap={"1.0": "abc123"},
+            )
+
+    def test_clone_url_rejects_empty_path(self) -> None:
+        with pytest.raises(pydantic.ValidationError):
+            VersionMapGitResolver(
+                provider="versionmap-git",
+                clone_url="https://git.test",  # type: ignore[arg-type]
+                versionmap={"1.0": "abc123"},
+            )
 
 
 class TestNotAvailableResolver:
--- a/src/fromager/packagesettings/__init__.py
+++ b/src/fromager/packagesettings/__init__.py
@@ -25,6 +25,7 @@ from ._resolver import (
     PyPIGitResolver,
     PyPIPrebuiltResolver,
     PyPISDistResolver,
+    VersionMapGitResolver,
     pep440_tag_matcher,
 )
 from ._settings import Settings, SettingsFile
@@ -82,6 +83,7 @@ __all__ = (
     "Variant",
     "VariantChangelog",
     "VariantInfo",
+    "VersionMapGitResolver",
     "default_update_extra_environ",
     "get_extra_environ",
     "pep440_tag_matcher",
--- a/tests/test_resolver.py
+++ b/tests/test_resolver.py
@@ -920,6 +920,64 @@ def test_resolve_versionmap_no_match() -> None:
         rslvr.resolve([Requirement("testpkg>=2.0")])
 
 
+def test_resolve_versionmap_git() -> None:
+    from fromager.versionmap import VersionMap
+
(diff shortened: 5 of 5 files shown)


src/fromager/packagesettings/_resolver.py
```diff
@@ -482,6 +482,63 @@ def resolver_provider(
...
+        if value.scheme not in {"https", "ssh"}:
+            raise ValueError(f"invalid scheme in url {value}")
+        if not value.path:
+            raise ValueError(f"url {value} has an empty path")
+        return value
+
+    def resolver_provider(
+        self, ctx: context.WorkContext, req_type: requirements_file.RequirementType
+    ) -> resolver.VersionMapProvider:
+        from ..versionmap import VersionMap
```
rd4398: Can we move this local import to the top?

src/fromager/packagesettings/_resolver.py
```diff
@@ -482,6 +482,63 @@ def resolver_provider(
...
+    """Mapping of version strings to git refs."""
+
+    @pydantic.field_validator("clone_url", mode="after")
+    @classmethod
+    def validate_clone_url(cls, value: pydantic.AnyUrl) -> pydantic.AnyUrl:
+        if value.scheme not in {"https", "ssh"}:
+            raise ValueError(f"invalid scheme in url {value}")
+        if not value.path:
+            raise ValueError(f"url {value} has an empty path")
+        return value
```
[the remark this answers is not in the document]
rd4398: This is a valid comment from code-rabbit and should be fixed.

docs/reference/config-reference.rst
```diff
@@ -41,6 +41,10 @@ Source Resolver
 .. autopydantic_model:: PyPIGitResolver
    :inherited-members: AbstractPyPIResolver, CooldownMixin
 
+.. autopydantic_model:: VersionMapGitResolver
+
+.. versionadded:: 0.79.0
```
rd4398: This needs to be indented so that it is rendered properly.

rd4398 (changes requested):
I have left a few comments. Also, please use the linked issue on GitHub as "closes". We should not mention anything related to downstream (AIPCC  in this case).

Also, I would appreciate if @tiran and / or @LalatenduMohanty takes a look as well.

tests/test_packagesettings_resolver.py
```diff
@@ -426,6 +427,50 @@ def test_resolver_provider(self, tmp_context: WorkContext) -> None:
...
+        assert r.build_sdist == BuildSDist.pep517
+        assert r.versionmap == {"1.0": "abad1dea", "1.1": "refs/tags/1.1"}
+
+    def test_resolver_provider(self, tmp_context: WorkContext) -> None:
+        r = _parse(self.YAML)
+        p = r.resolver_provider(tmp_context, _REQ_TYPE)
+        assert isinstance(p, resolver.VersionMapProvider)
+        clone_url = "https://git.test/project/repo.git"
+        assert p.version_map["1.0"] == f"git+{clone_url}@abad1dea"
+        assert p.version_map["1.1"] == f"git+{clone_url}@refs/tags/1.1"
```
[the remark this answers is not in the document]
jskladan: The `build_sdist` field on `VersionMapGitResolver` is a plain data attribute — it stores the user's config choice (`pep517` or `tarball`). It is **not consumed by `resolver_provider()`** and doesn't flow into `VersionMapProvider`. The build strategy is read later by the bootstrapper/build pipeline, completely separate from the resolver path.

The existing test coverage already verifies both behaviors of this class:

- `test_parse` (line 446): verifies `r.build_sdist == BuildSDist.pep517` (config parsing)
- `test_resolver_provider` (line 449): verifies URL rewriting into `VersionMapProvider`

There is no "build_sdist is ignored downstream" bug to regress against, because `resolver_provider()` is not supposed to use `build_sdist` at all. Adding a `build_sdist: tarball` variant to `test_resolver_provider` would only test that an unrelated field doesn't affect URL generation — not a meaningful regression test.

jskladan:
@rd4398 everything should be addressed. I'm also not sure what to do about the fialing coverage report though. 

I ran
```bash
hatch run test:coverage erase
hatch run test:test --cov --cov-report=term-missing --cov-config=pyproject.toml -x -q
```
on both my branch and the current main, and the branch **adds 70 statements and 4 branches** (new code in the resolver), while actually **reducing misses by 1**. Coverage is identical at 86% on both branches.

| Metric | `main` | `AIPCC-17610` | Delta |
|--------|--------|---------------|-------|
| **Total Stmts** | 14,087 | 14,157 | +70 |
| **Total Miss** | 1,617 | 1,616 | **-1** |
| **Total Branches** | 2,212 | 2,216 | +4 |
| **Partial Branches** | 271 | 270 | **-1** |
| **Overall Coverage** | **86%** | **86%** | **0%** |
| **Tests passed** | 800 | 806 | +6 |


I don't see any coverage regression on my part, but I obviously don't have all the e3e data the CI coverage job combines for the `--fail-under=60` threshold.

rd4398:
@jskladan this needs a rebase on top of main. We fixed this coverage issue couple of weeks back 
cc @jlarkin09

src/fromager/packagesettings/_resolver.py
```diff
@@ -482,6 +483,61 @@ def resolver_provider(
...
+        self, ctx: context.WorkContext, req_type: requirements_file.RequirementType
+    ) -> resolver.VersionMapProvider:
+        clone_url = str(self.clone_url)
+        url_map = {
+            ver: f"git+{clone_url}@{ref}" for ver, ref in self.versionmap.items()
+        }
+        version_map = VersionMap(url_map)  # type: ignore[arg-type]
+        return resolver.VersionMapProvider(
+            version_map=version_map,
+            package_name=None,
```
Lanceypantsy, line 535: **[Recommendation]**
- `src/fromager/packagesettings/_resolver.py` (~line 535, `resolver_provider` passes `package_name=None`)
- `tests/test_resolver.py` (~lines 938 & 972, tests pass `package_name="testpkg"`)

The code invocation always passes `package_name=None`. But the tests always pass `package_name="testpkg"`. So the tests are checking a situation that never actually happens. If the "no name" case is broken, every test still passes and nobody notices. 

Fix: add a test that uses the code the way it really runs — with `package_name=None` —
and actually resolves a package to make sure it works.

src/fromager/packagesettings/_resolver.py
```diff
@@ -233,7 +234,7 @@ class PyPIGitResolver(AbstractPyPIResolver):
     def validate_clone_url(cls, value: pydantic.AnyUrl) -> pydantic.AnyUrl:
         if value.scheme not in {"https", "ssh"}:
             raise ValueError(f"invalid scheme in url {value}")
-        if not value.path:
+        if not value.path or value.path == "/":
```
Lanceypantsy, line 237: **[Recommendation]**

- `src/fromager/packagesettings/_resolver.py` (~line 237, `PyPIGitResolver.validate_clone_url`)
- `src/fromager/packagesettings/_resolver.py` (~lines 518–523, `VersionMapGitResolver.validate_clone_url`)

The exact same check exists in two classes, word for word. This PR had to fix the same bug in both copies.

Fix: put the check in one shared helper function and have both classes call it.

Lanceypantsy, line 237: **[Recommendation]**

The PR's stated goal is adding a new resolver, but it also alters the existing `PyPIGitResolver` by adding the `value.path == "/"` rejection — a behavior change to shipped code. The diff only adds an empty-path test to the new class, so the fix on `PyPIGitResolver` ships with no regression test and a later refactor could silently revert it.

Fix: split the `PyPIGitResolver` change into its own commit/PR with a dedicated test, or keep
it here but add `test_clone_url_rejects_root_path` under the `PyPIGitResolver` suite.

src/fromager/packagesettings/_resolver.py
```diff
@@ -482,6 +483,61 @@ def resolver_provider(
...
+            raise ValueError(f"url {value} has an empty path")
+        return value
+
+    def resolver_provider(
+        self, ctx: context.WorkContext, req_type: requirements_file.RequirementType
+    ) -> resolver.VersionMapProvider:
+        clone_url = str(self.clone_url)
+        url_map = {
+            ver: f"git+{clone_url}@{ref}" for ver, ref in self.versionmap.items()
```
Lanceypantsy, line 530: **[Recommendation]** 

  The ref values from `versionmap` are dropped straight into `f"git+{clone_url}@{ref}"` with no validation. A ref containing `@`, whitespace, or a newline could produce a malformed URL. It's operator-controlled
  config so the risk is low, but a quick check would add defense-in-depth.

  Validate refs against a simple allowlist in a `field_validator("versionmap")`.

  **After:**
  ```python
  @pydantic.field_validator("versionmap", mode="after")
  @classmethod
  def validate_refs(cls, value: dict[str, str]) -> dict[str, str]:
      for ref in value.values():
          if not re.fullmatch(r"[\w./-]+", ref):
              raise ValueError(f"invalid git ref {ref!r}")
      return value

src/fromager/packagesettings/_resolver.py
```diff
@@ -482,6 +483,61 @@ def resolver_provider(
...
+
+    provider: typing.Literal["versionmap-git"]
+
+    clone_url: pydantic.AnyUrl
+    """Git clone URL (``https`` or ``ssh`` scheme)."""
+
+    build_sdist: BuildSDist = BuildSDist.pep517
+    """Source distribution build method."""
+
+    versionmap: dict[str, str]
```
Lanceypantsy, line 513: **[Recommendation]** — `src/fromager/packagesettings/_resolver.py` (~line 513, `versionmap` field) and `tests/test_packagesettings_resolver.py`

  An empty `versionmap: {}` passes validation, produces an empty `VersionMap`, and only fails later at resolve time with an opaque "no match" instead of a clear config error. There's also no positive test that the
  `ssh` scheme is accepted (only the http-rejection negative), and no test for the empty-map case.

  Reject an empty map with `min_length=1`, and add the two missing tests.

  **Before:**
  ```python
  versionmap: dict[str, str]
  ```

  **After:**
  ```python
  versionmap: dict[str, str] = pydantic.Field(min_length=1)
  ```

Lanceypantsy: There should also be a unit test case that ensure the empty `versionmap` dictionary fails.

src/fromager/packagesettings/_resolver.py
```diff
@@ -482,6 +483,61 @@ def resolver_provider(
         )
 
 
+class VersionMapGitResolver(AbstractResolver):
+    """Resolve version from a version map, build sdist from git clone.
+
+    The ``versionmap-git`` provider maps known version numbers to known git
+    refs (commit SHAs or ref paths such as ``refs/tags/1.1``).  It clones a
+    git repo at the configured ref and builds an sdist with PEP 517.
```
Lanceypantsy, line 491: **[Nit]** 

  The docstring says the provider "builds an sdist with PEP 517," which reads as fixed. But `build_sdist` is a configurable field where `pep517` is only the default — the wording will mislead anyone configuring a
  non-default build method.

  **Before:**
  ```python
  git repo at the configured ref and builds an sdist with PEP 517.
  ```

  **After:**
  ```python
  git repo at the configured ref and builds an sdist using the configured
  ``build_sdist`` method (default: PEP 517).
  ```

src/fromager/packagesettings/_resolver.py
Lanceypantsy, line 513: **[Nit]** 

  The field name `versionmap` breaks snake_case consistency with its siblings (`clone_url`, `build_sdist`), and the code internally uses `version_map`/`VersionMap`. Because the field name *is* the user-facing YAML
  key, the single-word form becomes a permanent API-consistency wart.

  **Before:**
  ```python
  versionmap: dict[str, str]
  ```
  
  **After:**
  ```python
  version_map: dict[str, str] = pydantic.Field(alias="versionmap")  # if the YAML key must stay
  ```

Lanceypantsy, line 513: **[Question]** 

  Are the `versionmap` keys validated as PEP 440 versions anywhere? They become version strings in `VersionMap`, so a typo key like `'v1.0'` or `'latest'` is accepted at config-load time and would only blow up
  later when parsed as a `Version`. If `VersionMap.__init__` doesn't validate keys eagerly, consider a `field_validator` calling `packaging.version.Version(k)` per key to fail fast.
  ```

src/fromager/packagesettings/_resolver.py
```diff
@@ -482,6 +483,61 @@ def resolver_provider(
...
+        return value
+
+    def resolver_provider(
+        self, ctx: context.WorkContext, req_type: requirements_file.RequirementType
+    ) -> resolver.VersionMapProvider:
+        clone_url = str(self.clone_url)
+        url_map = {
+            ver: f"git+{clone_url}@{ref}" for ver, ref in self.versionmap.items()
+        }
+        version_map = VersionMap(url_map)  # type: ignore[arg-type]
```
Lanceypantsy, line 532: **[Question]** — `src/fromager/packagesettings/_resolver.py` (~line 532, `resolver_provider`)

  `VersionMap(url_map)  # type: ignore[arg-type]` — the `type: ignore` implies `VersionMap` is declared to accept something other than `dict[str, str]`. Runtime tests show a plain `str`→`str` mapping works, so the
  contract seems satisfied, but the ignore hides whether `VersionMap` does any key/value coercion this call bypasses. Can you confirm `VersionMap.__init__`'s real signature rather than suppressing the type error
  — or add the proper type so the ignore can be dropped?
  ```

src/fromager/packagesettings/_resolver.py
```diff
@@ -482,6 +483,61 @@ def resolver_provider(
...
 
 
+class VersionMapGitResolver(AbstractResolver):
+    """Resolve version from a version map, build sdist from git clone.
+
+    The ``versionmap-git`` provider maps known version numbers to known git
+    refs (commit SHAs or ref paths such as ``refs/tags/1.1``).  It clones a
+    git repo at the configured ref and builds an sdist with PEP 517.
+
+    .. versionadded:: 0.79.0
```
Lanceypantsy, line 493: Why are we adding docstring annotation about when something was added? This is what release notest and git history are for.

PR #1199, opened 2026-06-17
feat(bootstrap): parallelize RESOLVE and PREPARE_SOURCE I/O with background threads

dhellmann:
## Summary

Add a thread pool for processing some tasks from the bootstrap work item stack before they are popped. Keep the stack processing as it is, allowing the backgrounded phases to get the results from the future object created by the thread pool. Apply this approach to the phases for resolving requirements and preparing source.

Ensure that background callables are module-level functions (no `self` capture) so threads cannot accidentally access mutable `Bootstrapper` state (`self.why`, `self._seen_requirements`, etc.)

Add a `_push_items()` helper to guarantee every `RESOLVE` and `PREPARE_SOURCE` item has a `bg_future` set, including the initial item in `bootstrap()`

Set the minimum size of the thread pool to 1 to avoid special cases.

## Test plan

- [x] All existing unit tests pass (`hatch run test:test tests/test_bootstrapper.py tests/test_bootstrapper_iterative.py`)
- [x] `TestPhaseResolve` tests updated to use pre-completed `Future` objects instead of patching `resolve_versions`
- [x] Loop tests updated to mock `_resolver.resolve` directly (the path background threads use)
- [x] `./e2e/test_bootstrap_parallel_git_url.sh` passes end-to-end

### DIFF
--- a/e2e/test_bootstrap_cache.sh
+++ b/e2e/test_bootstrap_cache.sh
@@ -42,11 +42,11 @@ fromager \
   bootstrap --cache-wheel-server-url="https://pypi.org/simple" "$DIST==$VER"
 
 EXPECTED_LOG_MESSAGES=(
-"$DIST: looking for existing wheel for version $VER with build tag () in"
-"$DIST: found existing wheel"
+"$DIST-$VER: looking for existing wheel for version $VER with build tag () in"
+"$DIST-$VER: found existing wheel"
 )
 for pattern in "${EXPECTED_LOG_MESSAGES[@]}"; do
-  if ! grep -q "$pattern" "$OUTDIR/bootstrap.log"; then
+  if ! grep -qF "$pattern" "$OUTDIR/bootstrap.log"; then
     echo "FAIL: Did not find log message $pattern in $OUTDIR/bootstrap.log" 1>&2
     pass=false
   fi
@@ -54,11 +54,11 @@ done
 $pass
 
 UNEXPECTED_LOG_MESSAGES=(
-"$DIST: checking if wheel was already uploaded to https://pypi.org/simple"
+"$DIST-$VER: checking if wheel was already uploaded to https://pypi.org/simple"
 )
 
 for pattern in "${UNEXPECTED_LOG_MESSAGES[@]}"; do
-  if grep -q "$pattern" "$OUTDIR/bootstrap.log"; then
+  if grep -qF "$pattern" "$OUTDIR/bootstrap.log"; then
     echo "FAIL: Found log message $pattern in $OUTDIR/bootstrap.log" 1>&2
     pass=false
   fi
@@ -105,14 +105,14 @@ done
 $pass
 
 UNEXPECTED_LOG_MESSAGES=(
-"$DIST: loading build sdist dependencies from build-sdist-requirements.txt"
-"$DIST: loading build backend dependencies from build-backend-requirements.txt"
-"$DIST: loading build system dependencies from build-system-requirements.txt"
+"$DIST-$VER: loading build sdist dependencies from build-sdist-requirements.txt"
+"$DIST-$VER: loading build backend dependencies from build-backend-requirements.txt"
+"$DIST-$VER: loading build system dependencies from build-system-requirements.txt"
 )
 
 for pattern in "${UNEXPECTED_LOG_MESSAGES[@]}"; do
   echo $pattern
-  if grep -q "$pattern" "$OUTDIR/bootstrap.log"; then
+  if grep -qF "$pattern" "$OUTDIR/bootstrap.log"; then
     echo "FAIL: found log message $pattern in $OUTDIR/bootstrap.log" 1>&2
     pass=false
   fi
--- a/e2e/test_bootstrap_prerelease.sh
+++ b/e2e/test_bootstrap_prerelease.sh
@@ -20,7 +20,7 @@ fromager \
 pass=true
 
 # Check for log message that the override is loaded
-if ! grep -q "flit_core: new toplevel dependency flit_core<2.0.1 resolves to 2.0" "$OUTDIR/bootstrap.log"; then
+if ! grep -q "flit_core-2.0: new toplevel dependency flit_core<2.0.1 resolves to 2.0" "$OUTDIR/bootstrap.log"; then
   echo "FAIL: flit_core did not resolve to 2.0 $OUTDIR/bootstrap.log" 1>&2
   pass=false
 fi
@@ -41,7 +41,7 @@ DEBUG_RESOLVER=true fromager \
 
 
 # Check for log message that the override is loaded
-if ! grep -q "flit_core: new toplevel dependency flit_core<2.0.1 resolves to 2.0rc3" "$OUTDIR/bootstrap.log"; then
+if ! grep -q "flit_core-2.0rc3: new toplevel dependency flit_core<2.0.1 resolves to 2.0rc3" "$OUTDIR/bootstrap.log"; then
   echo "FAIL: flit_core did not resolve to 2.0rc3 $OUTDIR/bootstrap.log" 1>&2
   pass=false
 fi
--- a/src/fromager/bootstrap_requirement_resolver.py
+++ b/src/fromager/bootstrap_requirement_resolver.py
@@ -118,6 +118,15 @@ class BootstrapRequirementResolver:
             pbi = self.ctx.package_build_info(req)
             pre_built = pbi.pre_built
 
+        # Check session cache BEFORE the git URL guard so that background
+        # threads can retrieve pre-cached git URL resolutions (populated by
+        # Bootstrapper.resolve_versions() on the main thread before bootstrap()
+        # is called) without hitting the ValueError.
+        cached_result = self.get_cached_resolution(req, pre_built)
+        if cached_result is not None:
+            logger.debug(f"resolved {req} from cache")
+            return list(cached_result) if return_all_versions else [cached_result[0]]
+
         # Git URL source resolution must be handled by Bootstrapper.
         # But git URL prebuilt resolution is allowed - we look for wheels on PyPI
         # (test mode fallback uses this path).
@@ -145,6 +154,32 @@ class BootstrapRequirementResolver:
         if return_all_versions:
             return matching
         return [matching[0]]
+
+    def get_cached_resolution(
+        self,
+        req: Requirement,
+        pre_built: bool,
+    ) -> list[tuple[str, Version]] | None:
+        """Return cached matching versions if this requirement was already resolved.
+
+        Returns ``None`` if the requirement has not been resolved yet, allowing
+        callers to distinguish between "no matching versions" and "not yet resolved".
+
+        Used by background threads to retrieve pre-cached git URL resolutions
+        populated by the main thread before entering the parallel section.
+
+        Args:
+            req: Package requirement
+            pre_built: Whether looking for prebuilt or source resolution
+
+        Returns:
+            List of (url, version) tuples if previously resolved, None otherwise.
+        """
+        rule_key = (str(req), pre_built)
+        with self._lock:
+            if rule_key in self._resolved_rules:
+                return self._get_matching_versions(req, pre_built)
+        return None
 
     def _resolve_and_extend(
         self,
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -33,6 +33,7 @@ from . import (
     resolver,
     server,
     sources,
+    threading_utils,
     wheels,
 )
 from .dependency_graph import DependencyGraph
@@ -47,7 +48,7 @@ logger = logging.getLogger(__name__)
 # package name, extras, version, sdist/wheel
 SeenKey = tuple[NormalizedName, tuple[str, ...], str, typing.Literal["sdist", "wheel"]]
 
-_DEFAULT_BG_THREADS: int = max(1, (os.cpu_count() or 2) // 2)
+_DEFAULT_BG_THREADS: int = max(1, threading_utils.get_cpu_count() // 2)
 
 
 @dataclasses.dataclass
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import concurrent.futures
 import contextlib
 import dataclasses
 import datetime
@@ -46,6 +47,8 @@ logger = logging.getLogger(__name__)
 # package name, extras, version, sdist/wheel
 SeenKey = tuple[NormalizedName, tuple[str, ...], str, typing.Literal["sdist", "wheel"]]
 
+_DEFAULT_BG_THREADS: int = max(1, (os.cpu_count() or 2) // 2)
+
 
 @dataclasses.dataclass
 class SourceBuildResult:
@@ -61,6 +64,27 @@ class SourceBuildResult:
     sdist_root_dir: pathlib.Path | None
     build_env: build_environment.BuildEnvironment | None
     source_type: SourceType
+
+
+@dataclasses.dataclass
+class PreparedSourceData:
+    """Result of background I/O pre-fetching returned to the main thread.
+
+    Fields are set in one of three combinations depending on the result type:
+
+    - Source (no cache hit): only ``sdist_root_dir`` is set.
+    - Source (cache hit): both ``sdist_root_dir`` and ``cached_wheel_filename`` are set.
+    - Prebuilt wheel: both ``wheel_filename`` and ``unpack_dir`` are set.
+    """
+
+    # Source path: set after download+unpack OR cache hit
+    sdist_root_dir: pathlib.Path | None = None
+    # Source path: set when the result came from the wheel cache
+    cached_wheel_filename: pathlib.Path | None = None
+    # Prebuilt path: downloaded wheel file
+    wheel_filename: pathlib.Path | None = None
+    # Prebuilt path: unpack directory (created by mkdir)
+    unpack_dir: pathlib.Path | None = None
 
 
 # Valid failure types for test mode error recording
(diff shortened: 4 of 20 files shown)


src/fromager/bootstrapper.py
```diff
@@ -143,6 +163,246 @@ class WorkItem:
...
+    wheel_url: str,
+) -> PreparedSourceData:
+    """Background-safe prebuilt download: no Bootstrapper state accessed."""
+    logger.info(f"{req_type} requirement {req} uses a pre-built wheel")
+    wheel_filename = wheels.download_wheel(req, wheel_url, ctx.wheels_prebuilt)
+    unpack_dir = ctx.work_dir / f"{req.name}-{resolved_version}"
+    unpack_dir.mkdir(parents=True, exist_ok=True)
+    server.update_wheel_mirror(ctx)
+    return PreparedSourceData(wheel_filename=wheel_filename, unpack_dir=unpack_dir)
+
```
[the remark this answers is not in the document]
dhellmann: this should be fixed in the latest version

src/fromager/bootstrapper.py
```diff
@@ -143,6 +163,246 @@ class WorkItem:
...
+            zipinfo.filename = filename
+            output_file = archive.extract(zipinfo, unpack_dir)
+            logger.info(f"extracted {output_file}")
+        logger.info(f"extracted build requirements from wheel into {unpack_dir}")
+        return unpack_dir
+    except Exception as e:
+        logger.info(f"could not extract build requirements from wheel: {e}")
+        for filename in req_filenames:
+            unpack_dir.joinpath(filename).unlink(missing_ok=True)
+        return None
```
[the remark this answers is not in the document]
dhellmann: fixed in the latest version

dhellmann:
# Claude Code Review: `background-tasks` branch

**One commit:** `feat(bootstrap): parallelize RESOLVE and PREPARE_SOURCE I/O with background threads`

## Overview

This PR adds a `ThreadPoolExecutor` to `Bootstrapper` to pre-fetch resolution and source download/unpack work in background threads while the main thread processes previously queued items. The design extracts I/O-heavy logic into module-level functions (no `Bootstrapper` state) that are safe to call from background threads. A `bg_future` field on `WorkItem` carries the pre-fetched result back to the main-thread phase handlers.

---

## Strengths

- **Good architectural discipline:** Background functions are deliberately stateless (no `Bootstrapper` access), closures capture only immutable values, and the invariant is clearly documented in docstrings.
- **`_drain_background_pool` for exclusive builds:** Correct barrier semantics — drains in-flight work before an exclusive build without losing already-started tasks (`cancel_futures=False`).
- **Fixes a pre-existing resource leak:** `_unpack_metadata_from_wheel` now uses `with zipfile.ZipFile(...) as archive:` (old code used bare `zipfile.ZipFile(...)` without a context manager).
- **`get_cached_resolution` pre-check in resolver:** The comment explains the race-avoidance motivation clearly.
- **Test updates are faithful:** `_make_resolved_future` + already-completed futures cleanly replace the `patch.object(bt, "resolve_versions", ...)` pattern.

---

## Issues

### Critical (Must Fix)

**1. `server.update_wheel_mirror(ctx)` called from background threads without synchronization**
- **Files:** `bootstrapper.py:441`, `bootstrapper.py:323`
- **Functions:** `_bg_prepare_prebuilt` and `_download_wheel_from_cache`
- **Issue:** Both background workers call `server.update_wheel_mirror(ctx)` concurrently. If two packages are resolved to prebuilt wheels simultaneously, both threads call this. Without seeing the implementation it's unclear if it's thread-safe, but `update_wheel_mirror` likely regenerates an on-disk index — a classic TOCTOU race.
- **Fix:** Either ensure `update_wheel_mirror` is idempotent+thread-safe, or move the call back to the main thread (e.g., after consuming `prepared` from the future in `_phase_prepare_source`).

### Important (Should Fix)

**2. Pool resource leak on exception**
- **File:** `bootstrapper.py:2064`
- **Issue:** `_bg_pool` is shut down only in `finalize()`. If `bootstrap()` raises an uncaught exception before the caller reaches `finalize()`, the thread pool and any in-flight futures leak. Threads will keep running until Python garbage-collects the executor.
- **Fix:** Add a `try/finally` in `bootstrap()` to call `self._bg_pool.shutdown(wait=False, cancel_futures=True)` on exception, or make `Bootstrapper` a context manager.

**3. `_bg_resolve` silently bypasses git URL handling in `resolve_versions`**
- **File:** `bootstrapper.py:386-400`
- **Issue:** `resolve_versions` handles git URL packages (lines ~580-590 in the original) before delegating to `_resolver.resolve`. The background `_bg_resolve` calls `_resolver.resolve` directly, which raises `ValueError` for git URL packages. The assumption is that git URL packages are always pre-resolved by `resolve_and_add_top_level`, but this invariant isn't enforced or asserted anywhere.
- **Impact:** If a git URL package somehow reaches `_phase_resolve` through a code path that bypasses pre-resolution, the error message will be confusing.
- **Fix:** Add an assertion or guard in `_bg_resolve` that `req` is not a git URL requirement, with a clear error message.

**4. No concurrency tests**
- **File:** `tests/test_bootstrapper_iterative.py`, `tests/test_bootstrapper.py`
- **Issue:** All tests use already-completed futures (`_make_resolved_future`). There are no tests that actually run items through the thread pool, verify ordering, or exercise the cancellation path. Concurrency bugs (like the `update_wheel_mirror` race) won't be detected.
- **Fix:** At minimum, add an integration-style test that runs a 2-package bootstrap with `num_bg_threads=2` and verifies the result is correct.

### Minor (Nice to Have)

**5. `PreparedSourceData` docstring is inaccurate**
- **File:** `bootstrapper.py:179`
- **Issue:** "Exactly one of (sdist_root_dir, wheel_filename) will be set" — but when a cached source wheel is found, both `sdist_root_dir` and `cached_wheel_filename` are set together (line 416-419).
- **Fix:** Revise to: "For prebuilt wheels, `wheel_filename`/`unpack_dir` are set. For source (including cache-hit), `sdist_root_dir` is set and `cached_wheel_filename` may also be set."

**6. Dead inline-fallback path in `_phase_prepare_source`**
- **File:** `bootstrapper.py:1646-1660`
- **Issue:** The comment says "falling back to inline I/O otherwise" but in practice `_push_items` always submits a future for PREPARE_SOURCE items, so `item.bg_future` will never be `None` here. The fallback is unreachable code.
- **Fix:** Either add a comment explaining it's a defensive fallback, or add an `assert item.bg_future is not None` consistent with `_phase_resolve`.

**7. Default `num_bg_threads` computed at import time in CLI decorator**
- **File:** `commands/bootstrap.py:1033`
- **Issue:** `default=max(1, (os.cpu_count() or 2) // 2)` in the `@click.option` decorator is evaluated once at module import. On systems where `cpu_count()` changes (containers with dynamic CPU limits), the default is stale.
- **Fix:** Use `default=None` with a callback, or just accept the minor limitation with a comment.

---

## Recommendations

1. **Verify `server.update_wheel_mirror` thread safety** before this merges — it's the highest-risk issue. If it involves `subprocess` or filesystem-level re-indexing, it almost certainly needs a lock.
2. Consider exposing the `num_bg_threads` parameter in the `Bootstrapper` constructor docs/changelog since it changes observable performance characteristics.
3. The `_push_items` loop over `reversed(items)` (line 782) submits the item that will be popped first — the comment explains this but a brief note tying it to LIFO stack ordering would help future readers.

---

## Assessment

**Ready to merge: With fixes**

The core parallelism design is sound — stateless background functions, captured-variable closures, and clean barrier semantics for exclusive builds. The `update_wheel_mirror` thread safety issue is the blocker; fixing it and adding at least one actual concurrency test would make this ready.

dhellmann:
# Revised Claude Code Review: `background-tasks` branch

**One commit:** `feat(bootstrap): parallelize RESOLVE and PREPARE_SOURCE I/O with background threads`

## Overview

This PR adds a `ThreadPoolExecutor` to `Bootstrapper` to pre-fetch resolution and source download/unpack work in background threads while the main thread processes previously queued items. The design extracts I/O-heavy logic into module-level functions (no `Bootstrapper` state) that are safe to call from background threads. A `bg_future` field on `WorkItem` carries the pre-fetched result back to the main-thread phase handlers.

---

## Strengths

- **Good architectural discipline:** Background functions are deliberately stateless (no `Bootstrapper` access), closures capture only immutable values, and the invariant is clearly documented in docstrings.
- **`_drain_background_pool` for exclusive builds:** Correct barrier semantics — drains in-flight work before an exclusive build without losing already-started tasks (`cancel_futures=False`).
- **Fixes a pre-existing resource leak:** `_unpack_metadata_from_wheel` now uses `with zipfile.ZipFile(...) as archive:` (old code used bare `zipfile.ZipFile(...)` without a context manager).
- **`get_cached_resolution` pre-check in resolver:** The comment explains the race-avoidance motivation clearly.
- **Test updates are faithful:** `_make_resolved_future` + already-completed futures cleanly replace the `patch.object(bt, "resolve_versions", ...)` pattern.

---

## Issues

### Critical (Must Fix)

**1. `server.update_wheel_mirror(ctx)` called from background threads without synchronization**
- **Files:** `bootstrapper.py:441`, `bootstrapper.py:323`
- **Functions:** `_bg_prepare_prebuilt` and `_download_wheel_from_cache`
- **Issue:** Both background workers call `server.update_wheel_mirror(ctx)` concurrently. If two packages are resolved to prebuilt wheels simultaneously, both threads call this. Without seeing the implementation it's unclear if it's thread-safe, but `update_wheel_mirror` likely regenerates an on-disk index — a classic TOCTOU race.
- **Fix:** Either ensure `update_wheel_mirror` is idempotent+thread-safe, or move the call back to the main thread (e.g., after consuming `prepared` from the future in `_phase_prepare_source`).
- **Verification: FALSE POSITIVE — no fix needed.** `update_wheel_mirror` in `server.py` is decorated with `@with_thread_lock()` (defined in `threading_utils.py`), which wraps every call in a module-level `threading.Lock()`. All concurrent calls serialize through that lock. The race condition does not exist. (Line numbers in the finding are also slightly off: the actual call sites are lines 404 and 286, not 441 and 323.)

### Important (Should Fix)

**2. Pool resource leak on exception**
- **File:** `bootstrapper.py:2064`
- **Issue:** `_bg_pool` is shut down only in `finalize()`. If `bootstrap()` raises an uncaught exception before the caller reaches `finalize()`, the thread pool and any in-flight futures leak. Threads will keep running until Python garbage-collects the executor.
- **Fix:** Add a `try/finally` in `bootstrap()` to call `self._bg_pool.shutdown(wait=False, cancel_futures=True)` on exception, or make `Bootstrapper` a context manager.
- **Verification: CONFIRMED — fix needed, and broader than stated.** `bootstrap()` has no `try/finally`. The caller in `commands/bootstrap.py` also has no `try/finally` — `finalize()` is unreachable if `bootstrap()` raises. Additionally, `lint_requirements.py` creates a `Bootstrapper` but never calls `finalize()` at all, leaking the pool on every invocation regardless of exceptions.

**3. `_bg_resolve` silently bypasses git URL handling in `resolve_versions`**
- **File:** `bootstrapper.py:386-400`
- **Issue:** `resolve_versions` handles git URL packages (lines ~580-590 in the original) before delegating to `_resolver.resolve`. The background `_bg_resolve` calls `_resolver.resolve` directly, which raises `ValueError` for git URL packages. The assumption is that git URL packages are always pre-resolved by `resolve_and_add_top_level`, but this invariant isn't enforced or asserted anywhere.
- **Impact:** If a git URL package somehow reaches `_phase_resolve` through a code path that bypasses pre-resolution, the error message will be confusing.
- **Fix:** Add an assertion or guard in `_bg_resolve` that `req` is not a git URL requirement, with a clear error message.
- **Verification: PARTIALLY REAL — severity overstated, but assertion is worth adding.** The invariant is implicitly enforced: `resolve_and_add_top_level` populates the resolver cache via `extend_known_versions` for all git URL requirements, and `BootstrapRequirementResolver.resolve()` checks `get_cached_resolution()` before the `ValueError` guard. The comment at that guard even explicitly documents this design. If a git URL did somehow reach `_bg_resolve` uncached, it would fail loudly (not silently) through the normal error-handling path. That said, adding an `assert not req.url` in `_bg_resolve` is a low-cost defensive improvement to document-by-contract.

**4. No concurrency tests**
- **File:** `tests/test_bootstrapper_iterative.py`, `tests/test_bootstrapper.py`
- **Issue:** All tests use already-completed futures (`_make_resolved_future`). There are no tests that actually run items through the thread pool, verify ordering, or exercise the cancellation path. Concurrency bugs (like the `update_wheel_mirror` race) won't be detected.
- **Fix:** At minimum, add an integration-style test that runs a 2-package bootstrap with `num_bg_threads=2` and verifies the result is correct.
- **Verification: CONFIRMED — fix needed.** No test in either file references `num_bg_threads`, `ThreadPoolExecutor`, `executor`, or `submit` (confirmed by grep). All `TestPhaseResolve` tests use pre-completed futures; all loop tests mock `_dispatch_phase` entirely. The entire thread-pool submission path (`executor.submit()`, future blocking, cancellation) is untested.

### Minor (Nice to Have)

**5. `PreparedSourceData` docstring is inaccurate**
- **File:** `bootstrapper.py:179`
- **Issue:** "Exactly one of (sdist_root_dir, wheel_filename) will be set" — but when a cached source wheel is found, both `sdist_root_dir` and `cached_wheel_filename` are set together (line 416-419).
- **Fix:** Revise to: "For prebuilt wheels, `wheel_filename`/`unpack_dir` are set. For source (including cache-hit), `sdist_root_dir` is set and `cached_wheel_filename` may also be set."
- **Verification: CONFIRMED — and worse than stated.** The class has four fields (`sdist_root_dir`, `cached_wheel_filename`, `wheel_filename`, `unpack_dir`) but the docstring only mentions two. The "exactly one" invariant is violated in two places: cached-source path sets both `sdist_root_dir` and `cached_wheel_filename` together; prebuilt path sets both `wheel_filename` and `unpack_dir` together. Three distinct cases: (1) source no cache-hit: only `sdist_root_dir`; (2) cached source wheel: `sdist_root_dir` + `cached_wheel_filename`; (3) prebuilt: `wheel_filename` + `unpack_dir`.

**6. Dead inline-fallback path in `_phase_prepare_source`**
- **File:** `bootstrapper.py:1646-1660`
- **Issue:** The comment says "falling back to inline I/O otherwise" but in practice `_push_items` always submits a future for PREPARE_SOURCE items, so `item.bg_future` will never be `None` here. The fallback is unreachable code.
- **Fix:** Either add a comment explaining it's a defensive fallback, or add an `assert item.bg_future is not None` consistent with `_phase_resolve`.
- **Verification: CONFIRMED.** `_bg_pool` is always non-`None` during the bootstrap loop, so `_push_items` always submits a future for PREPARE_SOURCE items, and `_get_background_work` always returns a callable for them. The `else` fallback branches are unreachable. `_phase_resolve` uses `assert item.bg_future is not None`; `_phase_prepare_source` silently falls back instead — an inconsistency that could mask future regressions. Prefer the assertion.

**7. Default `num_bg_threads` computed at import time in CLI decorator**
- **File:** `commands/bootstrap.py:1033`
- **Issue:** `default=max(1, (os.cpu_count() or 2) // 2)` in the `@click.option` decorator is evaluated once at module import. On systems where `cpu_count()` changes (containers with dynamic CPU limits), the default is stale.
- **Fix:** Use `default=None` with a callback, or just accept the minor limitation with a comment.
- **Verification: TECHNICALLY ACCURATE but not worth fixing.** The default is indeed computed at import time (confirmed at lines 125 and 534). However, dynamic CPU limit changes while a Python CLI process is running are not a realistic scenario — container orchestrators set CPU limits before process start. The default is already conservative (`cpu_count // 2`, min 1) and users can always override with `--bg-threads N`. Adding a callback would increase complexity for no practical benefit. Accept with a comment if anything is done at all.

---

## Recommendations

1. **Verify `server.update_wheel_mirror` thread safety** before this merges — it's the highest-risk issue. If it involves `subprocess` or filesystem-level re-indexing, it almost certainly needs a lock.
2. Consider exposing the `num_bg_threads` parameter in the `Bootstrapper` constructor docs/changelog since it changes observable performance characteristics.
3. The `_push_items` loop over `reversed(items)` (line 782) submits the item that will be popped first — the comment explains this but a brief note tying it to LIFO stack ordering would help future readers.

---

## Assessment

**Ready to merge: With fixes**

The core parallelism design is sound — stateless background functions, captured-variable closures, and clean barrier semantics for exclusive builds.

After verification, the priority order changes:

- **Critical #1 is a false positive** — `update_wheel_mirror` is already protected by `@with_thread_lock()`. No change needed.
- **The real blocker is Important #2** — the thread pool leaks on exception, and `lint_requirements.py` leaks it unconditionally. Fix by making `Bootstrapper` a context manager or adding `try/finally` in the relevant callers.
- **Important #4 (no concurrency tests)** is confirmed and should still be addressed.
- **Important #3** is not a bug — the design intentionally pre-caches git URL resolutions — but adding an assertion would be a cheap defensive improvement.
- **Minors #5 and #6** are confirmed and straightforward to fix.
- **Minor #7** is not worth changing.

dhellmann:
# Fix Plan: background-tasks branch review issues

## Context

The `background-tasks` branch adds `ThreadPoolExecutor`-based background I/O to `Bootstrapper`.
Code review identified 7 issues; verification confirmed 5 need fixing (#1 was a false positive,
#7 not worth changing). This plan addresses the 5 confirmed issues in priority order.

---

## Issue #2 — Pool resource leak (Important)

**Problem:** `_bg_pool` is shut down only in `finalize()`. Two callers have leaks:
- `commands/bootstrap.py` calls `bootstrap()` then `finalize()` with no `try/finally` — if `bootstrap()` raises, `finalize()` is never reached
- `lint_requirements.py` creates `Bootstrapper` but never calls `finalize()` at all

**Fix:** Make `Bootstrapper` a context manager.

### Changes

**`src/fromager/bootstrapper.py`** — add after `finalize()`:

```python
def __enter__(self) -> "Bootstrapper":
    return self

def __exit__(self, exc_type: type | None, exc_val: BaseException | None, exc_tb: object) -> None:
    if self._bg_pool is not None:
        self._bg_pool.shutdown(wait=False, cancel_futures=True)
        self._bg_pool = None
```

Note: `wait=False` on exception path — don't block on cleanup. Normal path still calls
`finalize()` (which does `wait=True`) before `__exit__` runs, so `_bg_pool` is already `None`.

**`src/fromager/commands/bootstrap.py`** — wrap `Bootstrapper` creation with `with`:

```python
# Before:
bt = bootstrapper.Bootstrapper(...)
...
bt.finalize()

# After:
with bootstrapper.Bootstrapper(...) as bt:
    ...
    bt.finalize()
```

**`src/fromager/commands/lint_requirements.py`** — wrap with `with`:

```python
# Before:
bt = bootstrapper.Bootstrapper(...)
# ... (no finalize call)

# After:
with bootstrapper.Bootstrapper(...) as bt:
    # ... (no finalize call needed — __exit__ handles cleanup)
```

---

## Issue #3 — Missing assertion in `_bg_resolve` (Minor defensive improvement)

**Problem:** `_bg_resolve` calls `_resolver.resolve()` directly, bypassing the git URL guard
in `resolve_versions`. The invariant (git URLs always pre-cached) is implicit and unenforced.

**Fix:** Add assertion at the top of `_bg_resolve`:

```python
def _bg_resolve(...) -> list[tuple[str, Version]]:
    """Background-safe resolution: no Bootstrapper state accessed."""
    assert not req.url, (
        f"git URL requirements must be pre-resolved before background dispatch: {req}"
    )
    ...
```

---

## Issue #5 — `PreparedSourceData` docstring inaccurate (Minor)

**Problem:** Docstring says "Exactly one of (sdist_root_dir, wheel_filename) will be set" but
there are 4 fields set in 3 distinct combinations.

**Fix:** Replace docstring:

```python
@dataclasses.dataclass
class PreparedSourceData:
    """Result of background I/O pre-fetching returned to the main thread.

    Fields are set in one of three combinations depending on the result type:

    - Source (no cache hit): only ``sdist_root_dir`` is set.
    - Source (cache hit): both ``sdist_root_dir`` and ``cached_wheel_filename`` are set.
    - Prebuilt wheel: both ``wheel_filename`` and ``unpack_dir`` are set.
    """
```

---

## Issue #6 — Dead fallback in `_phase_prepare_source` (Minor)

**Problem:** `_phase_prepare_source` uses `if item.bg_future is not None:` with a silent
fallback, while `_phase_resolve` uses `assert item.bg_future is not None`. The fallback is
unreachable dead code that could mask future regressions.

**Fix:** Replace the conditional block with an assertion matching `_phase_resolve`'s style.
Remove the dead fallback code entirely.

```python
# Replace:
prepared: PreparedSourceData | None = None
if item.bg_future is not None:
    prepared = item.bg_future.result()

# With:
assert item.bg_future is not None
prepared: PreparedSourceData = item.bg_future.result()
```

Also remove the unreachable inline-I/O fallback branches (the `else` blocks).

---

## Issue #4 — No concurrency tests (Important)

**Problem:** All tests use pre-completed futures (`_make_resolved_future`). The thread pool
submission path (`executor.submit`, future blocking, cancellation) is completely untested.

**Fix:** Add one test in `tests/test_bootstrapper_iterative.py` that uses a real
`ThreadPoolExecutor` to exercise `_push_items` → `executor.submit` → `future.result()` for a
RESOLVE item.

**Approach:**
- Create `Bootstrapper(tmp_context, num_bg_threads=1)`
- Mock `bt._resolver.resolve` to return a fixed synthetic result (no network)
- Create a RESOLVE `WorkItem` and call `bt._push_items(stack, [item])`
- Assert `item.bg_future` is a real `concurrent.futures.Future` (not pre-set)
- Call `item.bg_future.result()` and assert it equals the mocked return value
- Verify the result is processed correctly through `bt._phase_resolve(item)`
- Clean up with `bt.finalize()`

---

## Critical Files

| File | Purpose |
|------|---------|
| `src/fromager/bootstrapper.py` | Add `__enter__`/`__exit__`; fix assertion (#3), docstring (#5), dead-code (#6) |
| `src/fromager/commands/bootstrap.py` | Wrap `Bootstrapper` with `with` statement |
| `src/fromager/commands/lint_requirements.py` | Wrap `Bootstrapper` with `with` statement |
| `tests/test_bootstrapper_iterative.py` | Add real-thread-pool test |

src/fromager/commands/bootstrap.py
```diff
@@ -190,43 +202,43 @@ def bootstrap(
...
+            logger.info("resolving top-level dependencies before building")
+            resolved_reqs: list[Requirement] = []
+            for req in to_build:
+                token = requirement_ctxvar.set(req)
+                result = bt.resolve_and_add_top_level(req)
+                if result is not None:
+                    resolved_reqs.append(req)
+                # If result is None, test_mode or multiple_versions recorded the failure
+                requirement_ctxvar.reset(token)
+
```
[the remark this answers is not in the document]
dhellmann: This issue is unrelated to this PR. I will fix it separately.

src/fromager/commands/lint_requirements.py
```diff
@@ -47,61 +47,60 @@ def lint_requirements(
...
-    )
+    ) as bt:
+        for path in input_files_path:
+            is_constraints: bool = path.name.endswith("constraints.txt")
+            parsed_lines = requirements_file.parse_requirements_file(path)
+            unique_entries: dict[tuple[str, str], Requirement] = {}
+            for line in parsed_lines:
+                try:
+                    requirement = Requirement(line)
+                    marker_key = str(requirement.marker) if requirement.marker else ""
```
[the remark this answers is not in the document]
dhellmann: This issue is unrelated to this PR, I will fix it separately.

src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -118,6 +118,15 @@ def resolve(
...
 
+        # Check session cache BEFORE the git URL guard so that background
+        # threads can retrieve pre-cached git URL resolutions (populated by
+        # Bootstrapper.resolve_versions() on the main thread before bootstrap()
+        # is called) without hitting the ValueError.
+        cached_result = self.get_cached_resolution(req, pre_built)
+        if cached_result is not None:
+            logger.debug(f"resolved {req} from cache")
+            return list(cached_result) if return_all_versions else [cached_result[0]]
+
```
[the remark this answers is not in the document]
dhellmann: This issue is unrelated to this PR, I will fix it separately.

dhellmann:
Local testing of this with a representatively large set of input packages shows performance improvements:

| | single-version mode | multi-version mode |
| --- | --- | --- |
| **before PR** | 46 min (full job) | 17 min (bootstrap only) | 
| **after PR** | 37 min (full job) | 13 min (bootstrap only) |

All tests used bootstrap-parallel. The single-version mode test ran through to the end, including building all packages after the initial bootstrap phase. The multi-version mode stopped after bootstrapping and did not complete all of the builds, but the PR changes have no effect on that stage of the work.

e2e/test_bootstrap_cache.sh
```diff
@@ -42,8 +42,8 @@ fromager \
   bootstrap --cache-wheel-server-url="https://pypi.org/simple" "$DIST==$VER"
 
 EXPECTED_LOG_MESSAGES=(
-"$DIST: looking for existing wheel for version $VER with build tag () in"
-"$DIST: found existing wheel"
+"$DIST-$VER: looking for existing wheel for version $VER with build tag () in"
+"$DIST-$VER: found existing wheel"
```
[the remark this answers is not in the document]
dhellmann: fixed

e2e/test_bootstrap_prerelease.sh
```diff
@@ -20,7 +20,7 @@ fromager \
 pass=true
 
 # Check for log message that the override is loaded
-if ! grep -q "flit_core: new toplevel dependency flit_core<2.0.1 resolves to 2.0" "$OUTDIR/bootstrap.log"; then
+if ! grep -q "flit_core-2.0: new toplevel dependency flit_core<2.0.1 resolves to 2.0" "$OUTDIR/bootstrap.log"; then
```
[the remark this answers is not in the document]
dhellmann: fixed

dhellmann:
I think this is finally ready for review.

src/fromager/commands/bootstrap.py
```diff
@@ -496,6 +508,14 @@ def write_constraints_file(
     default=None,
     help="Reject package versions published more than this many days ago.",
 )
+@click.option(
+    "--bg-threads",
+    "num_bg_threads",
+    type=click.IntRange(min=1),
+    default=max(1, (os.cpu_count() or 2) // 2),
```
rd4398: This calculation is being repeated three times.
  - Bootstrapper.__init__ parameter default
  - bootstrap click option default
  - bootstrap_parallel click option default
Can we please consider extracting a module level constant so that it is easier to maintain in one place in the future?

src/fromager/bootstrapper.py
rd4398, line 1167: nit: duplication between background and inline paths

` _download_prebuilt` (instance method, line 908) and `_bg_prepare_prebuilt` (module-level, line 355) perform the same work — download wheel, create unpack dir, update mirror. The instance method is now only reached as a fallback when `bg_future` is None or didn't produce a prebuilt result.

Similarly, `_find_cached_wheel` (instance method, line 926) is now a thin delegation to the module-level `_find_cached_wheel`.

Could _download_prebuilt also delegate to _bg_prepare_prebuilt (extracting wheel_filename and unpack_dir from the returned PreparedSourceData) to avoid the duplicated logic?

src/fromager/bootstrapper.py
```diff
@@ -144,6 +166,248 @@ class WorkItem:
...
+    logger.info(f"{BootstrapPhase.RESOLVE} for {req_type} requirement")
+    return bg_resolver.resolve(
+        req=req,
+        req_type=req_type,
+        parent_req=parent_req,
+        return_all_versions=return_all_versions,
+    )
+
+
+def _bg_prepare_source(
```
rd4398, line 383: `_bg_prepare_source` calls `sources.download_source()` and `sources.prepare_source()`, which write into `ctx.sdists_downloads` and `ctx.sdists_builds`.  I think multiple background threads downloading different packages can race on these shared directories. `_bg_prepare_prebuilt` calls `wheels.download_wheel()` writing to `ctx.wheels_prebuilt`, and then calls `server.update_wheel_mirror()`  which is locked, but the download itself is not.

If two background threads download different packages whose filenames happen to create conflicting intermediate paths inside sdists_downloads, or if prepare_source extracts into work_dir with overlapping temporary names, this could corrupt state.

rd4398 (dismissed):
This looks overall good. I added few comments / suggestions.

src/fromager/bootstrapper.py
dhellmann: Good catch to flag this — the concern is a false positive, but I've added inline comments to document the reasoning explicitly.

Three complementary guarantees make the concurrent access safe:

1. **`_seen_requirements` guard (main thread)**: Before any `_push_items()` call, the Bootstrapper checks `_seen_requirements`. The same requirement is never submitted to the thread pool twice, so no two background threads ever process the same package simultaneously.

2. **Unique paths per package**: Every path used by background threads embeds `{name}-{version}`:
   - `download_source()`: writes to `ctx.sdists_downloads / {name}-{version}.tar.gz` (or `.zip`) — unique per package
   - `prepare_source()`: extracts to `ctx.work_dir / {req_name}-{version}` — unique per package
   - `wheels.download_wheel()`: filename comes from the wheel URL (e.g. `pkg-1.0-py3-none-any.whl`) — unique per package
   - `_bg_prepare_prebuilt`: `unpack_dir = ctx.work_dir / f"{req.name}-{resolved_version}"` — unique per package

3. **Atomic operations**: `download_url()` downloads to `outfile + ".tmp"` then renames atomically. `mkdir(exist_ok=True)` is atomic. `update_wheel_mirror()` is already locked.

I've added inline comments to both `_bg_prepare_source` and `_bg_prepare_prebuilt` documenting this.

src/fromager/commands/bootstrap.py
dhellmann: Fixed in the latest version.

src/fromager/bootstrapper.py
dhellmann: Fixed in the latest version.

src/fromager/bootstrapper.py
rd4398: Thank you for adding comments  and explaining in detail. This makes sense.

rd4398:
@LalatenduMohanty can you please review as well? We should merge this soon and include it in release as it increases performance of fromager bootstrap significantly.

LalatenduMohanty:
Looking in to this PR now.

src/fromager/bootstrapper.py
```diff
@@ -46,6 +47,8 @@
 # package name, extras, version, sdist/wheel
 SeenKey = tuple[NormalizedName, tuple[str, ...], str, typing.Literal["sdist", "wheel"]]
 
+_DEFAULT_BG_THREADS: int = max(1, (os.cpu_count() or 2) // 2)
+
```
LalatenduMohanty, line 52: We should use get_cpu_count() instead of os.cpu_count() to respect container CPU limits as there's already a function for exactly this in packagesettings/_pbi.py. 

May be we can have this as a follow up task

```
  1. Move get_cpu_count() from packagesettings/_pbi.py to threading_utils.py
  2. Have _pbi.py, bootstrapper.py, and download_sequence.py all import from there
  3. Replace os.cpu_count() with get_cpu_count() in the _DEFAULT_BG_THREADS calculation
```

src/fromager/bootstrapper.py
```diff
@@ -144,6 +168,254 @@ class WorkItem:
...
+    work_dir: pathlib.Path,
+    req: Requirement,
+    resolved_version: Version,
+) -> pathlib.Path:
+    unpack_dir = work_dir / f"{req.name}-{resolved_version}"
+    unpack_dir.mkdir(parents=True, exist_ok=True)
+    return unpack_dir
+
+
+def _unpack_metadata_from_wheel(
```
LalatenduMohanty: This function is little hard to read, The function name says "unpack metadata" but it actually extracts build requirement files (not general metadata) from fromager-specific paths inside a
  wheel. A reader has to reverse-engineer the intent from the implementation.


I understand that this is existing code and The PR moved it from being a method to a module-level function (for thread safety) without changing its  body. May be another follow up task?

LalatenduMohanty: Should we add a docstring as part of this PR?

LalatenduMohanty (changes requested):

src/fromager/bootstrapper.py
dhellmann: Fixed in 3bbf590ce7b9f7c5e39e4278c28d5b99cc56013e

src/fromager/bootstrapper.py
dhellmann: Fixed in 9a65aa652f6598b749999b648c8d807b26a165a1

LalatenduMohanty (approved):

PR #1217, opened 2026-06-24
feat(wheels): add configurable build tag hook for wheel filenames

### LINKED ISSUE #1181: Implement wheel_build_tag hook for unique wheel file names
## Summary

Implement the accepted proposal from PR #1066 (`docs/proposals/wheel-build-tag-hook.md`).
Tracking issue: #1059

## Requirements

1. **Global settings**: Add a `wheels` section to the global settings file with a `build_tag_hook` field (`ImportString` via Pydantic) that points to a callable.

   ```yaml
   wheels:
     build_tag_hook: "mypackage.hooks:build_tag_hook"
   ```

2. **Hook signature**: The configured callable receives `ctx`, `req`, `version`, and `wheel_tags` (all keyword-only) and returns `Sequence[str]` — ordered suffix segments like `["el9.6", "rocm7.1", "torch2.10.0"]`.

   ```python
   def build_tag_hook(
       *,
       ctx: context.WorkContext,
       req: Requirement,
       version: Version,
       wheel_tags: frozenset[Tag],
   ) -> typing.Sequence[str]:
       ...
   ```

3. **Suffix construction**: Join returned segments with `_` and append to the existing numeric build tag (e.g. `8` → `8_el9.6_rocm7.1`).

4. **Validation**: Each segment must match `[a-zA-Z0-9.]` only. Fail the build on invalid characters or hook exceptions.

5. **No-op default**: When `build_tag_hook` is unset, no suffix is appended (current behavior preserved).

6. **No wheel content access**: The hook must not depend on wheel content, build env, or ELF info — it must produce the same result whether building fresh or retrieving from cache.

## References

- Proposal: #1066
- Original issue: #1059

shifa-khan:
Allow downstream projects to append environment-specific suffixes (e.g. OS, accelerator) to wheel build tags via a user-defined hook. The hook is configured in settings.yaml under `wheels.build_tag_hook` and receives the build context, requirement, version, and wheel tags. Cache lookups and build-tag validation are updated to account for suffixed wheels. No behavior change when the hook is not configured.

Closes: #1181

### DIFF
--- a/src/fromager/bootstrapper.py
+++ b/src/fromager/bootstrapper.py
@@ -926,26 +926,34 @@ class Bootstrapper:
         search_in: pathlib.Path,
     ) -> tuple[pathlib.Path | None, pathlib.Path | None]:
         pbi = self.ctx.package_build_info(req)
-        expected_build_tag = pbi.build_tag(resolved_version)
+        base_build_tag = pbi.build_tag(resolved_version)
         logger.info(
-            f"looking for existing wheel for version {resolved_version} with build tag {expected_build_tag} in {search_in}"
+            f"looking for existing wheel for version {resolved_version} with build tag {base_build_tag} in {search_in}"
         )
         wheel_filename = finders.find_wheel(
             downloads_dir=search_in,
             req=req,
             dist_version=str(resolved_version),
-            build_tag=expected_build_tag,
+            build_tag=base_build_tag,
         )
         if not wheel_filename:
             return None, None
 
-        # Re-validate build tag from the actual wheel metadata because
-        # finders.find_wheel matches by filename prefix, which may not
-        # enforce an exact build tag match.
-        _, _, build_tag, _ = wheels.extract_info_from_wheel_file(req, wheel_filename)
-        if expected_build_tag and expected_build_tag != build_tag:
+        # Re-validate build tag from the actual wheel metadata.
+        # Compute the full expected tag (including any hook suffix)
+        # now that we have the wheel's tags.
+        _, _, actual_build_tag, wheel_tags = wheels.extract_info_from_wheel_file(
+            req, wheel_filename
+        )
+        expected_build_tag = wheels.get_build_tag(
+            ctx=self.ctx,
+            req=req,
+            version=resolved_version,
+            wheel_tags=wheel_tags,
+        )
+        if expected_build_tag and expected_build_tag != actual_build_tag:
             logger.info(
-                f"found wheel for {resolved_version} in {wheel_filename} but build tag does not match. Got {build_tag} but expected {expected_build_tag}"
+                f"found wheel for {resolved_version} in {wheel_filename} but build tag does not match. Got {actual_build_tag} but expected {expected_build_tag}"
             )
             return None, None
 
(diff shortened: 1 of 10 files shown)


src/fromager/bootstrapper.py
```diff
@@ -973,19 +981,23 @@ def _download_wheel_from_cache(
...
+            _, _, actual_build_tag, wheel_tags = wheels.extract_info_from_wheel_file(
                 req, wheelfile_name
             )
-            if expected_build_tag and expected_build_tag != build_tag:
+            expected_build_tag = wheels.get_build_tag(
+                ctx=self.ctx,
+                req=req,
+                version=resolved_version,
+                wheel_tags=wheel_tags,
+            )
```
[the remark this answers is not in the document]
shifa-khan: `_download_wheel_from_cache` uses except ResolverException (line 1016), not except Exception. A ValueError from the hook would propagate just fine. And `_look_for_existing_wheel` has no try/except at all.

src/fromager/packagesettings/_models.py
```diff
@@ -72,6 +72,36 @@ class SbomSettings(pydantic.BaseModel):
...
+    The callable receives keyword-only arguments ``ctx``, ``req``,
+    ``version``, and ``wheel_tags``, and returns ``Sequence[str]``.
+    """
+
+    @pydantic.field_validator("build_tag_hook", mode="before")
+    @classmethod
+    def _colon_to_dot(cls, v: typing.Any) -> typing.Any:
+        if isinstance(v, str) and ":" in v:
+            return v.replace(":", ".", 1)
+        return v
```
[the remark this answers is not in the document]
shifa-khan: Adding `inspect.signature` validation to reject `os.path.join` at load time adds complexity and can break legitimate callables (partials, C extensions). A wrong callable produces a clear TypeError on first call, that's sufficient.

src/fromager/packagesettings/_models.py
```diff
@@ -72,6 +72,36 @@ class SbomSettings(pydantic.BaseModel):
...
+
+      wheels:
+        build_tag_hook: "mypackage.hooks:build_tag_hook"
+
+    .. versionadded:: 0.82.0
+    """
+
+    model_config = MODEL_CONFIG
+
+    build_tag_hook: pydantic.ImportString[typing.Callable[..., typing.Any]] | None = (
```
tiran, line 88: Let's define a more detailed callable:

```python                                                                                                                                                                                                        
BuildTagHookCallable = typing.Callable[
    ["context.WorkContext", Requirement, Version, frozenset[Tag]],
    typing.Sequence[str],
]
```

the define the hook as:

```
build_tag_hook: pydantic.ImportString[BuildTagHookCallable] | None = None
```

src/fromager/packagesettings/_models.py
tiran, line 102: You don't need to modify the string. `ImportString` supports the `:` syntax.

```python
>>> import pydantic
>>> class WheelSettings(pydantic.BaseModel):
...     build_tag_hook: pydantic.ImportString
...
>>> WheelSettings(build_tag_hook="fromager.context:WorkContext")
WheelSettings(build_tag_hook=<class 'fromager.context.WorkContext'>)
```

src/fromager/packagesettings/_models.py
```diff
@@ -72,6 +72,36 @@ class SbomSettings(pydantic.BaseModel):
...
 
+class WheelSettings(pydantic.BaseModel):
+    """Global wheel settings
+
+    ::
+
+      wheels:
+        build_tag_hook: "mypackage.hooks:build_tag_hook"
+
+    .. versionadded:: 0.82.0
```
tiran, line 83: Fromager 0.88.0 is out, the new feature will land in 0.89.0 or later.

```suggestion
    .. versionadded:: 0.89.0
```

src/fromager/wheels.py
```diff
@@ -39,12 +40,60 @@
...
 FROMAGER_ELF_REQUIRES = "fromager-elf-requires.txt"
 FROMAGER_BUILD_REQ_PREFIX = "fromager"
 
 
+def get_build_tag(
+    *,
+    ctx: context.WorkContext,
+    req: Requirement,
+    version: Version,
+    wheel_tags: frozenset[Tag] | None = None,
```
tiran, line 56: Under which circumstances is `wheel_tags` None? All wheels must have at least one wheel tag.

tests/test_packagesettings.py
tiran, line 1: AI tends to write a lot of test cases. Every test case comes with a setup cost and makes the test unnecessary slow. I always tell Claude to combine tests to reduce the amount of test cases.

src/fromager/finders.py
tiran, line 1: The finder API is an old, badly designed part of Fromager. It tries to be clever, but gets it wrong in some cases. We may need to reconsider and redesign this part of Fromager, first. :/

smoparth: @tiran I've opened an issue to track this ([#1290](https://github.com/python-wheel-build/fromager/issues/1290)). Could you take a look when you have a moment and let me know if it covers everything, or add any additional thoughts on what you have in mind for the redesign?

rd4398:
@shifa-khan @jlarkin09 can we close this PR in favor of  https://github.com/python-wheel-build/fromager/pull/1273?

PR #1218, opened 2026-06-25
build(test): add pytest-xdist for parallel test execution

tiran:
# Pull Request Description

## What

Add `pytest-xdist` to test dependencies and configure 3 parallel workers via `addopts`. Testing showed that 4 or more workers does not provide measurable benefits.

See https://pytest-xdist.readthedocs.io/en/stable/

## Why

Speed up local testing. On my machine, unit tests now takes less than 4 seconds instead of 12 seconds.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,6 +71,7 @@ mypy = [
 test = [
     "coverage[toml]!=4.4,>=4.0",
     "pytest",
+    "pytest-xdist",
     "requests-mock",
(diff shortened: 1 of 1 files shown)


rd4398 (approved):
Looks good to me! For some reason CI is failing :(

tiran:
Yes, there is an issue with coverage and xdist. I have a local fix, but I haven't pushed it yet.

I'm creating a new PR.

PR #1219, opened 2026-06-25
build(test): add pytest-cov for xdist-compatible coverage

tiran:
# Pull Request Description

## What

`coverage run` only instruments the main process. With pytest-xdist, test workers are spawned as separate processes and their coverage is lost. Replace `coverage run -m pytest` with `pytest --cov`, which uses pytest-cov to instrument xdist workers correctly.

## Why

Follow-up to #1218


rd4398 (approved):
LGTM!

PR #1223, opened 2026-06-29
fix(ci): set COVERAGE_FILE so unit coverage uploads correctly

### LINKED ISSUE #1222: fix(ci): coverage gate broken on main since pytest-cov switch
Since #1219 (commit 7bfd36d), the coverage report job fails on every PR with:

```
Coverage failure: total of 37 is less than fail-under=60
```

**Root cause:** The test command changed from `coverage run -m pytest` to `pytest --cov`. `pytest --cov` writes a single `.coverage` file, but the CI upload step uses `path: .coverage.*` which requires a dotted suffix (e.g. `.coverage.<hostname>.<pid>`). The unit test coverage data is never uploaded — only e2e data reaches `coverage combine`, producing 37% instead of the expected ~93%.

**Reproducer:** Any PR triggers the failure — see [this run](https://github.com/python-wheel-build/fromager/actions/runs/28369296375/job/84046132399).

**Suggested fix:** use `pytest --cov` with options that produce parallel-mode files.

LalatenduMohanty:
The switch from `coverage run -m pytest` to `pytest --cov` (#1219)
caused pytest-cov to write a single `.coverage` file. The CI upload
step uses the glob `.coverage.*`, which requires a suffix after the
dot — so the unit test coverage data was never uploaded. Only e2e
coverage data (which still produces `.coverage.*` files via subprocess
instrumentation) reached the combine step, dropping overall coverage
from ~93% to 37% and failing the `--fail-under=60` gate.
    
Fix by setting `COVERAGE_FILE` with a per-job suffix so pytest-cov
writes directly to e.g. `.coverage.unit.py3.12.ubuntu-latest`, which
the upload glob matches.
    
Also remove `tests/` from `[tool.coverage.run].source` to measure
only production code coverage.

Closes: #1222

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -133,11 +133,10 @@ lint-requirements = "fromager.commands.lint_requirements:lint_requirements"
 branch = true
 parallel = true
 relative_files = true
-source = ["fromager", "tests/"]
+source = ["fromager"]
 
 [tool.coverage.paths]
 source = ["src/fromager", ".hatch/**/site-packages/fromager"]
-tests = ["tests/"]
 
 [tool.coverage.report]
 show_missing = true
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -57,6 +57,7 @@ jobs:
         run: hatch run test:test --with-network
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          COVERAGE_FILE: .coverage.unit.py${{ matrix.python-version }}.${{ matrix.os }}
 
       - name: Upload coverage
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a  # v7


pyproject.toml
```diff
@@ -133,11 +133,10 @@ lint-requirements = "fromager.commands.lint_requirements:lint_requirements"
 branch = true
 parallel = true
 relative_files = true
-source = ["fromager", "tests/"]
```
dhellmann, line 136: Do we need/want this change? If we don't make it, then the test files will be run by the unit test job and we'll see those numbers included in the output.

dhellmann (approved):

pyproject.toml
LalatenduMohanty: This change is unrelated to the actual issue, it changes what shows up in the coverage report and could lower the reported percentage. Do we want to know the code coverage of our tests. IMO we only care about fromager code. If you still think we should not change it, let me know and I will revert the change

dhellmann: The coverage for our tests should always be 100% because we should be running the tests. So maybe it's not important to include them, except to test that assumption that we do run them.

LalatenduMohanty: Right, the test coverage of tests inflates the overall coverage.

PR #1224, opened 2026-06-29
fix(candidate): use exact dist-info path in `get_metadata_for_wheel()`

### LINKED ISSUE #1146: get_metadata_for_wheel() may return wrong metadata
`get_metadata_for_wheel` returns the first match for `name.endswith(".dist-info/METADATA")`. This is a bad assumption. A wheel can contain multiple `.dist-info` directories. It's common to include dist-info directories of vendored packages.

`get_metadata_for_wheel` should only look at `dist-info` directories in the root of the wheel. Or better: parse the URL, split the name, and use `{dist_name}-{dist_version}.dist-info/METADATA`.

rd4398:
`get_metadata_for_wheel()` returned the first zip entry matching `*.dist-info/METADATA`, which could be a vendored package's metadata when a wheel bundles multiple dist-info directories. Derive the correct `{name}-{version}.dist-info/METADATA` path from the wheel URL instead and read it directly.

Closes: #1146

### DIFF
--- a/src/fromager/candidate.py
+++ b/src/fromager/candidate.py
@@ -1,8 +1,10 @@
 import dataclasses
 import datetime
 import logging
+import posixpath
 import typing
 from io import BytesIO
+from urllib.parse import urlparse
 from zipfile import ZipFile
 
 from packaging.metadata import Metadata
@@ -10,6 +12,7 @@ from packaging.requirements import Requirement
 from packaging.utils import BuildTag, canonicalize_name
 from packaging.version import Version
 
+from .pkgmetadata.pep376 import dist_info_name
 from .request_session import session
 
 logger = logging.getLogger(__name__)
@@ -141,12 +144,15 @@ def get_metadata_for_wheel(
     logger.debug(f"Downloading full wheel to extract metadata: {url}")
     data = session.get(url).content
     with ZipFile(BytesIO(data)) as z:
-        for n in z.namelist():
-            if n.endswith(".dist-info/METADATA"):
-                metadata_content = z.read(n)
-                # Parse metadata directly using packaging.metadata.Metadata
-                # (avoiding circular import with dependencies module)
-                return Metadata.from_email(metadata_content, validate=validate)
+        metadata_path = _wheel_metadata_path(url)
+        try:
+            metadata_content = z.read(metadata_path)
+        except KeyError as err:
+            raise ValueError(f"Could not find {metadata_path} in wheel: {url}") from err
+        return Metadata.from_email(metadata_content, validate=validate)
 
-    # If we didn't find the metadata, raise an error
-    raise ValueError(f"Could not find METADATA file in wheel: {url}")
+
+def _wheel_metadata_path(url: str) -> str:
+    """Derive the METADATA path inside a wheel from its URL."""
+    filename = posixpath.basename(urlparse(url).path)
+    return f"{dist_info_name(filename)}/METADATA"
--- a/src/fromager/wheels.py
+++ b/src/fromager/wheels.py
@@ -17,7 +17,6 @@ from packaging.requirements import Requirement
 from packaging.tags import Tag
 from packaging.utils import (
     BuildTag,
-    canonicalize_name,
     parse_wheel_filename,
 )
 from packaging.version import Version
@@ -33,6 +32,7 @@ from . import (
     sbom,
     sources,
 )
+from .pkgmetadata.pep376 import verbatim_dist_name
 
 if typing.TYPE_CHECKING:
     from . import build_environment, context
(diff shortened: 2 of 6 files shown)


src/fromager/candidate.py
```diff
@@ -141,12 +143,27 @@ def get_metadata_for_wheel(
...
-
-    # If we didn't find the metadata, raise an error
-    raise ValueError(f"Could not find METADATA file in wheel: {url}")
+        metadata_path = _wheel_metadata_path(url)
+        try:
+            metadata_content = z.read(metadata_path)
+        except KeyError:
+            raise ValueError(
+                f"Could not find {metadata_path} in wheel: {url}"
+            ) from None
```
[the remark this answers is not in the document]
rd4398: fixed in latest commit

src/fromager/candidate.py
```diff
@@ -141,12 +143,27 @@ def get_metadata_for_wheel(
...
+    """Derive the METADATA path inside a wheel from its URL.
+
+    Per the wheel spec, the dist-info directory is
+    ``{name}-{version}.dist-info`` where *name* and *version* come from the
+    first two ``-``-separated segments of the wheel filename.
+    """
+    filename = posixpath.basename(urlparse(url).path)
+    parts = filename.split("-")
+    if len(parts) < 3:
+        raise ValueError(f"Cannot parse wheel filename from URL: {url}")
```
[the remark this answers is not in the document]
rd4398: fixed in latest commit

src/fromager/candidate.py
```diff
@@ -141,12 +143,27 @@ def get_metadata_for_wheel(
...
-    raise ValueError(f"Could not find METADATA file in wheel: {url}")
+        metadata_path = _wheel_metadata_path(url)
+        try:
+            metadata_content = z.read(metadata_path)
+        except KeyError as err:
+            raise ValueError(f"Could not find {metadata_path} in wheel: {url}") from err
+        return Metadata.from_email(metadata_content, validate=validate)
+
+
+def _wheel_metadata_path(url: str) -> str:
```
smoparth, line 155: Consider using **parse_wheel_filename**() from packaging.utils instead of manually parsing the wheel filename. Note that it returns canonicalized names, so use filename.split("-", 1)[0] for the dist name (needed for case-sensitive ZIP lookups):
```

filename = posixpath.basename(urlparse(url).path)
_, dist_version, _, _ = parse_wheel_filename(filename)
dist_name = filename.split("-", 1)[0]
return f"{dist_name}-{dist_version}.dist-info/METADATA"
```

smoparth (dismissed):

src/fromager/candidate.py
tiran: Fromager has four implementations of mapping a wheel filename to a dist-info directory. The PR is adding the third implementation of extracting dist name and version from a wheel filename. I would prefer to have a single implementation of these algorithms.

rd4398: Done, latest commit now has:

  - New dist_info_name() in candidate.py — single source of truth for {name}-{version}.dist-info, uses parse_wheel_filename() for validation, preserves verbatim casing
  - wheels.py — extract_info_from_wheel_file() now delegates to dist_info_name() instead of reimplementing name extraction
  - dependencies.py — _get_metadata_from_wheel() replaced 4 lines of inline parsing with one dist_info_name() call; removed unused parse_wheel_filename import
  - Tests — added TestDistInfoName class with direct tests; kept existing URL-integration and wheel-extraction tests

src/fromager/wheels.py
```diff
@@ -143,20 +142,11 @@ def extract_info_from_wheel_file(
...
-        wheel_file.name
-    )
-    dist_name = wheel_file.name.split("-", 1)[0]
-    if dist_name_normalized != canonicalize_name(dist_name):
-        # sanity check, should never fail
-        raise ValueError(f"{dist_name_normalized} does not match {dist_name}")
-    return (dist_name, dist_version, build_tag, wheel_tags)
+    dist_name, dist_version_str = wheel_file.name.split("-", 2)[:2]
+    _, _, build_tag, wheel_tags = parse_wheel_filename(wheel_file.name)
+    return (dist_name, Version(dist_version_str), build_tag, wheel_tags)
```
tiran: Call `parse_wheel_filename` first to validate the wheel filename. This results in a better error message than `ValueError: not enough values to unpack (expected 2, got 1)` for a completely broken wheel file name.

You can use the version object from `parse_wheel_filename`:

```suggestion
    _, version, build_tag, wheel_tags = parse_wheel_filename(wheel_file.name)
    dist_name = wheel_file.name.split("-", 1)[0]
    return (dist_name, version, build_tag, wheel_tags)
```

src/fromager/candidate.py
```diff
@@ -141,12 +143,28 @@ def get_metadata_for_wheel(
...
-    raise ValueError(f"Could not find METADATA file in wheel: {url}")
+        metadata_path = _wheel_metadata_path(url)
+        try:
+            metadata_content = z.read(metadata_path)
+        except KeyError as err:
+            raise ValueError(f"Could not find {metadata_path} in wheel: {url}") from err
+        return Metadata.from_email(metadata_content, validate=validate)
+
+
+def dist_info_name(wheel_filename: str) -> str:
```
tiran: `fromager.candidate` is a weird place for PEP 376 dist-info handling. How about `fromager.pkgmetadata.pep376` instead?

Background: We have defined helper methods where they were used first, then imported them into other modules later. It's sometimes causing circular import issues.

rd4398: Hmm, makes sense. Done in latest commit

src/fromager/wheels.py
rd4398: Good idea, done in latest commit.

rd4398:
@tiran I have addressed the latest feedback and this ready for review again.

src/fromager/wheels.py
```diff
@@ -142,20 +142,11 @@ def extract_info_from_wheel_file(
...
+    (e.g. ``MarkupSafe``, not ``markupsafe``).
     """
-    # parse_wheel_filename normalizes the dist name, however the dist-info
-    # directory uses the verbatim distribution name from the wheel file.
-    # Packages with upper case names like "MarkupSafe" are affected.
-    dist_name_normalized, dist_version, build_tag, wheel_tags = parse_wheel_filename(
-        wheel_file.name
-    )
+    _, version, build_tag, wheel_tags = parse_wheel_filename(wheel_file.name)
     dist_name = wheel_file.name.split("-", 1)[0]
```
LalatenduMohanty: This is a duplicate logic we have in src/fromager/pkgmetadata/pep376.py

LalatenduMohanty: Should we add verbatim_dist_name() helper to pep376.py

LalatenduMohanty (changes requested):

src/fromager/wheels.py
rd4398: Good idea!
Latest commit has following changes:
  - pep376.py — added verbatim_dist_name() that validates via parse_wheel_filename and returns the original-cased dist name
  - wheels.py — extract_info_from_wheel_file() now calls verbatim_dist_name() instead of inlining wheel_file.name.split("-", 1)[0]
  - pkgmetadata/__init__.py — exported verbatim_dist_name in __all__

LalatenduMohanty (approved):

PR #1227, opened 2026-07-03
refactor(downloads): centralize download helpers in new module

tiran:
# Pull Request Description

## What

Create `fromager.downloads` as a single import point for all download helpers, avoiding cyclic import issues between `sources`, `wheels`, and `resolver`.

Move `download_url` and `extract_filename_from_url` from `sources` and `resolver` into the new module.

Add `download_sdist` and `download_wheel` that validate filenames with `parse_sdist_filename` / `parse_wheel_filename` before downloading and verify archives are non-empty afterwards.

Add `download_git_source` for cloning from pip VCS URLs using `git_clone_fast` with destination directory conflict detection.

Improve `download_url` temp file handling: use `NamedTemporaryFile` in the target directory with `os.rename` for atomic moves.

Replace `_download_source_check` and `_download_wheel_check` with the new `download_sdist` and `download_wheel` functions.

## Why

Make the code reusable without triggering cyclic import issues.

### DIFF
--- a/src/fromager/wheels.py
+++ b/src/fromager/wheels.py
@@ -12,7 +12,6 @@ import zipfile
 
 import elfdeps
 import tomlkit
-import wheel.wheelfile  # type: ignore
 from packaging.requirements import Requirement
 from packaging.tags import Tag
 from packaging.utils import (
@@ -24,6 +23,7 @@ from packaging.version import Version
 
 from . import (
     dependencies,
+    downloads,
     external_commands,
     metrics,
     overrides,
@@ -31,7 +31,6 @@ from . import (
     requirements_file,
     resolver,
     sbom,
-    sources,
 )
 
 if typing.TYPE_CHECKING:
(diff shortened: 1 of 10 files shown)


src/fromager/wheels.py
```diff
@@ -454,30 +453,20 @@ def download_wheel(
     wheel_url: str,
     output_directory: pathlib.Path,
 ) -> pathlib.Path:
-    wheel_filename = output_directory / resolver.extract_filename_from_url(wheel_url)
+    wheel_filename = output_directory / downloads.extract_filename_from_url(wheel_url)
     if not wheel_filename.exists():
         logger.info(f"downloading pre-built wheel {wheel_url}")
-        wheel_filename = _download_wheel_check(req, output_directory, wheel_url)
+        wheel_filename = downloads.download_wheel(
```
rd4398, line 459: The old `_download_wheel_check` validated using `wheel.wheelfile.WheelFile(wheel_filename)`, which opens the zip and checks wheel-specific structure. The new `download_wheel` only does:

```
  with zipfile.ZipFile(filepath) as zf:
      if not zf.namelist():
          raise zipfile.BadZipFile(f"empty wheel file: {filepath}")
```
This catches empty zips but won't catch structurally invalid wheels (e.g., a zip that has files but no .dist-info/WHEEL). The `parse_wheel_filename` call before download validates the name, not the content.

Was this intentional? The validation is slightly weaker now.

rd4398 (commented):
Overall looks good. I have added a comment for a clarification question.

src/fromager/wheels.py
tiran: I have re-added the WheelFile check with some improvements. The original code did not close the file. WheelFile also does not check for the presence of `METADATA`. It only checks for `RECORD`.

tiran:
@rd4398 I have addressed your comment and rebased the PR to include the CodeQL fix. The doc strings now also explains how much validation the functions perform.

rd4398 (approved):
This looks good! Thanks

PR #1229, opened 2026-07-06
feat(gitutils): add `git_clean` to remove untracked files

tiran:
# Pull Request Description

## What

add `git_clean` to remove untracked files from a working tree

## Why

Some projects write files when the PEP 517 hook `prepare_metadata_for_build_wheel` is called, e.g. projects that unconditionally compile code in setup.py. Add a `git_clean` wrapper that runs `git clean -xdf` to restore a working tree to a pristine state after such hooks.

### DIFF
--- a/src/fromager/gitutils.py
+++ b/src/fromager/gitutils.py
@@ -140,6 +140,19 @@ def git_clone_fast(
         logger.debug("no .gitmodules file")
 
 
+def git_clean(source_dir: pathlib.Path) -> None:
+    """Remove untracked files and directories from a git working tree.
+
+    Runs ``git clean -xdf`` in *source_dir* to remove all untracked and
+    ignored files, restoring the working tree to a pristine state.
+
+    .. versionadded:: 0.90.0
+    """
+    cmd: list[str] = ["git", "clean", "-xdf"]
+    logger.debug("cleaning untracked files in %s", source_dir)
+    external_commands.run(cmd, cwd=str(source_dir))
+
+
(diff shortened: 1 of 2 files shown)


src/fromager/gitutils.py
```diff
@@ -140,6 +140,19 @@ def git_clone_fast(
         logger.debug("no .gitmodules file")
 
 
+def git_clean(source_dir: pathlib.Path) -> None:
```
smoparth, line 143: QQ: prepare_metadata_for_build_wheel is called for both git-cloned sources and extracted sdist tarballs. In the follow-up that wires this in, will callers be responsible for checking whether the source is a git repo before calling git_clean, or would it be cleaner to have git_clean itself no-op when source_dir isn't a git working tree (early return if .git doesn't exist)?

smoparth (approved):
LGTM!

rd4398 (approved):
LGTM!

src/fromager/gitutils.py
tiran: Actually we have three distinct cases:

- sdist tarball with correct layout and a PKG-INFO file
- git checkout
- git archival tarball with incorrect layout and no PKG-INFO file

Long term, we should stop building a sdist from a plain git archive tarball. It does not work when the build backend assumes the presence of a SCM/VCS. Upstream projects don't use git archive tarballs to build sdists and wheels.

Let's keep this function simple and propagate the error.

PR #1235, opened 2026-07-07
feat(resolver): add download method to all resolver classes

tiran:
# Pull Request Description

## What

Add `_download()` to `AbstractResolver` that dispatches to the
appropriate download helper based on a mandatory `download_kind`
argument.  Each concrete resolver implements a public `download()`
that delegates with its specific `DownloadKind`.

- `PyPIDownloadResolver` accepts `download_kind` as a mandatory
  Pydantic field (restricted to `sdist | tarball`)
- `NotAvailableResolver` and `AbstractHookResolver` override
  `download()` to raise explicitly
- Replace the old `download_kind` ClassVar with a `download_kinds`
  frozenset ClassVar describing which kinds each resolver can return
- Replace `resolves_prebuilt_wheel` property with a ClassVar
- Validate `download_kind` against `download_kinds` in `_download()`

## Why

Include the download logic in the resolver.

### DIFF
--- a/src/fromager/packagesettings/_resolver.py
+++ b/src/fromager/packagesettings/_resolver.py
@@ -4,12 +4,13 @@ import datetime
 import enum
 import inspect
 import logging
+import pathlib
 import re
 import typing
 
 import pydantic
 
-from .. import resolver
+from .. import downloads, resolver
 from ..candidate import Cooldown
 from ._typedefs import MODEL_CONFIG
 
@@ -17,6 +18,7 @@ if typing.TYPE_CHECKING:
     from packaging.requirements import Requirement
 
     from .. import context, requirements_file
+    from ..candidate import Candidate
 
 logger = logging.getLogger(__name__)
 
@@ -42,6 +44,9 @@ class DownloadKind(enum.StrEnum):
         return self is not DownloadKind.not_available
 
 
+DownloadKindSet = frozenset[DownloadKind]
+
+
 class AbstractResolver(pydantic.BaseModel):
     """Abstract base class for resolvers"""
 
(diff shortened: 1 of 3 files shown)


src/fromager/packagesettings/_resolver.py
```diff
@@ -135,6 +201,15 @@ def resolver_provider(
             cooldown=self._cooldown,
         )
 
+    def download(
```
rd4398, line 213: Every concrete class defines download(), but AbstractResolver itself doesn't declare it. Can we consider adding a base definition something like:

```
def download(
      self,
      ctx: context.WorkContext,
      req: Requirement,
      candidate: Candidate,
  ) -> tuple[pathlib.Path, DownloadKind]:
      """Download the resolved artifact."""
      raise NotImplementedError
```
This would make the API contract explicit and give mypy something to check against.

src/fromager/packagesettings/_resolver.py
```diff
@@ -205,7 +293,12 @@ class PyPIDownloadResolver(AbstractPyPIResolver):
     """
 
     build_sdist: typing.ClassVar[BuildSDist | None] = BuildSDist.tarball
-    download_kind: typing.ClassVar[DownloadKind] = DownloadKind.tarball
+    download_kinds: typing.ClassVar[DownloadKindSet] = frozenset(
+        {DownloadKind.sdist, DownloadKind.tarball}
+    )
+
+    download_kind: typing.Literal[DownloadKind.sdist, DownloadKind.tarball]
```
rd4398, line 309: I think this is a breaking change.  Can we consider

  -   Noting it with a .. versionchanged:: directive in the docstring?

rd4398 (commented):
Overall looks good. I added couple of comments / suggestions

src/fromager/packagesettings/_resolver.py
tiran: The new code was introduced after the latest release.

src/fromager/packagesettings/_resolver.py
tiran: Good idea! I have added a dummy function that raises NotImplementedError

src/fromager/packagesettings/_resolver.py
rd4398: Ah, right, makes sense.

rd4398 (approved):
This looks good! Thank you

Issue #1214, opened 2026-06-24
[Bug]: fromager might "lie" about the reason of dependency resolution issue


frantisekz:
### Environment

- Fromager version: 0.88.0
- Python version: 3.12
- OS / architecture: el9
- Container or local: container

### Bug description

**Expected behavior:**
fromager stating the exact reason of the broken dependency chain.

**Actual behavior:**
fromager might "lie" (or not tell the full/exact truth and facts) about why some dependency/chain isn't resolvable.

**Steps to reproduce / logs:**
Have ENABLE_REPEATABLE_BUILD_MODE: true; break the resolvability of indirect dependency, observe fromager reporting eg.

```09:05:45 ERROR its-hub: Unable to resolve requirement specifier flashinfer-python==0.6.8.post1 with constraint flashinfer-python==0.6.11.post2 using PyPI resolver (searching at https://pypi.org/simple): found no match for flashinfer-python==0.6.8.post1 using PyPI resolver (searching at https://pypi.org/simple), searching for sdists, ignoring pre-release versions because found no match for flashinfer-python==0.6.8.post1 using PyPI resolver (searching at https://pypi.org/simple), searching for sdists, ignoring pre-release versions```

while the culprit would be having indirect its-hub -> reward-hub -> vllm broken on the vllm arc. fromager should in this case state the resolvability is broken due to vllm, not because of its-hub.

**Comment by LalatenduMohanty** [comment]
## Investigation findings

### What happens (v0.88.0)

The error message in the report:

```
ERROR its-hub: Unable to resolve requirement specifier flashinfer-python==0.6.8.post1
  with constraint flashinfer-python==0.6.11.post2 ...
```

has two problems:

1. **Wrong prefix**: The `its-hub:` prefix is misleading — the failing package is `flashinfer-python`, not `its-hub`.
2. **Missing dependency chain**: The user has to guess that the path is `its-hub → reward-hub → vllm → flashinfer-python`. Fromager knows this chain internally but never includes it in the error.

### Root cause

**Why the prefix is wrong (v0.88.0)**

In v0.88.0, `commands/bootstrap.py` wrapped the entire bootstrap call for each top-level requirement inside `requirement_ctxvar.set(req)` — intentionally without `try/finally` so the top-level name would survive in the log on failure. When a transitive dependency like `flashinfer-python` failed deep in the tree, the context variable reset back to the outer value (`its-hub`), which then appeared as the log prefix.

[...]

PR #1226, opened 2026-07-02
refactor(bootstrapper): decompose monolithic bootstrapper into a subpackage

dhellmann:
## Summary

This branch refactors the monolithic `bootstrapper.py` (2278 lines) into a well-structured subpackage with focused modules, making the code easier to navigate, test, and extend.

### What changed

- **Extracted phase class hierarchy** (`Phase` and its subclasses) out of `Bootstrapper` into individual modules, one per bootstrap phase
- **Converted `bootstrapper.py` to a subpackage** (`bootstrapper/`) with 10 focused modules: `_phase.py`, `_resolve.py`, `_start.py`, `_prepare_source.py`, `_prepare_build.py`, `_build.py`, `_process_install_deps.py`, `_complete.py`, `_work_item.py`, `_cache.py`
- **Moved logic to the right owners**: `do_build` and build helpers → `Build`, install-dep extraction → module-level function in `_process_install_deps`, `processing_build_requirement` state → `WorkItem`
- **Cleaned up `Bootstrapper`**: inlined `_create_build_env`, moved exclusive-build pool drain into the loop, made internal members public where phase classes need them, exposed `_resolver` as a property, wrapped `_failed_versions` behind `has_failed_version()`
- **Renamed phase classes** to drop the redundant `Item` suffix (`PhaseItem`→`Phase`, `BuildItem`→`Build`, `ResolveItem`→`Resolve`, etc.) and renamed files to match

### Commit sequence

1. `design: phase classes` — design doc / plan
2. `refactor: extract PhaseItem class hierarchy` — initial extraction
3. Series of encapsulation commits making `Bootstrapper` members accessible to phase classes
4. Logic migration commits (do_build, install deps, WorkItem state)
5. `refactor: convert bootstrapper.py to a subpackage`
6. `refactor: rename phase classes and files, drop "Item" suffix`

## Test plan

- [x] All existing tests pass (`hatch run test:test`)
- [x] mypy clean (`hatch run mypy:check`)
- [x] lint clean (`hatch run lint:check`)

## Advice to reviewers

There are no new features here, but there are a lot of code changes. It may be easier to review them by looking at the commit history individually instead of trying to understand the entire changeset as one big diff.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1241, opened 2026-07-08
Include dependency chain in resolution error messages

raravind007:
## Summary

Partially addresses #1214 — see #1242 for the remaining piece (misleading
log line prefix).

Resolution failures were only enriched with dependency-chain context in
`_phase_resolve()`, and only for `ResolverException`. As pointed out in
review, that missed several cases: a plain `RuntimeError("Could not
resolve any versions...")` in the same phase, and any failure recorded
in test mode (`_record_test_mode_failure`) or multiple-versions mode
(`_record_failed_version`) — both received the raw, un-enriched
exception.

`_handle_phase_error()` is the single centralized handler for errors
from every bootstrap phase, so the enrichment now lives there instead:
`_enrich_resolution_error()` attaches the dependency chain (from
`WorkItem.why_snapshot`, via `_dependency_chain_suffix()`) to any
RESOLVE-phase failure with a non-empty chain, regardless of exception
type, before test-mode/multiple-versions recording or the final raise.
For example:

`Unable to resolve requirement specifier flashinfer-python==0.6.8.post1 ... (dependency chain: its-hub==1.0 -> reward-hub==2.0 -> vllm==3.0 -> flashinfer-python==0.6.8.post1)`

- The enriched exception is constructed and raised without a `from`
  clause. Python still sets `__context__` to the original exception,
  keeping it available for traceback/debugging, while leaving
  `__cause__` unset, so `__main__._format_exception()` — which only
  follows `__cause__` — doesn't print the original message a second
  time via `"... because ..."`.
- `_format_exception()` itself is unchanged in this PR. A related,
  pre-existing duplication bug in that formatter (independent of this
  feature) is tracked separately as #1243.
- Top-level failures (no chain) and non-RESOLVE-phase failures are
  returned unchanged, preserving the original exception object's
  identity.

## Test plan

- [x] `pytest tests/test_bootstrapper_iterative.py` — `_enrich_resolution_error`
      covered in isolation (`ResolverException`, `RuntimeError`, unknown
      exception fallback, no-chain/non-RESOLVE passthrough, and the
      `__cause__`/`__context__` behavior), plus integration tests proving
      the enrichment reaches test-mode and multiple-versions recording,
      and the normal-mode `RuntimeError` path.
- [x] `pytest tests/test_external_commands.py`, `pytest tests/test_resolver.py`
- [x] `ruff check` / `ruff format --check`

### DIFF
--- a/src/fromager/bootstrapper/_bootstrapper.py
+++ b/src/fromager/bootstrapper/_bootstrapper.py
@@ -14,6 +14,7 @@ import typing
 from packaging.requirements import Requirement
 from packaging.utils import NormalizedName, canonicalize_name
 from packaging.version import Version
+from resolvelib.resolvers import ResolverException
 
 from .. import (
     bootstrap_requirement_resolver,
@@ -1127,6 +1128,54 @@ class Bootstrapper:
             unsatisfied.add(dep)
         return unsatisfied
 
+    def _dependency_chain_suffix(self, wi: WorkItem) -> str:
+        """Describe the requirement chain that pulled in a failed dependency.
+
+        Resolution failures are logged with whatever top-level package
+        happens to be active in the logging context at the time, which can
+        make an unrelated package look like the culprit when the actual
+        conflict is several levels down the dependency tree (see #1214).
+        This spells out the real chain, top-down, so the message is
+        accurate even when the log prefix is not.
+
+        Callers should only invoke this when ``wi.why_snapshot`` is
+        non-empty (i.e. not a top-level requirement).
+        """
+        chain = " -> ".join(
+            f"{req.name}=={version}" for _, req, version in wi.why_snapshot
+        )
+        return f" (dependency chain: {chain} -> {wi.req})"
+
+    def _enrich_resolution_error(self, item: Phase, err: Exception) -> Exception:
+        """Attach the dependency chain to a RESOLVE-phase failure.
+
+        Returns ``err`` unchanged for non-RESOLVE phases or top-level
+        requirements (no chain to add). Otherwise returns a *new* exception
+        of the same shape with the chain appended. The caller
+        (``_handle_phase_error()``) raises this new exception with no
+        `from` clause while still inside the original exception's active
+        handler; Python then sets the new exception's `__context__` to the
+        original one (keeping it available for traceback/debugging) while
+        leaving `__cause__` unset, so `__main__._format_exception()` -
+        which only follows `__cause__` - doesn't print the original message
+        a second time via "... because ...". See #1243 for the pre-existing,
+        independent duplication bug in that formatter.
+
+        `type(err)(msg)` isn't guaranteed to work for every exception
+        class, so only the two shapes that actually reach this code path
+        are reconstructed directly; anything else falls back to a
+        `RuntimeError` that keeps the original type name in the message.
+        """
+        wi = item.work_item
+        if not isinstance(item, Resolve) or not wi.why_snapshot:
+            return err
+        msg = f"{err}{self._dependency_chain_suffix(wi)}"
+        if isinstance(err, ResolverException):
+            return ResolverException(msg)
+        if isinstance(err, RuntimeError):
+            return RuntimeError(msg)
+        return RuntimeError(f"{type(err).__name__}: {msg}")
+
     def _handle_phase_error(
         self,
         item: Phase,
@@ -1140,6 +1189,7 @@ class Bootstrapper:
         wi = item.work_item
         # Resolution failures: recoverable in test mode and multiple versions mode
         if isinstance(item, Resolve):
+            err = self._enrich_resolution_error(item, err)
             if self.test_mode:
                 self.record_test_mode_failure(wi.req, None, err, "resolution")
                 if self.multiple_versions:
@@ -1158,7 +1208,7 @@ class Bootstrapper:
                     f"failed during {type(item).phase} phase",
                 )
                 return []
-            raise
+            raise err
 
         # Test mode: try prebuilt fallback for build-related phases
         if self.test_mode:
--- a/tests/test_bootstrapper_iterative.py
+++ b/tests/test_bootstrapper_iterative.py
@@ -27,6 +27,7 @@ import pytest
 from packaging.requirements import Requirement
 from packaging.utils import canonicalize_name
 from packaging.version import Version
+from resolvelib.resolvers import ResolverException
 
 from fromager import bootstrapper, build_environment
 from fromager.bootstrapper._build import Build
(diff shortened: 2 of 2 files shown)


raravind007:
Ran an AI-assisted code review (self-review) against this diff before/after opening the PR, which caught a couple of real issues, now fixed in the second commit:

- The original version suppressed the exception cause (`raise ... from None`) to avoid `__main__._format_exception()` printing the resolver message twice via `"... because ..."`. That broke this project's own documented convention (AGENTS.md: "Chain exceptions: `raise ValueError(...) from err`") and silently discarded the original traceback that debug/test-mode logging relies on.
- Fixed by keeping `from err` (preserving the real traceback/cause chain) and instead teaching `_format_exception()` to skip the `"because"` clause when the cause's message is already embedded in the outer message — the actual duplication this needed to avoid, fixed at the right layer instead of a local workaround.
- Also dropped a small redundant empty-string-as-boolean round trip between `_dependency_chain_suffix()` and its caller.
- Rebased to add the DCO `Signed-off-by` trailer to both commits (the first was missing it) — hence the force-push; branch still has the same two logical commits.

Happy to share the fuller review notes if useful.

src/fromager/__main__.py
```diff
@@ -299,8 +299,15 @@ def main(
...
 def _format_exception(exc: BaseException) -> str:
     if exc.__cause__:
+        exc_str = str(exc)
         cause = _format_exception(exc.__cause__)
-        return f"{exc} because {cause}"
+        # Skip the "because" clause when the cause's message is already
+        # embedded in the outer exception's own message (e.g. an exception
+        # re-raised with extra context appended to the original text) —
+        # otherwise the cause would be printed a second time verbatim.
+        if cause and cause in exc_str:
```
rd4398: This looks very fragile.  A safer check would be to verify the outer message starts with the cause text, since the pattern is always f"{err}{suffix}":

```
  if cause and exc_str.startswith(cause):
      return exc_str
```
Or even more precise: we could check that the outer message equals cause + some suffix. The startswith variant is still general enough for any append context to original message pattern while avoiding false positives from coincidental substring matches

rd4398 (changes requested):
I don't think this fully solves https://github.com/python-wheel-build/fromager/issues/1214

There are two separate issues that need to be addressed:
1. The error message body doesn't show the dependency path (PR fixes this)
The PR appends (dependency chain: its-hub==1.0 -> reward-hub==2.0 -> vllm==3.0 -> flashinfer-python==0.6.8.post1) to the ResolverException message. That helps a user trace the real path.
2. The its-hub: log prefix is still wrong (PR does NOT fix this)
The misleading prefix comes from `requirement_ctxvar`.

I have also added a suggestion for current PR. Can you take a look at this and wither send a second PR or change the original issue description and create a follow up card? cc @LalatenduMohanty for awareness

src/fromager/__main__.py
raravind007: Thanks for the `startswith` suggestion, Rohan! I applied that, but while adding a regression test using the actual nested resolver-style exception shape, I found one more issue. :)

The check should compare against the immediate cause's raw message, `str(exc.__cause__)`, rather than the recursively formatted cause. When the cause itself has a cause, `_format_exception(exc.__cause__)` can already include its own `" because ..."` suffix, which no longer matches the prefix of the outer `f"{err}{suffix}"` message.

I updated the formatter to use `str(exc.__cause__)` for the dedupe check, while still using the recursively formatted cause when the `"because"` clause is actually emitted. I also added a regression test for the real resolver.py → `_phase_resolve()` wrapping shape. Pushed this as a separate commit so the correction is clear in the PR history.

LalatenduMohanty (changes requested):
Nice fix — the `_dependency_chain_suffix()` helper and the test coverage are solid. A couple of structural suggestions.

  ### 1. Move enrichment from `_phase_resolve()` to `_handle_phase_error()`

  `_handle_phase_error()` is the centralized error handler — the natural place for error enrichment. The current placement in `_phase_resolve()` has three gaps:

  | Gap | Why it matters |
  |-----|---------------|
  | Only catches `ResolverException` | `RuntimeError("Could not resolve any versions")` at line 1638 gets no chain |
  | Test mode doesn't benefit | `_record_test_mode_failure(err)` captures `str(err)` in `failed_packages[].exception_message`, but `err` is un-enriched |
  | Same for `multiple_versions` mode | `_record_failed_version()` also misses the chain |

  ### 2. Drop the `_format_exception()` changes

  The `startswith` dedup in `_format_exception()` fixes a pre-existing duplication bug independent of this feature — filed as #1243 for a follow-up PR. Removing it from this
  PR simplifies the change and keeps the git history focused.

  Without the dedup, use `from None` instead of `from err` to avoid the duplicated "because" output. The traceback loss can be mitigated by logging `exc_info` in
  `_handle_phase_error()` before re-raising — you're still inside the `with req_ctxvar_context(item.req)` block there.

LalatenduMohanty:
@raravind007 Also the fix should be one commit and if you decide to fix #1243 then two commits.

raravind007:
Hello Lala, applied both points and many thanks for the detailed breakdown.

Moved enrichment into _handle_phase_error() via a new _enrich_resolution_error() helper, applied before test-mode / multiple-versions recording and the final raise. It now covers the RuntimeError("Could not resolve any versions...") case too, not just ResolverException. The helper uses explicit isinstance checks with a RuntimeError fallback for anything else, since type(err)(msg) is not safe for arbitrary exception classes.

Dropped the _format_exception() changes entirely — reverted `__main__.py` and its tests back to the original. That fix is now left for #1243, as you filed it.

On from None + manual exc_info logging vs. plain raise err: I went with plain raise err — no from clause — instead. Since the enriched exception is newly constructed inside the active exception handler, Python sets its __context__ to the original exception automatically while leaving __cause__ unset. That keeps the original exception available for traceback/debugging, while _format_exception() remains unchanged and does not print the message twice because it only follows `__cause__.`

I added a regression test for that behavior directly:

assert final.__cause__ is None
assert final.__context__ is original
assert __main__._format_exception(final).count("found no match") == 1

Happy to switch to your literal from None + explicit debug logging suggestion if you would rather keep the enrichment logic and traceback-preservation logic more explicitly separate. Just a judgment call, not a strong preference.

raravind007:
Yes, that's the update I meant. I pushed f9ba1e3, which moves the dependency-chain enrichment into _handle_phase_error(), removes the _phase_resolve()-specific ResolverException wrapper, and reverts the _format_exception() / test_external_commands.py changes so that formatter work stays in #1243.

Tightened the helper's docstring: __context__ is set when _handle_phase_error() raises the new exception (still inside the original exception's active handler), not when the helper constructs it — the helper itself never raises anything.

raravind007:
Updated again: the PR now only includes the #1214 dependency-chain enrichment work. I removed the _format_exception() changes from the final diff, kept that for #1243, moved enrichment into _handle_phase_error(), and fixed the helper docstring wording around when __context__ is set.

I also squashed the branch down to one focused commit as requested. Tests and lint still pass.I hope we are set now.:)

raravind007:
Rebased onto main to resolve the merge conflict caused by the bootstrapper package split in #1226. Since bootstrapper.py no longer exists as a single file, I manually ported the change rather than resolving the old file conflict mechanically.

_dependency_chain_suffix() and _enrich_resolution_error() now live on Bootstrapper in src/fromager/bootstrapper/_bootstrapper.py. The enrichment condition in _handle_phase_error() now uses isinstance(item, Resolve) instead of the previous item.phase == BootstrapPhase.RESOLVE.

The behavior and test coverage are otherwise unchanged from the previous version. The PR is still one commit, and tests, lint, and mypy all pass.

rd4398 (approved):
This looks good to me! I will wait for @LalatenduMohanty to approve as well since he requested changes.

LalatenduMohanty (approved):

LalatenduMohanty:
@mergify update

LalatenduMohanty:
@raravind007 Thank you!

PR #1249, opened 2026-07-10
perf(bootstrapper): reduce disk I/O with throttled stack writes, async graph writes, deferred build-order write

dhellmann:
Three file-write hotspots in the bootstrap loop were generating O(n²) disk I/O on the main thread:

1. `_record_stack_state` was called on every loop iteration, serializing and writing the full stack to bootstrap-stack.json. Throttled to at most once every 5 seconds (force-flush in `finalize()`).

2. `add_to_graph` → `write_to_graph_to_file()` wrote the full graph on every dependency edge. Graph writes are now moved to a single-threaded background `_write_pool` via `_write_graph_async()`, keeping serialization on the main thread for a consistent snapshot but eliminating blocking I/O from the critical path.

3. `add_to_build_order` rewrote build-order.json on every package. The disk write is removed entirely; data stays in `_build_stack` in memory and is written once in `finalize()`.

Also renames `write_to_graph_to_file()` → `write_graph_file()` in `WorkContext` for clarity.

Tests updated to reflect deferred build-order write (call `finalize()` before reading the file) and the stack-write throttle. Two new tests added: one verifying the throttle suppresses rapid writes, one verifying `finalize()` drains the write pool and produces both output files.

### DIFF
--- a/src/fromager/bootstrapper/_bootstrapper.py
+++ b/src/fromager/bootstrapper/_bootstrapper.py
@@ -3,12 +3,14 @@ from __future__ import annotations
 import concurrent.futures
 import contextlib
 import datetime
+import io
 import json
 import logging
 import operator
 import pathlib
 import shutil
 import tempfile
+import time
 import typing
 
 from packaging.requirements import Requirement
@@ -109,6 +111,16 @@ class Bootstrapper:
         self._build_order_filename = self.ctx.work_dir / "build-order.json"
         self._stack_filename = self.ctx.work_dir / "bootstrap-stack.json"
         logger.info("recording bootstrap stack state to %s", self._stack_filename)
+
+        # Single-threaded pool for background file writes (serialized, never interleave)
+        self._write_pool: concurrent.futures.ThreadPoolExecutor | None = (
+            concurrent.futures.ThreadPoolExecutor(
+                max_workers=1, thread_name_prefix="fromager-writes"
+            )
+        )
+        # Throttle stack-state writes to at most once per interval
+        self._last_stack_write: float = 0.0
+        self._STACK_WRITE_INTERVAL: float = 5.0
 
         # Track failed packages in test mode (list of typed dicts for JSON export)
         self.failed_packages: list[FailureRecord] = []
@@ -856,7 +868,7 @@ class Bootstrapper:
             pre_built=pbi.pre_built,
             constraint=self.ctx.constraints.get_constraint(req.name),
         )
-        self.ctx.write_to_graph_to_file()
+        self._write_graph_async()
 
     def _sort_requirements(
         self,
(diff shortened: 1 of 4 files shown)


src/fromager/bootstrapper/_bootstrapper.py
```diff
@@ -954,21 +966,31 @@ def add_to_build_order(
         if req.url:
             info["source_url"] = req.url
         self._build_stack.append(info)
-        with open(self._build_order_filename, "w") as f:
-            # Set default=str because the why value includes
-            # Requirement and Version instances that can't be
-            # converted to JSON without help.
-            json.dump(self._build_stack, f, indent=2, default=str)
+
+    def _schedule_write(self, path: pathlib.Path, content: str) -> None:
```
rd4398, line 981: `_schedule_write` submits `path.write_text` to the thread pool but discards the Future. If the write fails (disk full, permissions, etc.), the error is silently swallowed without log or exception. Can we consider wrapping the background write with error logging, or collecting the futures and checking .result() in finalize()? So something like:

```
def _schedule_write(self, path: pathlib.Path, content: str) -> None:
      if self._write_pool is not None:
          fut = self._write_pool.submit(path.write_text, content, "utf-8")
          fut.add_done_callback(self._check_write_error)
```
or

```
def _check_write_error(self, fut: concurrent.futures.Future) -> None:
      if exc := fut.exception():
          logger.error("background file write failed: %s", exc)
```

tests/test_bootstrapper.py
```diff
@@ -779,6 +784,46 @@ def test_record_stack_state_overwrites_each_call(tmp_context: WorkContext) -> No
...
+    first_mtime = bt._stack_filename.stat().st_mtime
+
+    # Second call is throttled — file should not change
+    bt._record_stack_state([_make_resolve_item("pkgc")])
+    second_mtime = bt._stack_filename.stat().st_mtime
+
+    assert first_mtime == second_mtime
+
+
+def test_finalize_writes_build_order_and_graph(tmp_context: WorkContext) -> None:
```
rd4398, line 805: This is only asserting build_order file but the name also mentions 'graph'. Can we assert graph file as well?

src/fromager/context.py
```diff
@@ -156,7 +156,8 @@ def uv_clean_cache(self, *reqs: Requirement) -> None:
         cmd.extend(req_list)
         external_commands.run(cmd, extra_environ=extra_environ)
 
-    def write_to_graph_to_file(self) -> None:
+    def write_graph_file(self) -> None:
```
rd4398: Non-blocker nit: I think this becomes a dead code now since bootstrapper uses `_write_graph_async() ` directly. It is still part of public API so I wonder whether we should remove it or keep it as it is.

rd4398 (commented):
This looks really good! I added few comments / suggestions

src/fromager/bootstrapper/_bootstrapper.py
dhellmann: Fixed in 88cf757052bc4b23f645c43db91494f41b21b0f3

tests/test_bootstrapper.py
dhellmann: Fixed in 88cf757052bc4b23f645c43db91494f41b21b0f3

src/fromager/context.py
dhellmann: I went ahead and remove that code in 88cf757052bc4b23f645c43db91494f41b21b0f3

dhellmann:
@coderabbitai review

rd4398 (approved):
This looks good! Thanks

PR #1253, opened 2026-07-13
feat(graph): add --dependencies-only and --dependents-only flags to graph subset

dhellmann:
## Summary

- Add `--dependencies-only` to `graph subset`: restricts output to the target package and its transitive dependencies, excluding any packages that depend on the target.
- Add `--dependents-only` to `graph subset`: restricts output to the target package and its transitive dependents, excluding the target's own dependencies.
- Using both flags together is an error.

## Implementation notes

When `--dependencies-only` is active, ROOT is not reachable via the normal upward traversal, so:
- ROOT is always added to the included node set to keep the graph rooted.
- ROOT's original edges are suppressed during the edge-copy pass (to avoid spurious edges from other top-level packages that happen to be transitive deps of the target).
- An artificial `ROOT → target` edge is added so the subset serializes correctly.

When `--dependents-only` is active, the upward traversal naturally reaches ROOT, so no special handling is needed.

## Test plan

- `test_graph_subset_dependencies_only` — target and its deps are included; dependents are not.
- `test_graph_subset_dependencies_only_no_dependents` — confirms dependents absent vs full subset.
- `test_graph_subset_dependents_only` — target and its dependents are included; target's deps are not.
- `test_graph_subset_dependents_only_and_dependencies_only_are_exclusive` — combining both flags exits non-zero with a clear error message.

### DIFF
--- a/src/fromager/commands/graph.py
+++ b/src/fromager/commands/graph.py
@@ -480,6 +480,12 @@ def why(
     default=False,
     help="Only include the package and its dependencies; exclude dependents (packages that depend on it).",
 )
+@click.option(
+    "--dependents-only",
+    is_flag=True,
+    default=False,
+    help="Only include the package and its dependents; exclude dependencies of the package.",
+)
 @click.argument(
     "graph-file",
     type=str,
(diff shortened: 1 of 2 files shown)


dhellmann:
@rd4398 @LalatenduMohanty I had a need for this feature this morning to prep some data for an internal conversation.

rd4398 (approved):
This looks good!

dhellmann:
> This looks good!

What's the process now for adding things to the merge queue? Do I click the button myself, for example?

rd4398:
@mergify rebase

PR #1260, opened 2026-07-19
refactor: remove deprecated `req.url` (build-from-URL) code paths

### LINKED ISSUE #1220: [Refactor]: Git URL path duplicates build dependency extraction with iterative phase machine
### Description

When bootstrapping a package from a git URL without a version tag (e.g., `stevedore @ git+https://...`), fromager performs build dependency extraction twice:

1. **During version discovery** — `_resolve_version_from_git_url` → `_get_version_from_package_metadata` → `_prepare_build_dependencies` calls PEP 517 hooks (`get_requires_for_build_wheel`, `get_requires_for_build_sdist`) to set up a build environment and extract the version from metadata.

2. **During normal phase processing** — After version discovery, the package enters the iterative phase machine. `_phase_prepare_source` and `_phase_prepare_build` call the same `dependencies.get_build_*_dependencies()` functions, invoking the same PEP 517 hooks again.

The dependency cache files (`build-system-requirements.txt`, `build-backend-requirements.txt`, `build-sdist-requirements.txt`) that would prevent re-extraction are written into a `TemporaryDirectory` during step 1 and destroyed when the git clone is moved to its final working directory. So the phase machine never finds them and re-invokes the hooks.

### What should happen vs what actually happens

```
EXPECTED (PyPI packages — single pass):

    resolve version (from PyPI)
         │
         ▼
    phase machine: PREPARE_SOURCE ─► extract build-system deps    ──► once
                   PREPARE_BUILD  ─► extract backend/sdist deps   ──► once
                   BUILD          ─► build wheel


ACTUAL (git URL without version tag — double pass):

    resolve version:
    ├── clone repo
    ├── extract build-system deps    ──► first time    ┐
    ├── extract backend/sdist deps   ──► first time    ├── cache files written
    ├── run metadata hook            ◄─────────────────┘   to tmpdir, then
    └── move clone, destroy tmpdir   ──► cache files LOST  destroyed
         │
         ▼
    phase machine: PREPARE_SOURCE ─► extract build-system deps    ──► second time
                   PREPARE_BUILD  ─► extract backend/sdist deps   ──► second time
                   BUILD          ─► build wheel
```

### Impact

- **Redundant PEP 517 hook calls** — `get_requires_for_build_wheel` and `get_requires_for_build_sdist` are called twice for every git URL package without a version tag.
- **Redundant dependency resolution** — Build dependencies are resolved twice (though `_has_been_seen()` prevents them from being fully rebuilt).
- **Two separate `BuildEnvironment` instances** are created — one temporary for metadata extraction, one permanent for the actual build.
- **Build-sdist dependencies are bootstrapped unnecessarily** during version discovery — they are not needed for `prepare_metadata_for_build_wheel`.

### Context

The git URL code path predates the iterative phase machine refactor (43bf6a8). The recursive-to-iterative conversion commit explicitly deferred unifying the two paths. Three methods exist solely for the git URL path:

- `_prepare_build_dependencies` — mirrors what PREPARE_SOURCE + PREPARE_BUILD do
- `_handle_build_requirements` — eager sequential bootstrap
- `_bootstrap_one` — single-requirement iterative loop wrapper

Note: The "version known" case (`git+https://...@1.2.3`) does **not** have this problem — `_get_version_from_package_metadata` is never called, and the phase machine handles everything in a single pass.

rd4398:
Remove PEP 508 direct reference URL support (`pkg @ git+https://...`) which was deprecated in proposal #1225. The `source` provider config system covers all use cases declaratively without special-casing.

This removes:
- 19 `if req.url:` branches across 6 source files
- `_resolve_version_from_git_url()`, `_get_version_from_package_metadata()`, `_prepare_build_dependencies()`, `_handle_build_requirements()` methods
- `SourceType.GIT` enum value and `get_top_level_requirement()` method
- `extend_known_versions()` public method and git URL guard in resolver
- 6 e2e test scripts, 11 unit tests, and all git URL documentation

Retained: `gitutils.py` and `downloads.download_git_source()` which are used by git-clone source providers (`pypi-git`, `github-tag-git`, etc.).

Closes: #1220
Closes: #1225


src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -321,29 +298,6 @@ def _get_matching_versions(
...
-        Used by Bootstrapper to cache git URL resolutions that are
-        handled externally (outside this resolver).
-
-        Args:
-            req: Package requirement (used for name and rule tracking)
-            pre_built: Whether this is a prebuilt or source resolution
-            result: List of (url, version) tuples to add
-        """
-        with self._lock:
-            self._extend_known_versions(req, pre_built, result)
```
dhellmann, line 345: Is the private method still used? It doesn't use the lock to be thread-safe, do we need it to be?

src/fromager/constraints.py
```diff
@@ -61,8 +61,6 @@ def add_constraint(self, unparsed: str) -> None:
         # validator properties: must have a specifier, must not have extras or URL
         if req.extras:
             raise InvalidConstraintError(f"Constraint {unparsed!r} has extras")
-        if req.url:
-            raise InvalidConstraintError(f"Constraint {unparsed!r} has an url")
```
dhellmann, line 65: This would still be an error, right? Do we protect against it somewhere else?

dhellmann (commented):
This is making the code so much simpler, thank you. I had a couple of small questions inline.

src/fromager/constraints.py
rd4398: yes, URL constraints are still rejected  and fall through to the `not req.specifier` check on line 64 since URL requirements never carry a version specifier.

src/fromager/bootstrap_requirement_resolver.py
rd4398: I think it is still used and thread-safety is fine. The lock is acquired in `resolve()` before calling `_resolve_and_extend()`, which calls `_extend_known_versions()`. The private method doesn't acquire the lock itself because the caller already holds it so I guess acquiring it again would result in a deadlock (it's not a reentrant lock).

The only thing that changed is that the public `extend_known_versions()` wrapper (which is removed) was the external entry point that did acquire the lock. Now that the only caller is internal and already holds the lock, the private method  feels the right level of abstraction. Does it make sense?

dhellmann: OK, I wonder if the private method needs to be a method? Is it called from anywhere other than `_resolve_and_extend`?

src/fromager/constraints.py
dhellmann: Sounds good

src/fromager/bootstrap_requirement_resolver.py
rd4398: It's not called from anywhere else. ` _extend_known_versions` has exactly one caller (_resolve_and_extend at line 226). It could be inlined there. I have done it in latest push

dhellmann: That looks good, thanks!

dhellmann:
I'm happy with this version, but I'll leave it open for the rest of the team to review.

src/fromager/constraints.py
smoparth: The req.url validation is more specific over !req.specifier. I would suggest to retain it for better debugging message.

src/fromager/bootstrapper/_bootstrapper.py
```diff
@@ -238,28 +233,6 @@ def resolve_versions(
             List of (url, version) tuples. Contains one item when
             return_all_versions=False, or all matching versions when True.
         """
-        if req.url:
```
smoparth, line 241: Now that the if req.url: handling is removed, if a user still passes mypkg @ git+https://... as a bootstrap input, the URL will be silently ignored and the requirement would flow into the normal resolver which resolves mypkg on PyPI and disregards the URL entirely. No error, no warning.

We should add a guard at the top of resolve_versions() that raises an error or emits a clear warning when req.url is set.

smoparth (changes requested):

src/fromager/constraints.py
rd4398: sure, done in latest push

src/fromager/bootstrapper/_bootstrapper.py
rd4398: Added in latest push

smoparth (approved):
LGTM!

tiran:
LGTM!

I like the fact that the deprecation simplifies one of the most complex parts of Fromager.

tiran (approved):

PR #1261, opened 2026-07-22
cleanup(bootstrapper): expose `DEFAULT_BG_THREADS` as public class constant

### LINKED ISSUE #1246: cleanup(bootstrapper): clarify public/private status of _DEFAULT_BG_THREADS
`_DEFAULT_BG_THREADS` is exported in `src/fromager/bootstrapper/__init__.py` via `__all__` but uses a leading underscore, sending mixed signals about whether it's public API. It's used by `src/fromager/commands/bootstrap.py` as a CLI default value.

Pre-existing naming from the original monolithic `bootstrapper.py`, surfaced during review of #1226.

**Option A:** Remove from `__all__` and keep the underscore. The `commands/` package is internal to fromager, so accessing it as a private name is fine.

**Option B:** Rename to `DEFAULT_BG_THREADS` (drop underscore) and keep in `__all__`, making it an explicit public API.

rd4398:
Make `DEFAULT_BG_THREADS` a public class constant on `Bootstrapper`
instead of a private module-level `_DEFAULT_BG_THREADS` in `_types.py`.
The CLI layer now references `Bootstrapper.DEFAULT_BG_THREADS` for the
`--bg-threads` default, keeping the thread-count knowledge with the
bootstrapper while giving it a clean public API surface.

Closes: https://github.com/python-wheel-build/fromager/issues/1246

### DIFF
--- a/src/fromager/bootstrapper/_bootstrapper.py
+++ b/src/fromager/bootstrapper/_bootstrapper.py
@@ -19,6 +19,7 @@ from .. import (
     bootstrap_requirement_resolver,
     progress,
     sources,
+    threading_utils,
 )
 from ..dependency_graph import DependencyGraph
 from ..log import req_ctxvar_context, requirement_ctxvar
@@ -31,7 +32,6 @@ from ._prepare_source import PrepareSource
 from ._process_install_deps import ProcessInstallDeps
 from ._resolve import Resolve
 from ._types import (
-    _DEFAULT_BG_THREADS,
     FailureRecord,
     FailureType,
     SeenKey,
@@ -52,6 +52,8 @@ class Bootstrapper:
     the dependency tree, then ``finalize()`` for cleanup and exit code.
     """
 
+    DEFAULT_BG_THREADS: int = max(1, threading_utils.get_cpu_count() // 2)
+
     def __init__(
         self,
         ctx: context.WorkContext,
@@ -61,7 +63,7 @@ class Bootstrapper:
         sdist_only: bool = False,
         test_mode: bool = False,
         multiple_versions: bool = False,
-        num_bg_threads: int = _DEFAULT_BG_THREADS,
+        num_bg_threads: int = DEFAULT_BG_THREADS,
     ) -> None:
         if test_mode and sdist_only:
             raise ValueError(
--- a/src/fromager/commands/bootstrap.py
+++ b/src/fromager/commands/bootstrap.py
@@ -20,6 +20,7 @@ from .. import (
     resolver,
     server,
 )
+from ..bootstrapper import Bootstrapper
 from .build import build_parallel
 from .graph import find_why, show_explain_duplicates
 
@@ -120,7 +121,7 @@ def _get_requirements_from_args(
     "--bg-threads",
     "num_bg_threads",
     type=click.IntRange(min=1),
-    default=bootstrapper._DEFAULT_BG_THREADS,
+    default=Bootstrapper.DEFAULT_BG_THREADS,
     show_default=True,
     help="Number of background threads for parallel I/O pre-fetching (min 1).",
 )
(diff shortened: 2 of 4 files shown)


src/fromager/commands/bootstrap.py
```diff
@@ -20,6 +20,7 @@
     resolver,
     server,
 )
+from ..bootstrapper._types import _DEFAULT_BG_THREADS
```
smoparth: Bootstrap command is still depending on an internal symbol it shouldn't know about (as per single underscore convention)

The underlying issue is that _DEFAULT_BG_THREADS is a user-facing default (what's a reasonable thread count for the bootstrap), not a bootstrapper internal mechanism. 

If we agree its the responsibility of command layer we can move the constant to commands/bootstrap.py
and make Bootstrapper accept num_bg_threads with no default. The CLI owns the logic it displays to users and passes it to bootstrapper, the bootstrap just uses what it's given.

or

Make it genuinely public. If there's a reason for the bootstrapper package to own this, drop the underscore and expose DEFAULT_BG_THREADS as a class constant on Bootstrapper so the CLI references bootstrapper.Bootstrapper.DEFAULT_BG_THREADS.

rd4398: Yeah this is tricky and I am inclined to make it public. I have amended the commit now to make it public.

LalatenduMohanty: @rd4398  the PR description does not reflect the direction this PR is taking. The commit message is correct though.

LalatenduMohanty (dismissed):
The PR looks good to me. Once you fix the PR description we should be able to merge it.

src/fromager/bootstrapper/_bootstrapper.py
```diff
@@ -61,7 +63,7 @@ def __init__(
         sdist_only: bool = False,
         test_mode: bool = False,
         multiple_versions: bool = False,
-        num_bg_threads: int = _DEFAULT_BG_THREADS,
+        num_bg_threads: int | None = None,
```
smoparth: int | None and the or fallback aren't needed. Click always passes a concrete int when the user omits --bg-threads. The or in max also leads to misunderstanding: a code using Bootstrap as API may pass num_bg_threads=0 and it would silently get the default instead of an error or min possible value to keep the ThreadPool working.

Simpler to express this in the signature itself:
num_bg_threads: int = DEFAULT_BG_THREADS,

and make self._num_bg_threads = max(1, num_bg_threads), the max(1, ...) stays as defensive validation since ThreadPoolExecutor needs at least 1 thread.

smoparth (commented):
Making DEFAULT_BG_THREADS as public is good choice. I just mentioned small nitpick above.

src/fromager/bootstrapper/_bootstrapper.py
```diff
@@ -76,7 +78,7 @@ def __init__(
         self.test_mode = test_mode
         self.multiple_versions = multiple_versions
         self.why: list[tuple[RequirementType, Requirement, Version]] = []
-        self._num_bg_threads = max(1, num_bg_threads)
+        self._num_bg_threads = max(1, num_bg_threads or self.DEFAULT_BG_THREADS)
```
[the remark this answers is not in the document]
smoparth: The end user cannot pass 0 as the click argument for this is restricted to min=1. Agree it might be possible if anyone calls it as API itself

src/fromager/bootstrapper/_bootstrapper.py
rd4398: Good catch, pushed the fix in latest commit

src/fromager/commands/bootstrap.py
rd4398: done, changed the description

LalatenduMohanty (approved):

PR #1262, opened 2026-07-23
Deduplicate embedded cause suffix in _format_exception

### LINKED ISSUE #1243: _format_exception() duplicates message when cause text is already embedded
### Bug description

**Expected behavior:**
Error messages should not repeat the same information.

**Actual behavior:**
`resolver.py:286-288` wraps exceptions with `raise ResolverException(f"...{original_msg}") from err` — embedding the cause's text in the outer message AND chaining it as `__cause__`. When `__main__._format_exception()` formats this, it appends `" because {cause}"`, producing a duplicated message:

```
Unable to resolve requirement specifier flashinfer-python==0.6.8.post1
  with constraint flashinfer-python==0.6.11.post2: found no match for
  flashinfer-python==0.6.8.post1 because found no match for
  flashinfer-python==0.6.8.post1
```

The `" because found no match..."` portion is redundant — the same text already appears after the colon.

**Steps to reproduce:**
Trigger any resolution failure where `find_all_matching_from_provider()` wraps a `ResolverException` with `from err` (the common case for constraint conflicts or missing packages).

### Proposed fix

Add a dedup check in `_format_exception()`: when the outer exception's message already starts with the cause's raw message (`str(exc.__cause__)`), skip the `" because ..."` clause. A `startswith` check matches the actual construction pattern (`f"...{original_msg}"`) without false-positiving on coincidental substrings.

This was originally part of #1241 but is being split out as an independent fix.

raravind007:
## Summary

Fixes #1243.

`_format_exception()` always appends `" because {cause}"` for any
chained exception. `resolver.py:285-288` wraps resolution failures as
`f"Unable to resolve requirement specifier ... using {provider}: {original_msg}"`
where `original_msg` is `str(err)`, then chains `from err` - so the
cause's own message is already embedded at the end of the outer
message before `_format_exception()` ever sees it, producing a
duplicated, hard-to-read error:

    Unable to resolve requirement specifier flashinfer-python==0.6.8.post1
    with constraint flashinfer-python==0.6.11.post2 using PyPI resolver:
    found no match for flashinfer-python==0.6.8.post1 because found no
    match for flashinfer-python==0.6.8.post1

While reproducing the issue against current `resolver.py`, I found that
the cause text is appended as `f"...: {original_msg}"`, so the concrete
relationship is a suffix rather than a prefix. This uses
`endswith(f": {cause_str}")` to match that exact construction and avoid
coincidental suffix matches - a bare `endswith(cause_str)` would
incorrectly dedupe something like `RuntimeError("failed to match")`
chained from `ValueError("match")`.

A naive suffix-dedupe would also silently truncate a deeper cause
chain: if the immediate cause itself has its own cause, deduping
against its raw message alone and returning early loses that deeper
"because ..." information. The fix formats the cause recursively first
and preserves anything beyond the raw immediate-cause text, so a 3+
level chain isn't truncated.

## Test plan

- [x] `pytest tests/test_external_commands.py` — covers the resolver-style
      duplication case, the unrelated-cause case (unaffected), the
      nested-cause preservation case, and the coincidental-suffix guard
      case, all with exact-string assertions.
- [x] `ruff check` / `ruff format --check`
- [x] `mypy -p fromager` — no new errors

### DIFF
--- a/src/fromager/__main__.py
+++ b/src/fromager/__main__.py
@@ -299,8 +299,26 @@ for cmd in commands.commands:
 
 def _format_exception(exc: BaseException) -> str:
     if exc.__cause__:
-        cause = _format_exception(exc.__cause__)
-        return f"{exc} because {cause}"
+        exc_str = str(exc)
+        cause_str = str(exc.__cause__)
+        formatted_cause = _format_exception(exc.__cause__)
+        # resolver.py wraps failures as f"...: {original_msg}" where
+        # original_msg is str(err) before chaining `from err`. Anchor on
+        # the ": " separator that construction actually uses, rather than
+        # a bare endswith(cause_str), so a short cause message can't
+        # coincidentally match the tail of an unrelated outer message
+        # (e.g. RuntimeError("failed to match") from ValueError("match")
+        # must NOT be deduped - it doesn't embed the cause deliberately).
+        # Preserve any additional cause-chain information the recursive
+        # formatting adds beyond the raw immediate cause - otherwise a
+        # deeper cause (the immediate cause's own cause) would be
+        # silently dropped.
+        if cause_str and exc_str.endswith(f": {cause_str}"):
+            if formatted_cause == cause_str:
+                return exc_str
+            if formatted_cause.startswith(f"{cause_str} because "):
+                return exc_str + formatted_cause[len(cause_str) :]
+        return f"{exc_str} because {formatted_cause}"
     return str(exc)
 
 
--- a/tests/test_external_commands.py
+++ b/tests/test_external_commands.py
@@ -176,3 +176,93 @@ def test_format_exception_formats_chained_exceptions() -> None:
         assert "Higher level error" in formatted
         assert "because" in formatted
         assert "Root cause" in formatted
+
+
+def test_format_exception_deduplicates_embedded_cause_suffix() -> None:
+    """Regression test for #1243.
+
+    resolver.py wraps failures as ``f"...: {original_msg}"`` where
+    ``original_msg`` is ``str(err)`` before chaining ``from err`` - the
+    cause's raw message ends up embedded at the *end* of the outer
+    message. Without deduping, ``_format_exception`` prints it twice via
+    "... because ...".
+    """
+    from resolvelib.resolvers import ResolverException
+
+    from fromager import __main__
+
+    try:
+        try:
+            raise ResolverException("found no match")
+        except ResolverException as err:
+            original_msg = str(err)
+            raise ResolverException(
+                f"Unable to resolve requirement: {original_msg}"
+            ) from err
+    except ResolverException as err:
+        formatted = __main__._format_exception(err)
+
+    assert formatted == "Unable to resolve requirement: found no match"
+
+
+def test_format_exception_still_shows_because_for_unrelated_cause() -> None:
+    """Unrelated causes (not embedded in the outer message) are still
+
+    joined with "because", so the dedupe check doesn't over-suppress.
+    """
+    from fromager import __main__
+
+    try:
+        try:
+            raise ValueError("Root cause")
+        except ValueError as e:
+            raise RuntimeError("Higher level error") from e
+    except RuntimeError as chained:
+        assert (
+            __main__._format_exception(chained)
+            == "Higher level error because Root cause"
+        )
+
+
+def test_format_exception_preserves_nested_cause_after_dedup() -> None:
+    """Deduping the immediate embedded cause must not drop a deeper cause.
+
+    If the outer message ends with the immediate cause's raw text, that
+    part is deduped - but the immediate cause may itself have its own
+    cause with information not present anywhere in the outer message,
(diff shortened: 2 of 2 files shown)


dhellmann:
What does the error output look like if we stop trying to do the formatting and appending of the "why" chain ourselves and we just let python raise the exception normally and then show the whole traceback? And what would it look like if we use `raise from` to build an exception chain using simple messages?

raravind007:
Good question — I tried the concrete alternatives against the exception shapes we have today.

**A. Current main after #1241**

For a transitive resolution failure, `_enrich_resolution_error()` creates a new exception whose message includes the dependency path, and `_handle_phase_error()` raises it without an explicit `from` clause. Python keeps the original exception in `__context__`, so the full traceback is still available to `exc_info=True`; it just is not exposed to our single-line `_format_exception()`, which only follows `__cause__`.

The user-facing line is roughly:

```
Unable to resolve requirement specifier flashinfer-python==0.6.8.post1
with constraint flashinfer-python==0.6.11.post2 using PyPI resolver:
found no match for flashinfer-python==0.6.8.post1
(dependency chain: its-hub==1.0 -> reward-hub==2.0 ->
vllm==3.0 -> flashinfer-python==0.6.8.post1)
```

**B. Simple messages + normal `raise ... from err` + native traceback**

If each exception adds only new information instead of embedding `str(err)`, the native chain is much cleaner:

```
ResolverException: found no match for flashinfer-python==0.6.8.post1

The above exception was the direct cause of the following exception:

ResolverException: could not resolve flashinfer-python==0.6.8.post1
(constraint flashinfer-python==0.6.11.post2)

The above exception was the direct cause of the following exception:

ResolverException: could not resolve flashinfer-python==0.6.8.post1,
required by vllm==3.0 <- reward-hub==2.0 <- its-hub==1.0
```

That gives full file/line information at each level, but it is much more verbose for normal CLI output.

**C. Simple messages + `raise ... from err` + the existing compact formatter**

This is the interesting result:

```
could not resolve flashinfer-python==0.6.8.post1, required by
vllm==3.0 <- reward-hub==2.0 <- its-hub==1.0 because
could not resolve flashinfer-python==0.6.8.post1
(constraint flashinfer-python==0.6.11.post2) because
found no match for flashinfer-python==0.6.8.post1
```

In that shape, the original unconditional `_format_exception()` works without any deduplication.

That suggests the underlying duplication in #1243 is not really caused by `_format_exception()` itself. `resolver.py` currently does both of these:

```python
original_msg = str(err)
raise ResolverException(f"...: {original_msg}") from err
```

so the same information exists both in the outer message and in `__cause__`.

A cleaner fix may therefore be to stop embedding `str(err)` in wrapper messages and let exception chaining carry the cause. We would still need to decide how we want to present the dependency `why_snapshot` — as one compact outer message, as #1241 does today, or as additional exception-chain levels.

Given that, I'm inclined to pause this `_format_exception()` PR rather than add dedup logic at the presentation layer. If you prefer, I can prototype the simpler-message / `raise from` approach and compare the final CLI output before changing the PR.

dhellmann:
Yes, I would like to see what the code changes and output look like if we stop trying to be clever and rely on more standard python implementation. _Maybe_ we want something in the command line tool that takes a chain of exceptions and formats those as a single error message _as well as_ showing the full set of tracebacks.

raravind007:
Prototyped this against the actual exception shapes in the codebase.

The duplication originates in the exception-construction pattern rather than requiring dedup logic in `_format_exception()`. The resolver wrapper currently both embeds `str(err)` into the new exception message and chains the same exception with `from err`. That means the cause text already exists twice before `_format_exception()` sees it. With the current, unmodified formatter, that reproduces #1243. If I remove the embedded `original_msg` but keep `from err`, the same formatter produces clean output with no deduplication logic.

#1241 has a related workaround one level up. `_enrich_resolution_error()` creates a replacement exception containing `{err}` plus the dependency chain, and `_handle_phase_error()` raises that replacement without an explicit `from` clause. That deliberately leaves `__cause__` unset, while the original exception remains available through `__context__`.

So I think the cleaner prototype is:

- Simplify the resolver wrapper so its message contains only the context it adds, and keep `raise ... from err`.
- Change the resolution-enrichment path to use normal explicit exception chaining as well, while preserving the `why_snapshot` information — the Python traceback alone cannot represent the package dependency graph.
- Compare two presentations of that same standard exception chain: the full native traceback and the compact CLI message produced by the existing simple formatter.

One useful detail is that `invoke_main()` already has those two presentation paths conceptually: `logger.debug(..., exc_info=True)` records the traceback and `logger.error(_format_exception(err))` emits the compact message. So the main experiment is whether standardizing the exception construction lets both work cleanly without special deduplication.

I'll prototype that code and post the actual diff plus both outputs before replacing the implementation in this PR.

One output question I'll include in the comparison: whether the existing `(dependency chain: ...)` representation should remain as-is or be expressed more naturally as `required by ...`. I'll keep that as presentation only; the dependency path itself still needs to be represented because it is not available from the Python traceback.

PR #1264, opened 2026-07-23
feat: auto-cap setuptools when setup.py uses removed APIs

### LINKED ISSUE #1263: Add global hook point for get_build_system_dependencies
## Problem

setuptools 81 removed `distutils.spawn(dry_run=...)` and `distutils.dir_util.remove_tree(dry_run=...)`. setuptools 82 removed `pkg_resources` entirely. Many packages on PyPI still use these in their `setup.py`, causing build failures when Fromager resolves an uncapped `setuptools` as a build dependency.

In the AIPCC wheels builder (fondue), **22 identical per-package plugins** exist solely to add a setuptools version cap via `get_build_system_dependencies`. Each plugin is a copy-paste of the same logic. Additionally, **~127 packages** with settings overrides have no protection at all.

## Proposed Solution

Instead of putting setuptools-specific detection logic into Fromager core, extend `fromager.hooks` to support `get_build_system_dependencies` as a global hook point. This lets downstream projects register hooks that post-process build dependencies for all packages.

### Changes in Fromager (~20 lines across two files)

**`hooks.py`:**
- Add `get_build_system_dependencies` to `GLOBAL_HOOK_NAMES`
- Add `run_get_build_system_dependencies_hooks()` following the same pattern as `run_post_build_hooks()`
- The hook receives the resolved requirements list and returns a modified list

**`dependencies.py`:**
- Inside `get_build_system_dependencies()`, after `overrides.find_and_invoke()` returns the dependency list (from a per-package plugin or the default), pass it through the global hook before `_filter_requirements()`

### How downstream projects use it

Downstream projects register hooks via `fromager.hooks` entry points, same as `post_build` today:

```toml
[project.entry-points."fromager.hooks"]
get_build_system_dependencies = "my_plugins.hooks.setuptools_cap:get_build_system_dependencies"
```

The hook signature:

```python
def get_build_system_dependencies(
    ctx: context.WorkContext,
    req: Requirement,
    sdist_root_dir: pathlib.Path,
    build_dir: pathlib.Path,
    requirements: list[str],
) -> list[str]:
    # post-process and return modified requirements
    ...
```

### Why this approach

- **Fromager stays generic** -- no setuptools version numbers or AST scanning in core
- **Opt-in** -- users register hooks or don't. No forced behavior
- **Extensible** -- downstream projects can add any build dependency fixups (setuptools caps, cmake constraints, custom build deps) without Fromager core changes
- **Follows existing patterns** -- same mechanism as `post_build` / `post_bootstrap` / `prebuilt_wheel` hooks

## Impact

In the fondue builder:
- **22 plugins eliminated** -- one global hook replaces all of them
- **3 plugins simplified** (llvmlite, torchvision, pyarrow) -- can remove setuptools constraint handling from their `get_build_system_dependencies` override
- **~127 packages gain protection** -- currently have no setuptools cap

## Previous approach

PR #1264 (now closed) put the setuptools detection logic directly in `default_get_build_system_dependencies`. Per feedback from @tiran, this was too specific to put in core. The global hooks approach keeps Fromager generic while giving downstream projects the extension point they need.

## Relationship to existing issues

Related to #928 (Reduce downstream plugin needs through YAML-driven configuration enhancements).

vshawrh:
## Summary

- Auto-detect `pkg_resources` imports and `dry_run` keyword arguments in `setup.py` via AST parsing
- Append `setuptools<82` or `setuptools<81` to build-system requirements when removed APIs are detected
- No cap added for packages without `setup.py` or clean `setup.py` files

Closes #1263

## Motivation

setuptools 81 removed `distutils.spawn(dry_run=...)` and `distutils.dir_util.remove_tree(dry_run=...)`. setuptools 82 removed `pkg_resources` entirely. Many packages on PyPI still use these in their `setup.py`, causing build failures when Fromager resolves an uncapped `setuptools`.

Currently every Fromager deployment must handle this per-package, either via a plugin overriding `get_build_system_dependencies` or via `project_override.update_build_requires`. In the AIPCC wheels builder, **22 identical plugins** exist solely to add this constraint, and **127 additional packages** have no protection at all.

## Implementation

The change adds a `_get_setuptools_constraint()` function to `dependencies.py` that:

1. Checks if `setup.py` exists in `sdist_root_dir`
2. Parses it with `ast.parse` (avoids false positives from comments/strings)
3. Walks the AST looking for `pkg_resources` imports and `dry_run=` keyword args
4. Returns the tighter constraint when both apply

`default_get_build_system_dependencies` calls this function and appends the constraint when needed.

## Regression analysis

| Scenario | Impact |
|----------|--------|
| Packages with no plugin | Gain automatic protection (net positive) |
| Packages with tighter `update_build_requires` (e.g., `setuptools<80`) | pip resolves to more restrictive cap. No conflict |
| Plugins that call `default_get_build_system_dependencies` internally | Inherit the cap automatically. Harmless duplicate if they also add their own |
| Plugins that fully replace the hook | Unaffected (default never called) |
| Packages without `setup.py` | No cap added. No change |
| `build-system-requirements.txt` cache | Generated fresh each build. No stale cache risk |

## Test plan

- [x] `test_pkg_resources_import` -- detects `import pkg_resources`
- [x] `test_pkg_resources_from_import` -- detects `from pkg_resources import ...`
- [x] `test_pkg_resources_submodule` -- detects `import pkg_resources.extern`
- [x] `test_dry_run_keyword` -- detects `dry_run=` keyword arg
- [x] `test_both_returns_tighter` -- returns `setuptools<81` when both found
- [x] `test_clean_setup_py` -- returns `None` for clean setup.py
- [x] `test_no_setup_py` -- returns `None` when no setup.py exists
- [x] `test_syntax_error` -- returns `None` gracefully on syntax errors
- [x] `test_pkg_resources_in_string_not_detected` -- no false positive from string literals
- [x] `test_default_build_system_deps_adds_setuptools_constraint` -- integration test through full `get_build_system_dependencies`

### DIFF
--- a/src/fromager/dependencies.py
+++ b/src/fromager/dependencies.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import ast
 import copy
 import logging
 import os
@@ -102,9 +103,74 @@ def default_get_build_system_dependencies(
     """Get build system requirements
 
     Defaults to ``[build-system] requires`` from ``pyproject.toml``.
+
+    When ``setup.py`` uses APIs removed in newer setuptools versions,
+    a version cap is appended automatically:
+
+    - ``setuptools<81`` when ``setup.py`` passes ``dry_run=`` keyword
+      arguments (removed in setuptools 81)
+    - ``setuptools<82`` when ``setup.py`` imports ``pkg_resources``
+      (removed in setuptools 82)
     """
     pyproject_toml = get_pyproject_contents(build_dir)
-    return typing.cast(list[str], get_build_backend(pyproject_toml)["requires"])
+    requires = list(
+        typing.cast(list[str], get_build_backend(pyproject_toml)["requires"])
+    )
+    constraint = _get_setuptools_constraint(sdist_root_dir)
+    if constraint:
+        logger.info(
+            "%s: auto-adding %s (setup.py uses removed APIs)", req.name, constraint
+        )
+        requires.append(constraint)
+    return requires
+
+
+def _get_setuptools_constraint(sdist_root_dir: pathlib.Path) -> str | None:
+    """Return a setuptools version cap if setup.py uses removed APIs.
+
+    - setuptools 81 removed ``distutils.spawn(dry_run=...)`` and
+      ``distutils.dir_util.remove_tree(dry_run=...)``
+    - setuptools 82 removed ``pkg_resources`` entirely
+
+    Parses the AST to avoid false positives from string matches in
+    comments or string literals.
+
+    Returns ``"setuptools<81"``, ``"setuptools<82"``, or ``None``.
+    The tighter constraint wins when both apply.
+    """
+    setup_py = sdist_root_dir / "setup.py"
+    if not setup_py.is_file():
+        return None
+    try:
+        source = setup_py.read_text(encoding="utf-8", errors="replace")
+        tree = ast.parse(source, filename=str(setup_py))
+    except (OSError, SyntaxError):
+        return None
+
+    findings: set[str] = set()
+    for node in ast.walk(tree):
+        if isinstance(node, ast.Import):
+            for alias in node.names:
+                if alias.name == "pkg_resources" or alias.name.startswith(
+                    "pkg_resources."
+                ):
+                    findings.add("pkg_resources")
+        elif isinstance(node, ast.ImportFrom):
+            if node.module is not None and (
+                node.module == "pkg_resources"
+                or node.module.startswith("pkg_resources.")
+            ):
+                findings.add("pkg_resources")
+        elif isinstance(node, ast.Call):
+            for kw in node.keywords:
+                if kw.arg == "dry_run":
+                    findings.add("dry_run")
+
+    if "dry_run" in findings:
+        return "setuptools<81"
+    if "pkg_resources" in findings:
+        return "setuptools<82"
+    return None
 
 
 def get_build_backend_dependencies(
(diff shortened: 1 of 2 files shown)


src/fromager/dependencies.py
```diff
@@ -102,9 +103,74 @@ def default_get_build_system_dependencies(
...
-    return typing.cast(list[str], get_build_backend(pyproject_toml)["requires"])
+    requires = list(
+        typing.cast(list[str], get_build_backend(pyproject_toml)["requires"])
+    )
+    constraint = _get_setuptools_constraint(sdist_root_dir)
+    if constraint:
+        logger.info(
+            "%s: auto-adding %s (setup.py uses removed APIs)", req.name, constraint
+        )
+        requires.append(constraint)
```
dhellmann, line 124: What if there is already a setuptools requirement in the list? Shouldn't we modify that requirement instead of just adding another one?

dhellmann (commented):
This is useful, but it feels somewhat specific to the set of packages we happen to be building right now. What sorts of options did you explore for making this a plugin or otherwise making it a more configurable behavior so that a user can choose to enable it or not?

src/fromager/dependencies.py
tiran: Yes, the code has to carefully merge constraints. If the upstream project or our downstream project overrides set a lower ceiling, then the new code must not raise the ceiling.

src/fromager/dependencies.py
```diff
@@ -102,9 +103,74 @@ def default_get_build_system_dependencies(
...
+    return requires
+
+
+def _get_setuptools_constraint(sdist_root_dir: pathlib.Path) -> str | None:
```
tiran, line 128: The function should look into `build_dir`, not `sdist_root_dir`.

tiran (changes requested):

vshawrh:
Closing in favor of a better approach: adding `get_build_system_dependencies` as a global hook point instead of putting setuptools-specific logic in core. New PR incoming.

PR #1265, opened 2026-07-24
refactor(build_environment): add sdist_root_dir and req to BuildEnvironment

tiran:
# Pull Request Description

## What

Replace the generic `parent_dir` parameter with `sdist_root_dir` and `req` so that `BuildEnvironment` has the context it needs for future sandboxing and environment filtering. All arguments are now keyword-only and follow the common `ctx, req, sdist_root_dir` order.

## Why

Part of the sandboxing proposal.

See: #1019

### DIFF
--- a/src/fromager/build_environment.py
+++ b/src/fromager/build_environment.py
@@ -78,12 +78,35 @@ class BuildEnvironment:
 
     def __init__(
         self,
+        *,
         ctx: context.WorkContext,
-        parent_dir: pathlib.Path,
+        req: Requirement,
+        sdist_root_dir: pathlib.Path,
     ):
+        """Create a build environment for an sdist.
+
+        Args:
+            ctx: the work context
+            req: the requirement being built
+            sdist_root_dir: root directory of the unpacked sdist
+        """
         self._ctx = ctx
-        self.path = parent_dir.absolute() / f"build-{platform.python_version()}"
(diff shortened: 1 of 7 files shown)


src/fromager/build_environment.py
```diff
@@ -78,13 +78,37 @@ class BuildEnvironment:
...
+            req: the requirement being built
+            sdist_root_dir: root directory of the unpacked sdist
+        """
         self._ctx = ctx
-        self.path = parent_dir.absolute() / f"build-{platform.python_version()}"
+        self._req = req
+        self._sdist_root_dir = sdist_root_dir
+        # abiflags adds 't' for free-threaded (nogil) builds
+        self.path = sdist_root_dir.parent.absolute().joinpath(
+            f"build-{platform.python_version()}{sys.abiflags}"
```
rd4398: Nit: Can we mention this `sys.abiflags` change in commit description or do a separate commit?

rd4398 (commented):
This looks great! I have left a non blocking comment

src/fromager/build_environment.py
tiran: I can make this a separate PR

rd4398 (approved):
This looks good to me! I am moving forward and approving this
cc @LalatenduMohanty @smoparth

PR #1266, opened 2026-07-24
feat(packagesettings): add `ExternalCommands` model for env filtering

tiran:
# Pull Request Description

## What

Add `ExternalCommands` Pydantic model with `keep_env` / `delete_env` pattern lists, a `filter_env()` method, and a `DEFAULT_KEEP_ENV` class variable for essential variables (HOME, PATH, LC_*, etc.).

Not yet wired into `external_commands.run()`.

## Why

See: #1083

### DIFF
--- a/src/fromager/packagesettings/_models.py
+++ b/src/fromager/packagesettings/_models.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import logging
 import os
 import pathlib
+import re
 import typing
 from collections.abc import Mapping
 
@@ -12,7 +13,7 @@ import pydantic
 import yaml
 from packaging.requirements import Requirement
 from packaging.utils import canonicalize_name
-from pydantic import AnyUrl, Field
+from pydantic import AnyUrl, Field, PrivateAttr, StringConstraints
 from pydantic_core import core_schema
 
 # from ._resolver import SourceResolver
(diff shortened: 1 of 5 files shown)


src/fromager/packagesettings/_models.py
```diff
@@ -72,6 +73,136 @@ class SbomSettings(pydantic.BaseModel):
...
+    parts: list[str] = []
+    for p in patterns:
+        if p.endswith("*"):
+            parts.append(re.escape(p[:-1]) + ".*")
+        else:
+            parts.append(re.escape(p))
+    return re.compile("|".join(parts))
+
+
+class ExternalCommands(pydantic.BaseModel):
```
rd4398, line 120: This is missing ` .. versionadded::` directives. Can we add that and set it to next release v0.92.0

src/fromager/packagesettings/_models.py
```diff
@@ -72,6 +73,136 @@ class SbomSettings(pydantic.BaseModel):
...
+                f"delete_env overlaps with keep_env / DEFAULT_KEEP_ENV: "
+                f"{sorted(overlap)}"
+            )
+        return self
+
+    def __bool__(self) -> bool:
+        """True when filtering is configured (``delete_env`` is non-empty)."""
+        return bool(self.delete_env)
+
+    def model_post_init(self, __context: typing.Any) -> None:
```
rd4398, line 194: Nit: can we add a docstring here like `Compile keep/delete regexes when ``delete_env`` is configured.`

rd4398 (commented):
This looks good! I have left couple of suggestions that are non blocking and can be done in a follow up

src/fromager/packagesettings/_models.py
```diff
@@ -72,6 +73,139 @@ class SbomSettings(pydantic.BaseModel):
...
+        if not self.delete_env:
+            return self
+        if "*" in self.delete_env and len(self.delete_env) > 1:
+            raise ValueError(
+                "delete_env: bare '*' must be the only entry, "
+                "additional patterns are redundant"
+            )
+        keep = set(self.DEFAULT_KEEP_ENV) | set(self.keep_env)
+        overlap = keep & set(self.delete_env)
+        if overlap:
```
smoparth, line 187: https://fromager.readthedocs.io/en/latest/proposals/filter-env.html#evaluation-order

The priority model on proposal says "If a variable matches the hard-coded always-keep set or an entry in keep_env, then the variable is kept". We should update the proposal. I think this strict approach is better for a security feature (fail loud)

src/fromager/packagesettings/_models.py
```diff
@@ -72,6 +73,139 @@ class SbomSettings(pydantic.BaseModel):
...
+
+    *patterns* must be non-empty.
+    """
+    parts: list[str] = []
+    for p in patterns:
+        if p.endswith("*"):
+            parts.append(re.escape(p[:-1]) + ".*")
+        else:
+            parts.append(re.escape(p))
+    return re.compile("|".join(parts))
```
smoparth: https://fromager.readthedocs.io/en/latest/proposals/filter-env.html#evaluation-order
The propsal says "All checks are case-insensitive and short-circuit". We should that proposal here as well. I think it should be case-sensitive as Env vars in Linux and macOS are case-sensitive but not in WIN and we don't support WIN.

src/fromager/packagesettings/_models.py
```diff
@@ -72,6 +73,139 @@ class SbomSettings(pydantic.BaseModel):
...
+# optionally ending with ``*`` (trailing wildcard).
+# DeleteEnvPattern additionally allows bare ``*`` (catch-all).
+KeepEnvPattern = typing.Annotated[
+    str,
+    StringConstraints(pattern=r"^[a-zA-Z_][a-zA-Z0-9_]*\*?$"),
+]
+
+DeleteEnvPattern = typing.Annotated[
+    str,
+    StringConstraints(pattern=r"^(\*|[a-zA-Z_][a-zA-Z0-9_]*\*?)$"),
```
smoparth, line 87: This enforces the pattern to look like a valid POSIX environment variable name but the proposal doesn't say that. I'm inclined with having this but worried if any package has an unusual env var needed or need filtering that doesn't follow POSIX standard. eg. BOT-PAT.

src/fromager/packagesettings/_models.py
```diff
@@ -72,6 +73,139 @@ class SbomSettings(pydantic.BaseModel):
...
+        delete_env:
+          - "CI_TOKEN"
+          - "AWS_*"
+
+    .. versionadded:: 0.92.0
+    """
+
+    model_config = MODEL_CONFIG
+
+    DEFAULT_KEEP_ENV: typing.ClassVar[tuple[str, ...]] = (
```
smoparth, line 142: Should `TERM` be in DEFAULT_KEEP_ENV?

tiran:
@mergifyio rebase

src/fromager/packagesettings/_models.py
```diff
@@ -72,6 +73,139 @@ class SbomSettings(pydantic.BaseModel):
...
+    def model_post_init(self, __context: typing.Any) -> None:
+        """Pydantic post init hook to initialize internal data structures"""
+        if self.delete_env:
+            self._keep_re = _compile_env_patterns(
+                self.DEFAULT_KEEP_ENV + tuple(self.keep_env)
+            )
+            if "*" not in self.delete_env:
+                self._delete_re = _compile_env_patterns(tuple(self.delete_env))
+
+    def filter_env(self, env: Mapping[str, str]) -> Mapping[str, str]:
```
smoparth, line 205: I think this is the 1st time I seen business logic tied in the data model here.
Just want to confirm: existing models here are pure data holders with validators and business logic on separate classes.

smoparth: I was initially thinking filter_env() and DEFAULT_KEEP_ENV would live in separate file or external_commands.py since the models here have always been data + validation only. But having logic along with data would simply things and maybe we can apply this to few of existing models too.

rd4398 (dismissed):
This looks good to me! I am moving forward and approving this.
cc @LalatenduMohanty @smoparth

tiran:
@mergifyio rebase

tiran:
@smoparth Thanks for the feedback! I have deviated slightly from my original proposal to address some design flaws. I like to protect users from misconfiguration. For example I included the POSIX compliance check for env var names to prevent typos like `delete_env: [BOT-PAT]`.

Here is a full diff between the initial commit and the current implementation (generated by Claude)

### Proxy variables removed from `DEFAULT_KEEP_ENV`

`http_proxy`, `https_proxy`, and `no_proxy` were removed from the
always-keep set. These variables can contain basic-auth credentials
(e.g. `http_proxy=http://user:password@proxy.example:3128/`) and
should not be passed to untrusted build scripts by default. Add them
to `keep_env` explicitly if your environment requires proxy access
during builds.

### Additional variables added to `DEFAULT_KEEP_ENV`

- `TERM` -- controls terminal capability detection; without it
  programs using ncurses/readline can produce garbled output.
- `TZ` -- timezone; affects timestamps in logs and build metadata.
- `TEMP`, `TMP`, `TMPDIR` -- honoured by Python's `tempfile` module
  and many build tools for temporary file placement.

### Non-POSIX environment variable keys are always removed

`filter_env()` unconditionally strips keys that are not valid POSIX
names (`[A-Za-z_][A-Za-z0-9_]*`). This removes keys with dashes,
dots, embedded spaces, or bash-exported function definitions
(`BASH_FUNC_*%%`).

In practice a user has to go out of their way to create such variables
-- common shells refuse them:

```console
$ export invalid-example=value
bash: export: `invalid-example=value': not a valid identifier
```

They can only appear when a program manipulates the environment
directly (e.g. via `os.environ` in Python or `setenv()` in C). No
legitimate build tool relies on non-POSIX keys.

### `delete_env` matching is case-insensitive

`delete_env` patterns are matched case-insensitively for convenience
and security. A pattern like `aws_*` removes `AWS_SECRET_ACCESS_KEY`
regardless of casing, so credentials cannot slip through due to
unexpected capitalisation.

`keep_env` and the always-keep set remain case-sensitive, matching the
exact variable names used in practice.

src/fromager/packagesettings/_models.py
```diff
@@ -72,6 +73,165 @@ class SbomSettings(pydantic.BaseModel):
...
+    for p in patterns:
+        if p.endswith("*"):
+            parts.append(re.escape(p[:-1]) + ".*")
+        else:
+            parts.append(re.escape(p))
+    flags = re.IGNORECASE if case_insensitive else 0
+    return re.compile("|".join(parts), flags)
+
+
+class ExternalCommands(pydantic.BaseModel):
```
smoparth, line 120: The key name external_commands in settings.yaml convey a broad scope subprocess settings (ex: timeouts, retries, wrappers, network_isolation config, etc). But right now it only handles env filtering, and __bool__ is also tied specifically to delete_env. If we later want to add other subprocess settings under this key, we'd need to rework __bool__ and possibly migrate the structure.

Should we use a nested key (e.g. external_commands.env) now, so the top-level key can evolve without breaking changes?

src/fromager/packagesettings/_models.py
```diff
@@ -72,6 +73,165 @@ class SbomSettings(pydantic.BaseModel):
...
+
+        Variables whose keys are not valid POSIX names are always
+        removed first.  Of the remaining variables, those matching
+        ``DEFAULT_KEEP_ENV`` or ``keep_env`` are always kept.  Of the
+        rest, those matching ``delete_env`` are removed.  Variables
+        matching neither list are kept.
+        """
+        # Remove keys that are not valid POSIX names, e.g. keys with
+        # dashes, dots, spaces, or bash function exports (BASH_FUNC_*%%).
+        env = {k: v for k, v in env.items() if _POSIX_ENV_KEY_RE.fullmatch(k)}
```
smoparth, line 216: Wiring code must check `if settings.external_commands:` before calling filter_env() to preserve backward compatibility. Without that guard, existing environments lose non-POSIX keys even with no filtering configured.

smoparth (commented):
LGTM! Left a non-blocking comment from maintenance perspective.

src/fromager/packagesettings/_models.py
tiran: Fair point. Let's get rid of `__bool__` then. It's not really necessary.

Also:

```python
>>> import this
The Zen of Python, by Tim Peters

...
Flat is better than nested.
...
```

src/fromager/packagesettings/_models.py
tiran: Non-compliant env keys have undefined behavior. It's okay to filter them out for all sandboxed calls.

smoparth (approved):
LGTM! Thanks for bearing with me and all detailed explanations ; )

rd4398 (approved):
LGTM!

PR #1272, opened 2026-07-24
docs(proposals): settings-driven hook for build-system dependency post-processing

vshawrh:
## Summary

Rewrites the proposal to use a settings-driven callable approach instead
of stevedore hooks.

- The hook is configured as a dotted import path in `settings.yaml`
- Loaded via Pydantic `ImportString` (same pattern as the proposed
  `build_tag_hook`)
- Runs after per-package overrides, before marker filtering
- Single callable, not stevedore-based

This replaces the earlier stevedore-based proposal that was rejected
because stevedore hooks are designed as fire-and-forget event listeners,
not behavior modifiers.

## Implementation PR

The corresponding implementation is in #1271.

Relates to #1263

### DIFF
--- a/docs/proposals/global-build-system-dependencies-hook.md
+++ b/docs/proposals/global-build-system-dependencies-hook.md
@@ -1,0 +1,212 @@
+# Settings-driven hook for build-system dependency post-processing
+
+- Author: Vikash Shaw
+- Created: 2026-07-24
+- Status: Proposed
+- GitHub issue: [#1263](https://github.com/python-wheel-build/fromager/issues/1263)
+- Proposal PR: [#1272](https://github.com/python-wheel-build/fromager/pull/1272)
+
+## What
+
+This proposal suggests adding a configurable hook in the global settings
+file that post-processes build-system dependencies for packages entering
+the source-build dependency-resolution path. The hook is specified as a
+dotted import path and loaded via Pydantic's `ImportString`, following
+the same pattern as the proposed `build_tag_hook`
+([wheel-build-tag-hook](wheel-build-tag-hook.md)).
+
+## Why
+
+Fromager currently provides two extension mechanisms:
+
+1. **Per-package plugins** (`fromager.project_overrides`): Override a
+   hook for a single package. When present, the plugin replaces the
+   default implementation entirely.
+
+2. **Global hooks** (`fromager.hooks`): Run for every package. Currently
+   support `post_build`, `post_bootstrap`, and `prebuilt_wheel`, which
+   are event callbacks that fire after an action has completed.
+
+Currently, no mechanism exists to post-process build-system dependencies
+across all packages. When a cross-cutting concern affects build
+dependencies for many packages, the only option today is to write
+identical per-package plugins for each one.
+
+### Motivating example
+
+setuptools 81 removed support for the `setup.py --dry-run` option and
+changed some related distutils/setuptools signatures. setuptools 82
+removed `pkg_resources` entirely. Many PyPI packages still reference
+these removed APIs in their `setup.py`, causing build failures when
+Fromager resolves an uncapped setuptools.
+
+In one downstream project, this led to **22 identical per-package
+plugins**, each scanning `setup.py` to detect removed API usage and
+appending a setuptools version cap. Every time a new package hits the
+same incompatibility, another identical plugin must be added. This does
+not scale well.
+
+### Why not `update_build_requires`?
+
+Fromager's YAML settings support `update_build_requires` for statically
+adding build dependencies. However, the setuptools cap is conditional
+and depends on what APIs a given `setup.py` actually uses. A static YAML
+entry would either over-constrain all packages or still require
+per-package entries, which reduces but does not eliminate the maintenance
+overhead.
+
+## How
+
+### Configuration
+
+The hook is configured in the global settings file as a dotted import
+path, using Pydantic's `ImportString` for validation and loading:
+
+```yaml
+build_system_dependencies_hook: "my_package.hooks:fix_build_deps"
+```
+
+When not set, no post-processing is applied and behavior is unchanged.
+
+### Hook signature
+
+```python
+def fix_build_deps(
+    *,
+    ctx: context.WorkContext,
+    req: Requirement,
+    sdist_root_dir: pathlib.Path,
+    build_dir: pathlib.Path,
+    requirements: list[str],
+) -> list[str]:
+    ...
+```
+
+The hook receives the current build-system requirements as a `list[str]`
+and must return a (possibly modified) `list[str]`.
+
+Before calling the hook, Fromager materializes the `Iterable[str]`
+returned by `overrides.find_and_invoke()` into a `list[str]`, so the
+hook always receives a concrete list.
+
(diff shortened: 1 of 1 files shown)


rd4398:
@vshawrh CI is failing here

docs/proposals/global-build-system-dependencies-hook.md
```diff
@@ -0,0 +1,234 @@
...
+[project.entry-points."fromager.hooks"]
+get_build_system_dependencies = "my_package.hooks:get_build_system_dependencies"
+```
+
+### Example hook
+
+A minimal hook that appends a constraint:
+
+```python
+def get_build_system_dependencies(
```
rd4398: Note: Today's global hooks (post_build, post_bootstrap, prebuilt_wheel) are fire-and-forget — they return nothing, they don't chain, and execution order doesn't matter. The proposed get_build_system_dependencies hook fundamentally changes this: hooks receive the previous hook's output and return a modified list.

docs/proposals/global-build-system-dependencies-hook.md
```diff
@@ -0,0 +1,234 @@
...
+    build_dir: pathlib.Path,
+    requirements: list[str],
+) -> list[str]:
+    ...
+```
+
+The hook receives the current requirements list and must return a
+(possibly modified) `list[str]`. When multiple hooks are registered,
+they chain: each receives the previous hook's output. Execution order
+follows stevedore's `HookManager` iteration order.
```
rd4398: HookManager doesn't guarantee order — it depends on entry point discovery, which varies across installations and Python versions. For side-effect hooks this is fine. For chained hooks where hook_A(hook_B(deps)) may differ from hook_B(hook_A(deps)), it's a problem. How should users control or reason about order?

docs/proposals/global-build-system-dependencies-hook.md
```diff
@@ -0,0 +1,234 @@
...
+    sdist_root_dir: pathlib.Path,
+    build_dir: pathlib.Path,
+    requirements: list[str],
+) -> list[str]:
+    ...
+```
+
+The hook receives the current requirements list and must return a
+(possibly modified) `list[str]`. When multiple hooks are registered,
+they chain: each receives the previous hook's output. Execution order
```
rd4398: If one hook in the chain raises an exception, does the entire chain fail? Does the build fail? Are subsequent hooks skipped? The existing hooks don't need to answer this because they're independent but chained hooks do.

docs/proposals/global-build-system-dependencies-hook.md
```diff
@@ -0,0 +1,234 @@
...
+
+Fromager's YAML settings support `update_build_requires` for statically
+adding build dependencies. However, the setuptools cap is conditional
+and depends on what APIs a given `setup.py` actually uses. A static YAML
+entry would either over-constrain all packages or require per-package
+entries, which has the same maintenance burden as plugins.
+
+## Proposed approaches
+
+### Option A: Core logic in Fromager (PR [#1264](https://github.com/python-wheel-build/fromager/pull/1264))
```
rd4398: I think we already rejected option A right? Can we just document why we rejected it?

rd4398 (commented):
I have left a few comments on the updated proposal. Also, this proposal sets a precedent and if accepted, we may want other override hooks the same treatment (e.g.,get_build_backend_dependencies, get_build_sdist_dependencies). I would prefer if @LalatenduMohanty and @tiran take a look at this as well.

docs/proposals/global-build-system-dependencies-hook.md
```diff
@@ -0,0 +1,234 @@
...
+same incompatibility, another identical plugin must be added. This does
+not scale well.
+
+### Why not `update_build_requires`?
+
+Fromager's YAML settings support `update_build_requires` for statically
+adding build dependencies. However, the setuptools cap is conditional
+and depends on what APIs a given `setup.py` actually uses. A static YAML
+entry would either over-constrain all packages or require per-package
+entries, which has the same maintenance burden as plugins.
```
LalatenduMohanty: I do not think per-package entries are same maintenance burden as plugins but they still need extra work to be done which can be handled in an easier way.

docs/proposals/global-build-system-dependencies-hook.md
```diff
@@ -0,0 +1,234 @@
...
+  `ast.parse` to detect removed API usage
+- Call it from `default_get_build_system_dependencies()` and append
+  the constraint if needed
+
+**Changes in downstream projects:**
+
+- None. Delete the 22 identical plugins and their entry points. The
+  packages build correctly without any downstream configuration,
+  plugin, or hook registration.
+
```
LalatenduMohanty: We should keep one solution here the one you are suggesting and move the others to alternative section.

docs/proposals/global-build-system-dependencies-hook.md
```diff
@@ -0,0 +1,234 @@
...
+## Proposed approaches
+
+### Option A: Core logic in Fromager (PR [#1264](https://github.com/python-wheel-build/fromager/pull/1264))
+
+Add the setuptools detection logic directly into
+`default_get_build_system_dependencies`. Fromager would automatically
+scan `setup.py` for `pkg_resources` imports and `dry_run` keyword
+arguments, and append the appropriate setuptools version cap to the
+build dependencies.
```
LalatenduMohanty: I understand setuptools are an special use-case but having a generic solution which can be used to solve similar problems would be something I would prefer.

docs/proposals/global-build-system-dependencies-hook.md
```diff
@@ -0,0 +1,234 @@
...
+  after per-package overrides return, before marker filtering
+
+**Changes in downstream projects:**
+
+- Add a hook implementation (one file with the setuptools detection
+  logic)
+- Register it as an entry point under `fromager.hooks`
+- Delete the 22 identical per-package plugins and their entry points
+- The packages still need to be listed in the downstream project's
+  requirements/collections
```
LalatenduMohanty: This sounds good to me. I agree whether we want to release the plugin as standalone installable package is a choice.

LalatenduMohanty (changes requested):

vshawrh:
@rd4398 @LalatenduMohanty I would like you to seriously consider and go through https://github.com/python-wheel-build/fromager/pull/1272/changes#diff-eac3b67ca88f82d425680bb6df4a0cecc7502118fe351a2498c35fca62eb74abR173 and consider this as well

docs/proposals/global-build-system-dependencies-hook.md
tiran: @rd4398 You are absolutely correct.

The stevedore hooks are event listener hooks. They are designed to act on events like "wheel is ready". The hooks are registered and enabled at installation time of a package. The hooks are not suited to change behavior.

tiran (changes requested):

vshawrh:
@tira @LalatenduMohanty @rd4398 I’ve updated the proposal. Please review the changes. If you have a solution or an alternative approach to this issue, please feel free to share it as well.

PR #1273, opened 2026-07-24
feat(wheels): add configurable build tag hook for wheel filenames

### LINKED ISSUE #1181: Implement wheel_build_tag hook for unique wheel file names

jlarkin09:
Implement the accepted proposal from docs/proposals/wheel-build-tag-hook.md (issue #1059, tracking issue #1181).

Add a `wheels.build_tag_hook` option in global settings that lets downstream projects append environment-specific suffixes (OS, accelerator, torch ABI) to wheel build tags via a user-defined callable. The hook receives ctx, req, version, and wheel_tags and returns suffix segments joined with `_`.

- Add `WheelSettings` model with `build_tag_hook: ImportString` to settings
- Add `get_build_tag()` and `_validate_build_tag_segments()` to wheels.py
- Update `add_extra_metadata_to_wheels()`, bootstrapper cache checks, and `_is_wheel_built()` to use computed build tags
- Minimal finder update to match suffixed build tag filenames
- Validate hook output: reject single strings, invalid chars, non-strings
- No behavior change when hook is not configured

Continues on PR: https://github.com/python-wheel-build/fromager/pull/1217
Closes: #1181

### DIFF
--- a/src/fromager/bootstrapper/_cache.py
+++ b/src/fromager/bootstrapper/_cache.py
@@ -86,22 +86,27 @@ def _look_for_existing_wheel(
     search_in: pathlib.Path,
 ) -> tuple[pathlib.Path | None, pathlib.Path | None]:
     pbi = ctx.package_build_info(req)
-    expected_build_tag = pbi.build_tag(resolved_version)
+    base_build_tag = pbi.build_tag(resolved_version)
     logger.info(
-        f"looking for existing wheel for version {resolved_version} with build tag {expected_build_tag} in {search_in}"
+        f"looking for existing wheel for version {resolved_version} with build tag {base_build_tag} in {search_in}"
     )
     wheel_filename = finders.find_wheel(
         downloads_dir=search_in,
         req=req,
         dist_version=str(resolved_version),
-        build_tag=expected_build_tag,
+        build_tag=base_build_tag,
     )
     if not wheel_filename:
         return None, None
-    _, _, build_tag, _ = wheels.extract_info_from_wheel_file(req, wheel_filename)
-    if expected_build_tag and expected_build_tag != build_tag:
+    _, _, actual_build_tag, wheel_tags = wheels.extract_info_from_wheel_file(
+        req, wheel_filename
+    )
+    expected_build_tag = wheels.get_build_tag(
+        ctx=ctx, req=req, version=resolved_version, wheel_tags=wheel_tags
+    )
+    if expected_build_tag and expected_build_tag != actual_build_tag:
         logger.info(
-            f"found wheel for {resolved_version} in {wheel_filename} but build tag does not match. Got {build_tag} but expected {expected_build_tag}"
+            f"found wheel for {resolved_version} in {wheel_filename} but build tag does not match. Got {actual_build_tag} but expected {expected_build_tag}"
         )
         return None, None
     logger.info(f"found existing wheel {wheel_filename}")
@@ -129,16 +134,20 @@ def _download_wheel_from_cache(
         results = resolver.find_all_matching_from_provider(provider, pinned_req)
         wheel_url, _ = results[0]
         wheelfile_name = pathlib.Path(urlparse(wheel_url).path)
+        _, _, actual_build_tag, wheel_tags = wheels.extract_info_from_wheel_file(
+            req, wheelfile_name
+        )
         pbi = ctx.package_build_info(req)
-        expected_build_tag = pbi.build_tag(resolved_version)
+        expected_build_tag = wheels.get_build_tag(
+            ctx=ctx, req=req, version=resolved_version, wheel_tags=wheel_tags
+        )
         logger.info(f"has expected build tag {expected_build_tag}")
         changelogs = pbi.get_changelog(resolved_version)
         logger.debug(f"has change logs {changelogs}")
 
-        _, _, build_tag, _ = wheels.extract_info_from_wheel_file(req, wheelfile_name)
-        if expected_build_tag and expected_build_tag != build_tag:
+        if expected_build_tag and expected_build_tag != actual_build_tag:
             logger.info(
-                f"found wheel for {resolved_version} in cache but build tag does not match. Got {build_tag} but expected {expected_build_tag}"
+                f"found wheel for {resolved_version} in cache but build tag does not match. Got {actual_build_tag} but expected {expected_build_tag}"
             )
             return None, None
 
--- a/src/fromager/commands/build.py
+++ b/src/fromager/commands/build.py
@@ -485,11 +485,25 @@ def _is_wheel_built(
             wheel_server_urls=wheel_server_urls,
         )
         logger.info("found candidate wheel %s", url)
-        pbi = wkctx.package_build_info(req)
-        build_tag_from_settings = pbi.build_tag(resolved_version)
-        build_tag = build_tag_from_settings if build_tag_from_settings else (0, "")
         wheel_basename = downloads.extract_filename_from_url(url)
-        _, _, build_tag_from_name, _ = parse_wheel_filename(wheel_basename)
+        _, _, build_tag_from_name, wheel_tags = parse_wheel_filename(wheel_basename)
+    except Exception:
+        logger.debug(
+            "could not locate prebuilt wheel %s-%s on %s",
+            dist_name,
+            resolved_version,
+            wheel_server_urls,
+            exc_info=True,
+        )
+        logger.info("could not locate prebuilt wheel")
+        return None
+    else:
+        # Compute expected build tag in the else clause so hook
+        # validation errors propagate instead of being swallowed.
+        expected_tag = wheels.get_build_tag(
+            ctx=wkctx, req=req, version=resolved_version, wheel_tags=wheel_tags
+        )
+        build_tag = expected_tag if expected_tag else (0, "")
         existing_build_tag = build_tag_from_name if build_tag_from_name else (0, "")
         if (
             existing_build_tag[0] > build_tag[0]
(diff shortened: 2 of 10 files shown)


rd4398:
@jlarkin09 The CI is failing here and this also needs rebase. Can you please rebase it?

src/fromager/packagesettings/_models.py
```diff
@@ -33,6 +33,39 @@
...
 
+class WheelSettings(pydantic.BaseModel):
+    """Global wheel build settings
+
+    ::
+
+      wheels:
+        build_tag_hook: "mypackage.hooks:build_tag_hook"
+
+    .. versionadded:: 0.92.0
```
rd4398: I think we tagged 0.92.0, so this needs to change to 0.93.0 everywhere

src/fromager/wheels.py
```diff
@@ -38,12 +39,67 @@
...
+    base_tag = pbi.build_tag(version)
+    if not base_tag:
+        return base_tag
+
+    hook = ctx.settings.build_tag_hook
+    if hook is None:
+        return base_tag
+
+    raw = hook(ctx=ctx, req=req, version=version, wheel_tags=wheel_tags)
+    if isinstance(raw, str | bytes):
```
rd4398, line 89: Can we add a test to verify bytes are also rejected? Currently, we only have test for `str`

tests/test_wheels.py
```diff
@@ -337,3 +339,155 @@ def test_validate_wheel_file(
     else:
         with pytest.raises(ValueError):
             wheels.validate_wheel_filename(req, version, wheel_file)
+
```
rd4398, line 342: We are missing test for hook exception propagation. The whole point of moving `get_build_tag()` outside the broad exception handler in _is_wheel_built is so hook errors propagate. But there's no test verifying that a hook raising (e.g.) `RuntimeError` actually surfaces to the caller rather than being swallowed.

```
def test_hook_exception_propagates(self, testdata_context, ...):
      def hook(**kwargs):
          raise RuntimeError("hook failure")
      ...
      with pytest.raises(RuntimeError, match="hook failure"):
          wheels.get_build_tag(...)
```

src/fromager/bootstrapper/_cache.py
```diff
@@ -129,16 +134,17 @@ def _download_wheel_from_cache(
...
-        expected_build_tag = pbi.build_tag(resolved_version)
+        _, _, actual_build_tag, wheel_tags = wheels.extract_info_from_wheel_file(
+            req, wheelfile_name
+        )
+        expected_build_tag = wheels.get_build_tag(
+            ctx=ctx, req=req, version=resolved_version, wheel_tags=wheel_tags
+        )
         logger.info(f"has expected build tag {expected_build_tag}")
-        changelogs = pbi.get_changelog(resolved_version)
-        logger.debug(f"has change logs {changelogs}")
```
rd4398, line 136: Is there a reason why we are removing this log line?

src/fromager/commands/build.py
```diff
@@ -485,39 +485,8 @@ def _is_wheel_built(
...
-                logger.info("wheel not found in local cache, preparing to download")
-                wheel_filename = None
-
-        if not wheel_filename:
-            # if the found wheel was on an external server, then download it
-            logger.info("downloading wheel from %s", url)
-            wheel_filename = wheels.download_wheel(req, url, wkctx.wheels_downloads)
-
-        return wheel_filename
+        _, _, build_tag_from_name, wheel_tags = parse_wheel_filename(wheel_basename)
```
rd4398, line 489: url, wheel_basename, build_tag_from_name, and wheel_tags are assigned inside the try block and used after the except. The except block returns None, so the code is technically correct, but the flow is fragile. An else clause on the try would make the intent explicit and survive future refactoring more safely:

```
  try:
      ...
      _, _, build_tag_from_name, wheel_tags = parse_wheel_filename(wheel_basename)
  except Exception:
      ...
      return None
  else:
      # build tag comparison lives here
      expected_tag = wheels.get_build_tag(...)
      ...
```

src/fromager/finders.py
```diff
@@ -162,30 +162,27 @@ def find_wheel(
...
+        build_tag_prefixes.append("")
+
+    name_variants = [
+        filename_prefix,
+        canonical_name,
+        req.name,
+        req.name.replace("-", "."),
+    ]
+
+    candidate_bases: list[str] = []
```
rd4398: Can we continue using set() here? The original code used a set to deduplicate candidate bases. The new code uses a list built with nested loops. If, say,filename_prefix == canonical_name (common for simple package names), the same prefix is searched twice. This is a minor performance concern

src/fromager/finders.py
rd4398, line 150: The new logic adds a "{build_tag[0]}_" prefix when build_tag[1] is empty, so that a base tag (2, "") can match suffixed  wheels like 2_el9.6-.... But `find_wheel` returns the first glob match and stops. If the downloads directory contains multiple suffixed wheels (e.g., 2_el9.6-... and 2_el9.6_rocm7.1-...), the function may return the wrong one. The caller in `_look_for_existing_wheel` then rejects it on exact-tag mismatch and returns (None, None) — the correct wheel is never tried.

This creates a chicken-and-egg: you need wheel_tags from a found wheel to compute the expected suffix (via the hook), but you need the suffix to find the right wheel. The current loose-find + exact-check approach only works reliably when there's at most one candidate. Can we make `find_wheel()` return multiple values? cc @LalatenduMohanty for his thoughts

rd4398 (commented):
I have left few review comments. I would prefer if @LalatenduMohanty takes a look as well.

Also a nit: The commit has 6 duplicate Signed-off-by / Co-authored-by trailer pairs. Can you please squash/clean up that?

src/fromager/bootstrapper/_cache.py
jlarkin09: My mistake. Will restore the changelog debug logging.

src/fromager/finders.py
jlarkin09: @Randori-Lance @tiran @smoparth any thoughts?

smoparth: chicken-and-egg problem can be avoided. wheel_tags is not needed to find the wheel, 
1st use the find_wheel uses pbi.build_tag(version) which is just the changelog-based tag (7, ""), no hook involved.
After find_wheel returns a match, parse its filename to get wheel_tags, then passes those to get_build_tag() for the full comparison. The hook runs on an already-found wheel, not before finding it.
Also the proposal mentions that get_build_tag hook must use the wheel compatibility tags only to decide if the wheel is pure python or not. I know its a minor detail which cannot be enforced in code logic but we need to communicate it properly as a Note.

smoparth: For the single match problem you're describing I created a issue #1290 and also we need to update our finders module, I'm currently thinking of few ways i'll add to the ticket #1290 sooner.

rd4398:
@tiran can you please review this PR?

rd4398:
@jlarkin09 CI continues to fail on this. Please take a look

jlarkin09:
CI looks to be passing after re run @rd4398

LalatenduMohanty (changes requested):
The PR looks good to me, however we need to add required docs for this feature.

docs/customization.md
```diff
@@ -393,6 +393,88 @@ $ tox -e cli -- canonicalize flit-core
...
+    Returns:
+        A sequence of suffix segments (alphanumeric + dots only).
+        Must not return a single string or bytes object.
+
+    Raises:
+        ValueError: If segments contain invalid characters or types
+    """
+    # Example: Return variant-specific suffix
+    return [ctx.variant]
+```
```
LalatenduMohanty, line 446: Two different hook mechanisms, no explanation

The existing "Process hooks" section (line 478) uses entry points:
[project.entry-points."fromager.hooks"]
post_build = "package_plugins.module:function"

This new hook uses an import string in YAML:
wheels:
  build_tag_hook: "myproject.hooks:build_tag_hook"

These are fundamentally different plugin mechanisms. The doc doesn't explain why, or even acknowledge the difference. A reader who knows the existing hooks will wonder why this one doesn't use entry points. 

One sentence of rationale would help e.g., "Unlike process hooks (which use entry points), ted directly in settings.yaml because it applies globally rather than per-package."

docs/customization.md
```diff
@@ -393,6 +393,88 @@ $ tox -e cli -- canonicalize flit-core
...
+
+    The segments are joined with underscores and appended to the numeric
+    build tag. For example, returning ["cpu"] produces the build tag:
+    {numeric_base}_cpu, while returning ["gpu", "cuda"] produces
+    {numeric_base}_gpu_cuda.
+
+    The hook must return identical suffix segments when called with the same
+    configuration, ensuring that wheels built on different machines with the
+    same build configuration have identical filenames. This allows wheel caches
+    to work correctly across builders.
```
LalatenduMohanty: The doc string is too large, can we make it concise?

LalatenduMohanty (changes requested):

PR #1277, opened 2026-07-27
refactor(build_environment): use `self.run()` in `_createenv()`

### LINKED ISSUE #1276: refactor(build_environment): _createenv() should use self.run() like other BuildEnvironment methods
**BuildEnvironment._createenv()** calls `external_commands.run()` directly, while every other method (install, get_sysconfig_path, get_distributions) uses `self.run()`. This forces `_createenv()` to manually pass self._ctx.network_isolation and self._ctx.settings.external_commands, **duplicating logic that self.run() already centralizes**.

self.run() calls get_venv_environ(), which sets PATH, VIRTUAL_ENV, and UV_* env vars pointing to the venv. Since _createenv() runs before the venv exists, those env vars reference a directory that doesn't exist yet but `uv venv` creates the venv and it doesn't need to run inside one. The extra env vars pointing to the not-yet-created directory are ignored by uv venv and have no effect.

smoparth:
`_createenv()` called `external_commands.run()` directly while every other method uses `self.run()`, duplicating ctx-access logic for `network_isolation`. The `--python` CLI flag passed to `uv venv` takes precedence over `UV_PYTHON` set by `get_venv_environ()`, so the swap is safe.

Having a single entry point for command execution simplifies auditing and future cross-cutting changes like environment filtering or network policy enforcement.

Closes: #1276

# Pull Request Description

## Why

### DIFF
--- a/src/fromager/build_environment.py
+++ b/src/fromager/build_environment.py
@@ -206,10 +206,7 @@ class BuildEnvironment:
             "--no-project",
             str(self.path),
         ]
-        external_commands.run(
-            cmd,
-            network_isolation=self._ctx.network_isolation,
-        )
+        self.run(cmd)
         logger.info("created build environment in %s", self.path)
 
     def install(self, reqs: typing.Iterable[Requirement]) -> None:


tiran (commented):
The `_createenv` function does not use `BuildEnvironment.run` for a reason. The `BuildEnvironment.run` assumes that the virtual env exists.

smoparth:
> The `_createenv` function does not use `BuildEnvironment.run` for a reason. The `BuildEnvironment.run` assumes that the virtual env exists.

Even I thought same. I understand by the time _createenv() runs VIRTUAL_ENV, PATH, and UV_PYTHON all point to the not-yet-created directory, but I verified that `uv venv` ignores the venv-related env vars set by get_venv_environ():

UV_PYTHON is overridden by the explicit **--python sys.executable CLI flag** 
VIRTUAL_ENV and PATH are not consulted by uv venv -- it creates the venv at the positional path argument

I tested this locally by running UV_PYTHON=/nonexistent/bin/python3 uv venv --python $(python3 -c "import sys; print(sys.executable)") --no-project /tmp/test-venv and it succeeds.

That said, if the design intent is for self.run() to only be used once the venv exists, that makes sense -- I mainly wanted to understand the reasoning behind keeping it separate. Happy to close the PR if you'd prefer it stays as-is.

PR #1171, opened 2026-05-27
docs: add sandboxing API proposal

tiran:
# Pull Request Description

## What

Proposal for a pluggable sandboxing API that lets users configure custom sandboxing solutions for build process confinement.

## Why

#1019

- [x] PR follows [CONTRIBUTING.md](https://github.com/python-wheel-build/fromager/blob/main/CONTRIBUTING.md) guidelines

PR #1278, opened 2026-07-27
feat(external_commands): wire `ExternalCommands` env filtering into `run()`

### LINKED ISSUE #1083: Add environment variable filtering for external commands

smoparth:
Add `env_filter` parameter to `external_commands.run()` and thread it through all call sites that execute untrusted third-party code (PEP 517 build hooks, wheel metadata) and  uv cache invalidation.

`os.environ` is filtered through `ExternalCommands.filter_env()` before `extra_environ` is applied, ensuring injected build variables are never stripped.

Closes: #1083 

# Pull Request Description

## Why

### DIFF
--- a/src/fromager/build_environment.py
+++ b/src/fromager/build_environment.py
@@ -186,6 +186,7 @@ class BuildEnvironment:
             cwd=cwd,
             extra_environ=extra_environ,
             network_isolation=network_isolation,
+            env_filter=self._ctx.settings.external_commands,
             log_filename=log_filename,
             stdin=stdin,
         )
@@ -209,6 +210,7 @@ class BuildEnvironment:
         external_commands.run(
             cmd,
             network_isolation=self._ctx.network_isolation,
+            env_filter=self._ctx.settings.external_commands,
         )
         logger.info("created build environment in %s", self.path)
 
--- a/src/fromager/context.py
+++ b/src/fromager/context.py
@@ -154,7 +154,11 @@ class WorkContext:
         req_list: list[str] = sorted(set(req.name for req in reqs))
         logger.debug("invalidate uv cache for %s", req_list)
         cmd.extend(req_list)
-        external_commands.run(cmd, extra_environ=extra_environ)
+        external_commands.run(
+            cmd,
+            extra_environ=extra_environ,
+            env_filter=self.settings.external_commands,
+        )
 
     def package_build_info(
         self, package: str | packagesettings.Package | Requirement
--- a/src/fromager/dependencies.py
+++ b/src/fromager/dependencies.py
@@ -554,6 +554,7 @@ def get_build_backend_hook_caller(
             cwd=cwd,
             extra_environ=extra_environ,
             network_isolation=ctx.network_isolation,
+            env_filter=ctx.settings.external_commands,
             log_filename=log_filename,
         )
 
--- a/src/fromager/external_commands.py
+++ b/src/fromager/external_commands.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import logging
 import os
 import pathlib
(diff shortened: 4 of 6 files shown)


LalatenduMohanty:
@smoparth Please mention ` Closes: #1083` in the PR description.

smoparth:
## Wiring decisions

### Wired (5 call sites) for untrusted third-party code

These call sites execute PEP 517 build-backend hooks or other third-party scripts, so they receive `env_filter=ctx.settings.external_commands`:

| Call site | File | Why |
|---|---|---|
| `BuildEnvironment.run()` | `build_environment.py` | Runs PEP 517 hooks inside the build venv |
| `BuildEnvironment._createenv()` | `build_environment.py` | Creates the build venv via `uv` |
| `_run_hook_with_extra_environ()` | `dependencies.py` | Invokes build-backend hooks (`get_requires`, etc.) |
| `_add_extra_metadata()` | `wheels.py` | Runs `wheel pack` on built wheels |
| `WorkContext.uv_clean_cache()` | `context.py` | Invalidates uv cache entries |

### Not wired (8 call sites) — trusted tools only

These call sites invoke our own trusted tools (`git`, `cargo`, `patch`) and do **not** run arbitrary third-party code. They have no access to a `ctx` object today, so wiring them would require plumbing `ctx` (or the `ExternalCommands` instance) through the entire call chain.

| Call site | File | Commands invoked |
|---|---|---|
| `clone_source()` (×3) | `gitutils.py` | `git clone`, `git checkout` |
| `clone_source_ref()` (×2) | `gitutils.py` | `git checkout`, `git submodule` |
| `git_clean()` | `gitutils.py` | `git clean` |
| `vendor_rust()` | `vendor_rust.py` | `cargo vendor` |
| `_apply_patch()` | `sources.py` | `patch -p1` |


### To achieve full filtering

If **all** subprocesses should be filtered including `git`, `cargo`, and `patch` — the `ctx` (or `ExternalCommands` instance) would need to be threaded through the following call chains:

- `sources.py` → `_apply_patch()`, called from `_apply_patches()` -- which already receives `ctx`; straightforward one-hop plumb.
- `gitutils.py` → `clone_source()` / `clone_source_ref()` / `git_clean()` -- these are standalone utility functions. `ctx` would need to be added as a parameter, and all callers in `sources.py` and `bootstrapper/` need to be updated.
- `vendor_rust.py` → `vendor_rust()`, called from `_prepare_source.py` -- `ctx` is available one level up but not passed into `vendor_rust()` today.

LalatenduMohanty:
Overall direction of the PR looks good to me.

tiran:
This approach defaults to insecure behavior. In #1171 I proposed to deprecate and eventually remove `external_commands.run` in favor of `WorkContext.run_*` and `BuildEnvironment.run_sandboxed` methods.

smoparth:
> This approach defaults to insecure behavior. In #1171 I proposed to deprecate and eventually remove `external_commands.run` in favor of `WorkContext.run_*` and `BuildEnvironment.run_sandboxed` methods.

Thanks for pointing to #1171, I missed that follow-up proposal. I've been going with the [filter-env proposal#api-changes](https://fromager.readthedocs.io/en/latest/proposals/filter-env.html#api-changes) and actually even worked on the data model of it earlier, but had a few questions I wanted to clarify before pushing code. By the time I got back to it, I saw you'd already pushed that part and now again I missed this context about the follow-up proposal.

Defaulting to secure behavior makes a lot of sense. I'm eager to work on Fromager, I think we should have an epic or an issue with linked sub-tasks to help with getting the correct context and picking up next steps.

PR #1282, opened 2026-07-30
fix(downloads): set 0o644 permissions on downloaded files

### LINKED ISSUE #1281: bug(downloads): NamedTemporaryFile creates wheels with 0o600 permissions, breaking external wheel servers
## Summary

Downloaded wheel files are created with `0o600` (`-rw-------`) permissions since the download refactor in 0.90.0, making them unreadable by external wheel servers (e.g. nginx) that run as a different user. This causes `403 Forbidden` errors when `uv pip install` tries to fetch wheels from the local package index.

## Root cause

`downloads.download_url()` uses `tempfile.NamedTemporaryFile(delete=False)` to write downloaded files atomically. `NamedTemporaryFile` creates files with mode **0o600** by default. After `os.rename()`, the final `.whl` file retains these restrictive permissions.

The previous implementation in `sources.download_url()` used `open(temp_file, "wb")`, which creates files with mode **0o644** (respecting the default umask), making them readable by all users.

**New code** (`src/fromager/downloads.py`):
```python
tmp = tempfile.NamedTemporaryFile(
    dir=destination_dir,
    prefix=f".tmp.{basename}.",
    delete=False,
)
# ... write ...
os.rename(temp_path, outfile)  # outfile inherits 0o600 permissions

smoparth:
`NamedTemporaryFile` creates files with mode 0o600, making them unreadable by external wheel servers (e.g. nginx) that run as a different user. This caused 403 Forbidden errors when `uv pip install` tried to fetch wheels from the local package index.

Use `os.fchmod` before close+rename to widen permissions to 0o644, preserving the atomicity and thread-safety benefits of `NamedTemporaryFile`.

Closes: #1281

# Pull Request Description

## Why

### DIFF
--- a/src/fromager/downloads.py
+++ b/src/fromager/downloads.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import logging
 import os
 import pathlib
+import stat
 import tarfile
 import tempfile
 import typing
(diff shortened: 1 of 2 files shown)


rd4398 (approved):
This looks good! Thanks

PR #1289, opened 2026-08-05
Fix: Multi version bootstrap age filter fallback for edge cases 

### LINKED ISSUE #1288: Multiple version bootstrap should not silently skip packages when all candidates are older than max-release-age
## Problem

When `--multiple-versions` is combined with `--max-release-age`, if all candidates for a package are older than the configured age window, the resolver returns an empty set and the package is silently dropped from the build.

This is problematic in two distinct scenarios:

### 1. Unconstrained package where all releases are older than the timeframe

A package like `cmarkgfm` may not have had a release in the last 30 days. In multi-version mode with `--max-release-age=30`:

- `find_all_matching_from_provider()` is called with `fallback_on_empty_age_filter=False`
- All candidates are filtered out by the age check
- The resolver returns an empty list
- The package is silently skipped from the build

**Expected behavior**: When age filtering removes all candidates for an unconstrained package, fromager should fall back to building the latest matching version from source rather than returning an empty set.

### 2. Constrained/pinned package where the pinned version is older than the timeframe

A package pinned in constraints (e.g., `boto3==1.35.88`) where that specific version was published more than 30 days ago:

- The constraint restricts candidates to exactly one version
- Age filtering removes it (too old)
- The resolver returns an empty list
- The pinned version is silently skipped from the build

**Expected behavior**: When a package is explicitly pinned via constraints, the pinned version should always be built regardless of its age. A constraint pin is an explicit user intent that should take precedence over the age filter.

## Relevant code

- `bootstrap_requirement_resolver.py:162`: `fallback_on_empty_age_filter=not self.multiple_versions` — multi-version mode opts out of the fallback, causing the empty set return
- `resolver.py:332-340`: When `fallback_on_empty_age_filter=False` and all candidates are older than the cutoff, `candidates_list` is set to `[]`
- `bootstrap_requirement_resolver.py:165-171`: Cache server fallback exists but only triggers when a cache server URL is configured, and even then returns at most one version

## Suggested behavior

| Scenario | Current behavior | Proposed behavior |
|---|---|---|
| Unconstrained, all releases older than max-age | Returns empty set, package silently skipped | Build the latest version from source |
| Constrained/pinned, pinned version older than max-age | Returns empty set, package silently skipped | Always build the pinned version — constraint pin should override age filter |

rd4398:
Commit 1: 

```
feat(resolver): add AgeFallback.NEWEST for multi-version age filter
Replace the boolean `fallback_on_empty_age_filter` parameter in
`find_all_matching_from_provider()` with an `AgeFallback` enum that
supports three strategies when max-release-age filtering removes all
candidates:

- `ALL`: keep every candidate (single-version default, unchanged)
- `NEWEST`: keep only the single newest candidate (new)
- `NONE`: return empty list

Multi-version mode now uses `AgeFallback.NEWEST` so that when a
dependency has no releases within the age window (e.g. bar from 2
years ago depended on by recent foo), the newest version is built
instead of failing. This prevents cascading failures where the
dependent package also fails to build.

```

Commit 2:

```
fix(resolver): constrained packages bypass max-release-age filtering
Packages with explicit constraints (e.g. `boto3==1.35.88` in
constraints files) now skip age filtering entirely in
`find_all_matching_from_provider()`. A constraint is explicit user
intent and should not be silently overridden by the age heuristic.

Previously in multi-version mode with `--max-release-age`, a
constrained package whose pinned version was older than the age
window would have all candidates removed, causing silent resolution
failure. In single-version mode the fallback kept all candidates
anyway, but emitted a misleading "keeping all to avoid empty
resolution" warning.

```

Closes https://github.com/python-wheel-build/fromager/issues/1288

### DIFF
--- a/src/fromager/resolver.py
+++ b/src/fromager/resolver.py
@@ -6,6 +6,7 @@
 from __future__ import annotations
 
 import datetime
+import enum
 import functools
 import logging
 import os
@@ -58,6 +59,14 @@ IGNORE_PLATFORM: str = "ignore"
 SUPPORTED_TAGS_IGNORE_PLATFORM: frozenset[Tag] = frozenset(
     Tag(t.interpreter, t.abi, IGNORE_PLATFORM) for t in SUPPORTED_TAGS
 )
+
+
+class AgeFallback(enum.StrEnum):
+    """Strategy when max-release-age filtering removes all candidates."""
+
+    ALL = "all"
+    NEWEST = "newest"
+    NONE = "none"
 
 
 @functools.lru_cache(maxsize=200)
@@ -237,7 +246,7 @@ def find_all_matching_from_provider(
     provider: BaseProvider,
     req: Requirement,
     max_age_cutoff: datetime.datetime | None = None,
-    fallback_on_empty_age_filter: bool = True,
+    age_fallback: AgeFallback = AgeFallback.ALL,
 ) -> list[tuple[str, Version]]:
     """Find all matching candidates from provider without full dependency resolution.
 
@@ -248,15 +257,15 @@ def find_all_matching_from_provider(
         provider: The provider to query for candidates.
         req: The requirement to match.
         max_age_cutoff: If set, reject candidates published before this time.
-            If all candidates are older than the cutoff, all are kept and
-            a warning is emitted to avoid empty resolution.  Age filtering
-            is skipped entirely for packages that have a constraint in the
-            provider's ``constraints`` object, since constraints represent
-            explicit user intent that should not be overridden by a heuristic.
-        fallback_on_empty_age_filter: If ``True`` (default), keep all
-            candidates when age filtering would produce an empty result.
-            If ``False``, return an empty list instead, allowing the
-            caller to implement its own fallback strategy.
+            Age filtering is skipped for packages that have an exact
+            ``==`` pin in the provider's ``constraints`` object, since
+            a pin represents explicit user intent that should not be
+            overridden by a heuristic.  Range constraints (``>=``,
+            ``<``, etc.) still go through age filtering normally.
+        age_fallback: Strategy when age filtering removes all candidates.
+            ``ALL`` (default) keeps every candidate with a warning.
+            ``NEWEST`` keeps only the single newest candidate.
+            ``NONE`` returns an empty list, letting the caller handle it.
 
     Returns list of (url, version) tuples sorted by version (highest first).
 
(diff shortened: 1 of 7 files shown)


andre-motta (dismissed):
LGTM. Lets test this with https://pypi.org/project/dead/ on walkerpass or similar before releasing!

rd4398:
The fix for CI failures are part of https://github.com/python-wheel-build/fromager/pull/1287
I will rebase the PR once above mentioned PR lands

LalatenduMohanty:
@rd4398 https://github.com/python-wheel-build/fromager/pull/1287 has merged

pavank63 (dismissed):
LGTM

**Tested on local system (x86_64)** with old (main: `0.92.1.dev13`) and new (PR: `0.92.1.dev15`) side-by-side using `--multiple-versions`.

### `tomli>=2.0,<=2.0.2` with `--max-release-age=1`

| Scenario | OLD (main) | NEW (PR) |
|---|---|---|
| Unconstrained | `all 3 candidate(s) older than 1 days` → empty, skipped | `falling back to newest version 2.0.2` → resolves |
| Constrained `tomli==2.0.0` | `all 1 candidate(s) older than 1 days` → pinned version skipped | `skipping age filter for constrained package` → resolves 2.0.0 |

### `dead>=1.0` (latest 2.1.0 uploaded 2025-02-08, ~543 days old)

| Scenario | OLD (main) | NEW (PR) |
|---|---|---|
| `--max-release-age=30` | `all 11 candidate(s) older than 30 days` → empty, skipped | `falling back to newest version 2.1.0` → resolves |
| `--max-release-age=365` | `all 11 candidate(s) older than 365 days` → empty, skipped | `falling back to newest version 2.1.0` → resolves |
| `--max-release-age=600` | `have 3 candidate(s) within 600 days` → resolves 3 | `have 3 candidate(s) within 600 days` → resolves 3 |
| `--max-release-age=30` + `dead==1.5.2` | `all 1 candidate(s) older than 30 days` → pinned version skipped | `skipping age filter for constrained package` → resolves 1.5.2 |

When versions fall within the age window, both behave identically. The PR only changes behavior when all candidates are outside the window.

src/fromager/resolver.py
```diff
@@ -323,6 +343,18 @@ def find_all_matching_from_provider(
                 req,
                 max_age_days,
             )
+        elif age_fallback == AgeFallback.NEWEST:
```
Lanceypantsy: This branch is reached when filtered is empty (i.e., all candidates failed the age check). filtered is built from candidates_list, so if candidates_list itself is empty — possible when the requirement specifier matches nothing — candidates_list[0] raises IndexError. There is no guard and no unit test for this path.

  Suggested fix: guard before the index access:

```
  elif age_fallback == AgeFallback.NEWEST:
      if not candidates_list:
          candidates_list = []
      else:
          newest = candidates_list[0]
          logger.info(...)
          candidates_list = [newest]
```

  Or just check candidates_list before assigning newest.

src/fromager/resolver.py
```diff
@@ -290,7 +301,16 @@ def find_all_matching_from_provider(
     # Materialize candidates so we can iterate more than once if filtering
     candidates_list = list(candidates)
 
-    if max_age_cutoff is not None:
+    # Constraints are explicit user intent — they override age filtering.
+    is_constrained = provider.constraints.get_constraint(req.name) is not None
```
Lanceypantsy: `is_constrained` is broader than "pinned"

`get_constraint` returns a match for any constraint type — range, exclusion, or exact pin — not just pins. So a loose constraint like `tomli>=1.0` also sets `is_constrained=True` and skips age filtering entirely, even though the docstring frames this as bypassing only for exact pins ("explicit user intent").

Failure case: user sets tomli>=1.0 (just to exclude 0.x) plus --max-release-age=365, expecting the age filter to still apply — but it's silently skipped since any constraint match short-circuits it.

Options:

1. Only bypass for exact pins (`constraint.specifier == "=="`), fall through to age filtering otherwise.
2. Keep current behavior, but rename "pinned" → "constrained" in docstring/logs and document the tradeoff.

tests/test_cooldown.py
```diff
@@ -965,3 +998,63 @@ def test_resolve_package_cooldown_toplevel_compound_specifier_not_exempt(
...
     )
     assert result is _COOLDOWN
+
+
+# ---------------------------------------------------------------------------
+# constraint bypass tests — constrained packages skip age filtering
+# ---------------------------------------------------------------------------
+
+
+def test_max_release_age_skips_filter_for_constrained_package(
```
Lanceypantsy: Constraint-bypass tests don't use production's fallback mode

Both new unit tests call find_all_matching_from_provider with `age_fallback=resolver.AgeFallback.NONE`, but production multi-version mode (bootstrap --multiple-versions + constraints file) uses `AgeFallback.NEWEST`. That combination — constrained package + NEWEST fallback — isn't unit tested; only covered implicitly by the (slow, non-isolating) e2e test.

**Gap**: if constraint bypass later starts interacting with fallback strategy (e.g. "skip bypass when fallback is NEWEST"), nothing here would catch the regression.

Suggested fix: add a unit test in `test_cooldown.py` with `age_fallback=resolver.AgeFallback.NEWEST` and a constrained provider, asserting the constrained version is returned rather than just the newest unconstrained one.

src/fromager/resolver.py
```diff
@@ -290,7 +301,16 @@ def find_all_matching_from_provider(
...
+
+    if max_age_cutoff is not None and is_constrained:
```
Lanceypantsy: is_constrained is computed unconditionally on every call, but it's only ever used inside if max_age_cutoff is not None branches. When max_age_cutoff is None (no age filter configured — the common case), the constraint lookup is wasted work.

  This is minor on its own, but find_all_matching_from_provider is called once per requirement per resolution pass, and get_constraint does a dict lookup on a canonicalized name. Not expensive, but the intent is clearer if the check is co-located with its use:

```
  if max_age_cutoff is not None:
      is_constrained = provider.constraints.get_constraint(req.name) is not None
      if is_constrained:
          logger.info(...)
      else:
          # existing filter logic
```

  Not blocking, but worth the two-line fix for clarity.

src/fromager/resolver.py
```diff
@@ -60,6 +61,14 @@
 )
 
 
+class AgeFallback(enum.StrEnum):
```
Lanceypantsy, line 64: NIT: 

**`StrEnum` is unnecessary here — plain `Enum` would be fine**

```python
class AgeFallback(enum.StrEnum):
    NONE = "none"
    ALL = "all"
    NEWEST = "newest"
```

`StrEnum` makes members compare equal to their string values (`AgeFallback.NEWEST == "newest"`), which is useful when values get serialized to config files, CLI args, or JSON. None of that applies here — `AgeFallback` is only ever used in internal comparisons, never as a raw string. This leaves an implicit interface open: callers could pass the bare string `"newest"` and it would silently work, bypassing the type system.

A plain `enum.Enum` makes the type boundary explicit with no downside.

src/fromager/resolver.py
```diff
@@ -60,6 +61,14 @@
...
+    """Strategy when max-release-age filtering removes all candidates."""
+
+    NONE = "none"
```
Lanceypantsy: NIT:

  `ALL` is the default value for `age_fallback in find_all_matching_from_provider` and represents the original behavior. Convention in Python enums is to put the default/most-common member first. Reordering to `ALL`, `NEWEST`, `NONE` would match the parameter default and read naturally from most- to least-permissive.

src/fromager/resolver.py
rd4398: The NEWEST branch is only reached when filtered is empty (all candidates failed the age check), but candidates_list is the original unfiltered list. It's guaranteed non-empty because provider.find_matches() raises `ResolverException` if zero candidates match the requirement specifier (line 829-832). The exception is re-raised at line 286-288. So by the time we reach the NEWEST branch, candidates_list always has at least one element.

src/fromager/resolver.py
rd4398: Every internal enum in the codebase uses StrEnum (RequirementType, SourceType, BuildSDist, DownloadKind, BootstrapPhase). Switching to Enum would break the established convention. I would prefer keeping it this way

src/fromager/resolver.py
rd4398: fixed in latest commit

src/fromager/resolver.py
rd4398: done in latest commit

tests/test_cooldown.py
rd4398: done in latest commit

src/fromager/resolver.py
rd4398: sounds good, fixed in latest commit

LalatenduMohanty (approved):

LalatenduMohanty:
I have approved the PR however my concern is that multi-version resolution now has three fallback layers stacked:

  1. Age filter with NEWEST fallback (new)
  2. Cache server fallback (pre-existing, only when cache URL configured)
  3. Log warning and return empty (pre-existing)

  That's a lot of "try this, else try that" for one resolution path, and it makes the behavior hard to predict for someone reading the code cold. 

Single-version mode has none of this . The two modes are diverging in resolution semantics, which could become a maintenance burden as more edge cases surface.  

I will create a follow up issue which we can look in to resolving this post this PR merge.

Issue #1291, opened 2026-08-06
Activate `source` field on PackageSettings/VariantInfo and add `source_resolver` PBI property


smoparth:
## Context

The new `SourceResolver` data model is complete in `_resolver.py` (PRs #1228, #1235), but the `source` field is still **commented out** on `PackageSettings` and `VariantInfo` in `_models.py`, and `PackageBuildInfo` has no `source_resolver` property. This is the foundation that all other wiring work depends on.

**Parent epic:** #1254 (Phase 1 — critical path, blocks all other phases)

## Problem

The new resolver config cannot be used at runtime because the `source` field isn't active on the settings models and there's no way to access it through `PackageBuildInfo`.

## Acceptance criteria

- [ ] `source: SourceResolver | None` field is active on both `PackageSettings` and `VariantInfo`
- [ ] `SourceResolver` is importable from `fromager.packagesettings`
- [ ] `PackageBuildInfo` has a `source_resolver` property that returns variant-level `source` if set, else package-level `source`, else `None`
- [ ] Loading YAML with `source:` config produces the correct `SourceResolver` instance
- [ ] `pbi.source_resolver` returns `None` when no `source:` is configured (legacy config unchanged)
- [ ] All existing tests pass (no legacy behavior change)

## Files

- `src/fromager/packagesettings/_models.py`
- `src/fromager/packagesettings/_pbi.py`
- `src/fromager/packagesettings/__init__.py`

## Related

- #1207 — Mutual exclusivity validation (can be done as part of this or separately)
- Depends on: nothing
- Blocks: Phase 2, 3, 5, 6, 7, 8, 9

LalatenduMohanty:
https://github.com/python-wheel-build/fromager/pull/1305 has merged

PR #1305, opened 2026-08-13
feat(packagesettings): activate source field and add source_resolver PBI property

LalatenduMohanty:
## Summary

- Uncomment and activate `source: SourceResolver | None` field on both `PackageSettings` and `VariantInfo`
- Export `SourceResolver` from the `fromager.packagesettings` public API
- Add `PackageBuildInfo.source_resolver` property with variant-then-package fallback (variant source if set, else package source, else `None`)

This is the foundation for all downstream resolver-config wiring (Phases 2–9 of the new resolver config epic).

Closes: #1291

### DIFF
--- a/tests/test_packagesettings.py
+++ b/tests/test_packagesettings.py
@@ -18,9 +18,12 @@ from fromager.packagesettings import (
     Package,
     PackageBuildInfo,
     PackageSettings,
+    PyPIPrebuiltResolver,
+    PyPISDistResolver,
     ResolverDist,
     Settings,
     SettingsFile,
+    SourceResolver,
     Variant,
     substitute_template,
 )
@@ -90,6 +93,7 @@ FULL_EXPECTED: dict[str, typing.Any] = {
         "use_pypi_org_metadata": True,
         "min_release_age": None,
     },
+    "source": None,
     "variants": {
         "cpu": {
             "annotations": {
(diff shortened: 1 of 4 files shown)


tests/test_packagesettings.py
```diff
@@ -1152,3 +1162,89 @@ def test_filter_env(
...
     ec = ExternalCommands(keep_env=keep, delete_env=delete)
     assert ec.filter_env(env) == expected
+
+
+# -- source field and source_resolver PBI property ----------------------------
+
+
+def test_source_resolver_importable() -> None:
+    """``SourceResolver`` is importable from ``fromager.packagesettings``."""
+    assert SourceResolver is not None
```
rd4398: Non blocker nit: This assertion is always going to be true for any successfully imported name. I wonder whether asserting something more meaningful, e.g. `assert callable(SourceResolver)` or checking it's a type/annotation will be more useful. If not, we can leave it as a smoke test.

rd4398 (dismissed):
This looks good! I have left a non-blocker nit but I am okay to merge this as it is.

smoparth (dismissed):
LGTM!

tests/test_packagesettings.py
LalatenduMohanty: You are right, fixing it.

LalatenduMohanty: done. PTAL

rd4398 (approved):
LGTM

Issue #1207, opened 2026-06-22
Add mutual exclusivity validator for source vs legacy fields


smoparth:
## Context

The new `SourceResolver` data model is complete in `_resolver.py` (PRs #1228, #1235). Once the `source` field is activated on `PackageSettings` and `VariantInfo` (Phase 1 — #1291), we need validation to prevent users from mixing old and new config styles.

The design proposal ([Migration section](https://fromager.readthedocs.io/en/latest/proposals/new-resolver-config.html#migration)) states:

> Top-level and variant-specific `source` settings are mutually exclusive with `download_source`, `resolver_dist`, `wheel_server_url`, and `pre_build` settings. It is an error to combine the new `source` settings with any of the old settings.

**Parent epic:** #1254 (Phase 1)

## Problem

Without validation, a user can set both `source` and legacy fields in the same YAML file without error:

```yaml
source:
  provider: pypi-sdist
  index_url: https://custom.example/simple
resolver_dist:
  sdist_server_url: https://other.example/simple
  include_wheels: true
```

The new `source` field would silently take precedence and the legacy `resolver_dist` is ignored. This is confusing and can lead to configuration mistakes that are hard to debug.

## Acceptance criteria

- [ ] Setting `source` alongside `download_source`, `resolver_dist`, `wheel_server_url`, or `pre_built` raises a clear validation error
- [ ] Validation works at both package level and variant level
- [ ] Existing configs (legacy-only or new-only) continue to load without errors
- [ ] Error message clearly tells the user which fields conflict

## Related

- Part of #1291 (Phase 1: Activate source field + PBI property)

LalatenduMohanty:
Closing as https://github.com/python-wheel-build/fromager/pull/1307 has merged.

PR #1307, opened 2026-08-21
feat(packagesettings): error when source and legacy settings are both set

LalatenduMohanty:
## Summary

- Add Pydantic model validators on `PackageSettings` and `VariantInfo` that reject configs combining the new `source` field with legacy settings
- Package-level `source` conflicts with `download_source` and `resolver_dist`; variant-level `source` conflicts with `wheel_server_url` and `pre_built`
- Uses `model_fields_set` to correctly detect explicitly provided legacy fields even when they have defaults

## Test plan

- [x] Parametrized tests for invalid package-level combos (source + download_source, source + resolver_dist, both, empty download_source)
- [x] Parametrized tests for invalid variant-level combos (source + wheel_server_url, source + pre_built)
- [x] Valid legacy-only and source-only configs still load
- [x] Error message includes conflicting field names
- [x] mypy and lint pass

Closes: #1207

### DIFF
--- a/src/fromager/packagesettings/_models.py
+++ b/src/fromager/packagesettings/_models.py
@@ -483,6 +483,21 @@ class VariantInfo(pydantic.BaseModel):
     source: SourceResolver | None = None
     """Source resolver and downloader"""
 
+    @pydantic.model_validator(mode="after")
+    def validate_source_exclusivity(self) -> typing.Self:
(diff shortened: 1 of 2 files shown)


rd4398 (approved):
This looks good! I will let @smoparth take a look as well

PR #1308, opened 2026-08-25
feat(settings): add version-specific prebuilt wheel configuration

### LINKED ISSUE #801: Add version-specific prebuilt wheel configuration support

andre-motta:
## Summary

- Add `VersionSpecificSettings` model and `versions` mapping on `VariantInfo` for per-version `pre_built` and `wheel_server_url` overrides
- Add `is_pre_built(version)` and `get_wheel_server_url(version)` methods on `PackageBuildInfo` with resolution order: version-specific YAML > variant default
- Existing `pre_built` and `wheel_server_url` properties delegate to the new methods (single code path)
- When version-specific settings differ from variant defaults, re-resolve the download URL in the START phase so URL type matches the actual build mode
- Thread version through all callers in bootstrapper, resolver, wheels, and build command

Example configuration:

```yaml
variants:
  tpu-ubi9:
    pre_built: false
    versions:
      "2.9.0.dev20250730":
        pre_built: true
        wheel_server_url: https://gitlab.example.com/simple
      "2.8.0":
        pre_built: false
```

Closes #801

## Test plan

- [x] `is_pre_built(version)` with matching, non-matching, inherited, and local version segments (`+cpu`)
- [x] `get_wheel_server_url(version)` with version-specific and fallback behavior
- [x] `get_wheel_server_urls` integration with version parameter
- [x] `build_tag` with version-specific prebuilt override
- [x] `before_none_versions` validator with bare/omitted YAML key
- [x] Wildcard version specifier (`==1.*`) handling
- [x] Re-resolution when version-specific settings differ from variant default
- [x] Updated mock setups in bootstrapper and resolver tests
- [x] Full test suite passes (938 passed), mypy clean, ruff clean

### DIFF
--- a/docs/concepts/package-settings.rst
+++ b/docs/concepts/package-settings.rst
@@ -78,6 +78,9 @@ layers override earlier ones:
    │  4. Variant overrides (within package YAML)      │
    │     (env vars, pre_built, wheel_server_url)      │
    │                                                  │
+   │  4a. Version-specific variant overrides          │
+   │     (per-version pre_built, wheel_server_url)    │
+   │                                                  │
    │  5. Version-specific patches and changelog       │
    │     (patches/<pkg>-<version>/, changelog entries)│
    │                                                  │
--- a/docs/concepts/package-settings.rst
+++ b/docs/concepts/package-settings.rst
@@ -88,6 +88,9 @@ layers override earlier ones:
    │     (update_extra_environ can mutate env vars)   │
    └──────────────────────────────────────────────────┘
 
+.. versionadded:: 0.95.0
+   Step 4a: version-specific variant overrides (``versions`` mapping).
+
 For environment variables specifically, the merge order within a
 single ``get_extra_environ()`` call is: parallel-jobs settings, build
 environment paths, package-level ``env``, then variant-level ``env``.
--- a/docs/concepts/package-settings.rst
+++ b/docs/concepts/package-settings.rst
@@ -88,9 +88,6 @@ layers override earlier ones:
    │     (update_extra_environ can mutate env vars)   │
    └──────────────────────────────────────────────────┘
 
-.. versionadded:: 0.95.0
-   Step 4a: version-specific variant overrides (``versions`` mapping).
-
 For environment variables specifically, the merge order within a
 single ``get_extra_environ()`` call is: parallel-jobs settings, build
 environment paths, package-level ``env``, then variant-level ``env``.
--- a/src/fromager/bootstrap_requirement_resolver.py
+++ b/src/fromager/bootstrap_requirement_resolver.py
@@ -24,6 +24,17 @@ if typing.TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
+def _extract_pinned_version(req: Requirement) -> Version | None:
+    """Return the version if *req* is pinned to exactly one (``==``).
+
+    Returns ``None`` for range specifiers, extras-only, or empty specifiers.
+    """
+    specs = list(req.specifier)
+    if len(specs) == 1 and specs[0].operator == "==":
+        return Version(specs[0].version)
+    return None
+
+
 class BootstrapRequirementResolver:
     """Resolve package requirements from PyPI or dependency graph during bootstrap.
 
@@ -113,7 +124,8 @@ class BootstrapRequirementResolver:
         # Determine pre_built if not specified (needed for cache key)
         if pre_built is None:
             pbi = self.ctx.package_build_info(req)
-            pre_built = pbi.pre_built
+            pinned = _extract_pinned_version(req)
+            pre_built = pbi.is_pre_built(pinned)
 
         rule_key = (str(req), pre_built)
 
@@ -181,8 +193,12 @@ class BootstrapRequirementResolver:
             results = cached_resolution
         elif pre_built:
             # Resolve prebuilt wheel
+            pinned = _extract_pinned_version(req)
             wheel_server_urls = wheels.get_wheel_server_urls(
-                self.ctx, req, cache_wheel_server_url=resolver.PYPI_SERVER_URL
+                self.ctx,
+                req,
+                cache_wheel_server_url=resolver.PYPI_SERVER_URL,
+                version=pinned,
             )
             results = wheels.resolve_all_prebuilt_wheels(
                 ctx=self.ctx,
--- a/src/fromager/bootstrap_requirement_resolver.py
+++ b/src/fromager/bootstrap_requirement_resolver.py
@@ -12,6 +12,7 @@ import typing
 
 from packaging.requirements import Requirement
 from packaging.utils import NormalizedName, canonicalize_name
+from packaging.version import InvalidVersion
 from packaging.version import Version
 
 from . import finders, resolver, sources, wheels
@@ -27,11 +28,15 @@ logger = logging.getLogger(__name__)
 def _extract_pinned_version(req: Requirement) -> Version | None:
     """Return the version if *req* is pinned to exactly one (``==``).
 
-    Returns ``None`` for range specifiers, extras-only, or empty specifiers.
+    Returns ``None`` for range specifiers, wildcard pins (``==1.*``),
+    extras-only, or empty specifiers.
     """
     specs = list(req.specifier)
-    if len(specs) == 1 and specs[0].operator == "==":
-        return Version(specs[0].version)
+    if len(specs) == 1 and specs[0].operator == "==" and "*" not in specs[0].version:
+        try:
+            return Version(specs[0].version)
+        except InvalidVersion:
+            return None
     return None
 
 
--- a/src/fromager/bootstrap_requirement_resolver.py
+++ b/src/fromager/bootstrap_requirement_resolver.py
@@ -12,8 +12,7 @@ import typing
 
 from packaging.requirements import Requirement
 from packaging.utils import NormalizedName, canonicalize_name
-from packaging.version import InvalidVersion
-from packaging.version import Version
+from packaging.version import InvalidVersion, Version
 
 from . import finders, resolver, sources, wheels
 from .dependency_graph import DependencyGraph
--- a/src/fromager/bootstrapper/_start.py
+++ b/src/fromager/bootstrapper/_start.py
@@ -122,11 +122,16 @@ class Start(Phase):
         wi.pbi_pre_built = pbi.is_pre_built(wi.resolved_version)
         wi.exclusive_build = pbi.exclusive_build
 
-        if wi.pbi_pre_built != pbi.pre_built:
+        version_url = pbi.get_wheel_server_url(wi.resolved_version)
+        variant_url = pbi.wheel_server_url
+        needs_re_resolve = wi.pbi_pre_built != pbi.pre_built or (
+            wi.pbi_pre_built and version_url != variant_url
+        )
+        if needs_re_resolve:
             logger.info(
-                f"{wi.req} {wi.resolved_version}: version-specific pre_built "
-                f"override ({wi.pbi_pre_built}) differs from variant default "
-                f"({pbi.pre_built}), re-resolving URL"
+                f"{wi.req} {wi.resolved_version}: version-specific override "
+                f"(pre_built={wi.pbi_pre_built}, url={version_url}) differs "
+                f"from variant default, re-resolving URL"
             )
             new_url = _re_resolve_url(
                 bt.ctx,
--- a/src/fromager/bootstrapper/_start.py
+++ b/src/fromager/bootstrapper/_start.py
@@ -96,36 +96,15 @@ class Start(Phase):
 
         wi.build_sdist_only = bt.sdist_only and not wi.is_build_requirement_context()
 
-        # Add to graph before the seen-check so every parent-to-dep edge
-        # is recorded, even when the package was already processed.
-        if wi.req_type != RequirementType.TOP_LEVEL:
-            bt.add_to_graph(
-                wi.req,
-                wi.req_type,
-                wi.resolved_version,
-                wi.source_url,
-                wi.parent,
-            )
-
-        if bt.has_been_seen(wi.req, wi.resolved_version, wi.build_sdist_only):
-            logger.debug(
-                f"redundant {wi.req_type} dependency {wi.req} "
-                f"({wi.resolved_version}, sdist_only={wi.build_sdist_only}) "
-                f"for {bt.explain}"
-            )
-            return []
-        bt.mark_as_seen(wi.req, wi.resolved_version, wi.build_sdist_only)
-
-        logger.info(
-            f"new {wi.req_type} dependency {wi.req} resolves to {wi.resolved_version}"
-        )
-
         # Must set pbi_pre_built before constructing PrepareSource so that
         # PrepareSource.background_work() immediately sees the correct value.
         pbi = bt.ctx.package_build_info(wi.req)
         wi.pbi_pre_built = pbi.is_pre_built(wi.resolved_version)
         wi.exclusive_build = pbi.exclusive_build
 
+        # Re-resolve URL before graph insertion so the graph stores the
+        # final URL, and before the seen-check so duplicate parents still
+        # get their edge recorded with the correct URL.
         version_url = pbi.get_wheel_server_url(wi.resolved_version)
         variant_url = pbi.wheel_server_url
         needs_re_resolve = wi.pbi_pre_built != pbi.pre_built or (
@@ -152,5 +131,29 @@ class Start(Phase):
                     f"{wi.req} {wi.resolved_version}: could not re-resolve URL "
                     f"for pre_built={wi.pbi_pre_built}, using original"
                 )
+
+        # Add to graph after URL finalization but before the seen-check
+        # so every parent-to-dep edge is recorded with the correct URL.
+        if wi.req_type != RequirementType.TOP_LEVEL:
+            bt.add_to_graph(
+                wi.req,
+                wi.req_type,
+                wi.resolved_version,
+                wi.source_url,
+                wi.parent,
+            )
+
+        if bt.has_been_seen(wi.req, wi.resolved_version, wi.build_sdist_only):
+            logger.debug(
+                f"redundant {wi.req_type} dependency {wi.req} "
+                f"({wi.resolved_version}, sdist_only={wi.build_sdist_only}) "
+                f"for {bt.explain}"
+            )
+            return []
+        bt.mark_as_seen(wi.req, wi.resolved_version, wi.build_sdist_only)
+
+        logger.info(
+            f"new {wi.req_type} dependency {wi.req} resolves to {wi.resolved_version}"
+        )
 
         return [PrepareSource(wi)]
--- a/src/fromager/bootstrapper/_start.py
+++ b/src/fromager/bootstrapper/_start.py
@@ -39,12 +39,15 @@ def _re_resolve_url(
             cache_wheel_server_url=cache_wheel_server_url,
             version=resolved_version,
         )
-        url, _ = wheels.resolve_prebuilt_wheel(
-            ctx=ctx,
-            req=pinned_req,
-            wheel_server_urls=wheel_server_urls,
-            req_type=req_type,
-        )
+        try:
+            url, _ = wheels.resolve_prebuilt_wheel(
+                ctx=ctx,
+                req=pinned_req,
+                wheel_server_urls=wheel_server_urls,
+                req_type=req_type,
+            )
+        except ExceptionGroup:
+            return None
         return str(url)
     else:
         pbi = ctx.package_build_info(req)
@@ -93,6 +96,17 @@ class Start(Phase):
 
         wi.build_sdist_only = bt.sdist_only and not wi.is_build_requirement_context()
 
+        # Add to graph before the seen-check so every parent-to-dep edge
+        # is recorded, even when the package was already processed.
+        if wi.req_type != RequirementType.TOP_LEVEL:
+            bt.add_to_graph(
+                wi.req,
+                wi.req_type,
+                wi.resolved_version,
+                wi.source_url,
+                wi.parent,
+            )
+
         if bt.has_been_seen(wi.req, wi.resolved_version, wi.build_sdist_only):
             logger.debug(
                 f"redundant {wi.req_type} dependency {wi.req} "
@@ -138,15 +152,5 @@ class Start(Phase):
                     f"{wi.req} {wi.resolved_version}: could not re-resolve URL "
                     f"for pre_built={wi.pbi_pre_built}, using original"
                 )
-
-        # Add to graph after re-resolution so the graph has the final URL
-        if wi.req_type != RequirementType.TOP_LEVEL:
-            bt.add_to_graph(
-                wi.req,
-                wi.req_type,
-                wi.resolved_version,
-                wi.source_url,
-                wi.parent,
-            )
 
         return [PrepareSource(wi)]
--- a/src/fromager/bootstrapper/_start.py
+++ b/src/fromager/bootstrapper/_start.py
@@ -3,15 +3,62 @@ from __future__ import annotations
 import logging
 import typing
 
+from packaging.requirements import Requirement
+from packaging.version import Version
+
+from .. import resolver, sources, wheels
 from ..requirements_file import RequirementType
 from ._phase import Phase
 from ._prepare_source import PrepareSource
 from ._types import BootstrapPhase
 
 if typing.TYPE_CHECKING:
+    from .. import context
     from ._bootstrapper import Bootstrapper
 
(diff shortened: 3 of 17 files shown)


src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -24,6 +24,17 @@
...
 
 
+def _extract_pinned_version(req: Requirement) -> Version | None:
+    """Return the version if *req* is pinned to exactly one (``==``).
+
+    Returns ``None`` for range specifiers, extras-only, or empty specifiers.
+    """
+    specs = list(req.specifier)
+    if len(specs) == 1 and specs[0].operator == "==":
+        return Version(specs[0].version)
```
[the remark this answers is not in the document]
andre-motta: Fixed, now returns None for wildcard pins

src/fromager/packagesettings/_models.py
```diff
@@ -294,6 +295,32 @@ def validate_update_build_requires(cls, v: list[str]) -> list[str]:
...
+          wheel_server_url: https://gitlab.example.com/simple
+    """
+
+    model_config = MODEL_CONFIG
+
+    wheel_server_url: str | None = None
+    """Alternative package index for this version's pre-built wheel"""
+
+    pre_built: bool | None = None
+    """Override pre-built flag for this version (None = inherit variant default)"""
```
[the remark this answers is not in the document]
andre-motta: Intentionally scoped out as per maintainer feedback on #800

src/fromager/packagesettings/_models.py
```diff
@@ -324,10 +355,32 @@ class VariantInfo(pydantic.BaseModel):
...
+    @pydantic.field_validator("versions", mode="before")
+    @classmethod
+    def before_none_versions(
+        cls,
+        v: dict[str, typing.Any] | None,
+        info: core_schema.ValidationInfo,
+    ) -> dict[str, typing.Any]:
+        if v is None:
+            return {}
+        return v
```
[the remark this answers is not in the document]
andre-motta: Fixed

src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -24,6 +24,21 @@
 logger = logging.getLogger(__name__)
 
 
+def _extract_pinned_version(req: Requirement) -> Version | None:
```
Lanceypantsy, line 27: **[Critical]** — `src/fromager/bootstrap_requirement_resolver.py` (lines 9–17)

  **What could go wrong:** A perfectly normal requirement like `torch==2.9.*` will crash the whole bootstrap run, because the new helper assumes anything using `==` is a single exact version.

  The helper treats any specifier whose only operator is `==` as a pinned version and immediately builds a `Version` from it. But a wildcard pin such as `torch==2.9.*` also uses the `==` operator, and
  `Version("2.9.*")` raises `InvalidVersion` (a `ValueError`). Nothing catches it, so resolution aborts. Wildcard pins are common in top-level and transitive requirements.

  Guard against the wildcard and treat any unparseable version as "not pinned" rather than letting it raise.

  **Before:**
  ```python
  def _extract_pinned_version(req: Requirement) -> Version | None:
      """Return the version if *req* is pinned to exactly one (``==``).

      Returns ``None`` for range specifiers, extras-only, or empty specifiers.
      """
      specs = list(req.specifier)
      if len(specs) == 1 and specs[0].operator == "==":
          return Version(specs[0].version)
      return None
  ```

  **After:**
  ```python
  def _extract_pinned_version(req: Requirement) -> Version | None:
      """Return the version if *req* is pinned to exactly one (``==``).

      Returns ``None`` for range specifiers, wildcard pins, extras-only,
      or empty specifiers.
      """
      specs = list(req.specifier)
      if len(specs) == 1 and specs[0].operator == "==" and "*" not in specs[0].version:
          try:
              return Version(specs[0].version)
          except InvalidVersion:
              return None
      return None
  ```

andre-motta: Fixed

src/fromager/bootstrap_requirement_resolver.py
```diff
@@ -113,7 +129,8 @@ def resolve(
         # Determine pre_built if not specified (needed for cache key)
         if pre_built is None:
             pbi = self.ctx.package_build_info(req)
-            pre_built = pbi.pre_built
+            pinned = _extract_pinned_version(req)
+            pre_built = pbi.is_pre_built(pinned)
```
Lanceypantsy, line 132: **[Critical]** 

  **What could go wrong:** For any requirement that isn't pinned with an exact `==` (e.g. `torch>=2.0`, or a plain `torch` whose version is pinned by a constraints file), the version-specific config is silently
  ignored during resolution — and worse, the "how do we resolve this" decision and the "is this pre-built" flag recorded later can end up disagreeing, causing build failures.

  Here's the split: `resolve()` decides whether to look for a source package or a pre-built wheel based on `is_pre_built(_extract_pinned_version(req))`. But `_extract_pinned_version` only returns a version for an
  exact `==` pin — so for `torch>=2.0` it returns `None`, and resolution falls back to the variant-wide default. Later, though, the graph node and the build step decide the pre-built flag from the *resolved*
  version via `is_pre_built(version)`. So the two stages can make opposite decisions. (Note constraints are applied inside the provider, not merged into `req.specifier`, so even a fully-constrained run still sees
  a bare requirement here.)

  Concrete failure — variant default is "source", but `2.9.0` is configured pre-built, requirement is `torch>=2.0`:
  - `resolve()` uses the source provider and looks for an sdist.
  - If the package is pre-built-only on the mirror (the motivating case — no sdist published), resolution finds nothing and the build fails.
  - If an sdist happens to exist, the graph node is stored with a source download URL but `pre_built=True`, corrupting the graph.

  The reverse also happens (variant default pre-built, a specific version configured as source): `resolve()` returns a wheel URL, but the node records `pre_built=False`, so a later phase tries to build that
  version from source against a wheel URL and fails.

  **Fix direction:** resolve the version first, then derive both the resolution strategy and the `pre_built` flag from that single resolved version — so one decision drives everything. If that's too large for this
  PR, at minimum document the "only exact `==` pins trigger version-specific behavior" limitation prominently and validate that configured versions are reachable.
  ```

src/fromager/packagesettings/_pbi.py
```diff
@@ -168,6 +168,34 @@ def pre_built(self) -> bool:
...
+        1. Plugin hook ``is_pre_built`` (if version given and hook exists)
+        2. Version-specific YAML setting
+        3. Variant-wide default
+
+        .. versionadded:: 0.90.0
+        """
+        vi = self._ps.variants.get(self.variant)
+        if vi is None:
+            return False
+        if version is not None:
```
Lanceypantsy, line 184: **[Critical]** — `src/fromager/packagesettings/_pbi.py` (lines 190–196 and 205–223)

  **What could go wrong:** The version lookup misses the exact versions the feature was built for. Accelerator wheels are versioned like `2.9.0+cpu` or `2.9.0+rocm6.2`, but a config key of `"2.9.0"` will never
  match them — so the version-specific pre-built setting and wheel-server URL are silently ignored, and the package falls back to the variant default.

  The two new methods do `vi.versions.get(pv)` using the raw `Version`. A `Version`'s identity includes its "local" segment (the part after `+`), so `Version("2.9.0+cpu")` is not equal to `Version("2.9.0")` and
  hashes differently — the dict lookup misses. The sibling methods in this same class (`get_changelog`, `build_tag`, `get_patches`) deliberately strip the local segment with `Version(version.public)` before
  looking things up; these two new methods don't, so they behave inconsistently with the rest of the class.

  Normalize the version to its public form before the lookup, matching the existing methods.

  **Before:**
  ```python
          if version is not None:
              hook_fn = overrides.find_override_method(self.package, "is_pre_built")
              if hook_fn is not None:
                  result = overrides.invoke(
                      hook_fn, version=version, variant=self.variant
                  )
                  if result is not None:
                      return bool(result)
              pv = typing.cast(PackageVersion, version)
              vs = vi.versions.get(pv)
  ```

  **After:**
  ```python
          if version is not None:
              hook_fn = overrides.find_override_method(self.package, "is_pre_built")
              if hook_fn is not None:
                  result = overrides.invoke(
                      hook_fn, version=version, variant=self.variant
                  )
                  if result is not None:
                      return bool(result)
              pv = typing.cast(PackageVersion, Version(version.public))
              vs = vi.versions.get(pv)
  ```

  Apply the same `Version(version.public)` normalization in `get_wheel_server_url` (line ~219). Add a test with a `+cpu` local-version to lock in the behavior.

src/fromager/packagesettings/_models.py
```diff
@@ -450,6 +451,32 @@ def validate_update_build_requires(cls, v: list[str]) -> list[str]:
...
 
 
+class VersionSpecificSettings(pydantic.BaseModel):
+    """Per-version overrides within a variant.
+
+    Allows overriding ``pre_built`` and ``wheel_server_url`` for
+    specific package versions.  When a field is ``None``, the
+    variant-wide default is used.
+
+    .. versionadded:: 0.90.0
```
Lanceypantsy: The current version is `0.94.0`, so if we're adding annotations about "version added," they should point to a future version, not a past version

Lanceypantsy:
**[Critical]** 

The new `is_pre_built` plugin hook — the most dynamic and error-prone part of the change — has no test at all. The YAML-based behavior is well covered, but the branch that finds a plugin, calls it, and interprets its result is never exercised, so a regression there would ship silently.

  The hook path involves several things that can quietly break: loading the plugin, stripping keyword arguments by signature inside `invoke`, coercing the result to a bool, and giving the hook precedence over the YAML setting. None of that is verified.

  Add a test that registers an override plugin (follow how existing hooks like `get_resolver_provider` or `prepare_source` are tested) and asserts:
  - the hook's return value wins over both the YAML version setting and the variant default, and
  - when the hook returns `None`, resolution falls through to the YAML/variant logic.

Lanceypantsy:
**[Critical]** — Documentation 

  **What could go wrong:** Both new user-facing features — the `versions:` YAML block and the `is_pre_built` plugin hook — are completely undocumented. A configuration option and a plugin hook that users can't find in the docs effectively don't exist for them.

src/fromager/packagesettings/_pbi.py
```diff
@@ -168,6 +168,34 @@ def pre_built(self) -> bool:
             return vi.pre_built
         return False
 
+    def is_pre_built(self, version: Version | None = None) -> bool:
```
Lanceypantsy, line 171: **[Recommendation]** — `src/fromager/packagesettings/_pbi.py` (lines 163–223)

  **What could go wrong:** The version-agnostic `pre_built` property and the new version-aware `is_pre_built(version)` method now sit side by side with nearly identical names — and the same pairing exists for `wheel_server_url` (property) vs `get_wheel_server_url(version)` (method). Nothing steers a caller to the correct one. A future call site (or a rebase that reintroduces `.pre_built`) will silently use the
  variant-wide default and ignore version overrides — a correctness bug that still type-checks and passes coarse tests.

  This PR itself had to hunt down and migrate about seven call sites from `.pre_built` to `.is_pre_built(version)`; the old property remains fully callable, so the trap is permanent.

  At a minimum, add a docstring note on both properties: "Version-agnostic; prefer `is_pre_built(version)` / `get_wheel_server_url(version)` when a version is known." Better: have the properties delegate to the methods with `version=None` so there's a single code path, or deprecate the bare properties.

Lanceypantsy: preferred solution, common code path with add'l parameter/argument w/ default=None

src/fromager/packagesettings/_pbi.py
andre-motta: Fixed in commit fd3d8df. Both is_pre_built and get_wheel_server_url now normalize with Version(version.public) before lookup, matching the pattern in get_changelog, build_tag, and get_patches. Added tests with +cpu and +local suffixes.

src/fromager/bootstrap_requirement_resolver.py
andre-motta: Fixed in commit fd3d8df. Added _re_resolve_url() in _start.py that re-resolves the URL when version-specific pre_built differs from variant default, using wheels.resolve_prebuilt_wheel or sources.get_source_provider as appropriate.

src/fromager/overrides.py
```diff
@@ -30,6 +30,7 @@
     "get_build_system_dependencies",
     "get_install_dependencies_of_sdist",
     "get_resolver_provider",
+    "is_pre_built",
```
Lanceypantsy: **[Recommendation]** 

  **What could go wrong:** The new `is_pre_built` hook doesn't follow the pattern every other hook in the project uses, which makes it harder to discover, impossible to auto-document, and inconsistent for plugin authors.

  All 13 other entries in `OVERRIDE_HOOK_NAMES` have a matching `default_<name>` function and are invoked through `overrides.find_and_invoke(..., default_fn=...)`. `is_pre_built` has no `default_is_pre_built` and instead uses a bespoke `find_override_method` + `invoke` + manual null-check. The docs generator (`autofromagerhook` in `docs/reference/hooks.rst`) keys off the `default_*` function — so without one, this hook literally can't be documented the standard way, which is exactly why it's missing from the docs.

  Add a `default_is_pre_built(version, variant, ...)` function and route the call through `find_and_invoke`, mirroring `get_resolver_provider`. That fixes the inconsistency and unblocks the docs entry.

andre-motta:
> **[Critical]**
> 
> The new `is_pre_built` plugin hook — the most dynamic and error-prone part of the change — has no test at all. The YAML-based behavior is well covered, but the branch that finds a plugin, calls it, and interprets its result is never exercised, so a regression there would ship silently.
> 
> The hook path involves several things that can quietly break: loading the plugin, stripping keyword arguments by signature inside `invoke`, coercing the result to a bool, and giving the hook precedence over the YAML setting. None of that is verified.
> 
> Add a test that registers an override plugin (follow how existing hooks like `get_resolver_provider` or `prepare_source` are tested) and asserts:
> 
> * the hook's return value wins over both the YAML version setting and the variant default, and
> * when the hook returns `None`, resolution falls through to the YAML/variant logic.

 Fixed in commit 5a169d5. Added test_is_pre_built_hook_overrides_yaml covering: hook returns True (overrides YAML False), hook returns False (overrides YAML True), hook returns None (defers to YAML), no hook found, and no version (hook not consulted).
 
>[Critical] — Documentation
>
>What could go wrong: Both new user-facing features — the versions: YAML block and the is_pre_built plugin hook — >are completely undocumented. A configuration option and a plugin hook that users can't find in the docs effectively >don't exist for them.



 Fixed in commit 5a169d5. Updated package-settings.rst merge order diagram to include version-specific variant overrides. Updated hooks-and-overrides.rst to mention is_pre_built hook.

Lanceypantsy:
**[Recommendation]** 

  **What could go wrong:** The function that actually decides which index a pre-built wheel is pulled from — `get_wheel_server_urls` with the new `version=` argument — has no direct test. Since choosing the wrong
  index is the whole point of the feature, an untested branch here could route a wheel to the wrong server without anything catching it.

  Two gaps:
  - There's no test that calls `get_wheel_server_urls(..., version=...)` directly, so the version-specific URL routing is only exercised (if at all) transitively.
  - The `before_none_versions` validator — which exists to turn an empty/`None` `versions:` block into `{}` — is never fed that input, so it may be dead code.

  Add a direct test for `get_wheel_server_urls` covering both the version-specific override and the fallback to the variant default, plus a loader test with an empty `versions:` block.

Lanceypantsy:
I have more nits, but given the size of the review, and that it isn't necessary to push this over the finish-line in a hurry, I'll let others provide feedback, and the author to address what's here so far, then give it another review.

src/fromager/packagesettings/_pbi.py
andre-motta: Done in cc6e590. pre_built now delegates to is_pre_built() and wheel_server_url delegates to get_wheel_server_url(), so there's one code path. Future callers can't accidentally bypass version-specific logic by using the property.

src/fromager/overrides.py
andre-motta: Done in cc6e590. Added _default_is_pre_built (returns None to defer to YAML config) and routed through find_and_invoke, consistent with all other override hooks. This also means autofromagerhook can discover and document it.

andre-motta:
> **[Recommendation]**
> 
> **What could go wrong:** The function that actually decides which index a pre-built wheel is pulled from — `get_wheel_server_urls` with the new `version=` argument — has no direct test. Since choosing the wrong index is the whole point of the feature, an untested branch here could route a wheel to the wrong server without anything catching it.
> 
> Two gaps:
> 
> * There's no test that calls `get_wheel_server_urls(..., version=...)` directly, so the version-specific URL routing is only exercised (if at all) transitively.
> * The `before_none_versions` validator — which exists to turn an empty/`None` `versions:` block into `{}` — is never fed that input, so it may be dead code.
> 
> Add a direct test for `get_wheel_server_urls` covering both the version-specific override and the fallback to the variant default, plus a loader test with an empty `versions:` block.

Done in cc6e590. Added three tests:
- test_get_wheel_server_urls_version_specific verifies version-specific URL override flows through the wheels module (default URL, version 2.9.0 override, and fallback for unknown version)
- test_variant_info_versions_none_yaml covers bare versions: key in YAML (parsed as None, coerced to empty dict)
- test_variant_info_versions_omitted covers omitting versions entirely (default empty dict)

src/fromager/bootstrapper/_start.py
```diff
@@ -72,6 +119,29 @@ def run(self, bt: Bootstrapper) -> list[Phase]:
...
+                wi.pbi_pre_built,
+                bt.cache_wheel_server_url,
+            )
+            if new_url is not None:
+                wi.source_url = new_url
+            else:
+                logger.warning(
+                    f"{wi.req} {wi.resolved_version}: could not re-resolve URL "
+                    f"for pre_built={wi.pbi_pre_built}, using original"
+                )
```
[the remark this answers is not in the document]
andre-motta: Done in 075ad4f. The re-resolve check now triggers on either pre_built mismatch OR wheel_server_url mismatch (when prebuilt). This handles the case where both variant and version use pre_built: true but point to different wheel servers.

docs/concepts/package-settings.rst
```diff
@@ -78,6 +78,9 @@ layers override earlier ones:
    │  4. Variant overrides (within package YAML)      │
    │     (env vars, pre_built, wheel_server_url)      │
    │                                                  │
+   │  4a. Version-specific variant overrides          │
+   │     (per-version pre_built, wheel_server_url)    │
```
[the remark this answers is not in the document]
andre-motta: Done in 075ad4f. Added .. versionadded:: 0.95.0 after the merge-order diagram in package-settings.rst and inline version note in hooks-and-overrides.rst.

tests/test_packagesettings.py
```diff
@@ -584,6 +602,144 @@ def test_global_changelog(testdata_context: context.WorkContext) -> None:
...
+    ) as mock_invoke:
+        pbi.is_pre_built()
+        mock_invoke.assert_not_called()
+
+
+def test_get_wheel_server_urls_version_specific(
+    testdata_context: context.WorkContext,
+) -> None:
+    """get_wheel_server_urls uses version-specific URL when version given."""
+    from fromager import wheels
```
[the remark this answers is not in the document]
andre-motta: Done in 075ad4f. Moved from fromager import wheels to the module-level import block.

src/fromager/bootstrapper/_start.py
```diff
@@ -3,17 +3,64 @@
...
+    resolved_version: Version,
+    pre_built: bool,
+    cache_wheel_server_url: str | None,
+) -> str | None:
+    """Re-resolve the download URL when version-specific pre_built differs.
+
+    Returns the new URL or ``None`` if re-resolution fails.
+    """
+    pinned_req = Requirement(f"{req.name}=={resolved_version}")
+    if pre_built:
```
rd4398, line 35: This can raise an unhandled `ExceptionGroup` from `wheels.resolve_prebuilt_wheel` (which propagates from `resolve_all_prebuilt_wheels` at wheels.py:548) if no matching wheel is found on the version
-specific server. The source branch correctly returns `None` when no results are found (line 46-47), but the prebuilt branch has no exception handling. The caller at line 142-147 expects either a URL or `None` and logs a warning on `None` — an uncaught exception would instead crash the bootstrap run.

Can we wrap the prebuilt resolution in a try/except block and return `None` on failure, matching the source branch's behavior?

docs/concepts/package-settings.rst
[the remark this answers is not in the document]
andre-motta: Done in e8e9fdf. Removed the .. versionadded:: directive from the concept doc. Per repo guidelines, these directives belong in API-level docs only, not in conceptual architecture docs.

src/fromager/bootstrapper/_start.py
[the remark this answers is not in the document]
andre-motta: Done in e8e9fdf. Moved add_to_graph() to after the re-resolution block so the dependency graph always records the final download URL.

tests/test_packagesettings.py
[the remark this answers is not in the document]
andre-motta: This from fromager import __main__ import is pre-existing code, not introduced by this PR. Out of scope for this change.

src/fromager/bootstrapper/_start.py
```diff
@@ -3,17 +3,64 @@
...
 from ._types import BootstrapPhase
 
 if typing.TYPE_CHECKING:
+    from .. import context
     from ._bootstrapper import Bootstrapper
 
 logger = logging.getLogger(__name__)
 
 
+def _re_resolve_url(
```
rd4398, line 22: This function is the most important part of PR and has no test coverage.

Can we add tests for

(1) re-resolution when version-specific `pre_built=True` overrides variant default `pre_built=False` (prebuilt path), 
(2) re-resolution when version-specific `pre_built=False` overrides variant default `pr
e_built=True` (source path), 
(3) re-resolution when only `wheel_server_url` differs, 
(4) failure handling when re-resolution returns `None` (falls back to original URL with warning), 
(5) no re-resolution when version-specific settings match variant defaults.

rd4398 (changes requested):
Overall looks good but needs two changes

src/fromager/bootstrapper/_start.py
andre-motta: Wrapped `resolve_prebuilt_wheel` in a try/except ExceptionGroup block that returns None, matching the source branch's behavior. 

The caller already handles None by falling back to the original URL with a warning.

src/fromager/bootstrapper/_start.py
andre-motta: Added 7 tests in TestReResolveUrl covering: prebuilt success, source success, ExceptionGroup failure, empty source results, wheel_server_url mismatch triggering re-resolution, no re-resolution when settings match, and fallback on failure.

src/fromager/bootstrapper/_start.py
```diff
@@ -72,6 +109,44 @@ def run(self, bt: Bootstrapper) -> list[Phase]:
...
+
+        # Add to graph after re-resolution so the graph has the final URL
+        if wi.req_type != RequirementType.TOP_LEVEL:
+            bt.add_to_graph(
+                wi.req,
+                wi.req_type,
+                wi.resolved_version,
+                wi.source_url,
+                wi.parent,
+            )
```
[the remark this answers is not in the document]
andre-motta: Good catch. Moved add_to_graph back before has_been_seen to restore the original ordering, so edges from a second parent are recorded even when the package was already processed. Added a regression test with the same requirement from two different parents.

PR #1310, opened 2026-09-01
docs(proposals): add SHA-256 digest capture and verification proposal

### LINKED ISSUE #1296: SHA-256 digest capture and verification for all downloads (Provenance Phase 1)
## Summary

fromager downloads sdists and wheels without computing or verifying any cryptographic hash. A compromised upstream can replace artifact bytes between runs and fromager will accept them silently. This is Phase 1 of #1238.

## Current state

- `download_url()` streams bytes to disk with no `hashlib` computation
- `#sha256=` URL fragments from PyPI indexes are carried in URL strings but never verified
- `pypi_simple.DistributionPackage.digests` is available but never read
- `Candidate`, `DependencyNode`, and build-order entries have no digest field

## Proposed changes

1. **Compute SHA-256 during download** — add streaming hash in `download_url()`, the single chokepoint for all HTTP downloads; return the digest alongside the file path
2. **Verify URL fragment hashes** — extract `#sha256=` from download URLs and hard-fail on mismatch with computed digest
3. **Read index digests** — use `dp.digests` from `pypi_simple` as a cross-check when creating `Candidate` objects
4. **Store digests in data structures** — add optional `digest` field to `Candidate`, `DependencyNode`/`DependencyNodeDict`, and build-order entries (backward-compatible via `NotRequired`/defaults)
5. **Write `provenance-index.json`** — aggregate per-package digests during `finalize()`, alongside `build-order.json`
6. **Re-download verification** — when a digest is already recorded, verify new downloads produce identical bytes (trust-on-first-use)

LalatenduMohanty:
Proposes computing, storing, and verifying SHA-256 digests for all downloaded artifacts to close the supply-chain gap where fromager accepts unverified bytes.

Closes: #1296

### DIFF
--- a/docs/proposals/digest-verification.md
+++ b/docs/proposals/digest-verification.md
@@ -1,0 +1,432 @@
+# Upstream and downstream source provenance
+
+- Author: Lalatendu Mohanty
+- Created: 2026-08-06
+- Status: Open
+- Issue: [#1296](https://github.com/python-wheel-build/fromager/issues/1296)
+  (child of [#1238](https://github.com/python-wheel-build/fromager/issues/1238))
+
+## Prerequisite
+
+This proposal assumes that Fromager builds wheels from the verified downstream
+sdist. The current implementation builds the wheel from the mutable prepared
+source tree, which may be modified during sdist creation—for example, by
+adding `PKG-INFO`.
+
+This is tracked separately in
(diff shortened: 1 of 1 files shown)


rd4398 (commented):
This looks good and I really like the idea. I will let @tiran take a look as well before clicking approve.

Also, we should set a timeline for this to be implemented. I would love to contribute to the implementation

PR #1314, opened 2026-09-03
feat(bootstrapper): write partial-failures.json for multi-version mode

ryanpetrello:
## Summary

- Write `partial-failures.json` to `work-dir/` when multi-version bootstrap has package failures
- Exit code stays 0 (the expected behavior with multi-version bootstrap is enabled)

This change gives analysis tools the metadata necessary to detect soft failures without manually parsing logs/stdout

## Test plan

- [x] Report written when multi-version has failures
- [x] No report when zero failures
- [x] No report when `multiple_versions=False`
- [x] Existing tests pass

### DIFF
--- a/src/fromager/bootstrapper/_bootstrapper.py
+++ b/src/fromager/bootstrapper/_bootstrapper.py
@@ -124,7 +124,8 @@ class Bootstrapper:
         self.failed_packages: list[FailureRecord] = []
 
         # Track failed versions in multiple_versions mode
-        self._failed_versions: dict[tuple[str, str], Exception] = {}
+        # Maps (package_name, version) -> (exception, detail)
+        self._failed_versions: dict[tuple[str, str], tuple[Exception, str]] = {}
 
     @property
     def resolver(self) -> bootstrap_requirement_resolver.BootstrapRequirementResolver:
@@ -1001,6 +1002,13 @@ class Bootstrapper:
             self.record_test_mode_failure(
                 wi.req, str(wi.resolved_version), err, "bootstrap"
             )
+            if self.multiple_versions:
+                self._record_failed_version(
+                    wi.req,
+                    str(wi.resolved_version),
+                    err,
+                    f"failed during {type(item).phase} phase",
+                )
             return []
 
         # Multiple versions mode: record failure, remove from graph, continue
@@ -1035,7 +1043,7 @@ class Bootstrapper:
     ) -> None:
         """Record a version failure in multiple versions mode."""
         pkg_name = canonicalize_name(req.name)
-        self._failed_versions[(pkg_name, version)] = err
+        self._failed_versions[(pkg_name, version)] = (err, detail)
         logger.warning(
             "%s==%s: %s: %s: %s",
             req.name,
@@ -1048,8 +1056,38 @@ class Bootstrapper:
     def _log_failed_versions_table(self) -> None:
         """Log a summary table of all failed versions."""
         logger.warning("%d version(s) failed to bootstrap:", len(self._failed_versions))
-        for (name, ver), exc in self._failed_versions.items():
+        for (name, ver), (exc, _detail) in self._failed_versions.items():
             logger.warning("  %s==%s: %s: %s", name, ver, type(exc).__name__, exc)
+
+    def _write_partial_failures_report(self) -> None:
+        """Write a JSON report of multi-version bootstrap failures.
+
+        Produces a ``partial-failures.json`` file in the work directory
+        so that downstream analysis tools can detect soft failures in
+        jobs that exited successfully.
+        """
+        failures_file = self.ctx.work_dir / "partial-failures.json"
+        failures = sorted(
+            [
+                {
+                    "name": name if ver == "unresolved" else f"{name}=={ver}",
+                    "version": None if ver == "unresolved" else ver,
+                    "phase": detail,
+                    "error_type": type(exc).__name__,
+                    "message": str(exc),
+                }
+                for (name, ver), (exc, detail) in self._failed_versions.items()
+            ],
+            key=lambda f: f["name"] or "",
+        )
+        with open(failures_file, "w") as f:
+            json.dump({"failures": failures}, f, indent=2)
+            f.write("\n")
+        logger.info(
+            "multi-version: wrote %d failure(s) to %s",
+            len(failures),
+            failures_file,
+        )
 
     def finalize(self) -> int:
         """Finalize bootstrap and return exit code.
@@ -1089,6 +1127,9 @@ class Bootstrapper:
 
         if self.multiple_versions and self._failed_versions:
             self._log_failed_versions_table()
+            self._write_partial_failures_report()
+        else:
+            (self.ctx.work_dir / "partial-failures.json").unlink(missing_ok=True)
 
         if not self.test_mode:
             return 0
--- a/src/fromager/resolver.py
+++ b/src/fromager/resolver.py
@@ -1181,7 +1181,7 @@ class GitHubTagProvider(GenericProvider):
         identifier: str,
     ) -> Iterable[Candidate]:
         headers = {"accept": "application/vnd.github+json"}
-        nexturl = self.api_url.format(self=self)
+        nexturl: str | None = self.api_url.format(self=self)
         while nexturl:
             resp = session.get(nexturl, headers=headers)
             resp.raise_for_status()
@@ -1288,7 +1288,7 @@ class GitLabTagProvider(GenericProvider):
         self,
         identifier: str,
     ) -> Iterable[Candidate]:
-        nexturl: str = self.api_url
+        nexturl: str | None = self.api_url
         created_at: datetime.datetime | None
         project_name = self.project_path.split("/")[-1]
         if self.override_download_url is None:
--- a/tests/test_bootstrapper.py
+++ b/tests/test_bootstrapper.py
@@ -426,9 +426,10 @@ def test_multiple_versions_continues_on_error(tmp_context: WorkContext) -> None:
                 pkg_name = canonicalize_name("testpkg")
                 version_str = "1.5"
                 assert (pkg_name, version_str) in bt._failed_versions
-                exc = bt._failed_versions[(pkg_name, version_str)]
+                exc, detail = bt._failed_versions[(pkg_name, version_str)]
                 assert isinstance(exc, ValueError)
                 assert str(exc) == "Simulated failure for version 1.5"
+                assert "failed during" in detail
 
         # Verify that failed version 1.5 is NOT in the dependency graph
         failed_key = f"{canonicalize_name('testpkg')}==1.5"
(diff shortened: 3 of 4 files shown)


ryanpetrello:
Example of this change while attempting to build `uv` in a container that doesn't have the rust toolchain:

```
~ podman run --rm -v ~/dev/fromager:/fromager:ro python:3.12 bash -c '
mkdir -p /tmp/test && cd /tmp/test
pip install -q /fromager 2>&1 | tail -3
fromager \
  --work-dir ./work-dir \
  --sdists-repo ./sdists-repo \
  --wheels-repo ./wheels-repo \
  bootstrap \
  --full-build \
  --multiple-versions \
  --max-release-age 7 \
  "uv" 2>&1 | tail -15
echo "=== EXIT CODE: $? ==="
echo "=== partial-failures.json ==="
cat work-dir/partial-failures.json 2>/dev/null || echo "Not found"
' 2>&1
```

...and the partial failures:

```
=== partial-failures.json ===
{
  "failures": [
    {
      "name": "uv==0.12.9",
      "phase": "failed during prepare-source phase",
      "error_type": "FileNotFoundError",
      "message": "[Errno 2] No such file or directory: 'cargo'"
    },
    {
      "name": "uv==0.12.8",
      "phase": "failed during prepare-source phase",
      "error_type": "FileNotFoundError",
      "message": "[Errno 2] No such file or directory: 'cargo'"
    },
    {
      "name": "uv==0.12.7",
      "phase": "failed during prepare-source phase",
      "error_type": "FileNotFoundError",
      "message": "[Errno 2] No such file or directory: 'cargo'"
    }
  ]
}
```

For examples' sake, let's say we try in an environment that has `cargo` but not `rustc` and then try again.  This new artifact explains _which versions_ failed, and in what phase:

```
=== partial-failures.json ===
{
  "failures": [
    {
      "name": "maturin==1.15.0",
      "phase": "failed during build phase",
      "error_type": "CalledProcessError",
      "message": "Command '['/tmp/test/work-dir/maturin-1.15.0/build-3.12.14/bin/python3', '/usr/local/lib/python3.12/site-packages/pyproject_hooks/_in_process/_in_process.py', 'build_wheel', '/tmp/tmp6kd4b_50']' returned non-zero exit status 1."
    },
    {
      "name": "uv==0.12.9",
      "phase": "failed during prepare-build phase",
      "error_type": "CalledProcessError",
      "message": "Command '['uv', 'pip', 'install', '--verbose', '--upgrade', '--only-binary', ':all:', '--index-url', 'http://127.0.0.1:36053/simple/', '--trusted-host', '127.0.0.1', 'maturin<2.0,>=1.0']' returned non-zero exit status 1."
    },
    {
      "name": "uv==0.12.8",
      "phase": "failed during prepare-build phase",
      "error_type": "CalledProcessError",
      "message": "Command '['uv', 'pip', 'install', '--verbose', '--upgrade', '--only-binary', ':all:', '--index-url', 'http://127.0.0.1:36053/simple/', '--trusted-host', '127.0.0.1', 'maturin<2.0,>=1.0']' returned non-zero exit status 1."
    },
    {
      "name": "uv==0.12.7",
      "phase": "failed during prepare-build phase",
      "error_type": "CalledProcessError",
      "message": "Command '['uv', 'pip', 'install', '--verbose', '--upgrade', '--only-binary', ':all:', '--index-url', 'http://127.0.0.1:36053/simple/', '--trusted-host', '127.0.0.1', 'maturin<2.0,>=1.0']' returned non-zero exit status 1."
    }
  ]
}
```

...and (as you'd expect), `bootstrap.log` explain the "why" for `maturin`:

```
...running build_ext\nrunning build_rust\nerror: can't find Rust compiler\n\nIf you are using an outdated pip version, it is possible a prebuilt wheel is available for this package but pip is not able to install from it. Installing from the wheel would avoid the need for a Rust compiler.\n\nTo update pip, run:\n\n    pip install --upgrade pip\n\nand then retry package installation.\n\nIf you did intend to build this package from source, try installing a Rust compiler from your system package manager and ensure it is on the PATH during installation. Alternatively, rustup (available at https://rustup.rs) is the recommended way to download and update the Rust compiler toolchain.
```

src/fromager/bootstrapper/_bootstrapper.py
```diff
@@ -1048,9 +1049,36 @@ def _record_failed_version(
...
         logger.warning("%d version(s) failed to bootstrap:", len(self._failed_versions))
-        for (name, ver), exc in self._failed_versions.items():
+        for (name, ver), (exc, _detail) in self._failed_versions.items():
             logger.warning("  %s==%s: %s: %s", name, ver, type(exc).__name__, exc)
 
+    def _write_partial_failures_report(self) -> None:
+        """Write a JSON report of multi-version bootstrap failures.
+
+        Produces a ``partial-failures.json`` file in the work directory
+        so that downstream CI analysis tools can detect soft failures in
```
mprpic: Nitpick: downstream analysis tools don't have to run in CI :-P

src/fromager/bootstrapper/_bootstrapper.py
```diff
@@ -1048,9 +1049,36 @@ def _record_failed_version(
...
+                "name": f"{name}=={ver}",
+                "phase": detail,
+                "error_type": type(exc).__name__,
+                "message": str(exc),
+            }
+            for (name, ver), (exc, detail) in self._failed_versions.items()
+            if ver != "unresolved"
+        ]
+        with open(failures_file, "w") as f:
+            json.dump({"failures": failures}, f, indent=2)
```
mprpic, line 1084: Can we sort here or above in the list comprehension so the files are identical for multiple runs on the same scenario?

ryanpetrello: Addressed.

src/fromager/bootstrapper/_bootstrapper.py
ryanpetrello: Updated.

src/fromager/bootstrapper/_bootstrapper.py
```diff
@@ -1048,9 +1049,39 @@ def _record_failed_version(
...
+        failures = sorted(
+            [
+                {
+                    "name": f"{name}=={ver}",
+                    "phase": detail,
+                    "error_type": type(exc).__name__,
+                    "message": str(exc),
+                }
+                for (name, ver), (exc, detail) in self._failed_versions.items()
+                if ver != "unresolved"
```
andre-motta: The `ver != "unresolved"` filter can produce a degenerate report. In multiple-versions mode, resolution failures are recorded with the version set to `"unresolved"` (see `_resolve_and_add_top_level` and `_handle_phase_error`). If those are the only failures, `finalize()` still writes `partial-failures.json`, but with an empty list: exit code 0, a requested package missing from the collection, and the report carries no information about it. In the mixed case, unresolved packages are silently dropped from the report while still appearing in the log table, so the report can look complete when it is not.

Resolution-only failures look like exactly the class of soft failures this artifact is meant to surface, so it would be good to pin down the contract here:

1. Include unresolved entries (for example, `name` without the `==version` suffix, or a separate `version` field that can be null), or
2. Keep the filter, skip writing the report when the filtered list is empty, and document that resolution failures are out of scope (accepting that they will not appear in mixed runs).

I would prefer option 1. Whichever is chosen, a test covering an unresolved-only `_failed_versions` would lock the behavior in.

Repro against this branch (mirrors the `tmp_context` fixture in `tests/conftest.py`):

```python
import pathlib
import tempfile

from fromager import bootstrapper, context

base = pathlib.Path(tempfile.mkdtemp())
(base / "overrides/patches").mkdir(parents=True)
ctx = context.WorkContext(
    active_settings=None,
    patches_dir=base / "overrides/patches",
    sdists_repo=base / "sdists-repo",
    wheels_repo=base / "wheels-repo",
    work_dir=base / "work-dir",
    variant="cpu",
)
ctx.setup()

bt = bootstrapper.Bootstrapper(ctx, multiple_versions=True)
# What the multi-version error paths record when a top-level
# package has no resolvable version (see _resolve_and_add_top_level):
bt._failed_versions = {
    ("somepkg", "unresolved"): (
        RuntimeError("no versions found for somepkg"),
        "failed to resolve",
    ),
}

print("exit code:", bt.finalize())
print((ctx.work_dir / "partial-failures.json").read_text())
```

```text
exit code: 0
{
  "failures": []
}
```

ryanpetrello: Yea, that's fair.

I was filtering them out because you get a bunch of duplicates/noise when resolution isn't the "root cause", but you're right about the case of a package that just isn't public or, say, a typo.

I've added a change to address this.

src/fromager/bootstrapper/_bootstrapper.py
```diff
@@ -1089,6 +1120,7 @@ def finalize(self) -> int:
 
         if self.multiple_versions and self._failed_versions:
             self._log_failed_versions_table()
+            self._write_partial_failures_report()
```
LalatenduMohanty, line 1130: When running locally we expect work-dir to be reused across bootstrap runs. So a previous partial-failures.json could remain after a later clean run. Should we remove/overwrite it on successful runs, or use timestamped reports  to help identify current report?

LalatenduMohanty (changes requested):

src/fromager/bootstrapper/_bootstrapper.py
LalatenduMohanty: I would suggest filename with timestamp is a better idea as I can compare the reports if required.

LalatenduMohanty: A timestamped failure report still cannot prove that the latest run succeeded, because successful runs produce no report. For automation, a success or a  current-run marker is required.

ryanpetrello: I think for the purposes of CI failure analysis here - the main use case we're solving for - it's probably simplest to just do cleanup afterwards.

I'll update the PR to do that.

ryanpetrello: @LalatenduMohanty I added a fallback case to clean up this file on success or if you re-run outside of multi-version mode.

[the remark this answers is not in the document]
rd4398: Looks like this is fixed. Resolving the thread

LalatenduMohanty: @ryanpetrello can we please add a test for this usecase ` creates an old partial-failures.json and verifies that a later clean run removes it.`

rd4398 (dismissed):
This looks good! Thank you for adding this. I will let @LalatenduMohanty do a final review since he requested changes.

LalatenduMohanty:
@ryanpetrello https://github.com/python-wheel-build/fromager/pull/1314#pullrequestreview-5115523903 seems to be a valid issue. 
here is the code flow

  1. A build phase fails, such as PrepareSource, PrepareBuild, or Build.
  2. _handle_phase_error() checks self.test_mode first at src/fromager/bootstrapper/_bootstrapper.py:987.
  3. Test mode tries a prebuilt fallback.
  4. If fallback fails or is unavailable, it calls record_test_mode_failure() and returns at line 1005.
  5. The later if self.multiple_versions: branch is never reached, so _failed_versions is not populated.
  6. finalize() writes partial-failures.json only when _failed_versions is non-empty at src/fromager/bootstrapper/_bootstrapper.py:1121.

  Therefore, with both flags enabled:

  --test-mode --multiple-versions

With both flags enabled, a build-phase failure that has no successful prebuilt fallback is recorded in the timestamped test-mode report but not in partial-failures.json.
 
If another resolution failure exists, the partial report may still be created, but it will still omit the build-phase failure.

andre-motta (dismissed):

ryanpetrello:
@LalatenduMohanty I believe I've addressed your remaining feedback.

LalatenduMohanty:
@ryanpetrello Thanks. There is one thing remaining which I somehow missed in my previous review, 
The new partial-failures.json artifact is not documented ( also docs/proposals/test-mode.md still says multi-version output is “Logs only") 

I am fine if you want to do that in a follow up PR.

LalatenduMohanty (approved):

LalatenduMohanty:
Created a follow up issue https://github.com/python-wheel-build/fromager/issues/1320

LalatenduMohanty:
https://github.com/python-wheel-build/fromager/pull/1321

PR #1323, opened 2026-09-10
Updating new-resolver-config branch

LalatenduMohanty:
# Pull Request Description

## Why

We need to keep the new-resolver-config uptodate with main branch


rd4398 (approved):
LGTM

I wonder we should auto-merge these updates in the future as the code is already merged on main branch

PR #1327, opened 2026-09-15
feat(sbom): merge Maturin components into wheel SBOM

### LINKED ISSUE #965: Use maturin's SBOM feature to track vendored Rust crates
maturin build backend can now generate SBOMs for all Python packages that use Maturin, see https://www.maturin.rs/sbom.html . It covers the waste majority of Python packages with a Rust component. Some packages may still use setuptools-rust, though.

mprpic:
# Pull Request Description

## What

Import non-excluded CycloneDX components from Maturin-generated wheel SBOMs into Fromager's canonical SPDX document. Relate bundled components to the wheel with CONTAINS, normalize local PURLs, and preserve the original CycloneDX files.

## Why

Maturin generates an SBOM for crates bundled in a Python wheel. These components should be represented in Fromager's SBOM as well.

Closes: #965


tiran:
Why is this necessary? AFAIK the Atlas tool can deal with Maturin's SBOMs. That's how William found out that one of our images was shipping uv with rustls.

rd4398 (approved):
This looks good! Thanks
@tiran @LalatenduMohanty please take a look at this PR when time permits

mprpic:
> Why is this necessary? AFAIK the Atlas tool can deal with Maturin's SBOMs. That's how William found out that one of our images was shipping uv with rustls.

@tiran It's a lot easier to look at one comprehensive SBOM of all components than have to learn how to parse the output of each individual build tool that may produce one. Fromager itself is a build tool so I find it acceptable that it would be within its scope to provide a unified view into the components of a wheel that it builds, by understanding the format of individual SBOMs that may already placed within a wheel by the tools it calls during the build process.

Down the line, we could even provide _more_ information in Fromager's SBOM if maturin (or other build tools) happen to lack a certain feature to add more information into an SBOM or make it more accurate.

I would also be surprised if Atlas processed maturin-generated SBOMs. Do you have a link to one?

mprpic:
> I would also be surprised if Atlas processed maturin-generated SBOMs. Do you have a link to one?

Here is actually an example of an external SBOM (not generated by Fromager) trying to be uploaded to Atlas and failing (line 293): https://konflux-ui.apps.kflux-prd-rh03.nnv1.p1.openshiftapps.com/ns/rhtap-releng-tenant/applications/calunga-v2-index-main/pipelineruns/managed-zx8qm/logs?task=upload-sboms-to-atlas

I think this further underlines the need for us to have one SBOMs that has a common format for SBOM data instead of forcing downstream tools to recognize many different files/formats.

PR #1328, opened 2026-09-15
fix(sources): sdist filenames are not PEP 625 normalized

### LINKED ISSUE #1317: sdist filenames are not PEP 625 normalized
## Problem

[PEP 625](https://peps.python.org/pep-0625/) requires a source distribution to be named `{name}-{version}.tar.gz`, where `name` is normalized per [PEP 503](https://peps.python.org/pep-0503/#normalized-names) and then has `-` replaced with `_`.

`default_build_sdist` builds the filename from the requirement string instead:

https://github.com/python-wheel-build/fromager/blob/4eeb0f415c237578b2b970c516bbf2cee75b409f/src/fromager/sources.py#L536

`req.name` is whatever spelling the requirement was written with, so the result is PEP 625 compliant only by coincidence — when the requested name happens to contain no separators or capitals.

`validate_sdist_filename` does not catch this, because it parses the name with `packaging.utils.parse_sdist_filename`, which canonicalizes before comparing. The non-normalized form validates cleanly.

## Example

From an index built with fromager:

```
published by the index    : colpali-engine-0.3.18.tar.gz
Name: in its own PKG-INFO : colpali_engine
published by upstream     : colpali_engine-0.3.18.tar.gz
PEP 625 requires          : colpali_engine-0.3.18.tar.gz
```

The filename disagrees with the spec, with the sdist's own metadata, and with the upstream artifact for the same version.

In that index 4,814 of 11,813 sdists, across 671 projects, are named this way. Where filenames are compliant it is incidental: `Cython` produces `cython-3.3.0.tar.gz` correctly only because the requested spelling had no separators.

## Impact

Low. `pip` and `uv` resolve and install these without complaint, since `parse_sdist_filename` canonicalizes. What it does affect:

- PyPI has enforced PEP 625 on upload since 2023, so these artifacts cannot be republished to a compliant index without being renamed.
- Filename-keyed comparison against a PyPI-derived set — mirror deduplication, cache keys, "do I already have this file" checks — does not match.

## Suggested fix

Normalize when constructing the filename:

```python
from packaging.utils import canonicalize_name

sdist_filename = (
    ctx.sdists_builds / f"{canonicalize_name(req.name).replace('-', '_')}-{version}.tar.gz"
)
```

jlarkin09:
- Import canonicalize_name from packaging.utils
- Apply PEP 503 canonicalization when constructing sdist filenames
- Add comprehensive test coverage for various name normalization cases

Closes #1317


mprpic (dismissed):
Nice 👍

LalatenduMohanty (dismissed):

LalatenduMohanty (commented):
The pre-commit job is failing because the changed line exceeds the 88-character line length limit.

The offending line in src/fromager/sources.py (line 537 in the PR):

    sdist_filename = ctx.sdists_builds / f"{canonicalize_name(req.name).replace('-', '_')}-{version}.tar.gz"

That's 108 characters — well over the project's configured limit of 88 (set in pyproject.toml). ruff format or black would flag/reformat this.

LalatenduMohanty (approved):

PR #1333, opened 2026-09-18
fix(resolver): warn when a version is skipped due to missing sdist

### LINKED ISSUE #1332: Multi-version bootstrap silently drops versions with no sdist
In multi-version bootstrap mode, versions that match the specifier but have no sdist are silently filtered out during candidate resolution. No warning is logged, no entry is added to partial-failures.json, and no record exists that the version was ever considered.

## Reproduction

Package `drr` has versions 1.0.2 through 1.0.5 on PyPI. Version 1.0.4 has only a wheel (no sdist).

```
$ fromager list-versions "drr>=1.0.2" --distribution-type both
Found 4 version(s)
1.0.2
1.0.3
1.0.4
1.0.5

$ fromager list-versions "drr>=1.0.2" --distribution-type sdist
Found 3 version(s)
1.0.2
1.0.3
1.0.5
```

When bootstrapping with `--multi-version` and `drr>=1.0.2`, only 1.0.2, 1.0.3, and 1.0.5 are processed. 1.0.4 is never mentioned in logs or in partial-failures.json.

## Root cause

`PyPIProvider.validate_candidate()` in `resolver.py` ([lines 945-957](https://github.com/python-wheel-build/fromager/blob/main/src/fromager/resolver.py#L945-L957)) silently rejects candidates that don't match the desired distribution type. The debug log is gated behind `DEBUG_RESOLVER`, which is off by default.

## Expected behavior

In multi-version mode, when a version matching the specifier is dropped because it lacks an sdist, Fromager should either:
1. Log a warning (e.g., "drr==1.0.4: skipped, no sdist available on PyPI")
2. Record it in partial-failures.json with a clear reason
3. Ideally, both.

This matters for downstream observability -- operators need to know when a version exists but was not built, and why.

## Contrast with pinned versions

When pinned (`drr==1.0.4`), Fromager correctly errors out: "found no match for drr==1.0.4... searching for sdists." The gap is only in multi-version mode where other versions satisfy the requirement and the missing one is silently skipped.

ryanpetrello:
Log a warning when a candidate version is filtered out because it has no sdist (wheel-only release). Previously this was gated behind `DEBUG_RESOLVER` and invisible at default verbosity.

## Why WARNING level?

1. **Fromager builds from source.** Its core job is to download sdists and build wheels. When a version exists on PyPI but has no sdist, Fromager cannot build it. That is a noteworthy situation worth exposing to the user by default.

2. **Silent filtering violates the principle of least surprise.** When a user resolves `pkg>=1.0` and gets back 3 versions instead of 4, there is no indication why that matching version was excluded. A warning makes the distinction between "version does not exist" and "version exists but has no sdist" explicit.

3. **This is infrequent in practice.** Wheel-only releases are uncommon. The warning fires once per affected version per resolution, not per candidate file. For most packages, this adds zero output.

## Reproduction

Package `drr` on PyPI: version 1.0.4 has only a wheel (no sdist).

```
$ fromager list-versions "drr>=1.0.2" --distribution-type both
Found 4 version(s): 1.0.2, 1.0.3, 1.0.4, 1.0.5

$ fromager list-versions "drr>=1.0.2" --distribution-type sdist
Found 3 version(s): 1.0.2, 1.0.3, 1.0.5
```

With this change, resolution logs:
```
WARNING: drr==1.0.4: no sdist available, only a wheel (skipped)
```

Fixes #1332


ryanpetrello:
```
~ .venv/bin/fromager -O $TMPDIR/fromager-test bootstrap --multiple-versions --max-release-age 120 --sdist-only "drr>=1.0.2"

09:34:29 WARNING drr: drr==1.0.4: no sdist available, only a wheel (skipped)
09:34:29 INFO drr: drr: found 3 candidate(s) matching drr>=1.0.2
09:34:29 INFO drr: resolved 3 version(s) for drr>=1.0.2
09:34:29 INFO drr: drr>=1.0.2 resolves to 1.0.5
09:34:29 INFO drr: drr>=1.0.2 resolves to 1.0.3
09:34:29 INFO drr: drr>=1.0.2 resolves to 1.0.2
```

rd4398:
Thank you for the fix! Can we add the test here to the code so that it runs in CI?

rd4398 (approved):
LGTM! Thanks for the fix

PR #1336, opened 2026-09-18
fix(bootstrap): warm the resolver cache before adding top-level nodes

### LINKED ISSUE #1334: Top-level pre-resolution leaves an unprocessed, edge-less node in graph.json when a second rule for the same package bypasses cooldown
With `--min-release-age` set, two top-level requirements for the same package (an unpinned rule followed by an exact `==` pin on a release younger than the cooldown window) leave a node in graph.json that bootstrap never processed: no sdist, no build-system edges. `build-parallel` then schedules that node in round 1 and fails to install its build backend from the local wheel server.

## Reproduction

requirements.txt, in this order:

```
codeflare-sdk
codeflare-sdk==0.39.1
```

At the time of the run, 0.39.1 (released 2026-09-15) was inside a 3-day cooldown and 0.39.0 was the newest release outside it.

```
$ fromager --min-release-age 3 bootstrap -r requirements.txt
...
codeflare-sdk: cooldown blocked 1 version(s): 0.39.1
codeflare-sdk: codeflare-sdk resolves to 0.39.0
codeflare-sdk: cooldown bypassed as the top-level requirement uses == pin
codeflare-sdk: codeflare-sdk==0.39.1 resolves to 0.39.1
...
codeflare-sdk-0.39.1: new toplevel dependency codeflare-sdk==0.39.1 resolves to 0.39.1
```

There is no `new toplevel dependency codeflare-sdk==0.39.0` line and no sdist for 0.39.0 is produced, but graph.json contains both nodes:

```
"codeflare-sdk==0.39.0": {"pre_built": false, "edges": []}          # only a ROOT edge points to it
"codeflare-sdk==0.39.1": {"pre_built": false, "edges": [{"key": "poetry-core==2.4.1", "req_type": "build-system", ...}, ...]}
```

```
$ fromager build-parallel work-dir/graph.json
round 1: starting to build 16 node(s): codeflare-sdk==0.39.0, flit-core==3.12.0, ..., poetry-core==2.4.1, ...
ERROR codeflare-sdk-0.39.0: Failed to install build-system dependency ... poetry-core>=1.6.0
```

Swapping the two lines (pin first) makes the problem disappear, so it is order-dependent.

## Root cause

1. `Bootstrapper.bootstrap()` calls `_resolve_and_add_top_level()` for each rule in file order, which resolves and immediately adds the result to the graph ([`_bootstrapper.py#L139-L189`](https://github.com/python-wheel-build/fromager/blob/0883ce2/src/fromager/bootstrapper/_bootstrapper.py#L139-L189)). It then pushes a `Resolve` item that resolves the same rule a second time ([`_bootstrapper.py#L277-L288`](https://github.com/python-wheel-build/fromager/blob/0883ce2/src/fromager/bootstrapper/_bootstrapper.py#L277-L288)).
2. The resolver caches known versions per package and filters them by specifier only, without re-applying the cooldown ([`bootstrap_requirement_resolver.py#L118-L136`](https://github.com/python-wheel-build/fromager/blob/0883ce2/src/fromager/bootstrap_requirement_resolver.py#L118-L136), [`#L295-L309`](https://github.com/python-wheel-build/fromager/blob/0883ce2/src/fromager/bootstrap_requirement_resolver.py#L295-L309)). That sharing is intentional (#1187). So the first resolution of the unpinned rule returns 0.39.0, the pinned rule adds 0.39.1 to the shared cache, and the second resolution of the unpinned rule returns 0.39.1.
3. `Start.run` skips `add_to_graph` for `TOP_LEVEL` because the node was "already added in _resolve_and_add_top_level" ([`_start.py#L47-L55`](https://github.com/python-wheel-build/fromager/blob/0883ce2/src/fromager/bootstrapper/_start.py#L47-L55)), and the second `Start(0.39.1)` is dropped by `has_been_seen`. Nothing removes or corrects the stale 0.39.0 node from step 1.

## Expected behavior

graph.json should only contain top-level nodes that bootstrap actually processed. Either the top-level graph nodes should be derived from the versions the `Resolve` phase settles on (after the shared cache is warm), or each version added in pre-resolution should be bootstrapped.

## Impact

Any package that is listed unpinned and, further down, pinned to a release inside the cooldown window hits this. The failure only surfaces when the wheel for the stale version is not already on the cache server, so it stayed hidden until a cache index was recreated.

andre-motta:
Fixes #1334.

Top-level requirements were resolved and written to the dependency graph one rule at a time. A later rule for the same package (an exact `==` pin that bypasses the release-age cooldown) widens the resolver's shared per-package cache, so the `Resolve` phase settled on a different version than the one already in the graph, and `Start` skips graph insertion for top-level requirements. The stale node stayed in graph.json with `edges: []` and `build-parallel` scheduled it in round 1.

Fix: resolve every top-level rule once before any of them is added to the graph, so the recorded version matches what the `Resolve` phase uses. The extra pass only hits the resolver cache on the second lookup, and errors are still reported by the existing code path.

Tests: regression for the unpinned-then-pinned case (fails on main with `{'pkg==0.39.0', 'pkg==0.39.1'} == {'pkg==0.39.1'}`) and a multiple-versions guard. `pytest`: 952 passed, 3 skipped. `ruff`, `mypy` clean.


andre-motta:
@coderabbitai review

src/fromager/bootstrapper/_bootstrapper.py
```diff
@@ -266,6 +266,22 @@ def bootstrap(self, requirements: list[Requirement]) -> None:
         Args:
             requirements: Top-level requirements to resolve and bootstrap.
         """
+        # Warm the resolver cache with every top-level rule before adding any
+        # of them to the graph: a later rule for the same package (e.g. an
+        # exact pin that bypasses the release-age cooldown) changes what an
+        # earlier rule resolves to. Errors are reported by
+        # _resolve_and_add_top_level below.
+        for req in requirements:
```
rd4398, line 274: The warm-up loop resolves every top-level requirement with `req_type=TOP_LEVEL`, which triggers cooldown bypass logic for == pins (via resolve_package_cooldown). However, the warm-up swallows all
exceptions with a bare `except Exception: pass`, including ValueError from PEP 508 direct-reference URLs (req.url). While `_resolve_and_add_top_level` handles these errors properly on the second call, the silent swallow means the warm-up could mask unexpected errors during debugging.

More importantly, the warm-up currently doesn't set the `requirement_ctxvar` context, so any
log messages emitted during warm-up resolution (e.g. 'cooldown bypassed as the top-level requirement uses == pin') will lack the package name context that the main loop provides via requirement_ctxvar.set(req). This inconsistency means duplicate log lines: once without context during warm-up, once with context during `_resolve_and_add_top_level`

Suggestion:

Set requirement_ctxvar around each warm-up call to maintain consistent log context, and
consider logging a debug message when catching the exception so silent failures are traceable:

rd4398 (changes requested):
This looks mostly good. I have left a comment about exception handling.

Thank you for the fix!

src/fromager/bootstrapper/_bootstrapper.py
andre-motta: Fixed in 9335af1. Each warm-up call now runs inside `req_ctxvar_context(req)`, so any log lines emitted during warm-up (such as the cooldown bypass message) carry the same package name context as the main loop. The bare `except Exception: pass` is replaced with a `logger.debug()` that records the deferred error, so failures are traceable in debug output while the second call in `_resolve_and_add_top_level` still reports them through the existing path.

rd4398 (approved):
LGTM! Thanks

rd4398:
@mergify rebase

andre-motta:
@mergify update

rd4398:
@LalatenduMohanty I am struggling to understand why mergify is not working here. Do you know what is going on?

PR #1341, opened 2026-09-22
build(packaging): modernize metadata, adopt PEP 735 dependency groups

tiran:
# Pull Request Description

## What

Move packaging metadata to current PyPA specs and relocate dev tooling out of published metadata:

- Use SPDX license expression: `license = "Apache-2.0"` plus `license-files`, and drop the deprecated `License ::` classifier (PEP 639). Requires `hatchling>=1.27` to emit Metadata-Version 2.4+.
- Move the `build`/`mypy`/`test`/`docs` sets from `[project.optional-dependencies]` to `[dependency-groups]` (PEP 735) so they no longer leak into published wheel metadata, and rewire every Hatch env from `features` to `dependency-groups`.
- Migrate the group consumers in the same change: `pip install .[build]` becomes `pip install --group build` in the release workflow.
- Require Hatch >= 1.18.0 via `tool.hatch.requires-hatch` (the version that enforces the field) in CI installs and CONTRIBUTING.
- Add a zizmor GitHub Actions security-audit job to the Check workflow.

Generated by applying the modernize-python-metadata and secure-python-release-pipeline skills from
https://github.com/tiran/agent-skills

## Why

Use modern Python packaging standards like PEP 735 and PEP 639.


pyproject.toml
```diff
@@ -289,7 +295,7 @@ check = [
 
 [tool.hatch.envs.docs]
 description = "Sphinx docs."
-features = ["docs"]
+dependency-groups = ["docs"]
```
LalatenduMohanty, line 298: The PR moves docs into [dependency-groups], but .readthedocs.yaml still requests extra_requirements: [docs].
That configuration means pip install '.[docs]'; it does not install the new dependency group.

LalatenduMohanty (changes requested):
docs/develop.md:132 still references [project.optional-dependencies.build]; update it to [dependency-groups].

.github/workflows/check.yaml
```diff
@@ -69,7 +69,26 @@ jobs:
...
+      # Needed to upload SARIF results to code scanning
+      security-events: write
+
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@cc914d7f3750a2d13d75c7f184a1060aa0e9d482  # v0.6.4
```
LalatenduMohanty, line 94: Whats our expectation from this?  zizmor points out potential security problems, but reviewers decide whether they should prevent merging?

PR #1342, opened 2026-09-22
fix(resolver): suppress cooldown warning for localhost indexes

ryanpetrello:
`localhost` simple indexes (e.g. fromager's local wheel server) never provide PEP 691 upload timestamps, so the release-age cooldown warning is noise.  This results in lots of `warn()` logs like this:

image

## Test plan

- [x] Existing cooldown tests pass (35/35)
- [x] Updated `test_local_wheel_server_allows_without_upload_time` to assert the warning is absent for localhost


src/fromager/resolver.py
```diff
@@ -911,6 +912,20 @@ def __init__(
...
+        Silently skips the cooldown check for localhost indexes, which never
+        provide upload timestamps.
+        """
+        if (
+            self.cooldown is not None
+            and candidate.upload_time is None
+            and urlparse(self.sdist_server_url).hostname in _LOCALHOST_NAMES
+        ):
+            return False
+        return super().is_blocked_by_cooldown(candidate)
```
LalatenduMohanty, line 927: Do we want to check cooldown for the local wheel server. The cooldown gate should have already checked before downloading the bits.  But the challenge is if we rerun with the same --ouput-dir stale wheels from a previous run will bypass the cooldown.

LalatenduMohanty (changes requested):

src/fromager/resolver.py
LalatenduMohanty: The current code misses external --build-wheel-server-url intentionally. I guess we do not want to suppress the warnings for this usecase?

ryanpetrello: 1.  No.  If you rerun with the same `--output-dir`, previously built wheels are served by the local wheel server without re-checking cooldown (that's already the case today). The local wheel server is PEP 503 only and will never provide upload timestamps, so the cooldown is fundamentally unenforceable there.  This PR only changes whether a warning is logged (because it's very noisy), not whether the cooldown is enforced.

2.  `--build-wheel-server-url` case is intentionally left alone. Keeping the warning is appropriate there because the user _should_ be warned that their _external_ wheel server is not providing timestamp data for cooldown enforcement.

LalatenduMohanty (approved):

